PR #17, opened 2017-07-18
DOC: update nginx container name

LaureVergeron:
Update nginx container name, and swarm-production/README file size in
 bash examples.


LaureVergeron:
@baslking : fixes #18 ; this one was forgotten, will merge

Issue #19, opened 2017-08-02
metadata and data ports are reversed


LaurenSpiegel:
metadata should be 9990 and data 9991.  the opposite is in the yml file.

PR #20, opened 2017-08-09
use correct ports

vrancurel:
Ports in docker-stack.yml are swapped between data and metadata.


philipyoo:
Mentioned in issue #19  :)

PR #21, opened 2017-09-08
ft: backbeat integration for zenko docker stack deployment

jonathan-gramain:
REVIEW ONLY

It's relying on some base images in zenko docker hub, and some backbeat and arsenal PRs not yet merged, so review only for now.

### DIFF
--- a/images/kafka/server.properties.backbeat
+++ b/images/kafka/server.properties.backbeat
@@ -1,0 +1,108 @@
+# see kafka.server.KafkaConfig for additional details and defaults
+
+############################# Server Basics #############################
+
+# The id of the broker. This must be set to a unique integer for each broker.
+broker.id=0
+
+# Switch to enable topic deletion or not, default value is false
+#delete.topic.enable=true
+
+############################# Socket Server Settings #############################
+
+# The address the socket server listens on. It will get the value returned from 
+# java.net.InetAddress.getCanonicalHostName() if not configured.
+#   FORMAT:
+#     listeners = listener_name://host_name:port
+#   EXAMPLE:
+#     listeners = PLAINTEXT://your.host.name:9092
+#listeners=PLAINTEXT://:9092
+
+# Hostname and port the broker will advertise to producers and consumers. If not set, 
+# it uses the value for "listeners" if configured.  Otherwise, it will use the value
+# returned from java.net.InetAddress.getCanonicalHostName().
+#advertised.listeners=PLAINTEXT://your.host.name:9092
+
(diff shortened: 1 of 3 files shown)


images/kafka/server.properties.backbeat
```diff
@@ -0,0 +1,108 @@
...
+num.io.threads=8
+
+# The send buffer (SO_SNDBUF) used by the socket server
+socket.send.buffer.bytes=102400
+
+# The receive buffer (SO_RCVBUF) used by the socket server
+socket.receive.buffer.bytes=102400
+
+# The maximum size of a request that the socket server will accept (protection against OOM)
+socket.request.max.bytes=104857600
```
LaurenSpiegel: how were these numbers determined?

swarm-production/docker-stack.yml
```diff
@@ -76,27 +79,73 @@ services:
...
+      mode: replicated
+      placement:
+        constraints:
+          - node.labels.io.zenko.type == storage
+
+  backbeat:
+    image: zenko/backbeat
+    networks:
+      - backend
+    command: npm start
```
LaurenSpiegel: is this supposed to start a task? how do we determine whether we are starting the replication task or the clueso task?

images/kafka/server.properties.backbeat
jonathan-gramain: Those values are from the default kafka config file, so I assume they are sensible default values (actually I started from Vianney's branch without looking closely at this file except the critical params but that does not waive the need for an explanation :smile: )

swarm-production/docker-stack.yml
jonathan-gramain: I added the `start` rule in https://github.com/scality/backbeat/pull/88, which starts both the queue populator and the replication processor. We would need to selectively start the services that we'll need at processing stage, say replication/clueso/both, though I don't know what's the best way to deal with this at this point. Any idea?

jonathan-gramain: Note that on the populator side I assume there will be one populator for all services, with pluggable filters for each service (again this has to be properly designed)

LaurenSpiegel: I like the idea of pluggable filters. Clueso has other tasks beyond the queue (creating the METADATA bucket for instance). Maybe there could be environment variables here for clueso or replication and if those are set, in backbeat we then start whatever processes beyond the populator are necessary?

ssalaues:
Since backbeat has been merged into both stacks, I will be closing this issue. Please feel free to reopen if needed.

PR #22, opened 2017-09-08
FT: expose UtAPI port in CloudServer

LaureVergeron:
- In current deployments, the 8100 port is not exposed by default in
 production stacks. This is limiting as it requires either executing
 commands from inside the CloudServer frontend container, or deploying
 an extra UTAPI service. This commit exposes the UTAPI port by default.

### DIFF
--- a/swarm-production/docker-stack.yml
+++ b/swarm-production/docker-stack.yml
@@ -41,6 +41,7 @@ services:
     image: scality/s3server
     ports:
       - "8000"
+      - "8100:8100"
     networks:
       - backend
       - frontend-dmz


LaureVergeron:
Utapi is in the node_modules of S3 Server, and Zenko deploys a redis container, which I believe acts as the Utapi Client, so I don't see why we wouldn't be able to use Utapi via S3 Server (that's what we do at customers);
As for the vault implementation, I know the Utapi team is working for a vaultless authentication (see https://github.com/scality/utapi/pull/149 ).
With these two informations, it looks like we could use Utapi via S3 Server, no?

PR #27, opened 2017-12-04
Update for Backbeat and Orbit

RachedBenMustapha:


swarm-production/docker-stack.yml
```diff
@@ -1,12 +1,12 @@
 ---
 
-version: "3.1"
+version: "3.4"
 
 services:
   s3-data:
-    image: scality/s3server
+    image: zenko/cloudserver:pensieve-0
```
LaurenSpiegel: are we going to maintain 2 docker images for open source?

swarm-production/docker-stack.yml
```diff
@@ -38,17 +39,21 @@ services:
...
     ports:
-      - "8000"
+      - "8001"
     networks:
-      - backend
-      - frontend-dmz
+      backend:
+        aliases:
```
LaurenSpiegel: why an alias?

swarm-production/docker-stack.yml
```diff
@@ -38,17 +39,21 @@ services:
...
+          - zenko-cloudserver-replicator
+      frontend-dmz:
     environment:
       DATA_HOST: s3-data
       METADATA_HOST: s3-metadata
       REDIS_HOST: cache
       ENDPOINT: "${ENDPOINT:-zenko}"
+      MANAGEMENT_ENDPOINT: 'https://api.zenko.io'
+      REMOTE_MANAGEMENT_DISABLE: "${REMOTE_MANAGEMENT_DISABLE:-0}"
```
LaurenSpiegel: why is the value in this format?

swarm-production/docker-stack.yml
```diff
@@ -38,17 +39,21 @@ services:
           - node.labels.io.zenko.type == storage
 
   s3-front:
-    image: scality/s3server
+    image: zenko/cloudserver:pensieve-0
     ports:
-      - "8000"
+      - "8001"
```
LaurenSpiegel: why changing from 8000?

swarm-production/docker-stack.yml
```diff
@@ -89,6 +94,90 @@ services:
         delay: "10s"
         monitor: "5s"
 
+  queue:
+    image: 'wurstmeister/kafka:1.0.0'
```
LaurenSpiegel: why not our kafka docker image?

swarm-production/docker-stack.yml
```diff
@@ -89,6 +94,90 @@ services:
...
+      KAFKA_ADVERTISED_PORT: 9092
+      KAFKA_ADVERTISED_HOST_NAME: queue
+      KAFKA_CREATE_TOPICS: 'backbeat-replication:1:1'
+      KAFKA_ZOOKEEPER_CONNECT: 'quorum:2181'
+    ports:
+      - 9092
+    networks:
+      - backend
+    volumes:
+      - "queue-journal:/kafka:rw"
```
LaurenSpiegel: do you have to have a volume for the journal as well as a volume for zk?

swarm-production/docker-stack.yml
```diff
@@ -89,6 +94,90 @@ services:
...
+      - "quorum-datalog:/datalog:rw"
+    deploy:
+      mode: replicated
+      replicas: 1
+      placement:
+        constraints:
+          - node.labels.io.zenko.type == storage
+
+  backbeat-producer:
+    image: 'zenko/backbeat:pensieve-2'
```
LaurenSpiegel: for search we have a separate image that is just one backbeat producer for everything. we need to reconcile this.

swarm-production/docker-stack.yml
LaurenSpiegel: does your image have S3DATA=multiple by default?  how will people configure the locations without pensieve?

swarm-production/docker-stack.yml
RachedBenMustapha: this one does auto-creation and auto-configuration while the scality one is vanilla (because configuration happens in the proprietary installer).

swarm-production/docker-stack.yml
RachedBenMustapha: I would not encourage using the same volume for different services

swarm-production/docker-stack.yml
RachedBenMustapha: I thought search would not be swarm-based? This service just moves events from MD to kafka so I don't see why it would be incompatible.

swarm-production/docker-stack.yml
RachedBenMustapha: S3DATA=multiple is hardcoded because it would not work with Orbit otherwise. The gradual switch from the old image to the new one was decided for Orbit - the source dockerfile is still the same.

swarm-production/docker-stack.yml
RachedBenMustapha: shell fallback

swarm-production/docker-stack.yml
RachedBenMustapha: allows talking to s3 through the backend network with a known endpoint (errors out with 400 otherwise)

swarm-production/docker-stack.yml
RachedBenMustapha: to make sure that incompatible versions won't work with this stack as this will be the number one problem. overall as long as it's coherent inside the stack the port does not matter

swarm-production/docker-stack.yml
LaurenSpiegel: as discussed, all of the locations are file at start up.

swarm-production/docker-stack.yml
LaurenSpiegel: we have a different branch of backbeat for search which we were intending to go into Zenko CE but we can update it later.

PR #36, opened 2018-02-07
ZENKO-9 Replicaset Compatibility

ssalaues:
I have left it disabled by default but can be enabled simply in the zenko/values.yml or with the `--set mongodb-replicaset.enabled=true` flag when deploying

### DIFF
--- a/helm/zenko/charts/cloudserver-front/templates/deployment.yaml
+++ b/helm/zenko/charts/cloudserver-front/templates/deployment.yaml
@@ -28,6 +28,16 @@ spec:
               value: "{{- printf "%s-%s" .Release.Name "s3-metadata" | trunc 63 | trimSuffix "-" -}}"
             - name: REDIS_HOST
               value: "{{- printf "%s-%s" .Release.Name "redis" | trunc 63 | trimSuffix "-" -}}"
+{{- if .Values.global.mongodbBackend.enabled }}
+            - name: S3METADATA
+              value: "mongodb"
+            - name: MONGODB_HOST
+              value: "{{ .Release.Name }}-mongodb-replicaset"
+            - name: MONGODB_PORT
+              value: "{{ .Values.global.mongodbBackend.port }}"
+            - name: MONGODB_RS
+              value: "{{ .Values.global.mongodbBackend.replicaSet }}"
+{{- end}}
 {{- if .Values.orbit.enabled }}
             - name: REMOTE_MANAGEMENT_DISABLE
               value: "0"
--- a/helm/zenko/charts/mongodb-replicaset/.helmignore
+++ b/helm/zenko/charts/mongodb-replicaset/.helmignore
@@ -1,0 +1,21 @@
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
+.DS_Store
+# Common VCS dirs
+.git/
+.gitignore
+.bzr/
+.bzrignore
+.hg/
+.hgignore
+.svn/
+# Common backup files
+*.swp
+*.bak
+*.tmp
+*~
+# Various IDEs
+.project
+.idea/
+*.tmproj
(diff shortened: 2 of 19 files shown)


helm/zenko/charts/cloudserver-front/templates/deployment.yaml
```diff
@@ -28,6 +28,16 @@ spec:
               value: "{{- printf "%s-%s" .Release.Name "s3-metadata" | trunc 63 | trimSuffix "-" -}}"
             - name: REDIS_HOST
               value: "{{- printf "%s-%s" .Release.Name "redis" | trunc 63 | trimSuffix "-" -}}"
+{{- if index .Values "mongodb-replicaset" }}
+            - name: S3METADATA
+              value: mongodb
+            - name: MONGODB_HOST
+              value: {{- printf "%s-mongodb-replicaset.%s.svc.cluster.local" .Release.Name .Release.Namespace }}
```
NicolasT: `%s-mongodb-replicaset` should do. No need to include the namespace, and `svc.cluster.local` can be deployment-dependent.

helm/zenko/charts/cloudserver-front/templates/deployment.yaml
```diff
@@ -28,6 +28,16 @@ spec:
...
+            - name: MONGODB_PORT
+              value: {{- template "mongodb-replicaset.port"  }}
```
NicolasT: I'm surprised this works, because this seems to be inter-chart code, which is not supposed to be possible 😕

helm/zenko/charts/mongodb-replicaset/.helmignore
```diff
@@ -0,0 +1,21 @@
+# Patterns to ignore when building packages.
```
NicolasT: Could you, in the commit message, specify very specifically
- From which source these files were imported
- What the revision was of the foreign tree from which these files were copied
- Optionally, what the revision was of `stable/mongodb-replicaset` or whatever that was copied in that foreign tree

helm/zenko/charts/s3-data/templates/deployment.yaml
```diff
@@ -32,6 +32,16 @@ spec:
               value: /data
             - name: LISTEN_ADDR
               value: 0.0.0.0
+{{- if index .Values "mongodb-replicaset" }}
```
NicolasT: Same comments as above

helm/zenko/requirements.yaml
```diff
@@ -0,0 +1,5 @@
+dependencies:
```
NicolasT: No need for this file, since the Chart was imported as a subchart

helm/zenko/values.yaml
```diff
@@ -27,6 +27,15 @@ cloudserver-front:
     keyId: deployment-specific-access-key
     secretKey: deployment-specific-secret-key
 
+#
+#
+#
+mongodb-replicaset:
+  enabled: false
+  replicaset: rs0
```
NicolasT: Could you leave a comment describing what this value means?

helm/zenko/charts/cloudserver-front/templates/deployment.yaml
ssalaues: I believe you are right and this isn't working how I intended necessarily. I will go back over this.

helm/zenko/charts/mongodb-replicaset/.helmignore
ssalaues: Commit message has been amended

ssalaues:
I believe all the previous issues have been addressed

vrancurel:
shall we merge?

vrancurel:
@ssalaues can you quickly enable mongo by default (and replica set also) ?

ssalaues:
Update: mongodb is now enabled by default however since the current cloudserver images are not updated with the mongodb backend, it may cause issues with people trying to test a deployment.

PR #37, opened 2018-02-10
various minor stuff

vrancurel:
Fixes container versions

### DIFF
--- a/helm/zenko/charts/cloudserver-front/templates/deployment.yaml
+++ b/helm/zenko/charts/cloudserver-front/templates/deployment.yaml
@@ -31,10 +31,8 @@ spec:
 {{- if .Values.global.mongodbBackend.enabled }}
             - name: S3METADATA
               value: "mongodb"
-            - name: MONGODB_HOST
-              value: "{{ .Release.Name }}-mongodb-replicaset"
-            - name: MONGODB_PORT
-              value: "{{ .Values.global.mongodbBackend.port }}"
+            - name: MONGODB_HOSTS
+              value: "{{ .Release.Name }}-mongodb-replicaset-0.{{ .Release.Name }}-mongodb-replicaset:27017,{{ .Release.Name }}-mongodb-replicaset-1.{{ .Release.Name }}-mongodb-replicaset:27017,{{ .Release.Name }}-mongodb-replicaset-2.{{ .Release.Name }}-mongodb-replicaset:27017"
             - name: MONGODB_RS
               value: "{{ .Values.global.mongodbBackend.replicaSet }}"
 {{- end}}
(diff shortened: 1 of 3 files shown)


helm/zenko/charts/cloudserver-front/templates/deployment.yaml
```diff
@@ -31,10 +31,8 @@ spec:
 {{- if .Values.global.mongodbBackend.enabled }}
             - name: S3METADATA
               value: "mongodb"
-            - name: MONGODB_HOST
-              value: "{{ .Release.Name }}-mongodb-replicaset"
-            - name: MONGODB_PORT
-              value: "{{ .Values.global.mongodbBackend.port }}"
+            - name: MONGODB_HOSTS
+              value: "{{ .Release.Name }}-mongodb-replicaset-0.zenko-mongodb-replicaset.default.svc.cluster.local:27017,{{ .Release.Name }}-mongodb-replicaset-1.zenko-mongodb-replicaset.default.svc.cluster.local:27017,{{ .Release.Name }}-mongodb-replicaset-2.zenko-mongodb-replicaset.default.svc.cluster.local:27017"
```
ssalaues: change any hardcoded "zenko" reference to {{ .Release.Name }} because it is deployment specific

ssalaues: Same goes with the "default" and "cluster" (each should have their own variable to use)

NicolasT: You don't need to suffix the namespace and svc.cluster.local etc. Resolving should happen in the 'right' domain already.

helm/zenko/charts/cloudserver-front/templates/deployment.yaml
```diff
@@ -31,10 +31,8 @@ spec:
 {{- if .Values.global.mongodbBackend.enabled }}
             - name: S3METADATA
               value: "mongodb"
-            - name: MONGODB_HOST
-              value: "{{ .Release.Name }}-mongodb-replicaset"
-            - name: MONGODB_PORT
-              value: "{{ .Values.global.mongodbBackend.port }}"
+            - name: MONGODB_HOSTS
+              value: "{{ .Release.Name }}-mongodb-replicaset-0.{{ .Release.Name }}-mongodb-replicaset:27017,{{ .Release.Name }}-mongodb-replicaset-1.{{ .Release.Name }}-mongodb-replicaset:27017,{{ .Release.Name }}-mongodb-replicaset-2.{{ .Release.Name }}-mongodb-replicaset:27017"
```
NicolasT: What you really want to do is
- Refactor this in a helper function in `_helpers.tpl`
- Use `printf`, `trunc` and `trimSuffix` accordingly, in said helper.
But OK for now I guess.

vrancurel:
thanks

PR #38, opened 2018-02-12
Use upstream Charts and requirements

NicolasT:
These commits:
- `git mv helm charts` to be aligned with other repositories and naming convention
- Remove the embedded `zookeeper` Chart and set up `requirements.yaml` to use https://github.com/scality/zenko-zookeeper v0.0.4
- Remove the embedded `prometheus` and `mongodb-replicaset` Charts and set up `requirements.yaml` to use the one from the stable channel
- Move all sub-charts into `charts/`, and update Zenko's `requirements.yaml` to use them accordingly


NicolasT:
We may want to consider to not let the `zenko` top-level Chart depend on (e.g.) `mongodb-replicaset`, but instead let `cloudserver-front` depend on it, as this is the 'most correct' dependency definition.

NicolasT:
TODO Due to use of `zenko-zookeeper`, the services using `zookeeper` may need changes to resolve the hosts etc.

NicolasT:
Zookeeper hosts configuration changes done.

NicolasT:
I'll update this to use https://github.com/scality/zenko-zookeeper 0.0.4 once released, such that I can test the end result on my cluster.

NicolasT:
Now `zenko-zookeeper` 0.0.4 was integrated, I tried to deploy on my cluster and all services start nicely.

charts/value-dev.yml
```diff
@@ -1,8 +1,8 @@
-zookeeper:
+zenko-zookeeper:
   Image: localhost:5000/zookeeper
   ImageTag: latest
   storageClass: local-storage
-  storage: 20G
+  storage: 10G
```
NicolasT: Note: before, this change was made inside the `zookeeper` Chart. Better to put it here.

charts/zenko/values.yaml
```diff
@@ -27,10 +27,6 @@ cloudserver-front:
     keyId: deployment-specific-access-key
     secretKey: deployment-specific-secret-key
 
-  mongodb:
```
NicolasT: These are the defaults in `cloudserver-front`.

PR #39, opened 2018-02-12
Fix MongoDB ReplicaSet sub-chart and integration

NicolasT:
In 6d6ae5bd9f044ad9e5e30b607b1fd6692df7ec24, the MongoDB ReplicaSet Chart from the upstream stable channel was imported, though with some minor modifications.

This commit
- Reverts those modification, such that the imported Chart now matches upstream
- Gets rid of the modifications made to the upstream Chart in 846e48b91485eec58c1a1908e409a91fff47eee3 because they're not necessary
- Changes the top-level Chart's `values.yaml` to get back the desired configuration, and updates changes to the `cloudserver-front` deployment accordingly.

Note `values.yaml` now contains a couple of values which are not strictly required to be present. Instead, later on, we could use 'import values' (https://github.com/kubernetes/helm/blob/master/docs/charts.md#using-the-child-parent-format).

### DIFF
--- a/helm/zenko/values.yaml
+++ b/helm/zenko/values.yaml
@@ -27,6 +27,10 @@ cloudserver-front:
     keyId: deployment-specific-access-key
     secretKey: deployment-specific-secret-key
 
+  mongodb:
+    enabled: true
+    replicaSet: rs0
+
 prometheus:
   rbac:
     create: true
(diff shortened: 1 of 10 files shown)


helm/zenko/values.yaml
```diff
@@ -43,15 +48,7 @@ prometheus:
...
-  mongodbBackend: 
-    enabled: true
-    replicaSet: rs0
-    # The 'rs0' is the default mongo replicaset name
-    replicas: 5
-    # The number of replicas to have in the cluster
-    port: 27107
-    # The default port number used by mongo
-
+mongodb-replicaset:
```
ssalaues: Shouldn't we keep some comments to describe what all these values mean?

NicolasT: Not sure, TBH. We're not inventing new configuration values, simply overriding whatever is passed by default in the mongodb-replicaset chart values.yaml, where they are (or should be) documented. Duplicating this here may make things go out of sync easily.

NicolasT:
@ssalaues Still approve?

ssalaues:
Looks good thanks for the proper import! I should've been more critical of the changes I made.

PR #45, opened 2018-02-14
Add the mongodb 3.6.2 image and fix value-dev.yml

ballot-scality:


ballot-scality:
Would like to have @NicolasT approval before merging it into his PR

charts/cloudserver-front/values.yaml
```diff
@@ -18,7 +18,7 @@ mongodb:
 replicaCount: 1
 image:
   repository: zenko/cloudserver
-  tag: 0.1.4
+  tag: 0.1.5
```
NicolasT: Over time, changes like these will also require to bump the version in `Chart.yaml`, but I guess we're good for now.

charts/value-dev.yml
```diff
@@ -1,6 +1,4 @@
 zenko-zookeeper:
-  Image: localhost:5000/zookeeper
```
NicolasT: How does @Zempashi feel about this? I removed it also at first, he thought it should stay. What's the final call?

charts/zenko/values.yaml
```diff
@@ -46,3 +46,8 @@ prometheus:
 mongodb-replicaset:
   replicaSet: rs0
   replicas: 5
+  image:
```
NicolasT: I assume this has been tested. The approach is right, so if it works LGTM.

charts/cloudserver-front/values.yaml
ballot-scality: yes you're right, I just tested it but I will bump the charts number as well.

charts/zenko/values.yaml
ballot-scality: Yes it does. I could successfully register my instance into orbit and create a bucket and

charts/cloudserver-front/values.yaml
NicolasT: Note bumping the number will not be sufficient. From memory:

- Bump the number in cloudserver-front's `Chart.yaml`
- Bump the requirement version in zenko's `requirements.yaml`
- Rebuild zenko's `requirement.lock` through `helm dep update`

charts/zenko/values.yaml
vrancurel: So far 3.6.2 is the latest stable so lets go for it

charts/value-dev.yml
vrancurel: What is the benefit of keeping a link to the local registry since we have an official image on dockerhub. It makes sense to also support having all zenko images into a local registry for some specific deployments.

charts/zenko/requirements.yaml
```diff
@@ -32,5 +32,5 @@ dependencies:
   version: "0.1.0"
   repository: "file://../s3-metadata"
 - name: cloudserver-front
-  version: "0.1.0"
+  version: "0.1.1"
```
NicolasT: This would introduce a bump of the Zenko Chart version as well, actually...

charts/value-dev.yml
NicolasT: @Zempashi's rationale was that a developer may wish to deploy using some (locally modified) version of the zenko-zookeeper image.

Zempashi: There is two kinds of dev here: dev on charts (and here it make sense to keep the images from the repositories) and dev on image (and it's a pain on kubernetes without something like above). We wan create two separate files, or now that an image is release: just manually override this when working on docker image. 
The rational for me to keep this as the default behaviour: if there's change an a docker file and the helm (to reflect change in the docker image). The helm install will fail on that branch (until the branch is 'released'). If it's acceptable go ahead to remove this.

PR #48, opened 2018-02-17
feat(chart): update to `zenko-zookeeper` 0.0.6

NicolasT:
Also update `value-dev.yml` accordingly, since the upstream changed some
value definitions.

See: https://github.com/scality/zenko-zookeeper/pull/7
See: https://github.com/scality/zenko-zookeeper/pull/7/commits/355ccd6d6c7c233034455baa8071ddf4e5d5a34d

### DIFF
--- a/charts/value-dev.yml
+++ b/charts/value-dev.yml
@@ -1,6 +1,11 @@
 zenko-zookeeper:
-  storageClass: local-storage
-  storage: 10G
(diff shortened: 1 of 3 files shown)


NicolasT:
Can reviewer(s) validate  the new `value-dev.yml` on the 'standard' K8s deployment?

PR #55, opened 2018-02-23
Readme Clarifications and Debug options for Cloudserver

ssalaues:
UPDATED PR:
-Added a debug option to cloudserver helm
-Updated the README for clarification for the custom Orbit endpoint

### DIFF
--- a/charts/README.md
+++ b/charts/README.md
@@ -30,9 +30,9 @@ $ helm install --name zenko --set ingress.enabled=true --set ingress.hosts[0]=ze
 
 See `values.yml` for more options.
 
-One option to highlight is the ability to enabled [Orbit] integration (which is
-disabled by default) and set a custom Orbit API endpoint (for development
-purposes). 
+One option to highlight is the ability to enable integration with the
+[Orbit] management UI (which is disabled by default) and set a custom
+Orbit API endpoint (for development purposes). 
 
 To enable Orbit use:
 
--- a/charts/README.md
+++ b/charts/README.md
@@ -32,13 +32,15 @@ See `values.yml` for more options.
 
 One option to highlight is the ability to enabled [Orbit] integration (which is
 disabled by default) and set a custom Orbit API endpoint (for development
-purposes). Use
+purposes). 
+
+To enable Orbit use:
 
 ```shell
 --set cloudserver-front.orbit.enabled=true
 ```
 
-and
+If a custom Orbit endpoint is required use:
 
 ```shell
 --set cloudserver-front.orbit.endpoint=https://dev.zenko.io
--- a/charts/cloudserver-front/templates/deployment.yaml
+++ b/charts/cloudserver-front/templates/deployment.yaml
@@ -31,6 +31,8 @@ spec:
               value: "{{- printf "%s-%s" .Release.Name "s3-data" | trunc 63 | trimSuffix "-" -}}"
             - name: REDIS_HOST
               value: "{{- printf "%s-%s" .Release.Name "redis" | trunc 63 | trimSuffix "-" -}}"
+            - name: LOG_LEVEL
+              value: {{ .Values.logging.level }}
 {{- if .Values.mongodb.enabled }}
             - name: S3METADATA
               value: "mongodb"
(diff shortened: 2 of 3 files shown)


ssalaues:
Could potentially merge into #54 instead as they are buy updates to the cloudserver helm otherwise will need to increment the helm chart version before merging into ft/k8s-helm.

charts/README.md
```diff
@@ -41,7 +41,7 @@ purposes). Use
 and
 
 ```shell
---set cloudserver-front.orbit.endpoint=https://dev.zenko.io
+--set cloudserver-front.orbit.endpoint=https://api.zenko.io
```
NicolasT: Well, no, now it's not really in line with the doc, which refers to a *custom* endpoint.

charts/cloudserver-front/templates/deployment.yaml
```diff
@@ -28,6 +28,10 @@ spec:
               value: "{{- printf "%s-%s" .Release.Name "s3-metadata" | trunc 63 | trimSuffix "-" -}}"
             - name: REDIS_HOST
               value: {{ template "redis.fullname" . }}
+            - name: ENDPOINT
```
NicolasT: I really don't get why this is required *when using Orbit*. Which value should one set this to? This feels more like a bug in Cloudserver rather than something to work-around here...

charts/README.md
ssalaues: You are totally correct however this wasn't clear at first which is why I changed it (Since we are doing development and using orbit, it was thought that we were to use this endpoint based on the documentation). I'll add clarification in the README.

Also yes this ```orbit.endpoint``` not to be confused with ```endpoint``` that was changed in the other commits.

charts/cloudserver-front/templates/deployment.yaml
ssalaues: Hmm we could not reproduce the error we were getting yesterday. However the original error in the logs were that the default "cloudserver-front" endpoint listed in the config.json wasn't reachable. 

I also confirmed with Rached that the ```ENDPOINT``` variable should NOT be necessary when using orbit. This commit will be deleted.

ssalaues:
Updated the PR and added clarifications

charts/cloudserver-front/values.yaml
```diff
@@ -11,6 +11,9 @@ credentials:
   keyId: deployment-specific-access-key
   secretKey: deployment-specific-secret-key
 
+# Logging options: info/debug/trace
+logLevel: info
```
NicolasT: Sorry for nit-pick'ing, but now I wonder whether we shouldn't use
```yaml
logging:
  # Options: info, debug, trace
  level: info
```
to allow for future extensions, and use similar configuration mechanisms for services with more 'complex' logging features.

On the other hand, there seems to be no common convention in https://github.com/kubernetes/charts/

ssalaues: Makes perfect sense. I will update this.

ssalaues:
@NicolasT I rebased off your branch and updated the PR to merge into #54

LaureVergeron:
Just a suggestion, LGTM otherwise

charts/README.md
```diff
@@ -32,13 +32,15 @@ See `values.yml` for more options.
 
 One option to highlight is the ability to enabled [Orbit] integration (which is
 disabled by default) and set a custom Orbit API endpoint (for development
-purposes). Use
+purposes). 
+
+To enable Orbit use:
```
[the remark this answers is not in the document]
LaureVergeron: Agreed.
2 more things:
- There  is a typo on that line 33 (to enableD vs to enable), but that's outside the scope of this, maybe?
- I think saying somewhere in this document that Orbit is a 'management UI' is good for newcomers who are not yet familiar with the stack, but that's just an idea.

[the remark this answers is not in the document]
NicolasT: Done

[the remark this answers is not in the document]
smaffulli: I agree, let's not assume that readers know what Orbit is. Especially because the link to Orbit is to the signup/signin page which doesn't really explain what Orbit is or what it does. This is a better page to link to: https://www.zenko.io/admin/

PR #63, opened 2018-03-13
Ft/backbeat-mongo

vrancurel:
Backbeat Mongo baseline

### DIFF
--- a/charts/backbeat-consumer/templates/deployment.yaml
+++ b/charts/backbeat-consumer/templates/deployment.yaml
@@ -19,7 +19,8 @@ spec:
         - name: {{ .Chart.Name }}
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
-          args: ["npm", "run", "queue_processor"] 
+          command: ["/usr/src/app/docker-entrypoint.sh"]
+          args: ["npm", "run", "queue_processor", "zenko"] 
           env:
             - name: ZOOKEEPER_AUTO_CREATE_NAMESPACE
               value: "1"
@@ -45,6 +46,8 @@ spec:
               value: "{{- printf "%s-s3-metadata" .Release.Name | trunc 63 | trimSuffix "-" -}}"
             - name: QUEUE_POPULATOR_DMD_PORT
               value: "9993"
+            - name: MONGODB_HOSTS
+              value: "{{ .Release.Name }}-mongodb-replicaset-0.{{ .Release.Name }}-mongodb-replicaset:27017,{{ .Release.Name }}-mongodb-replicaset-1.{{ .Release.Name }}-mongodb-replicaset:27017,{{ .Release.Name }}-mongodb-replicaset-2.{{ .Release.Name }}-mongodb-replicaset:27017"
           # TODO livenessProbe:
           # TODO readinessProbe:
           resources:
--- a/charts/backbeat-consumer/values.yaml
+++ b/charts/backbeat-consumer/values.yaml
@@ -4,7 +4,7 @@
 replicaCount: 1
 image:
   repository: zenko/backbeat
-  tag: pensieve-2
+  tag: 0.1.4
   pullPolicy: IfNotPresent
 resources: {}
   # We usually recommend not to specify default resources and to leave this as a conscious
--- a/charts/backbeat-producer/templates/deployment.yaml
+++ b/charts/backbeat-producer/templates/deployment.yaml
@@ -31,6 +31,8 @@ spec:
               value: "{{- printf "%s-s3-metadata" .Release.Name | trunc 63 | trimSuffix "-" -}}"
             - name: QUEUE_POPULATOR_DMD_PORT
               value: "9993"
+            - name: MONGODB_HOSTS
+              value: "{{ .Release.Name }}-mongodb-replicaset-0.{{ .Release.Name }}-mongodb-replicaset:27017,{{ .Release.Name }}-mongodb-replicaset-1.{{ .Release.Name }}-mongodb-replicaset:27017,{{ .Release.Name }}-mongodb-replicaset-2.{{ .Release.Name }}-mongodb-replicaset:27017"
           # TODO livenessProbe:
           # TODO readinessProbe:
           resources:
--- a/charts/backbeat-producer/values.yaml
+++ b/charts/backbeat-producer/values.yaml
@@ -4,7 +4,7 @@
 replicaCount: 1
 image:
   repository: zenko/backbeat
-  tag: pensieve-2
+  tag: 0.1.4
   pullPolicy: IfNotPresent
 resources: {}
   # We usually recommend not to specify default resources and to leave this as a conscious
--- a/charts/cloudserver-front/templates/deployment.yaml
+++ b/charts/cloudserver-front/templates/deployment.yaml
@@ -28,6 +28,8 @@ spec:
               value: "{{- printf "%s-%s" .Release.Name "s3-metadata" | trunc 63 | trimSuffix "-" -}}"
             - name: REDIS_HOST
               value: {{ template "redis.fullname" . }}
+            - name: ENDPOINT
+              value: {{ .Values.endpoint }}
 {{- if .Values.mongodb.enabled }}
             - name: S3METADATA
               value: "mongodb"
@@ -44,8 +46,6 @@ spec:
 {{- else }}
             - name: REMOTE_MANAGEMENT_DISABLE
               value: "1"
-            - name: ENDPOINT
-              value: {{ .Values.endpoint }}
             - name: SCALITY_ACCESS_KEY_ID
               valueFrom:
                 secretKeyRef:
--- a/charts/cloudserver-front/templates/deployment.yaml
+++ b/charts/cloudserver-front/templates/deployment.yaml
@@ -29,7 +29,7 @@ spec:
             - name: REDIS_HOST
               value: {{ template "redis.fullname" . }}
             - name: ENDPOINT
-              value: {{ .Values.endpoint }}
+              value: "{{ .Release.Name }}-cloudserver-front,{{ .Values.endpoint }}"
 {{- if .Values.mongodb.enabled }}
             - name: S3METADATA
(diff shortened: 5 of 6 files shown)


charts/backbeat-consumer/templates/deployment.yaml
```diff
@@ -45,6 +46,8 @@ spec:
               value: "{{- printf "%s-s3-metadata" .Release.Name | trunc 63 | trimSuffix "-" -}}"
             - name: QUEUE_POPULATOR_DMD_PORT
               value: "9993"
+            - name: MONGODB_HOST
+              value: "{{ .Release.Name }}-mongodb-replicaset-0.{{ .Release.Name }}-mongodb-replicaset:27017,{{ .Release.Name }}-mongodb-replicaset-1.{{ .Release.Name }}-mongodb-replicaset:27017,{{ .Release.Name }}-mongodb-replicaset-2.{{ .Release.Name }}-mongodb-replicaset:27017"
```
NicolasT: Should really get SRV-based lookup working. How should we get single-node deployments to work?!

charts/backbeat-consumer/values.yaml
```diff
@@ -4,8 +4,8 @@
 replicaCount: 1
 image:
   repository: zenko/backbeat
-  tag: pensieve-2
-  pullPolicy: IfNotPresent
+  tag: 0.1.3
+  pullPolicy: Always
```
NicolasT: Why?

charts/backbeat-producer/templates/deployment.yaml
```diff
@@ -31,6 +31,8 @@ spec:
               value: "{{- printf "%s-s3-metadata" .Release.Name | trunc 63 | trimSuffix "-" -}}"
             - name: QUEUE_POPULATOR_DMD_PORT
               value: "9993"
+            - name: MONGODB_HOST
+              value: "{{ .Release.Name }}-mongodb-replicaset-0.{{ .Release.Name }}-mongodb-replicaset:27017,{{ .Release.Name }}-mongodb-replicaset-1.{{ .Release.Name }}-mongodb-replicaset:27017,{{ .Release.Name }}-mongodb-replicaset-2.{{ .Release.Name }}-mongodb-replicaset:27017"
```
NicolasT: Same

charts/backbeat-producer/values.yaml
```diff
@@ -4,8 +4,8 @@
 replicaCount: 1
 image:
   repository: zenko/backbeat
-  tag: pensieve-2
-  pullPolicy: IfNotPresent
+  tag: 0.1.3
+  pullPolicy: Always
```
NicolasT: Same

charts/cloudserver-front/templates/deployment.yaml
```diff
@@ -28,6 +28,8 @@ spec:
               value: "{{- printf "%s-%s" .Release.Name "s3-metadata" | trunc 63 | trimSuffix "-" -}}"
             - name: REDIS_HOST
               value: {{ template "redis.fullname" . }}
+            - name: ENDPOINT
```
NicolasT: Unclear why this should be set when Orbit is enabled.

charts/cloudserver-front/values.yaml
```diff
@@ -6,7 +6,7 @@ orbit:
   endpoint: https://api.zenko.io
 # When 'orbit.enabled' is 'true', these aren't used, please use
 # https://zenko.io to manage your deployment
-endpoint: zenko.local
+endpoint: zenko-cloudserver-front
```
NicolasT: Why?

charts/zenko/values.yaml
```diff
@@ -22,7 +22,7 @@ cloudserver-front:
     enabled: false
   # When 'orbit.enabled' is 'true', these aren't used, please use
   # https://zenko.io to manage your deployment
-  endpoint: zenko.local
+  endpoint: zenko-cloudserver-front
```
NicolasT: Why?

charts/backbeat-consumer/templates/deployment.yaml
vrancurel: Yes for now we mimic what we have in cloudserver,  and we do most tests in replication mode. Could you tackle the mono/multi instance problem in a following dedicated PR which will address the problem globally including cloudserver ?

charts/backbeat-consumer/values.yaml
vrancurel: I think it was a dev setting. @JianqinWang can you plz update the pullPolicy ?

charts/cloudserver-front/values.yaml
vrancurel: Apparently something in the code is rejecting URLs with '.' (dots) in hostnames. @JianqinWang is it a Scality function or something else we can change because '.' (dots) in host name in pretty frequent

JianqinWang: The default settings weren't working, so we were changing these values to get a deployment running

NicolasT: So, what's not clear to me: is `endpoint` a *hostname*, or is it a *hostname + dnsdomainname*? I believe it's likely to be the latter, so values with dots (like the default) *must* be accepted.

Strictly speaking, a *hostname* can't have a dot in it.

JianqinWang: `zenko-cloudserver-front` is the hostname of the cloudserver, but we can use hostname + dnsdomainname - we don't think the '.' is actually causing issues, but rather 'zenko.local' wasn't resolving to anything.

NicolasT: Oh, right. Now what I don't get is why this value should be set to the server hostname, or should resolve at all (it's just used to dispatch vhosts, no?).
Also, hard-coding `zenko-cloudserver-front` wouldn't be right because nothing says that's the hostname for all deployments (it wouldn't be for mine, for example).

NicolasT: Basically: can anyone explain what `ENDPOINT` *means* and how it's *used*?

charts/backbeat-consumer/values.yaml
```diff
@@ -4,8 +4,8 @@
 replicaCount: 1
 image:
   repository: zenko/backbeat
-  tag: pensieve-2
-  pullPolicy: IfNotPresent
+  tag: 0.1.4
+  pullPolicy: Always
```
NicolasT: No need to change the `pullPolicy`.

charts/backbeat-producer/values.yaml
```diff
@@ -4,8 +4,8 @@
 replicaCount: 1
 image:
   repository: zenko/backbeat
-  tag: pensieve-2
-  pullPolicy: IfNotPresent
+  tag: 0.1.4
+  pullPolicy: Always
```
NicolasT: No need to change the `pullPolicy`.

charts/cloudserver-front/values.yaml
JianqinWang: It shouldn't be hard coded - this branch was originally just to save progress/changes while testing deployments.
We set `endpoint` here: https://github.com/scality/Zenko/blob/ba6e26902ac23f71ca0c38adb41b74a3b4163e61/charts/cloudserver-front/templates/deployment.yaml#L47
which gets passed to here in S3: https://github.com/scality/S3/blob/5731a65ab400017f3be45639050d535e1d5ca177/docker-entrypoint.sh#L9
The code in the docker-entrypoint script changes the S3 `config.json`. In this case, it adds the hostname `zenko-cloudserver-front` to the restEndpoints.
I will remove this change so it's not hardcoded to `zenko-cloudserver-front`.

NicolasT: Ah, thanks for the pointers/clarification/intel!

Looking at this, it seems that indeed, the `endpoint` is used to map a vhost (as used by the client, and set in the HTTP `Host` header) to an internal 'region'. As such, the value depends on how the service is exposed to the outside world (with DNS etc), not to the host where the service is running (though, granted, if you test from 'within' a K8s cluster, where DNS for the service object is provided, it may make sense to set it to the service object name).

I would however prefer to keep the `zenko.local` default, since we'll likely expose the service using an Ingress object, with a given host selector, and the ingress object configuration and the `endpoint` value must be kept in sync, as explained in the `README.md`.

Finally, given this, I'm even more confused as to why `endpoint` should be set when running an Orbit-managed system: in that case, Orbit should push the endpoint configuration into CloudServer, and no initial value should be required. If it is, that seems like a bug (or maybe design mistake).

charts/cloudserver-front/templates/deployment.yaml
ssalaues: I updated the endpoint to always have "Release.Name"-cloudserver-front. There is a PR to update the docker entry point script to accept csv style strings for this var.

ssalaues: https://github.com/scality/S3/pull/1170

charts/backbeat-consumer/values.yaml
```diff
@@ -4,7 +4,7 @@
 replicaCount: 1
 image:
   repository: zenko/backbeat
-  tag: pensieve-2
+  tag: 0.1.4
```
ssalaues: This will need to change because we will need a new S3 docker image from this S3 PR https://github.com/scality/S3/pull/1170

vrancurel: I rebuild the image

vrancurel: actually it is cloudserver, not backbeat

ssalaues:
The cloudserver image will need to be updated for this helm to function properly https://github.com/scality/S3/pull/1170

vrancurel:
tested with 
```
helm install --name zenko --set ingress.enabled=true --set ingress.hosts[0]=zenko.local --set cloudserver-front.orbit.enabled=true -f value-dev.yml zenko
```

and it works (no need to specify '--set cloudserver-front.endpoint=zenko-cloudserver-front' anymore)

charts/cloudserver-front/templates/deployment.yaml
NicolasT: Cool, thx! :+1:

PR #64, opened 2018-03-22
Minikube Values

ssalaues:
Minikube values.yml file for light or single machine deployments. It sets up 1 zookeeper and 3 mongos (minimum for a replicaset).

### DIFF
--- a/charts/minikube.md
+++ b/charts/minikube.md
@@ -1,0 +1,98 @@
+### Requirements
+- [kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl/)
+- [minikube](https://github.com/kubernetes/minikube/#installation)
+  - [virtualbox](https://www.virtualbox.org/wiki/Downloads)
+- [helm](https://github.com/kubernetes/helm#install)
+- [git](https://git-scm.com/book/en/v2/Getting-Started-Installing-Git)
+
+The following assumes you have minikube (which requires virtualbox or other virtualization options),
+kubectl, and helm installed (see links above). Once minikube, kubectl, and helm are installed, 
+start minikube with at least version 1.9 of kubernetes and perferably with 4GB of RAM (although the
+default value of 2GB should work, we recommend 4GB or more) and enable the minikube ingress addon for communication.
+```shell
+minikube start --kubernetes-version=v1.9.0 --memory 4096
+minkube addon enable ingress
+```
+###### For installations requiring Role Based Access Control see [RBAC](#installation-using-rbac)
+
+Once minikube has started, run the helm initialization. 
+```
+helm init --wait
+```
+#### Older versions of helm may not work with the --wait flag
+
+Clone the repo and go into the charts directory
+```shell
+git clone https://github.com/scality/Zenko.git
+cd ./Zenko/charts
+```
+
+Once you have the repo cloned you can retrieve all dependencies:
+
+```shell
+$ helm repo add zenko-zookeeper https://scality.github.io/zenko-zookeeper/charts
+"zenko-zookeeper" has been added to your repositories
+
+$ helm dependency build zenko/
+Hang tight while we grab the latest from your chart repositories...
+...Successfully got an update from the "zenko-zookeeper" chart repository
+...Successfully got an update from the "stable" chart repository
+Update Complete. ⎈Happy Helming!⎈
+Saving 1 charts
+Downloading zenko-zookeeper from repo https://scality.github.io/zenko-zookeeper/charts
+Deleting outdated charts
+```
+
+With your dependencies built, you can run the following shell command to deploy a single node zenko stack with orbit enabled.
+```shell
+helm install --name zenko --set prometheus.rbac.create=false -f single-node-values.yml zenko
+```
+#### NOTE that Orbit is enabled by default with these values
+
+To view the Kubernetes dashboard type the following and will launch the dashboard in your default browser:
+```shell
+minikube dashboard
+```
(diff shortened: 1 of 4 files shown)


charts/minikube.md
```diff
@@ -0,0 +1,55 @@
...
+Once minikube has started, run the helm initialization. 
+#### NOTE that once you helm init, it may take a couple minutes for the Tiller services to fully deploy on your minikube
+```
+helm init
+```
+
+Clone the repo and go into the charts directory
+```shell
+git clone https://github.com/scality/Zenko.git
+cd Zenko/charts
```
NicolasT: This likely won't work right now as-is...

charts/minikube.md
```diff
@@ -0,0 +1,55 @@
...
+...Successfully got an update from the "stable" chart repository
+Update Complete. ⎈Happy Helming!⎈
+Saving 1 charts
+Downloading zenko-zookeeper from repo https://scality.github.io/zenko-zookeeper/charts
+Deleting outdated charts
+```
+
+With your dependencies built, you can run the following shell command to deploy the zenko stack with orbit enabled.
+```shell
+helm install --name zenko --set ingress.enabled=true  --set cloudserver-front.orbit.enabled=true -f single-node-values.yml zenko
```
NicolasT: Why pass `--set` values when you provide a tailored YAML configuration anyway?

charts/single-node-values.yml
```diff
@@ -0,0 +1,63 @@
+# Default values for zenko.
```
NicolasT: I wouldn't copy the existing `values.yaml` as you did (or at least partially). This file doesn't contain 'default values'. It's overlayed with the default values from the Chart's `values.yaml`.

charts/single-node-values.yml
```diff
@@ -0,0 +1,63 @@
...
+    keyId: deployment-specific-access-key
+    secretKey: deployment-specific-secret-key
+
+zenko-zookeeper:
+  servers: 1
+  resources:
+    requests:
+      cpu: 500m
+      memory: 256Mi
+  heap: "512M"
```
NicolasT: Then `resources.requests.memory` should be the same, no?

charts/single-node-values.yml
```diff
@@ -0,0 +1,63 @@
...
+    create: false
+
+  alertmanager:
+    enabled: false
+
+  kubeStateMetrics:
+    enabled: false
+
+  nodeExporter:
+    enabled: false
```
NicolasT: All of this should be the case already.

charts/minikube.md
ssalaues: You're right I was thinking too far ahead. I'll add a checkout to ft/k8s-helm for when this gets merged.

charts/minikube.md
ssalaues: I hadn't considered that. I'll set these as the defaults in the single node values.

ssalaues:
Updated PR with simplified values (only necessary values included). Updated the documentation to reflect changes.

charts/single-node-values.yml
```diff
@@ -0,0 +1,24 @@
...
+  servers: 1
+  resources:
+    requests:
+      cpu: 500m
+      memory: 256Mi
+  heap: "256M"
+
+prometheus:
+  rbac:
+    create: false
```
NicolasT: Purely because of this, I don't really like this file to be called `single-node-values.yaml`, because indeed it would fail to deploy on a single-node MetalK8s cluster, or most other single-node deployments of a recent K8s version, including GKE.

Could, instead

- *or* Minikube be deployed with RBAC enabled (`--extra-config=apiserver.Authorization.Mode=RBAC`), and things still work,
- *or* this file be renamed to something like `minikube-values.yaml`
- *or* better suggestions :smile:

charts/single-node-values.yml
```diff
@@ -0,0 +1,24 @@
+ingress:
+  enabled: true
+  hosts:
+    - ""
+cloudserver-front:
+  orbit:
+    enabled: true
```
NicolasT: Shouldn't this become the default in the `cloudserver-front` Chart, because it's basically broken without this right now? Maybe the option should even be removed until things are fixed again. WDYT @vrancurel

charts/single-node-values.yml
ssalaues: Yeah I'll play with the RBAC on minikube this way we don't potentially need another values.yml

ssalaues: So I tested with RBAC and seems to work just fine but it add more steps plus makes the current steps seem a bit more cumbersome.

So this would be the modified steps plus the new ones.
```
minikube start --kubernetes-version=v1.9.0 --memory 4096 --extra config=apiserver.Authorization.Mode=RBAC

kubectl create clusterrolebinding add-on-cluster-admin --clusterrole=cluster-admin --serviceaccount=kube-system:tiller

helm init --service-account tiller --wait
```
Which I'm okay with doing but might slightly take away from the "simplicity" perceptions. 

Another option would be to append the instructions in a new section "Installing with RBAC enabled". This way we can take out the prometheus.rbac.false from the yaml file (making it more generically applicable) and simply add it as a --set in the helm install options.

charts/single-node-values.yml
vrancurel: until we have a clear view it is not bad to be explicit

ssalaues:
-Updated documentation for optional RBAC setup
-Took out prometheus.rbac.create from single node yml
-Updated the zenko chart notes to show the endpoint in case of the single node

ssalaues:
All comments addressed

PR #77, opened 2018-04-16
Add initial E2E testing infrastructure

NicolasT:


tests/.dockerignore
```diff
@@ -0,0 +1,18 @@
+clueso-tests/
```
ssalaues: Not exactly pertinent to this PR but at some point "clueso-tests" should probably be updated to something more generic like "metadata-search-tests"

NicolasT: Agree, though indeed not something for this PR :)

PR #79, opened 2018-04-17
bf: fix deployment defaults

rahulreddy:


charts/s3-data/values.yaml
```diff
@@ -4,13 +4,13 @@
...
-  tag: pensieve-0
+  tag: 0.1.9
   pullPolicy: IfNotPresent
 service:
   name: s3-data
   type: ClusterIP
-  internalPort: 9992
-  externalPort: 9992
+  internalPort: 9991
+  externalPort: 9991
```
NicolasT: If this changes, I would kind-of expect a related change somewhere else? Or is 9991 hard-coded in some other service? If that's the case, why have it configurable here?

`internalPort` shouldn't have been configurable from the start, actually. That's something we should fix.

PR #80, opened 2018-04-17
BF: Fix: cloud server port in backbeat

anurag4DSB:
changed the port number from 8001 to 8000 in deployments.yaml for backbeat-consumer


anurag4DSB:
Closing pull request as this is being handled in https://github.com/scality/Zenko/pull/79

PR #86, opened 2018-04-30
ft: helm chart for Backbeat API stateless container

giacomoguiulfo:


charts/backbeat-api/templates/deployment.yaml
```diff
@@ -0,0 +1,48 @@
...
+        release: {{ .Release.Name }}
+    spec:
+      containers:
+        - name: {{ .Chart.Name }}
+          image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
+          ports:
+            - name: backbeat-api
+              containerPort: 8900
+          env:
```
ssalaues: Can you add the ```REMOTE_MANAGEMENT_DISABLE```orbit variable (just like S3) as it was added to Backbeat as well.

ssalaues: https://github.com/scality/backbeat/blob/bf08d9600a7d62ab725ce7e13e789cfbdf2166ea/lib/management.js#L336

charts/backbeat-api/templates/deployment.yaml
```diff
@@ -0,0 +1,48 @@
...
+            - name: ZOOKEEPER_AUTO_CREATE_NAMESPACE
+              value: "1"
+            - name: ZOOKEEPER_CONNECTION_STRING
+              value: "{{- printf "%s-zenko-zookeeper:2181" .Release.Name | trunc 63 | trimSuffix "-" -}}"
```
jonathan-gramain: I would rather let this error out at kub level than truncate the value, since it has zero chance to work if truncated (port or port+host will be incorrect so zookeeper will be unreachable) and the error may be harder to diagnose in such case (to be confirmed as I don't know what will actually happen).

charts/backbeat-api/templates/_helpers.tpl
```diff
@@ -0,0 +1,32 @@
...
+{{/*
+Expand the name of the chart.
+*/}}
+{{- define "backbeat-api.name" -}}
+{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" -}}
+{{- end -}}
+
+{{/*
+Create a default fully qualified app name.
+We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
```
jonathan-gramain: As commented below, I would think it's better to fail early if names are > 64 chars to avoid any further hard to diagnose issues, but it's good if we can have this check done around here anyway (but failing instead of truncating).

charts/backbeat-api/templates/deployment.yaml
```diff
@@ -0,0 +1,48 @@
...
+              value: "6379"
+            - name: HEALTHCHECKS_ALLOWFROM
+              value: '0.0.0.0/0'
+          args: ["npm", "start"]
+          livenessProbe:
+            tcpSocket:
+              port: backbeat-api
+            initialDelaySeconds: 5
+          resources:
+{{ toYaml .Values.resources | indent 12 }}
```
jonathan-gramain: minor: if such thing is at all possible, maybe a template `foreach` style loop over the resources could be more readable/maintainable than specifying an indentation level and passing the template with no indentation.

charts/backbeat-api/templates/deployment.yaml
```diff
@@ -0,0 +1,48 @@
...
+    {{- with .Values.nodeSelector }}
+      nodeSelector:
+{{ toYaml . | indent 8 }}
```
jonathan-gramain: same here

charts/backbeat-api/templates/deployment.yaml
```diff
@@ -0,0 +1,48 @@
...
+            - name: ZOOKEEPER_CONNECTION_STRING
+              value: "{{- printf "%s-zenko-zookeeper:2181" .Release.Name | trunc 63 | trimSuffix "-" -}}"
+            - name: KAFKA_HOSTS
+              value: "{{- printf "%s-kafka-0.%s-kafka:9092" .Release.Name .Release.Name | trunc 63 | trimSuffix "-" -}}"
+            - name: REDIS_HOST
+              value: "{{- printf "%s-%s" .Release.Name "redis" | trunc 63 | trimSuffix "-" -}}"
+            - name: REDIS_PORT
+              value: "6379"
+            - name: HEALTHCHECKS_ALLOWFROM
+              value: '0.0.0.0/0'
```
jonathan-gramain: Probably this should be limited eventually to relevant monitoring hosts only, but ok for now.

charts/backbeat-api/templates/deployment.yaml
giacomoguiulfo: Possibly, however that part was generated by helm. It's probably a convention in the community.

charts/backbeat-api/Chart.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
+description: A Helm chart for Kubernetes
```
NicolasT: Needs to be filled in

charts/backbeat-api/Chart.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
```
NicolasT: There's more useful metadata that can be added (refs to project pages, authors,...)

charts/backbeat-api/templates/_helpers.tpl
NicolasT: No, this is how it's done in every chart (not just ours) out there.

charts/backbeat-api/templates/deployment.yaml
```diff
@@ -0,0 +1,50 @@
...
+  labels:
+    app: {{ template "backbeat-api.name" . }}
+    chart: {{ template "backbeat-api.chart" . }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+spec:
+  replicas: {{ .Values.replicaCount }}
+  template:
+    metadata:
+      labels:
```
NicolasT: Could you try deploying this using `apps/v1` `apiVersion` and check whether things still work?

charts/backbeat-api/templates/deployment.yaml
NicolasT: If we ever want to limit inter-service access, we'd better specify [NetworkPolicies](https://kubernetes.io/docs/concepts/services-networking/network-policies/) rather than provide partial implementations of some kind of firewall in every single service.

charts/backbeat-api/templates/deployment.yaml
```diff
@@ -0,0 +1,50 @@
...
+              value: "{{- printf "%s-kafka-0.%s-kafka:9092" .Release.Name .Release.Name | trunc 63 | trimSuffix "-" -}}"
+            - name: REDIS_HOST
+              value: "{{- printf "%s-%s" .Release.Name "redis" | trunc 63 | trimSuffix "-" -}}"
+            - name: REDIS_PORT
+              value: "6379"
+            - name: HEALTHCHECKS_ALLOWFROM
+              value: '0.0.0.0/0'
+          args: ["npm", "start"]
+          livenessProbe:
+            tcpSocket:
```
NicolasT: Isn't there some HTTP endpoint to query?

charts/backbeat-api/templates/deployment.yaml
NicolasT: As Giacomo said, this is how it's done in every chart (not just ours) out there, so no.

charts/backbeat-api/templates/deployment.yaml
NicolasT: As Giacomo said, this is how it's done in every chart (not just ours) out there, so no.

charts/backbeat-api/templates/service.yaml
```diff
@@ -0,0 +1,19 @@
...
+  labels:
+    app: {{ template "backbeat-api.name" . }}
+    chart: {{ template "backbeat-api.chart" . }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+spec:
+  type: {{ .Values.service.type }}
+  ports:
+    - port: {{ .Values.port }}
+      targetPort: 8900
```
NicolasT: Use named ports.

charts/backbeat-api/templates/service.yaml
```diff
@@ -0,0 +1,19 @@
...
+  name: {{ template "backbeat-api.fullname" . }}
+  labels:
+    app: {{ template "backbeat-api.name" . }}
+    chart: {{ template "backbeat-api.chart" . }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+spec:
+  type: {{ .Values.service.type }}
+  ports:
+    - port: {{ .Values.port }}
```
NicolasT: I still fail to see why we'd want to make the port of our `Service`s 'configurable', unless we retrieve them from the environment or DNS in other software connecting to them.

charts/backbeat-api/values.yaml
```diff
@@ -0,0 +1,23 @@
+# Default values for backbeat-api.
```
NicolasT: Where's `orbit.enabled` etc?

charts/single-node-values.yml
```diff
@@ -2,6 +2,19 @@ ingress:
   enabled: true
   hosts:
     - ""
+
+backbeat-api:
```
NicolasT: These should not be required here, but just use the default from the Chart.

charts/single-node-values.yml
```diff
@@ -2,6 +2,19 @@ ingress:
...
+  orbit:
+    enabled: true
+
+backbeat-consumer:
```
NicolasT: These should not be required here, but just use the default from the Chart.

charts/single-node-values.yml
```diff
@@ -2,6 +2,19 @@ ingress:
...
+  orbit:
+    enabled: true
+
+backbeat-producer:
```
NicolasT: These should not be required here, but just use the default from the Chart.

charts/zenko/requirements.yaml
```diff
@@ -16,6 +16,9 @@ dependencies:
   repository: "https://kubernetes-charts.storage.googleapis.com/"
 
 # Charts managed in this repository
+- name: backbeat-api
+  version: "0.1.0"
+  repository: "file://../backbeat-api"
 - name: backbeat-consumer
```
NicolasT: Not sure how this works? As in, `backbeat-consumer` and `backbeat-producer` chart versions are bumped as well...

charts/zenko/values.yaml
```diff
@@ -24,6 +24,17 @@ e2e:
     tag: latest
     pullPolicy: Always
 
+backbeat-api:
```
NicolasT: These should not be required here, but just use the default from the Chart.

charts/zenko/values.yaml
```diff
@@ -24,6 +24,17 @@ e2e:
...
+  orbit:
+    enabled: false
+
+backbeat-consumer:
```
NicolasT: These should not be required here, but just use the default from the Chart.

charts/zenko/values.yaml
```diff
@@ -24,6 +24,17 @@ e2e:
...
+  orbit:
+    enabled: false
+
+backbeat-producer:
```
NicolasT: These should not be required here, but just use the default from the Chart.

NicolasT:
So, I'm wondering now: the intent is to merge `backbeat-producer` and `backbeat-consumer` in a single Chart. So, why doesn't this one go in there as well? Is this chart of any use *by itself*?

@shivanipradhan @giacomoguiulfo @vrancurel @rahulreddy

NicolasT:
> The backbeat-api chart will be used by different other backbeat-based services.

That's not the point. The question is: is there any using in *running* a *backbeat-api* service when there's no *backbeat-\** behind?

To me it seems *backbeat-api* is part of the `backbeat` suite, so part of the `backbeat` Chart.

charts/backbeat/values.yaml
```diff
@@ -1,12 +1,25 @@
 # Default values for backbeat.
 # This is a YAML-formatted file.
 # Declare variables to be passed into your templates.
+orbit:
+  enabled: true
```
ssalaues: Can you expose this to the zenko/values.yaml as well? You will have to make a backbeat section of the zenko ```values.yaml``` file

NicolasT: There's no need to 'expose' this in `zenko/values.yaml`. All dependencies' `values` are available as-is.

charts/backbeat/templates/api/service.yaml
```diff
@@ -0,0 +1,19 @@
...
+  labels:
+    app: {{ template "backbeat.name" . }}
+    chart: {{ template "backbeat.chart" . }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+spec:
+  type: {{ .Values.api.service.type }}
+  ports:
+    - port: 8900
+      targetPort: 8900
```
NicolasT: Use named ports instead.

charts/backbeat/templates/api/service.yaml
```diff
@@ -0,0 +1,19 @@
...
+      protocol: TCP
+      name: {{ .Values.api.service.name }}
```
NicolasT: Why call the port after the service? The service is (e.g.) `backbeat-api`. The port is `http`, or `api`, or whatever.
Furthermore, there's no need to make the port name configurable. On the contrary, this makes it easier to configure things in a 'broken' way.

charts/backbeat/templates/api/deployment.yaml
```diff
@@ -0,0 +1,59 @@
...
+    app: {{ template "backbeat.name" . }}-api
+    chart: {{ template "backbeat.chart" . }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+spec:
+  replicas: {{ .Values.api.replicaCount }}
+  template:
+    metadata:
+      labels:
+        app: {{ template "backbeat.name" . }}
```
NicolasT: This needs to be fixed.

charts/backbeat/templates/api/deployment.yaml
```diff
@@ -0,0 +1,59 @@
...
+        release: {{ .Release.Name }}
+    spec:
+      containers:
+        - name: {{ .Chart.Name }}
```
NicolasT: Use constant names

charts/backbeat/values.yaml
```diff
@@ -1,12 +1,25 @@
...
 
 image:
   repository: zenko/backbeat
-  tag: 0.1.4
+  tag: 0.1.5
   pullPolicy: IfNotPresent
 
+api:
```
NicolasT: Put this after the `consumer` definition, or move the comments on `resources` here, such that they appear at the first occurrence of `resources`.

charts/cloudserver-front/templates/deployment.yaml
```diff
@@ -31,6 +31,10 @@ spec:
               value: "{{- printf "%s-%s" .Release.Name "s3-data" | trunc 63 | trimSuffix "-" -}}"
             - name: REDIS_HOST
               value: "{{- printf "%s-%s" .Release.Name "redis" | trunc 63 | trimSuffix "-" -}}"
+            - name: CRR_METRICS_HOST
```
NicolasT: Actually, we may want to change our code to simply use the environment variables Kubernetes already provides to look up service hosts/addresses and ports, instead of needing to always re-export it. Anyway, not a blocker for this PR.

charts/backbeat/values.yaml
ssalaues: While yes they are all available, an advantage of having certain values in zenko/values.yaml is so that users can easily identify the most relevant settings of how Zenko is being deployed. Which is why I was asking for it to be 'exposed' so that it can easily be identified that not only cloudserver-front has Orbit management but backbeat as well for anyone wanting to make their own changes. 

Not a make or break but just a suggestion.

giacomoguiulfo:
Made the requested/suggested changes @NicolasT @ssalaues.

charts/backbeat/values.yaml
rahulreddy: I agree, the actual configuration of enabling/disabling Orbit should be a global not per component.

charts/backbeat/templates/api/service.yaml
```diff
@@ -0,0 +1,19 @@
...
+  labels:
+    app: {{ template "backbeat.name" . }}-api
+    chart: {{ template "backbeat.chart" . }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+spec:
+  type: {{ .Values.api.service.type }}
+  ports:
+    - name: api
+      port: 8900
```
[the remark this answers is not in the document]
vrancurel: Yes it should be exposed in Values.yaml

giacomoguiulfo:
Right now backbeat-api will crash showing this error https://github.com/scality/Arsenal/pull/490. However, the fix has already been merged. We just need to update docker image. To temporarily fix it, you just need to do a PUT to have some metrics and prevent it from crashing on empty metrics.

charts/backbeat/templates/consumer/deployment.yaml
```diff
@@ -22,6 +22,8 @@ spec:
           command: ["/usr/src/app/docker-entrypoint.sh"]
           args: ["npm", "run", "queue_processor", "zenko"]
```
philipyoo: I know outside the scope of this PR, but why is "zenko" provided as an argument?

ssalaues: I remember it was needed when @JianqinWang and I were updating the charts to work with mongodb but not sure if it's still needed.

PR #87, opened 2018-05-02
charts: Unify Backbeat chart

NicolasT:
No need to have separate `backbeat-consumer` and `backbeat-producer`
charts: the Backbeat project is a self-contained application, providing
multiple services. Instead of splitting it in two charts, use one which
generates two distinct `Deployment`s, one for the `consumer`, one for
the `producer`.

See: https://github.com/scality/Zenko/pull/86#issuecomment-386026890
See: ZENKO-278


charts/backbeat/templates/consumer/deployment.yaml
```diff
@@ -1,26 +1,26 @@
...
+        app: {{ template "backbeat.name" . }}
         release: {{ .Release.Name }}
     spec:
       containers:
         - name: {{ .Chart.Name }}
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
           command: ["/usr/src/app/docker-entrypoint.sh"]
-          args: ["npm", "run", "queue_processor", "zenko"] 
+          args: ["npm", "run", "queue_processor", "zenko"]
```
rahulreddy: Is the expectation that we run multiple containers within the same chart (looking for direction) - one for `queue_populator` which produces entries into Kafka and `queue_processor` which consumes entries from Kafka and writes to MongoDB.

charts/backbeat/values.yaml
```diff
@@ -0,0 +1,35 @@
+# Default values for backbeat.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+image:
+  repository: zenko/backbeat
+  tag: 0.1.4
```
giacomoguiulfo: In case it's decided to put backbeat-api as a part of this unified chart, then well have to use the image with ```tag: 0.1.5```.

charts/backbeat/templates/consumer/deployment.yaml
NicolasT: Instead of having 2 charts (`backbeat-producer`, `backbeat-consumer`) each creating one `Deployment` which in turn runs one `Pod` with one container, there's now one chart (`backbeat`) which on install creates two `Deployment`s each running one `Pod` with one container.

Every `Deployment` is self-standing, e.g. changing the number of `consumer` replicas doesn't (and shouldn't) impact the number of `producer` replicas.

Rationale: there's no use in deploying `backbeat-consumer` without `backbeat-producer` and vice-versa, so combining them and deploying them as a single unit makes sense.

Keep in mind Charts have no direct relation to what *runs* in a cluster! E.g. the Prometheus chart will create a whole load of `Deployment`s, `DaemonSet`s and whatnot.

charts/backbeat/values.yaml
NicolasT: Of course, if this lands, the `api` part would be part of this chart, as another `Deployment` (I guess). That may indeed imply the version of the image needs to be bumped, so be it :)

charts/backbeat/templates/producer/deployment.yaml
```diff
@@ -36,8 +36,16 @@ spec:
           # TODO livenessProbe:
           # TODO readinessProbe:
           resources:
-{{ toYaml .Values.resources | indent 12 }}
-    {{- if .Values.nodeSelector }}
+{{ toYaml .Values.producer.resources | indent 12 }}
+    {{- with .Values.consumer.nodeSelector }}
```
giacomoguiulfo: It should be ```{{- with .Values.producer.nodeSelector }}``` instead of ```{{- with .Values.consumer.nodeSelector }}```.

NicolasT: Good catch! Should've `grep`ed :) Will fix.

charts/backbeat/templates/consumer/deployment.yaml
rahulreddy: There has been some discussions over the last two days (I am hoping that you were in the loop as well). We are working on building charts by use cases (lifecycle, replication, ingestion etc, backbeat api etc.). Keeping that in mind, I think this chart should start the API server which serves healthchecks and CRR metrics routes (for now, we will expand this provide more functionality).

NicolasT: @rahulreddy I am, partially at least, but at the same time think this is not at all the right direction.

Nonetheless, it doesn't change the validity of this PR given state of things *right now*, so I'd like to move forward and merge.

rahulreddy: I agree, let's merge this.

PR #91, opened 2018-05-03
ft: Unify charts with Orbit

ssalaues:
This allows Orbit to be enabled by default which is our normal for swarm and cloudserver.

### DIFF
--- a/charts/README.md
+++ b/charts/README.md
@@ -35,10 +35,10 @@ One option to highlight is the ability to enable integration with the
 [Orbit] management UI (which is disabled by default) and set a custom
 Orbit API endpoint (for development purposes). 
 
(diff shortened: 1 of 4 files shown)


vrancurel:
just asking: why do we need to enable it by default ?

ssalaues:
@vrancurel It's because our default in s3 cloudserver is to have it enabled along with having enabled by default in the docker stack as well. Was speaking with @rahulreddy and was saying this should be our default configuration for Zenko.

NicolasT:
@ssalaues May want to amend the typo in the commit message.

ssalaues:
Oops! Good catch. Fixed the typo in commit.

PR #98, opened 2018-05-15
Run helm test on EVE CI

Zempashi:


eve/main.yml
```diff
@@ -0,0 +1,69 @@
...
+    - ShellCommand: &install_tiller
+        name: install helm (tiller into kubernetes)
+        command: >-
+          helm init --wait
+        env:
+          TILLER_NAMESPACE: '%(prop:testNamespace)s'
+    - ShellCommand:
+        name: Install needed repo
+        command: >
+          helm repo add zenko-zookeeper https://scality.github.io/zenko-zookeeper/charts
```
giacomoguiulfo: It should now be: ```helm repo add incubator http://storage.googleapis.com/kubernetes-charts-incubator```.

eve/main.yml
```diff
@@ -0,0 +1,69 @@
...
+        command: >
+          helm repo add incubator http://storage.googleapis.com/kubernetes-charts-incubator
+        env:
+          TILLER_NAMESPACE: '%(prop:testNamespace)s'
+    - ShellCommand:
+        name: Retrieve dependency
+        command: >-
+           helm dep build charts/zenko
+        env:
+          TILLER_NAMESPACE: '%(prop:testNamespace)s'
```
ssalaues: Might be a good idea to have "haltOnFailure: true" on all the steps since any of these steps will fail if the previous step fail.

Currently looks like it's failing on this step but I believe it's because we have been making many changes to the repo and the branch need to be rebased.

eve/main.yml
```diff
@@ -0,0 +1,69 @@
...
+    - ShellCommand:
+        name: Install Zenko !
+        command: >-
+          helm upgrade zenko-test --namespace %(prop:testNamespace)s
+          --install charts/zenko --wait
+          --set prometheus.rbac.create=false
```
ssalaues: The following will need rbac disabled as well
`
  --set zenko-queue.rbac.enabled=false
  --set redis-ha.rbac.create=false 
`

NicolasT: All disabling of RBAC is a bit troublesome to me. I know it's likely required right now, but not something we can maintain forever: in the end, 'real' deployments will have RBAC enabled, and if all we test is with it disabled, bugs can remain uncaught all too easily.

ssalaues: Oh yeah that is a really valid point. We should bring it up in the next Releng meeting.

PR #99, opened 2018-05-16
ft: ZENKO-310 Update Kafka and Zookeeper Dependencies

ssalaues:
This updates the Zenko requirements to use the upstream Kafka and
Zookeeper along with changing the deployment names to use the
backbeat-queue and backbeat-quorum alias, respectively.


ssalaues:
There are still changes that need to be made like the hardcoded 2181 port in the helper function. I'm having issues with templating because of the '-' in the backbeat-queue or backbeat-quorum names. Using the index function of Go templating doesn't seem to be working so any advice would be appreciated =)

charts/zenko/templates/_helpers.tpl
```diff
@@ -14,3 +14,12 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this
 {{- $name := default .Chart.Name .Values.nameOverride -}}
 {{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" -}}
 {{- end -}}
+
+{{/*
+Create the default url for the backbeat quorum service
+*/}}
+{{- define "zookeeper.url" -}}
+{{- $zookeeperConnect := printf "%s-%s:2181" .Release.Name .Values.zookeeper.url }}
```
ssalaues: Trying to get the port from .Values.backbeat-queue.zookeeper.port has issues because of the '-' in the name. Trying to use the index function as suggested [here](https://github.com/kubernetes/helm/issues/2192), however I get a 'error calling index: index of nil pointer' error with code ```{{- $port := index .Values "backbeat-queue" "zookeeper" "port" -}}```. It seems that for some reason it's not recognizing the values.

Any suggestions?

charts/backbeat/templates/consumer/deployment.yaml
```diff
@@ -25,9 +25,9 @@ spec:
             - name: ZOOKEEPER_AUTO_CREATE_NAMESPACE
               value: "1"
             - name: ZOOKEEPER_CONNECTION_STRING
-              value: "{{- printf "%s-zenko-zookeeper:2181" .Release.Name | trunc 63 | trimSuffix "-" -}}"
+              value: "{{- printf "%s-backbeat-quorum:2181" .Release.Name | trunc 63 | trimSuffix "-" -}}"
```
rahulreddy: Let's use `zenko-quorum`

charts/backbeat/templates/consumer/deployment.yaml
```diff
@@ -25,9 +25,9 @@ spec:
...
             - name: KAFKA_HOSTS
-              value: "{{- printf "%s-kafka-0.%s-kafka:9092" .Release.Name .Release.Name | trunc 63 | trimSuffix "-" -}}"
+              value: "{{- printf "%s-kafka-0.%s-backbeat-queue:9092" .Release.Name .Release.Name | trunc 63 | trimSuffix "-" -}}"
```
rahulreddy: And `zenko-queue` here.

charts/zenko/requirements.yaml
```diff
@@ -14,14 +9,21 @@ dependencies:
 - name: redis
   version: "1.1.12"
   repository: "https://kubernetes-charts.storage.googleapis.com/"
+- name: kafka
+  version: "0.7.1"
```
rahulreddy: Can we upgrade Kafka to 1.0.x?

NicolasT: This is the Chart version, not the `appVersion`: https://github.com/kubernetes/charts/blob/753273e6d2e06f656e7eb8730fbbda6da13a495f/incubator/kafka/Chart.yaml#L6

charts/zenko/templates/_helpers.tpl
giacomoguiulfo:
```php
{{- $zookeeperConnect := printf "%s-%s:%d" .Release.Name .Values.zookeeper.url (int64 .Values.zookeeper.port) }}
```

charts/zenko/requirements.yaml
ssalaues: Exactly, as Nicolas said, this is the chart version but I believe the Kafka version used in the Confluent 4.0 docker image used in this chart is Kafka 1.0

charts/backbeat/templates/consumer/deployment.yaml
giacomoguiulfo: Shouldn't it be ```backbeat-queue-0``` instead of ```kafka-0```?

ssalaues: Updating now. Thanks!

ssalaues:
Updated based off requested changes. There are remaining issues with deployment and seems like the backbeat-queue (kafka) is still deploying it's own zookeeper along with the one mandated by the Zenko chart resulting int 2 zookeeper deployments

charts/zenko/values.yaml
```diff
@@ -55,10 +55,32 @@ mongodb-replicaset:
...
+    "num.partitions": 1                   # - default number of log partitions per topic
+    "auto.create.topics.enable": true     # - enable auto creation of topic on the server
+    "min.insync.replicas": 2              # - min number of replicas that must acknowledge a write
+  prometheus:
+    jmx: 
+      enabled: true
+    kafka:
+      enabled: true
+zookeeper:
+  enabled: false # Disables zenko-queue from deploying it's own quorum
```
jonathan-gramain: typo -> `its own quorum`

ssalaues:
Fixed the double zookeeper deployment issue.

ssalaues:
Manually tested replication via Orbit with the new queue/quorum (works!). Will update the documentation to reflect the changes.

charts/minikube.md
```diff
@@ -31,22 +31,29 @@ cd ./Zenko/charts
...
+Downloading zookeeper from repo http://storage.googleapis.com/kubernetes-charts-incubator
 Deleting outdated charts
 ```
 
 With your dependencies built, you can run the following shell command to deploy a single node zenko stack with orbit enabled.
 ```shell
-helm install --name zenko --set prometheus.rbac.create=false -f single-node-values.yml zenko
+helm install --name zenko \
+  --set prometheus.rbac.create=false \
+  --set backbeat-queue.rbac.enabled=false \
```
giacomoguiulfo: It should be ```--set zenko-queue.rbac.enabled=false```

ssalaues:
Fixed the typo @giacomoguiulfo Giacomo found

PR #102, opened 2018-05-17
ft: helm chart for Redis Sentinel

giacomoguiulfo:
This PR adds redis sentinel to the Zenko stack. It is used (currently) by backbeat replication and backbeat api.

### DIFF
--- a/charts/single-node-values.yml
+++ b/charts/single-node-values.yml
@@ -19,3 +19,7 @@ mongodb-replicaset:
   replicaSet: rs0
   replicas: 3
 
+redis-ha:
+  replicas:
+    servers: 1
+    sentinels: 1
--- a/charts/zenko/requirements.yaml
+++ b/charts/zenko/requirements.yaml
@@ -14,10 +14,12 @@ dependencies:
   repository: "http://storage.googleapis.com/kubernetes-charts-incubator"
   alias: zenko-queue
 - name: zookeeper
-  version: "1.0.0" 
+  version: "1.0.0"
   repository: "http://storage.googleapis.com/kubernetes-charts-incubator"
   alias: zenko-quorum
-
+- name: redis-ha
+  version: "2.1.2"
+  repository: "https://kubernetes-charts.storage.googleapis.com/"
 
 # Charts managed in this repository
 - name: backbeat
--- a/charts/zenko/values.yaml
+++ b/charts/zenko/values.yaml
@@ -67,7 +67,7 @@ zenko-queue:
     "auto.create.topics.enable": true     # - enable auto creation of topic on the server
     "min.insync.replicas": 2              # - min number of replicas that must acknowledge a write
   prometheus:
-    jmx: 
+    jmx:
       enabled: true
     kafka:
       enabled: true
@@ -88,3 +88,10 @@ redis:
     enabled: false
   metrics:
     enabled: true
+
+redis-ha:
+  rbac:
+    create: true
+  replicas:
+    servers: 3
+    sentinels: 3
--- a/charts/backbeat/templates/api/deployment.yaml
+++ b/charts/backbeat/templates/api/deployment.yaml
@@ -33,7 +33,7 @@ spec:
             - name: KAFKA_HOSTS
               value: "{{- printf "%s-zenko-queue:9092" .Release.Name | trunc 63 | trimSuffix "-" -}}"
             - name: REDIS_HOST
-              value: "{{- printf "%s-%s" .Release.Name "redis" | trunc 63 | trimSuffix "-" -}}"
+              value: "{{- printf "%s-%s" .Release.Name "redis-ha-master-svc" | trunc 63 | trimSuffix "-" -}}"
             - name: REDIS_PORT
               value: "6379"
             - name: HEALTHCHECKS_ALLOWFROM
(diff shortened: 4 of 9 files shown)


charts/single-node-values.yml
```diff
@@ -19,3 +19,7 @@ mongodb-replicaset:
   replicaSet: rs0
   replicas: 3
 
+redis-ha:
```
ssalaues: If we can get away with simply using the normal redis cache for the single node deployments, it would help to keep requirements down.

charts/zenko/requirements.yaml
```diff
@@ -14,10 +14,12 @@ dependencies:
   repository: "http://storage.googleapis.com/kubernetes-charts-incubator"
   alias: zenko-queue
 - name: zookeeper
-  version: "1.0.0" 
+  version: "1.0.0"
   repository: "http://storage.googleapis.com/kubernetes-charts-incubator"
   alias: zenko-quorum
-
+- name: redis-ha
```
ssalaues: You could add a condition to this chart so it could be disabled in the single node values.

charts/zenko/values.yaml
```diff
@@ -88,3 +88,10 @@ redis:
     enabled: false
   metrics:
     enabled: true
+
+redis-ha:
```
ssalaues: Can add an ```enabled: true``` that can be referenced in the requirements.yaml

charts/zenko/values.yaml
```diff
@@ -88,3 +88,10 @@ redis:
...
+  rbac:
+    create: true
+  replicas:
+    servers: 5
```
ssalaues: We switched to 3 replicas for our default deployments.

giacomoguiulfo: Fixed.

charts/single-node-values.yml
giacomoguiulfo: We will need to use globals to solve that issue properly.

charts/zenko/requirements.yaml
giacomoguiulfo: Same here.

charts/zenko/values.yaml
giacomoguiulfo: Same here.

charts/single-node-values.yml
ssalaues: Yeah this could be setup later on if we end up having resourcing issues.

charts/zenko/values.yaml
```diff
@@ -88,3 +88,10 @@ redis:
     enabled: false
   metrics:
     enabled: true
+
+redis-ha:
+  rbac:
+    create: true
+  replicas:
+    servers: 3
```
philipyoo: Hm, the more I look into this redis-ha chart and looking at this value, it sounds like a redis cluster with multiple redis sentinels as its setup. I know you asked me about this offline, and it sounded ok in practice since we introduce a single endpoint for redis. I'm not too sure the impact of using a cluster, but maybe someone else will have an answer

giacomoguiulfo: This setup has (by default) 3 redis-sentinels and 3 redis-servers. The redis sentinels allow monitoring, serve notifications, start failover processes and are configuration providers for the redis-servers (the acutal keystore). The Redis documentation indicates that at least 3 sentinels are required for a robust deployment. Redis servers are under a master/slave architecture, and the master is exposed through a kuberntes service, which is the endpoint for Backbeat.

vrancurel: I am a bit confused here, we must support clusters with 3 or 5 nodes. If you deploy with 5 servers, 3 is obviously not a good default setting.

How could we support the 2 types of cluster generically @NicolasT @ssalaues ? E.g. having a global value clusterSize=3 or 5 and then we could derive the proper settings for all stateful components

ssalaues: This will still support 5 nodes, in those cases where we have a 5 node deployment, the main zenko/values.yaml would need to be modified for all the appropriate replicas (simply changing 3 to 5). I don't believe any other settings would need to be changed. So we inherently already support 1, 3, or 5+ node clusters thanks to the Kubernetes replica and stateful sets.

However it would be nice to have a single global value for the cluster size however I don't believe there is a way to do this (at least with a mixture of our charts and upstream charts).

NicolasT: > I am a bit confused here, we must support clusters with 3 or 5 nodes. If you deploy with 5 servers, 3 is obviously not a good default setting.

Why's that? What happens when we have a 20-node K8s cluster?

We should *not* have globals depending on cluster size. It's up to the person deploying the solution to pick what's suitable. We could however have a library of `values.yml` files for various standard architectures.

ssalaues: If it makes it simpler for SE's and others to deploy, it wouldn't be a bad idea to have a library of values.yml.

charts/single-node-values.yml
rahulreddy: I think for single nodes having 1 sentinel is the right approach (this is how it's implemented in Federation as well). This helps if the deployment scales to multiple nodes.

charts/zenko/values.yaml
vrancurel: so do we go for 3 as the default ?

PR #104, opened 2018-05-18
bf: add backbeat producer init container

giacomoguiulfo:
This will prevent the backbeat-producer from hanging while opening the zookeeper connection by using an init container that will check if there are any Kafka brokers ready.

### DIFF
--- a/charts/backbeat/templates/producer/deployment.yaml
+++ b/charts/backbeat/templates/producer/deployment.yaml
@@ -15,6 +15,15 @@ spec:
         app: {{ template "backbeat.name" . }}
         release: {{ .Release.Name }}
     spec:
+      initContainers:
+      - name: check-dependencies-readiness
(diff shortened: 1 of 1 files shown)


Zempashi:
The kubernetes way of doing things is to crash if requirements are not met (and kubernetes will restart periodically). Especially for code we own, we are not forced to do such hacks

NicolasT:
What @Zempashi said. Our software should be able to handle outages of services it consumes (at least by crash(loop)ing). Waiting for them to be available, in this implementation even infinitely long (!) is not the right approach.

giacomoguiulfo:
I agree, will open a ticket for a fix in backbeat instead. The idea for this essentially came from this article https://medium.com/@xcoulon/initializing-containers-in-order-with-kubernetes-18173b9cc222 to start containers in order. An init container like this can still lead to a faster deployment but that's probably out of the scope for now. Closing this PR.

PR #112, opened 2018-05-22
ft: add replication status to backbeat

ssalaues:
ZENKO-394
This adds the Replication Status Processor deployment to the backbeat
chart which is needed to write the status back to CloudServer.

### DIFF
--- a/charts/backbeat/templates/replication/status_deployment.yaml
+++ b/charts/backbeat/templates/replication/status_deployment.yaml
@@ -1,0 +1,60 @@
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  name: {{ template "backbeat.fullname" . }}-status
+  labels:
+    app: {{ template "backbeat.name" . }}-replication
+    chart: {{ template "backbeat.chart" . }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+spec:
+  replicas: {{ .Values.replication.status.replicaCount }}
+  template:
+    metadata:
+      labels:
+        app: {{ template "backbeat.name" . }}-replication
+        release: {{ .Release.Name }}
+    spec:
+      containers:
+        - name: replication-status
+          image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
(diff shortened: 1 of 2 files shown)


charts/backbeat/templates/replication/status_deployment.yaml
```diff
@@ -0,0 +1,68 @@
...
+              value: "{{- printf "%s-cloudserver-front" .Release.Name | trunc 63 | trimSuffix "-" -}}"
+            - name: EXTENSIONS_REPLICATION_SOURCE_S3_PORT
+              value: "80"
+            - name: EXTENSIONS_REPLICATION_DEST_AUTH_TYPE
+              value: service
+            - name: EXTENSIONS_REPLICATION_DEST_AUTH_ACCOUNT
+              value: service-replication
+            - name: EXTENSIONS_REPLICATION_DEST_BOOTSTRAPLIST
+              value: "{{- printf "%s-cloudserver-front:80" .Release.Name | trunc 63 | trimSuffix "-" -}}"
+            - name: MONGODB_HOSTS
```
jonathan-gramain: The replication status processor does not access mongodb directly, so it should not be necessary to provide this var (we only have the single queue populator component accessing the mongo log and populating kafka with log entries for other components to use).

charts/backbeat/values.yaml
```diff
@@ -47,3 +47,12 @@ replication:
...
     affinity: {}
+
+  status:
+    replicaCount: 1
+    
+    resources: {}
+    nodeSelector: {}
+    tolerations: []
+    affinity: {}
+
```
jonathan-gramain: minor: missing newline at end of file

jonathan-gramain:
@rahulreddy `REMOTE_MANAGEMENT_DISABLE` is needed since the replication status processor is also initializing management layer otherwise. I'm good to remove the others.

ssalaues:
Updated to reflect the comments.

charts/backbeat/values.yaml
ssalaues: Added the newline to the file. Thank you!

PR #113, opened 2018-05-22
ft: add service tests

giacomoguiulfo:
These are very basic service tests. They essentially test that service works as an endpoint that properly connects to the underlying pods. These tests can be useful for catching bad port configurations.

An example of an official Kubernetes chart service test: https://github.com/kubernetes/charts/blob/master/stable/redis-ha/templates/tests/test-redis-sentinel-service.yaml

These tests are run as part of helm test. So to execute them just do:
```helm test zenko``` (assuming zenko is the name of your release) **OR**
```helm test zenko --cleanup``` to cleanup containers automatically.

### DIFF
--- a/charts/backbeat/templates/tests/test-api-service.yml
+++ b/charts/backbeat/templates/tests/test-api-service.yml
@@ -1,0 +1,18 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: {{ template "backbeat.fullname" . }}-api-service-test
+  annotations:
+    "helm.sh/hook": test-success
+spec:
+  containers:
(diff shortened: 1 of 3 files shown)


NicolasT:
TBH, I'd rather see the Zenko E2E test-suite to be extended rather than having random tests in various charts with different ways to report failures etc.

ssalaues:
Yeah definitely, the E2E test suite really does need to be expanded. I believe @tmacro has been working on that specifically. These really are mostly just for quick sanity checks when working on individual charts that may become redundant in the future as the test-suite is extended.

PR #118, opened 2018-05-25
DOC: minikube tutorial updates

LaureVergeron:
- highlight minikube uses SSL and self-signed certs
 - include orbit registration process
 - cosmetic changes


charts/minikube.md
```diff
@@ -60,30 +63,66 @@ helm install --name zenko \
...
+certificate.
+
+See the example below
+
+```shell
+$ cat /etc/hosts
+127.0.0.1   localhost
+127.0.1.1   machine
+
+192.168.99.100 minikube
```
ssalaues: Just a heads up that in this configuration the hosts file modification isn't necessary and should be able to connect directly using the IP address because of the way we setup the hosts in `single-node-values.yml`. Not a blocker but just incase we wanted to add that.

charts/minikube.md
```diff
@@ -60,30 +63,66 @@ helm install --name zenko \
...
-minikube dashboard
+$ minikube dashboard
 ```
 #### NOTE that once you helm install, it may take several minutes for all the pods to load and stabilize
 
 The endpoint can now be accessed via the kubernetes cluster ip (run ```minikube ip``` to display the cluster ip)
 
+## Testing your deployment
+
+To test your minikube deployment, please refer to our client configuration guide,
```
ssalaues: Since Orbit is enabled by default, it might be a good idea to have a link to `docs/orbit_registration.md` here.

docs/orbit_registration.md
```diff
@@ -0,0 +1,142 @@
...
+
+rs0:PRIMARY> use metadata
+switched to db metadata
+
+rs0:PRIMARY> db.PENSIEVE.findOne({_id: "auth/zenko/remote-management-token"})
+{
+    "_id" : "auth/zenko/remote-management-token",
+    "value" : {
+        "instanceId" : "f1a846d6-b115-4f15-b0e6-5a9c17e7ae01",
+        "issueDate" : "2018-05-25T15:45:56Z",
```
ssalaues: maybe something more like this? It takes up less screen space when reading and gets rid of unnecessary info for finding the ID =)
```
{
     "_id" : "auth/zenko/remote-management-token",
     "value" : {
         "instanceId" : "f1a846d6-b115-4f15-b0e6-5a9c17e7ae01",
         "issueDate" : "2018-05-25T15:45:56Z",
         ...
      }
}
```

charts/minikube.md
```diff
@@ -60,30 +63,66 @@ helm install --name zenko \
...
+use_https = True
+
+$ s3cmd ls  --no-check-certificate
+
+$
+```
+
+If you get an empty response to your first list, it means communication went
+through properly with Zenko hosted in minikube. You may now configure it using
+Orbit, once you [register your instance](../docs/orbit_registation.md).
```
jonathan-gramain: There's a typo in the link ("registation")

docs/orbit_registration.md
```diff
@@ -0,0 +1,142 @@
...
+A new tab should open in your browser, taking you to the Kubernetes dashboard.
+
+##### For MetalK8s:
+```
+$ kubectl proxy
+```
+While the tunnel is up and running, access the dashboard at
+[http://localhost:8001/api/v1/namespaces/kube-system/services/https:kubernetes-dashboard:/proxy/](http://localhost:8001/api/v1/namespaces/kube-system/services/https:kubernetes-dashboard:/proxy/).
+
+### Step 2: finding the Zenko instance ID
```
jonathan-gramain: :+1:

docs/orbit_registration.md
ssalaues: A dummy instance ID that can't be registered might be a good idea so people can't register the example ID.

docs/orbit_registration.md
```diff
@@ -0,0 +1,142 @@
...
+the pods running for that service. On the top right menu, click "Logs", and then
+do a search of "instance ID" - you should find a line in the logs giving you a
+long string of random characters: this is your instance id, copy it. If you
+cannot find such a line in the logs, try option B.
+
+#### Option B: from the Metadata MongoDB cluster
+
+From the dashboard, click on the mongodb-replicaset service, and then on the
+pod that is mongodb-replicaset-**0**. This is important as the other pods are
+slaves to that master pod 0, so you will not be able to query the MongoDB
```
vrancurel: Are you sure the master is always the pod 0 ?

charts/minikube.md
```diff
@@ -34,6 +34,9 @@ Once you have the repo cloned you can retrieve all dependencies:
 $ helm repo add incubator http://storage.googleapis.com/kubernetes-charts-incubator
 "incubator" has been added to your repositories
 
+$ helm repo add kubernetes-charts-incubator http://storage.googleapis.com/kubernetes-charts-incubator
```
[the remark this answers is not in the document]
LaureVergeron: Thanks, rebased too late on master probably :) removed

charts/minikube.md
LaureVergeron: I can add a mention ; the main thing, to me, is that it's easier then for Orbit config because endpoint definition DNS style is human-readable and less error prone than IPs. I'll add a note explaining why this choice is made in the tutorial, good catch!

charts/minikube.md
LaureVergeron: Added a link here, and rephrased the second one at the end of the "testing" Paragraph

docs/orbit_registration.md
LaureVergeron: Yes, of course ;) You reviewed really fast, which is cool, so I hadn't crashed the instance from which this id came - it's done, registration should now be disabled

Shortened paragraph about mongo query

docs/orbit_registration.md
LaureVergeron: This is empirical, of course, but it's always pod0 that comes up first at start time(as should the mongo master), and on my fairly numerous deployments, it's always been the master. Now I guess if you crash it, the master might change? But at start time, it should be pod 0, yes.

ssalaues: This is true, pod0 will always come up first in a fresh deployment and generally be the master. However in many situations (long running stacks, VM host is rebooted, stack updated, etc) it is very likely that pod0 will no longer be the replicaset master as any of the others are able to take this role. It may be good to have a note that this is typically correct but not always the case.

docs/orbit_registration.md
LaureVergeron: Isn't that a bug then? I'll change the id, but might be worth opening an Orbit issue for "Destroyed stack ids are not forgotten", no?

ssalaues: I'm not sure if it's intended to be this way but you're right it'd be worth opening an issue.

docs/orbit_registration.md
LaureVergeron: Added a "TIP" kind of note, let me know if it works for you :)

PR #121, opened 2018-06-05
ft: add proxy support for egress to clouds

giacomoguiulfo:
This adds the ability to specify optional HTTP and HTTPS proxy endpoints for Cloudserver on its values.yml.

### DIFF
--- a/charts/cloudserver-front/templates/deployment.yaml
+++ b/charts/cloudserver-front/templates/deployment.yaml
@@ -43,6 +43,21 @@ spec:
               value: "{{ .Release.Name }}-cloudserver-front,{{ .Values.endpoint }}"
             - name: HEALTHCHECKS_ALLOWFROM
               value: "{{ .Values.allowHealthchecksFrom }}"
+{{- if .Values.proxy.http }}
+            - name: http_proxy
+              value: "{{ .Values.proxy.http }}"
+            - name: HTTP_PROXY
+              value: "{{ .Values.proxy.http }}"
+            - name: https_proxy
+              value: "{{- if .Values.proxy.https }}{{ .Values.proxy.https }}{{- else }}{{ .Values.proxy.http }}{{- end }}"
+            - name: HTTPS_PROXY
+              value: "{{- if .Values.proxy.https }}{{ .Values.proxy.https }}{{- else }}{{ .Values.proxy.http }}{{- end }}"
+{{- else if .Values.proxy.https }}
+            - name: https_proxy
+              value: "{{ .Values.proxy.https }}"
+            - name: HTTPS_PROXY
+              value: "{{ .Values.proxy.https }}"
+{{- end }}
 {{- if .Values.mongodb.enabled }}
             - name: S3METADATA
               value: "mongodb"
--- a/charts/cloudserver-front/values.yaml
+++ b/charts/cloudserver-front/values.yaml
@@ -33,6 +33,13 @@ image:
   tag: 0.2.1
   pullPolicy: IfNotPresent
 
+proxy:
+  # If you want to use an HTTP proxy, add the respective endpoints after
+  # 'http:' and/or 'https:'. If the HTTP proxy endpoint is set but the HTTPS
+  # one isn't, the HTTP proxy will be used for HTTPS traffic as well.
+  http: ""
+  https: ""
+
 service:
   type: ClusterIP
   port: 80


charts/cloudserver-front/values.yaml
```diff
@@ -29,6 +29,14 @@ image:
   tag: 0.2.1
   pullPolicy: IfNotPresent
 
+env: {}
+  # If you want to specify additional environmental variables, like http proxy
+  # endpoints, remove the curly braces after 'env:' and the variables below.
+  # http_proxy:
```
NicolasT: Not fond of this approach. In the most likely case, one would need to set **4** different custom variables, all with the value, which leaves a lot of room for error (typos, whatnot).

Please reconsider.

charts/cloudserver-front/templates/deployment.yaml
```diff
@@ -27,6 +27,10 @@ spec:
             - name: http
               containerPort: 8000
           env:
+{{- range $key, $value := .Values.env }}
+            - name: {{ $key | replace "." "_" }}
```
NicolasT: Even when sticking to this approach (cfr. https://github.com/scality/Zenko/pull/121/files#r193228860), this is wrong. When one sets the key to something with a dot, this may have some reason. If it should be an `_`, one should use an `_` in the values definition. Rewriting this behind the user's back seems odd, and can easily cause confusion.

charts/cloudserver-front/values.yaml
dora-korpar: Technically, the user would only need to set one of each `http` and `https`, as the code checks for both all lowercase and uppercase

NicolasT: @dora-korpar In that case, I guess only 2 examples should be provided.
@xaf-scality What do you think?

rahulreddy: I agree with @dora-korpar . We should allow setting only `HTTP_PROXY` or `HTTPS_PROXY` and case management will be handled by CloudServer.

NicolasT: My point was that this allows to set arbitrary environment variables, and happens to document 2 important ones. In the ticket I described having 2 dedicated configuration options, where even setting only one of them would be sufficient in most cases to have the system running.

The fact that this one (or 2) options are rendered into however many environment variables using whatever names for the container environment (i.e. in the template logic) is irrelevant at that point.

charts/cloudserver-front/templates/deployment.yaml
giacomoguiulfo: It's just a protection since most shells don't support dots in env variables. Found it on some other Kubernetes charts: https://github.com/kubernetes/charts/blob/master/incubator/zookeeper/templates/statefulset.yaml#L69

charts/cloudserver-front/values.yaml
giacomoguiulfo: This approach doesn't look the best for me too. The approach @NicolasT suggested was better, but it introduced unnecessary templating. However, I later figured a simple workaround using '{}' braces. Updating the PR now.

charts/cloudserver-front/values.yaml
```diff
@@ -29,6 +29,12 @@ image:
   tag: 0.2.1
   pullPolicy: IfNotPresent
 
+proxy:
+  # If you want to use an http proxy, remove the curly braces
+  # after 'http' and/or 'https', and add the respective endpoints.
+  http: {}
```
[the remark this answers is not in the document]
giacomoguiulfo: If an empty string is used, the variables would be exported as empty strings. Also {{- if .Values.proxy.https }} will be true for an empty string. Commenting them out will cause a templating error because ".Values.proxy.https" won't exist. These are the reasons why I initially went with the other solution.

charts/cloudserver-front/templates/deployment.yaml
```diff
@@ -43,6 +43,21 @@ spec:
...
+{{- if .Values.proxy.http }}
+            - name: http_proxy
+              value: "{{ .Values.proxy.http }}"
+            - name: HTTP_PROXY
+              value: "{{ .Values.proxy.http }}"
+            - name: https_proxy
+              value: "{{- if .Values.proxy.https }}{{ .Values.proxy.https }}{{- else }}{{ .Values.proxy.http }}{{- end }}"
+            - name: HTTPS_PROXY
+              value: "{{- if .Values.proxy.https }}{{ .Values.proxy.https }}{{- else }}{{ .Values.proxy.http }}{{- end }}"
+{{- else if .Values.proxy.https }}
```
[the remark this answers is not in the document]
giacomoguiulfo: No, if one configures them differently they will be different.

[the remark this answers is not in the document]
NicolasT: *facepalm*

charts/cloudserver-front/values.yaml
```diff
@@ -29,6 +29,12 @@ image:
   tag: 0.2.1
   pullPolicy: IfNotPresent
 
+proxy:
+  # If you want to use an http proxy, add the respective endpoints after
+  # 'http:' and/or 'https:'
```
NicolasT: Maybe add a note explaining that, if a HTTP proxy is set but no HTTPS proxy, the HTTP one will be used for HTTPS traffic anyway.

charts/cloudserver-front/values.yaml
```diff
@@ -29,6 +29,13 @@ image:
   tag: 0.2.1
   pullPolicy: IfNotPresent
 
+proxy:
+  # If you want to use an http proxy, add the respective endpoints after
+  # 'http:' and/or 'https:'. If the http proxy endpoint is set but the https
+  # one isn't, the http proxy will be used for https traffic anyway.
```
NicolasT: Change `http` into `HTTP`, `https` to `HTTPS` (in the comment, not the actual variable names, nor references to the variable names in the comment, of course), and change `anyway` into `as well`, and I'm all happy and such.

PR #125, opened 2018-06-12
ft/ZENKO-487 Add ability to stand up end to end tests locally

tmacro:
Test away with just a `make local-run`!


tests/.gitignore
```diff
@@ -12,3 +12,5 @@ dist/
 venv/
 zenko_e2e.egg-info/
 .secrets.env
+.local.env
+.local.make
```
jonathan-gramain: did you mean `local.make`?

tmacro: I changed it to keep it inline with the other local config files

PR #126, opened 2018-06-13
chore(ingress): set nginx max-body-size default to 100m

Zempashi:


ssalaues:
Could you rebase off master so we can run the updated CI and fix conflicts? 

Also is there a reason why we couldn't set it to 0m (disabling the checking of body sizes) or is that a bad idea?

Zempashi:
@ssalaues done ;)

Zempashi:
@ssalaues yes security wise, I imagine someone able to put enough body to make nginx go out of memory and crash multiple times, which will make all ingress controller being in crashloopbackoff resulting in a denial of service.

PR #127, opened 2018-06-13
bf: enabled site level metrics

philipyoo:
goes with https://github.com/scality/backbeat/pull/322

### DIFF
--- a/charts/backbeat/templates/api/deployment.yaml
+++ b/charts/backbeat/templates/api/deployment.yaml
@@ -38,6 +38,12 @@ spec:
               value: "6379"
             - name: HEALTHCHECKS_ALLOWFROM
               value: '0.0.0.0/0'
+            - name: EXTENSIONS_REPLICATION_SOURCE_AUTH_ACCOUNT
+              value: service-replication
+            - name: EXTENSIONS_REPLICATION_DEST_BOOTSTRAPLIST
+              value: "{{- printf "%s-cloudserver-front:80" .Release.Name | trunc 63 | trimSuffix "-" -}}"
+            - name: MONGODB_HOSTS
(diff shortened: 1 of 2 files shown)


charts/backbeat/templates/api/deployment.yaml
```diff
@@ -38,6 +38,12 @@ spec:
               value: "6379"
             - name: HEALTHCHECKS_ALLOWFROM
               value: '0.0.0.0/0'
+            - name: EXTENSIONS_REPLICATION_SOURCE_AUTH_ACCOUNT
+              value: service-replication
+            - name: EXTENSIONS_REPLICATION_DEST_BOOTSTRAPLIST
+              value: "{{- printf "%s-cloudserver-front:80" .Release.Name | trunc 63 | trimSuffix "-" -}}"
+            - name: MONGODB_HOSTS
+              value: "{{ .Release.Name }}-mongodb-replicaset-0.{{ .Release.Name }}-mongodb-replicaset:27017,{{ .Release.Name }}-mongodb-replicaset-1.{{ .Release.Name }}-mongodb-replicaset:27017,{{ .Release.Name }}-mongodb-replicaset-2.{{ .Release.Name }}-mongodb-replicaset:27017"
```
NicolasT: We should really start resolving the headless service to discover the MongoDB hosts instead of constructing it like this... @ssalaues

charts/backbeat/values.yaml
```diff
@@ -6,7 +6,7 @@ orbit:
 
 image:
   repository: zenko/backbeat
-  tag: 0.2.4
+  tag: 8.0.0
```
jonathan-gramain: Was this intended to bump to 8.0.0?

ssalaues: Yeah we are starting from 8 for any new images =)

PR #137, opened 2018-06-15
Add guide to install Zenko on Google Kubernetes Engine

smaffulli:
This guide adds details on how to install Zenko on an existing K8s
cluster hosted on Google Cloud.

### DIFF
--- a/charts/gke.md
+++ b/charts/gke.md
@@ -1,0 +1,96 @@
+# Install Zenko On Google Cloud Kubernetes Engine
+
+Zenko can be installed on a Kubernetes cluster managed on Google Cloud
+(GKE). Start a new cluster on Kubernetes following the instructions on [Google
+Cloud
+documentation](https://cloud.google.com/kubernetes-engine/docs/quickstart).
+
+To run Zenko you'll need a cluster with 3 nodes, 2vCPU and 7.5GB RAM
+each. Once the cluster is running, connect to it and install Helm on
+it.
+
+## Create Role For Tiller
+
+Google Kubernetes Engine requires Role Based Access Control to be
+setup. The first step is to create a `serviceaccount` for `tiller`:
+
+```shell
+$ kubectl create serviceaccount tiller --namespace kube-system
+```
+
+Check that the correct context is set:
+
+```shell
+$ kubectl get nodes
+NAME                                       STATUS   ROLES    AGE VERSION
+gke-cluster-1-default-pool-9ad69bcf-4g2n   Ready    <none>   1m  v1.8.10-gke.0
+gke-cluster-1-default-pool-9ad69bcf-frj5   Ready    <none>   1m  v1.8.10-gke.0
+gke-cluster-1-default-pool-9ad69bcf-rsbt   Ready    <none>   1m  v1.8.10-gke.0
+```
+
+## Install Helm on GKE Cluster
+
+Helm is not available by default on GKE and needs to be installed.
+
+```shell
+$ curl https://raw.githubusercontent.com/kubernetes/helm/master/scripts/get > get_helm.sh
+$ bash ./get_helm.sh
+```
+
+Once that's completed, start Helm:
+
+```shell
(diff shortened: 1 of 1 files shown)


charts/gke.md
```diff
@@ -0,0 +1,87 @@
...
+Downloading zookeeper from repo http://storage.googleapis.com/kubernetes-charts-incubator
+Deleting outdated charts
+```
+
+With your dependencies built, you can run the following shell command
+to deploy a single node Zenko stack with Orbit enabled.
+
+
+```shell
+$ helm install --name zenko -f single-node-values.yml zenko
```
giacomoguiulfo: Just curious, why make a single node deployment is we have 3 nodes available? 🤔

smaffulli: eh, maybe because I'm a dork? :) /me fixing it

charts/gke.md
```diff
@@ -0,0 +1,87 @@
...
+Downloading zookeeper from repo http://storage.googleapis.com/kubernetes-charts-incubator
+Deleting outdated charts
+```
+
+With your dependencies built, you can run the following shell command
+to deploy a single node Zenko stack with Orbit enabled.
+
+
+```shell
+$ helm install --name zenko zenko
```
giacomoguiulfo: Typically you may want to enable the ingress on Zenko like this:
```helm install --name zenko --set ingress.enabled=true zenko```. Which does require a running ingress controller. I think GKE runs an GCE ingress controller by default but I can't find an official reference to it yet.

Comment (smaffulli) [reply charts/gke.md]: Better be explicit then. I amended the commit.

File: charts/gke.md [hunk 3]
```diff
@@ -0,0 +1,87 @@
+# Install Zenko On Google Cloud Kubernetes Engine
+
+Zenko can be installed on a cluster managed on Google Cloud (GKE).
```
NicolasT: `Kubernetes cluster`?

charts/gke.md
```diff
@@ -0,0 +1,87 @@
...
+Google Kubernetes Engine requires Role Based Access Control to be
+setup. The first step is to create a `serviceaccount` for `tiller`:
+
+```shell
+$ kubectl create serviceaccount tiller --namespace kube-system
+```
+
+## Install Helm on GKE Cluster
+
+Strangely, Helm is not available by default on GKE and needs to be
```
NicolasT: Why `Strangely`? Also, given a `ServiceAccount` is created already, the 'strangely' comes a bit 'late'.

charts/gke.md
```diff
@@ -0,0 +1,87 @@
...
+installed.
+
+```shell
+$ curl https://raw.githubusercontent.com/kubernetes/helm/master/scripts/get > get_helm.sh
```
NicolasT: `curl -Lo https://...`, no need to redirect. Also, I'd rather not tell our users to grab scripts from the internet which then download a tarball and extract it. Why not simply point to the Helm install docs?

charts/gke.md
```diff
@@ -0,0 +1,87 @@
...
+$ bash ./get_helm.sh
+```
+
+Once that's completed, start Helm:
+
+```shell
+$ helm init --service-account tiller --wait
```
NicolasT: Make sure the right kubecontext is set up etc. first, e.g. by asking the user to run `kubectl get nodes` before and validate the output.

smaffulli:
@ballot-scality you're preaching to the converted here :) I simply used the format that was already in this folder. Happy to repurpose in ReST. Let's add a ticket to change all .md into .rst :)

charts/gke.md
smaffulli: Not a fan of installing from random shell script but that's what Helm install docs recommends as one of the installation method. Check https://github.com/kubernetes/helm/blob/master/docs/install.md

ssalaues:
Awesome so we'll pass it off to the docs team to convert to RST =)

PR #139, opened 2018-06-18
Setting up nightly stage on CI

tcarmet:
Successfully tested, nightly images are properly pulled and installed on the Zenko repo. Here's a build link https://eve.devsca.com/github/scality/zenko/#/builders/4/builds/664

Feel free to ask some changes and everything

### DIFF
--- a/eve/ci-values.yml
+++ b/eve/ci-values.yml
@@ -1,0 +1,40 @@
+
+ingress:
+  enabled: true
+
+prometheus:
+  rbac:
+    create: false
+  serviceAccounts:
+    alertmanager:
+      name: default
+      create: false
+    nodeExporter:
+      name: default
+      create: false
+    kubeStateMetrics:
+      name: default
+      create: false
+    pushgateway:
+      name: default
+      create: false
+    server:
+      name: default
+      create: false
+
+zenko-queue:
+  rbac:
+    enabled: false
+
+redis-ha:
+  rbac:
+    create: false
+  serviceAccount:
+    create: false
+    name: default
+
+cloudserver-front:
+  image:
+    pullPolicy: Always
+  ci:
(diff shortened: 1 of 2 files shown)


tcarmet:
The pre-merge build is not working, but I think it is due to backbeat's new version (if I rollback it works)

Maybe @tmacro knows a little bit more about this

Zempashi:
In the theory of releng engeneering "science" there is no so called "nightly". There is only a pipeline:
```
test 1 => action 1 => test2 => action2 => test3 =>action3
```
On ring, we have something similar:
```
pre-merge => merge => post-merge
```
The fact that post-merge run nightly is for performance sake. We have to take care that having the nightly will not shot ourself
"oh the nightly hasn't run today, I can't work", "I will wait for the nightly to run to do something specific and long running"

tcarmet:
Sure thing!
I'm personnally not a fan of nightly builds (because you have to wait the next morning for results) and you're right about the "Oh I can't work, no build was triggered" I've been there.

I think the nightly build was requested to have some kind of early news about the end2end tests, and avoid having bad news just before a release, or making a PR on this project.

I'm closing this PR until I have further discussion with everyone (you, SF, releng)

Zempashi:
No no, don't close this PR. You misunderstand me. Just call it differently that nightly. We are just bare human, and believe it to occur at night (asynchronously, which is bad), but nothing technical prevent it run synchronously at each merge/tag/whatever_but_not_commit

tcarmet:
OH got you! sorry about that @Zempashi :smile:  

I'll rename the stage and also make some modification into the tags on s3, backbeat..., using the sha1 or something like that. Remove the nightly name to avoid confusions, and I'll reopen a PR according to that.

PR #140, opened 2018-06-18
feat(zenko-505): E2E Finishing touches

tmacro:


ssalaues:
Looks good but it seems like the replication test is failing. Is it the same one enabled in this PR?

PR #145, opened 2018-06-21
pushing and using e2e docker image on each build  

tcarmet:
I was afraid that using always latest tag on the e2e test image will not actually test zenko as we're not applying the same code if we modify the tests.

So I'm using the private registry to : 
```shell
build the image -> tag with the git sha1 -> push it to the private registry -> and pull it for testing
```

So the concept work! which is nice. But as suspected the tests are failing for now. I'll continue the work if you guys agree on this!

Thanks !

### DIFF
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -5,6 +5,13 @@ branches:
     stage: "pre-merge"
 
 models:
+  - Git: &git_pull
+      name: git pull
+      repourl: "%(prop:git_reference)s"
+      mode: full
+      method: clobber
+      retryFetch: true
+      haltOnFailure: true
   - Upload: &upload_artifacts
       source: tests/artifacts
       urls:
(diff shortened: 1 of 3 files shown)


giacomoguiulfo:
@tcarmet

eve/main.yml
```diff
@@ -5,10 +5,20 @@ branches:
...
+      method: clobber
+      retryFetch: true
+      haltOnFailure: true
   - Upload: &upload_artifacts
       source: tests/artifacts
       urls:
         - "*"
+  - SetProperty: &docker_image_name
+      property: docker_image_name
+      value: '%(secret:private_registry_url)s/zenko/zenko-e2e'
```
sylvain-killian: the secret value will be stored in the build properties, a visible by all in the build tabs.

eve/main.yml
```diff
@@ -93,8 +97,32 @@ stages:
...
+        name: Private Registry Login
+        command: >
+          docker login
+          -u '%(secret:private_registry_username)s'
+          -p '%(secret:private_registry_password)s'
+          '%(secret:private_registry_url)s'
+    - SetProperty: *docker_image_name
+    - ShellCommand:
+        name: build end2end image
+        command: make DOCKER_IMAGE_NAME='%(prop:docker_image_name)s' DOCKER_IMAGE_TAG='%(prop:commit_short_revision)s' container-image
```
sylvain-killian: cosmetic: could use > notation to split this line into multiple shorter lines

tcarmet: Good point, done!
Thanks

tmacro:
@tcarmet We've been having some weirdness with the backend buckets that I believe is caused by object name conflicts when multiple tests are running simultaneously. I'd rather keep the changes in it's own PR for easier review, plus PR is needed ASAP

tcarmet:
Ok @tmacro for me it's ok to merge this PR. As it avoid breaking things in the future

tcarmet:
Maybe ping @vrancurel ?

PR #148, opened 2018-06-22
bf: ZENKO-423 Remove deep healthcheck from livenessProbe

anurag4DSB:
- LivenessProbe is for checking if the service is up: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/ 
- We do not need deep health check for that because deep health check also checks the backend.
- Use Case for bug: If someone deletes their AWS bucket which was configured in orbit, kubernetes will try to restart the cloud server container because /_/healthcheck/deep will respond with an error.
- To check the service status we only need /_/healthcheck

### DIFF
--- a/charts/cloudserver-front/templates/deployment.yaml
+++ b/charts/cloudserver-front/templates/deployment.yaml
@@ -95,7 +95,7 @@ spec:
           args: ['npm', 'run', 'start_s3server']
           livenessProbe:
             httpGet:
-              path: /_/healthcheck/deep
+              path: /_/healthcheck
               port: http
           resources:
 {{ toYaml .Values.resources | indent 12 }}


NicolasT:
> LivenessProbe is for checking if the service is up

Rather: "LivenessProbe is for checking if the service is up *and able to properly serve requests*"

- Cloudserver instances are dispatched to through a `Service`
- As long as an instance returns successful to a liveness check, it will be added as a backend for said `Service`

'liveness' of Cloudserver can be defined in 2 ways:
- It's up and running and can serve requests, even returning HTTP 500 when e.g. MongoDB is not available
- It's up and running but can not serve requests because MongoDB is not available, and you don't want your service to return a HTTP 500 to the client

I would argue scenario 2 above is better than the first one: if one instance of Cloudserver is unable to access MongoDB (or is heavily overloaded, or whatever), you want to take it out of the `Service`, otherwise 1 out of N (= # of Cloudserver instances) of client requests will fail, for no good reason: they could be handled properly by any of the other instances!

So, as far as I'm concerned, a liveness check should be 'deep', i.e. including checks about availability of backing services.

Now, maybe the current definitions of Cloudserver's healthcheck modes are simply off: there seems to be one to check whether the service is up (non-deep), none to check whether a service is able to handle requests *if the environment is in a good state* (where existence of some remote bucket has no impact: the service may not be able to serve *specific* requests, related to that bucket, but *is* able to serve other requests!), and one for end-users which could take the existence of all backend buckets into account.

To summarize:
- When adding the liveness check, I assumed the service would only validate whether its dependencies were available, no per-request specifics
- I believe this should still be added, or backend checks should be added to the non-deep check

PR #149, opened 2018-06-22
[DO NOT MERGE] doc: Reconcile documentation files

NicolasT:
This branch reworks the structure of files in this repository related to documentation. This is
- Work in Progress
- Merely suggestions, for now


smaffulli:
thanks for leading the charge Nicolas. I have scheduled a meeting with William next week to draft a comprehensive table of content for Zenko docs.  Let's wait for that meeting's output before making a decision on content structure. FWIW, this that you made makes lots of sense to me.

NicolasT:
It's not done. When you wrote the GKE docs, I figured we have too many places in which we document the `helm dep build; helm install; ...` dance. So here's what I'm doing next: extract these instructions, turn the Minikube and GKE guide into a 'Create a K8s environment' sorta guides only, so you end up with a K8s cluster which has Helm set up, add a similar page describing MetalK8s and linking to its docs, then have all of them at the end refer to a more generic 'Set up Zenko on K8s' document. There's way too much duplication right now.

NicolasT:
Note: this PR is not meant to improve any of the existing documents, they're simply copied around.

smaffulli:
Yep, that's the idea of sketching out a comprehensive Table Of Content with William next week... a master plan of what documentation needs to exist in order to have a good Zenko 1.0 release. Some pieces already exist, scattered around as you noticed, and some will need to be produced.

smaffulli:
This is the draft structure we came up with: https://docs.google.com/document/d/1hC60HewoRgtKec6a12DkpGWyDi1T5ylyPNjk-lkvxSw/edit#

basically docs/ should contain 4 `guides`: installation, operations, reference API, developers. What do you think of renaming `deployment` to `installation` and merging this? we can add stubs for the other directories/files we need to author.

NicolasT:
TBH I'm not very fond of *Installation*, because where does *Upgrade* go then? In *Operations*? Since I believe it's good practice (and common practice in the Kubernetes/Helm world) to blur the line between *installation* and *upgrade* (and have the same procedure for both, basically), this may be better of with some different vocabulary.

@nyemade-harmon @smaffulli @wabernatScality

smaffulli:
how about using `getting-started` to describe the *first* installation? and use `operations` to include all deployment options in detail (new installs and upgrades?)

PR #150, opened 2018-06-23
feat: extend support without orbit

giacomoguiulfo:
#### This adds several features to Zenko when Orbit is disabled:
- Support for using different cloud backends.
- Replication.
- Multiple Cloudserver users.

#### Documentation

Instrucctions on the setup can be found in ```charts/setup-no-orbit.md```.

#### Upgrades

If related configuration values are changed in ```values.yml```, a ```helm dependency build zenko && helm upgrade zenko zenko``` command (assuming zenko is the release name) will restart the pods with the updated configuration.

### DIFF
--- a/charts/backbeat/values.yaml
+++ b/charts/backbeat/values.yaml
@@ -1,8 +1,12 @@
 # Default values for backbeat.
 # This is a YAML-formatted file.
 # Declare variables to be passed into your templates.
-orbit:
-  enabled: true
+
+global:
+  orbit:
+    enabled: true
+  locationConstraints: {}
+  replicationEndpoints: []
 
 image:
   repository: zenko/backbeat
--- a/charts/zenko/values.yaml
+++ b/charts/zenko/values.yaml
@@ -24,19 +24,25 @@ ingress:
     #   hosts:
     #     - zenko.example.com
 
+global:
+  orbit:
+    enabled: true
+    endpoint: https://api.zenko.io
+    # When 'orbit.enabled' is 'true', these aren't used, please use
+    # https://zenko.io to manage your deployment
+  locationConstraints: {}
+  replicationEndpoints: []
+
 cloudserver:
   replicaCount: *nodeCount
   replicaFactor: 10
   mongodb:
     replicas: *nodeCount
-  orbit:
-    enabled: true
-  # When 'orbit.enabled' is 'true', these aren't used, please use
-  # https://zenko.io to manage your deployment
   endpoint: zenko.local
-  credentials:
-    keyId: deployment-specific-access-key
-    secretKey: deployment-specific-secret-key
+  users: {}
+    # accountName:
+      # access:
+      # secret:
 
 backbeat:
   replication:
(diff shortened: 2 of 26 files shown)


vrancurel:
Impressive ! all access keys are in secrets ?

vrancurel:
I hope access keys are fake in the doc ;)

charts/backbeat/templates/replication/consumer_deployment.yaml
```diff
@@ -11,6 +11,8 @@ spec:
   replicas: {{ .Values.replication.consumer.replicaCount }}
   template:
     metadata:
+      annotations:
```
ssalaues: It doesn't seem that this should be here if orbit isn't enabled. Should prob add a conditional for this.

charts/backbeat/templates/replication/producer_deployment.yaml
```diff
@@ -11,6 +11,8 @@ spec:
   replicas: {{ .Values.replication.producer.replicaCount }}
   template:
     metadata:
+      annotations:
```
ssalaues: same here for orbit conditional

charts/backbeat/templates/replication/status_deployment.yaml
```diff
@@ -11,6 +11,8 @@ spec:
   replicas: {{ .Values.replication.status.replicaCount }}
   template:
     metadata:
+      annotations:
```
ssalaues: same here for orbit conditional

charts/cloudserver-front/templates/deployment.yaml
```diff
@@ -15,17 +15,30 @@ spec:
       release: {{ .Release.Name }}
   template:
     metadata:
+      annotations:
```
ssalaues: same here for orbit conditional

charts/backbeat/values.yaml
```diff
@@ -1,8 +1,6 @@
 # Default values for backbeat.
 # This is a YAML-formatted file.
 # Declare variables to be passed into your templates.
-orbit:
```
ssalaues, line 4: Instead of completely removing `.Values.orbit` consider moving it to `.Values.global.orbit` in each of the sub charts (the Zenko parent values will still overwrite these)

giacope: Done! 👍

giacomoguiulfo:
@NicolasT Any thoughts before we merge this?

charts/zenko/values.yaml
```diff
@@ -24,17 +24,56 @@ ingress:
...
     #     - zenko.example.com
 
+global:
+  orbit:
+    enabled: true
+    endpoint: https://api.zenko.io
+    # When 'orbit.enabled' is 'true', these aren't used, please use
+    # https://zenko.io to manage your deployment
+  locationConstraints: {}
+    # awsbackend:
```
giacope: @ssalaues Do you think we should leave these comments under `locationConstraints`?

PR #151, opened 2018-06-25
ft: ZENKO-143 chart for backbeat GC

jonathan-gramain:
Create a new chart for the Garbage Collector service (internal service
for transient source, and later lifecycle transition policies, abort
MPU, etc.)

### DIFF
--- a/charts/backbeat/templates/gc/deployment.yaml
+++ b/charts/backbeat/templates/gc/deployment.yaml
@@ -1,0 +1,54 @@
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  name: {{ template "backbeat.fullname" . }}-gc-consumer
+  labels:
+    app: {{ template "backbeat.name" . }}-gc
+    chart: {{ template "backbeat.chart" . }}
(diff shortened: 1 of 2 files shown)


charts/backbeat/templates/gc/deployment.yaml
```diff
@@ -0,0 +1,54 @@
...
+spec:
+  replicas: {{ .Values.garbageCollector.consumer.replicaCount }}
+  template:
+    metadata:
+      labels:
+        app: {{ template "backbeat.name" . }}-gc
+        release: {{ .Release.Name }}
+    spec:
+      containers:
+        - name: {{ template "backbeat.fullname" . }}-gc-consumer
```
ssalaues: For the container name we can just use `consumer` or `gc-consumer` the entire backbeat name doesn't need to be templated in here as well since it exists under the `{{ template "backbeat.fullname" . }}-gc-consumer` pod. Otherwise LGTM

jonathan-gramain: ok, will change!

PR #152, opened 2018-06-25
feat: add support to change the number of mongodb replicas

giacomoguiulfo:
Before, the environmental variable MONGODB_HOSTS (which both Backbeat and Cloudserver-front use) was a hardcoded string of 3 MongoDB replica sets, like this:

```
{{ .Release.Name }}-mongodb-replicaset-0.{{ .Release.Name }}-mongodb-replicaset:27017,{{ .Release.Name }}-mongodb-replicaset-1.{{ .Release.Name }}-mongodb-replicaset:27017,{{ .Release.Name }}-mongodb-replicaset-2.{{ .Release.Name }}-mongodb-replicaset:27017
```

These changes will generate the proper MONGODB_HOSTS string based on the number of replicas sets specified in the charts/values.yaml file.

These changes also introduce the "nodeCount" anchor value, which serves as a base value for the number of replicas for the various deployments (MongoDB, Redis-HA, Zenko-Queue and Zenko-Quorum). Typically, this will be the number of nodes in a Kubernetes cluster, but of course, the user can modify individual values as they wish.

The single-node-values.yml was also updated and now it will only deploy 1 MongoDB replica set instead of 3.

### DIFF
--- a/charts/zenko/values.yaml
+++ b/charts/zenko/values.yaml
@@ -1,6 +1,12 @@
 # Default values for zenko.
 # This is a YAML-formatted file.
 # Declare variables to be passed into your templates.
+
+# By default, MongoDB, Redis-HA, Zenko-Queue, and Zenko-Quorum
+# will use this value for their replica count. Typically, this
+# is equivalent to the number of nodes in a Kubernetes Cluster.
+nodeCount: &nodeCount 3
+
 ingress:
   enabled: false
   # Used to create an Ingress record.
@@ -19,6 +25,8 @@ ingress:
     #     - zenko.example.com
 
 cloudserver-front:
+  mongodb:
+    replicas: *nodeCount
   orbit:
     enabled: true
   # When 'orbit.enabled' is 'true', these aren't used, please use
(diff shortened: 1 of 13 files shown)


charts/zenko/values.yaml
```diff
@@ -28,30 +36,30 @@ cloudserver-front:
...
+  replicas: *nodeCount
 
 zenko-queue:
 ## Extensive list of configurables can be found here:
 ## https://github.com/kubernetes/charts/blob/master/incubator/kafka/values.yaml
-  replicas: 3
+  replicas: *nodeCount
   rbac:
     enabled: true
   configurationOverrides:
```
giacomoguiulfo: @ssalaues Do we need to set ```offsets.topic.replication.factor``` to nodeCount as well?

ssalaues: Replication of 3 and insync of 2 should still provide enough HA for two failures. It's a user configurable option so if higher redundancy is needed.

edit: For anyone confused about the above statement, the point being made was that these should be safe values for both 3 and 5 node clusters which are the typical sizes we deploy.

PR #153, opened 2018-06-26
feat(zenko-437): Add liveness and readiness probes to backbeat

tmacro:


charts/backbeat/templates/replication/consumer_deployment.yaml
```diff
@@ -52,8 +52,14 @@ spec:
               value: "{{- printf "%s-%s" .Release.Name "redis-ha-master-svc" | trunc 63 | trimSuffix "-" -}}"
             - name: REDIS_PORT
               value: "6379"
-          # TODO livenessProbe:
-          # TODO readinessProbe:
+          livenessProbe:
+            httpGet:
+              path: /_/health/liveness
+              port: 4042
+          readinesProbe:
```
JianqinWang: `readinessProbe`?

charts/backbeat/templates/replication/producer_deployment.yaml
```diff
@@ -37,8 +37,14 @@ spec:
               value: "6379"
             - name: LOG_LEVEL
               value: {{ .Values.logging.level }}
-          # TODO livenessProbe:
-          # TODO readinessProbe:
+          livenessProbe:
+            httpGet:
+              path: /_/health/liveness
+              port: 4042
+          readinesProbe:
```
JianqinWang: same as above

charts/backbeat/templates/replication/status_deployment.yaml
```diff
@@ -46,8 +46,14 @@ spec:
               value: "6379"
             - name: LOG_LEVEL
               value: {{ .Values.logging.level }}
-          # TODO livenessProbe:
-          # TODO readinessProbe:
+          livenessProbe:
+            httpGet:
+              path: /_/health/liveness
+              port: 4042
+          readinesProbe:
```
JianqinWang: same as above

charts/backbeat/templates/lifecycle/conductor_deployment.yaml
```diff
@@ -48,6 +48,14 @@ spec:
               value: "{{ .Values.lifecycle.rules.noncurrentVersionExpiration.enabled }}"
             - name: EXTENSIONS_LIFECYCLE_RULES_ABORT_INCOMPLETE_MPU_ENABLED
               value: "{{ .Values.lifecycle.rules.abortIncompleteMPU.enabled }}"
+          livenessProbe:
+            httpGet:
+              path: /_/health/liveness
+              port: 4042
```
NicolasT: Use named ports instead of numeric ports (this applies to all occurrences...)

tmacro: Instead of using named ports, which would require that they be exposed, move the config into values.yml

PR #154, opened 2018-06-27
feat: add grafana chart

giacomoguiulfo:
Grafana chart deployed as a part of the Zenko Stack.

This chart deploys with two sidecars. One of them is responsible for adding data sources while the other one adds dashboards to the main Grafana container. This is done by reading all the config maps within the same namespace that have a specific label (see ```zenko/values.yaml``` for a better reference).

Additionally, there is already a config map that has the proper configuration to set Prometheus as a data source. It can be found on ```zenko/templates/datasource.yaml```.

This PR also adds annotations for Cloudserver and Backbeat API in order to allow Prometheus to collect metrics and therefore be able to make the Grafana Dashboards.

Finally, a basic dashboard for cloudserver is also provided, in order to prove that the setup works (and to show some graphs, of course).

This chart is been pulled from the official Kubernetes charts. See https://github.com/kubernetes/charts/tree/master/stable/grafana for further reference.

## IMPORTANT NOTE:

The Grafana tests have been disabled temporarily. This is due to the fact that the chart sidecars don't install properly on CI because of privilege issues. Therefore the chart is not been installed on CI.

Both Grafana and Prometheus can be disabled (they are enabled by default) in case the user prefers to use his own charts, which could potentially be outside the Zenko chart scope (like in Metal-k8s where Grafana and Prometheus charts deployed as a part of it).


NicolasT:
Even though it's not ready to go in as-is, very happy to see progress on this project! :+1:

charts/backbeat/templates/api/service.yaml
```diff
@@ -1,6 +1,10 @@
 apiVersion: v1
 kind: Service
 metadata:
+  annotations:
+{{- if .Values.metrics.enabled }}
```
NicolasT: In most charts, there's an `annotations` section in the `service` section of `values` where one can put arbitrary annotations. In our case, I'd fill this with the Prometheus SD annotations by default (unlike in most charts where this is empty), but follow the (naming) conventions of other charts.

charts/backbeat/values.yaml
```diff
@@ -1,6 +1,14 @@
 # Default values for backbeat.
 # This is a YAML-formatted file.
 # Declare variables to be passed into your templates.
+
+metrics:
```
NicolasT: No need for a `metrics` section, see comment above.

charts/backbeat/values.yaml
```diff
@@ -1,6 +1,14 @@
...
+  enabled: true
```
NicolasT: I see no reason to make this 'optional'.

charts/backbeat/values.yaml
```diff
@@ -1,6 +1,14 @@
...
+  annotations:
+    prometheus.io/scrape: "true"
+    prometheus.io/port: "8000"
```
NicolasT: Does Prometheus support named ports in annotations? If so, that'd be a better approach.

charts/cloudserver-front/templates/service.yaml
```diff
@@ -1,6 +1,10 @@
 apiVersion: v1
 kind: Service
 metadata:
+  annotations:
+{{- if .Values.metrics.enabled }}
```
NicolasT: Same as above.

charts/cloudserver-front/values.yaml
```diff
@@ -2,6 +2,13 @@
 # This is a YAML-formatted file.
 # Declare variables to be passed into your templates.
 
+metrics:
```
NicolasT: Same as above (2x)

charts/zenko/requirements.yaml
```diff
@@ -20,6 +20,10 @@ dependencies:
 - name: redis-ha
   version: "2.1.2"
   repository: "https://kubernetes-charts.storage.googleapis.com/"
+- name: grafana
+  version: "1.11.6"
+  repository: "https://kubernetes-charts.storage.googleapis.com/"
+  condition: grafana.enabled
```
NicolasT: Why make this conditional? Prometheus deployment isn't conditional either, and there's not much use for Prometheus without Grafana.

charts/zenko/templates/datasource.yaml
```diff
@@ -0,0 +1,18 @@
+{{- if .Values.grafana.sidecar.datasources.enabled -}}
```
NicolasT: No need to make this configurable.

charts/zenko/values.yaml
```diff
@@ -95,4 +95,25 @@ redis-ha:
     create: true
   replicas:
     servers: *nodeCount
-    sentinels: *nodeCount
\ No newline at end of file
+    sentinels: *nodeCount
+
+grafana:
+  enabled: false
```
NicolasT: Don't even make it optional.

charts/zenko/values.yaml
```diff
@@ -95,4 +95,25 @@ redis-ha:
...
+  # When 'grafana.enabled' is true, a grafana chart is deployed as part of the Zenko stack
+  ingress:
+    enabled: true
```
NicolasT: No. Don't enable `Ingress`es by default, since there's no way to know the host to use etc.

charts/zenko/values.yaml
```diff
@@ -95,4 +95,25 @@ redis-ha:
...
+    datasources:
+      enabled: true
+      # Every config map with the following label will be used as datasource. By default,
+      # promtheus is set as datasource under zenko/templates/datasources.yaml
+      label: grafana-datasource
+    dashboards:
+      enabled: true
+      # Likewise, every config map with the following label will be used as a source for
+      # a dashboard.
+      label: grafana-dasboard
```
NicolasT: Please add a dashboard with this PR, to make sure this functionality actually works.

charts/zenko/values.yaml
```diff
@@ -95,4 +95,25 @@ redis-ha:
...
+  adminUser: deployment-specific-admin-user
```
NicolasT: Just `admin` should do.

charts/zenko/values.yaml
```diff
@@ -95,4 +95,25 @@ redis-ha:
...
+  adminPassword: deployment-specific-admin-password
```
NicolasT: No. Don't hard-code 'default' passwords. If the Grafana chart doesn't somehow set up some random password in a secret unless overridden, then that should be added upstream. Otherwise, we should simly use that functionality.

charts/backbeat/templates/api/service.yaml
giacomoguiulfo: Do we want to have service level annotations or pod level annotations? What do you think it's best?

charts/zenko/values.yaml
giacomoguiulfo: Yes, it does have support for it.

charts/backbeat/values.yaml
giacomoguiulfo: Looking at many other charts, it doesn't seem that it supports named ports.

giacomoguiulfo:
@NicolasT Made all the requested changes 🙂.
However, I was wondering if it's better to have the ```annotations``` at the pod level or the service level? Right now I have set them at the pod level.

I also added a dashboard for Cloudserver-front. It is not the best looking right now, but it's a starting point, and this proves that the grafana sidecar setup works.

I will add non-destructive e2e tests soon.

charts/backbeat/templates/api/service.yaml
NicolasT: Service-level. Don't do pod-level when there's a service whose endpoints can be resolved.

NicolasT:
Thanks!

Let Prometheus monitor a `Service`, don't set annotations on `Pod`s. Prometheus will resolve the `Service` to its `Endpoints` and monitor them accordingly, which is useful because it makes sure `Pod`s which aren't added to a `Service` (because their `readinessCheck` or `livenessCheck` fails, for example) won't be monitored (and why would they? They won't be accessed by any client going through the `Service` anyway!), which reduces the risk of skewing the metrics.

charts/backbeat/templates/api/deployment.yaml
```diff
@@ -11,6 +11,10 @@ spec:
   replicas: {{ .Values.api.replicaCount }}
   template:
     metadata:
+    {{- if .Values.podAnnotations }}
```
NicolasT: See other comments in the PR ;-)

charts/backbeat/values.yaml
```diff
@@ -16,6 +16,11 @@ logging:
 mongodb:
   replicas: 3
 
+podAnnotations:
```
NicolasT: Same

charts/cloudserver-front/templates/dashboard.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: sample-grafana-dashboard
+  labels:
+     grafana-dashboard: "true"
```
NicolasT: This will need the standard Helm chart labels.

charts/zenko/templates/datasource.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ .Release.Name }}-grafana-datasource
+  labels:
```
NicolasT: Also needs Helm labels

charts/zenko/values.yaml
```diff
@@ -95,4 +95,22 @@ redis-ha:
...
   replicas:
     servers: *nodeCount
-    sentinels: *nodeCount
\ No newline at end of file
+    sentinels: *nodeCount
+
+grafana:
+  ingress:
+    enabled: false
+    hosts:
```
NicolasT: Is this whole `ingress` section required anyway? I'd say no... just use the upstream defaults, no need to override.

charts/zenko/values.yaml
```diff
@@ -95,4 +95,22 @@ redis-ha:
...
+      enabled: true
+      # Every config map with the following label will be used as datasource. By default,
+      # promtheus is set as datasource under zenko/templates/datasources.yaml
+      label: grafana-datasource
+    dashboards:
+      enabled: true
+      # Likewise, every config map with the following label will be used as a source for
+      # a dashboard.
+      label: grafana-dashboard
+  adminUser: admin
```
NicolasT: This is also the default, no need to override

charts/zenko/templates/datasource.yaml
giacomoguiulfo: Done 👍

giacomoguiulfo:
@NicolasT fixed everything and added ```non-destructive``` e2e tests.

charts/cloudserver-front/templates/dashboard.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: sample-grafana-dashboard
```
NicolasT: This needs to be based on the `release` name, like any other object created by a chart.

charts/cloudserver-front/templates/dashboard.yaml
```diff
@@ -0,0 +1,13 @@
...
+  labels:
+    app: {{ template "cloudserver-front.name" . }}
+    chart: {{ template "cloudserver-front.chart" . }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+    grafana-dashboard: "true"
```
NicolasT: Maybe this label should also depend on the release name, otherwise we'll end up with dashboards from different releases shared across various Grafana instances.

charts/zenko/values.yaml
```diff
@@ -95,4 +95,17 @@ redis-ha:
...
+    datasources:
+      enabled: true
+      # Every config map with the following label will be used as datasource. By default,
+      # promtheus is set as datasource under zenko/templates/datasources.yaml
+      label: grafana-datasource
+    dashboards:
+      enabled: true
+      # Likewise, every config map with the following label will be used as a source for
+      # a dashboard.
+      label: grafana-dashboard
```
NicolasT: Guess this will become a tricky one again cfr. comment above... Maybe the `tpl` function can be of use.

tests/zenko_e2e/grafana/test_dashboard.py
```diff
@@ -0,0 +1,17 @@
+import requests
```
NicolasT: Pretty sure some of this code wouldn't pass a `tox` run with linters, but that's currently completely broken anyway *sigh* so never mind I guess.

charts/zenko/values.yaml
giacomoguiulfo: @NicolasT How exactly can this value be templated to include the release name? Grafana doesn't use partials to set this value so it cannot be overwritten through a define in ```_helpers.tpl``` as far as I know.

charts/cloudserver-front/templates/dashboard.yaml
```diff
@@ -0,0 +1,13 @@
...
+  name: {{ template "cloudserver-front.fullname" . }}-grafana-dashboard
+  labels:
+    app: {{ template "cloudserver-front.name" . }}
+    chart: {{ template "cloudserver-front.chart" . }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+    grafana-dashboard: "true"
+data:
+  cloudserver-front.json: |
+{{ .Files.Get "dashboards/cloudserver-front.json" | indent 4 }}
```
NicolasT: Use `.Files.Glob` instead https://github.com/kubernetes/helm/blob/master/docs/chart_template_guide/accessing_files.md#configmap-and-secrets-utility-functions

charts/zenko/values.yaml
NicolasT: Good point. One of the shortcomings of Helm when it comes to chart composition.

charts/cloudserver-front/templates/dashboard.yaml
giacomoguiulfo: Done!

giacomoguiulfo:
@NicolasT Added a test to check if Prometheus is monitoring Cloudserver.

PR #158, opened 2018-06-29
feat: add zenko-filer chart

giacomoguiulfo:
This adds the Zenko-Filer chart to Zenko, which you allows exporting buckets through NFS 🚀.

Zenko-Filer is made up of 3 components:
- Zenko-NFSD: Zenko's NFS server
- Zenko-MD: Zenko Metadata Gateway
- Portmap: RPC Server (rpcbind)

All of these are deployed in a 3 container pod (see ```zenko-filer/templates/deployment.yaml```).

NFS server and exports configuration is managed through a configmap, which reads from a file ```zenko-filer/nfsd.conf``` that can be modified (i.e to add new NFS exports). An upgrade (like ```helm upgrade``` for example), will then restart the NFS server with the updated configuration.

### DIFF
--- a/charts/zenko-filer/templates/deployment.yaml
+++ b/charts/zenko-filer/templates/deployment.yaml
@@ -1,0 +1,122 @@
+apiVersion: apps/v1beta2
+kind: Deployment
+metadata:
+  name: {{ template "zenko-filer.fullname" . }}
+  labels:
+    app: {{ template "zenko-filer.name" . }}
+    chart: {{ template "zenko-filer.chart" . }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+spec:
+  replicas: {{ .Values.replicaCount }}
+  selector:
+    matchLabels:
+      app: {{ template "zenko-filer.name" . }}
+      release: {{ .Release.Name }}
+  template:
+    metadata:
+      annotations:
+        checksum/config: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
+      labels:
+        app: {{ template "zenko-filer.name" . }}
+        release: {{ .Release.Name }}
+    spec:
+      containers:
+        - name: {{ .Values.nfsd.name }}
+          image: "{{ .Values.nfsd.image.repository }}:{{ .Values.nfsd.image.tag }}"
+          imagePullPolicy: {{ .Values.nfsd.image.pullPolicy }}
+          ports:
+            - name: nfs
+              containerPort: 2049
+          volumeMounts:
+            - name: portmap-data
+              mountPath: /run
+            - name: metadata-data
+              mountPath: /conf
+            - name: nfsd-config
+              mountPath: /template
+          env:
+            - name: CONFIGURATION_DIRECTORY
+              value: "/conf"
+            - name: CONFIGURATION_TEMPLATE
+              value: "/template/nfsd.conf"
+            - name: ZENKO_MD_HOST
+              value: "127.0.0.1"
+            - name: ZENKO_MD_PORT
+              value: "9000"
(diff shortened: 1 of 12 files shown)


charts/zenko-filer/values.yaml
```diff
@@ -0,0 +1,46 @@
...
+
+md:
+  name: "zenko-md"
+  image:
+    repository: zenko/zenko-md
+    tag: edge
+    pullPolicy: IfNotPresent
+
+portmap:
+  name: "portmap"
```
jonathan-gramain: `zenko-portmap`?

charts/zenko-filer/templates/deployment.yaml
```diff
@@ -0,0 +1,122 @@
...
+        - name: {{ .Values.nfsd.name }}
+          image: "{{ .Values.nfsd.image.repository }}:{{ .Values.nfsd.image.tag }}"
+          imagePullPolicy: {{ .Values.nfsd.image.pullPolicy }}
+          ports:
+            - name: nfs
+              containerPort: 2049
+          volumeMounts:
+            - name: portmap-data
+              mountPath: /run
+            - name: metadata-data
```
jonathan-gramain: would it be better to name it `metadata-conf`? Not sure just asking

charts/zenko-filer/templates/deployment.yaml
```diff
@@ -0,0 +1,122 @@
...
+            - name: ZENKO_MD_PORT
+              value: "9000"
+            - name: REDIS_SENTINEL_BOOTSTRAP
+              value: "{{- printf "%s-%s:26379" .Release.Name "redis-ha-sentinel" | trunc 63 | trimSuffix "-" -}}"
+            - name: REDIS_CLUSTER_NAME
+              value: "mymaster"
+            - name: NFSD_LOG_FILE
+              value: "/logs/scality-nfsd.log"
+          args: ['scality.nfsd', '-N', 'INFO', '-F', '-f', '/conf/scality-nfsd.conf', '-L', '/dev/stdout']
+          # TODO: livenessProbe:
```
jonathan-gramain: livenessProbe and readinessProbe are defined below, comments seem outdated

charts/zenko-filer/values.yaml
```diff
@@ -0,0 +1,46 @@
...
+# Declare variables to be passed into your templates.
+replicaCount: 1
+
+nfsd:
+  name: "zenko-nfsd"
+  image:
+    repository: zenko/zenko-nfsd
+    tag: edge
+    pullPolicy: IfNotPresent
+  loglevel: INFO
```
jonathan-gramain: shall we use capital letters? I think all other charts use lowercase, and a structure like

```
logging:
    level: info
```

charts/zenko-filer/templates/deployment.yaml
giacomoguiulfo: Probably yes, those were just the names that came to my mind. Do we want to change it now or later on? @jonathan-gramain

charts/zenko-filer/values.yaml
giacomoguiulfo: It would probably require a change in the nfsd code, ```AFAIK``` nfs ganesha log variables are in all-caps. We can still make the division of ```logging``` and ```level```.

charts/zenko-filer/values.yaml
giacomoguiulfo: 👍

charts/zenko-filer/templates/deployment.yaml
giacomoguiulfo: They are defined below for the ```zenko-md``` container there aren't any for the ```nfsd``

PR #160, opened 2018-06-29
bugfix: ZENKO-616 Use local cache for metrics

bennettbuchanan:


bennettbuchanan:
@bert-e wait

charts/backbeat/templates/replication/status_deployment.yaml
```diff
@@ -44,6 +44,10 @@ spec:
               value: "{{- printf "%s-%s" .Release.Name "redis" | trunc 63 | trimSuffix "-" -}}"
```
giacomoguiulfo: Both REDIS and REDIS_LOCALCACHE are pointing to the same Redis Service. Ideally the REDIS one should be pointing to Redis-HA service (if it's even needed). This was a misconfiguration from before and should be fixed.

giacomoguiulfo:
Just a suggestion, but maybe we can set an alias to the redis chart to something like localcache or redis-localcache (similar to what queue and quorum are to Kafka and Zookeeper), in order to be more explicit and coherent. What do you think @rahulreddy ?

rahulreddy:
I agree, we can look as technical debt and fix it at an appropriate time.

PR #170, opened 2018-07-02
DOC: adding template to show Zenko forums in pages

smaffulli:
This template adds the possibility to discuss documentation on
Zenko forum. At the bottom of each docs page an iframe is loaded
showing a prompt to start a new thread in case there is not one
already. If there is already a thread open, the associated
discussion will be loaded.


smaffulli:
Ok, I enabled building this branch on RTD so we can test this more properly. The main thing to be decided is how to handle discussions pages on the forum: each URL will have one topic opened on forum.zenko.io, in the `Docs` category (currently visible only to admins). This means that each version of each page may have one topic assigned. For example, `https://zenko.readthedocs.io/en/latest/root.hml` and `https://zenko.readthedocs.io/en/v1/root.hml` both may have a discussion linked at the bottom (or a chance to create one). Is that what we want? The main drawback of this approach is that there may be many (many) new topics in the Docs category.

Also, if we publish documentation on multiple domains (i.e. https://zenko.readthedocs.io and https://zenko.io/docs) we'll multiply topics even more.

Thoughts?

smaffulli:
After thinking about it, I came to the conclusion that having one thread per page per version is not a bad thing after all. Can this be merged?

PR #171, opened 2018-07-03
ft: ZENKO-645 add additional orbit managment vars

alexanderchan-scality:
allows for testing with dev orbit endpoints

### DIFF
--- a/charts/README.md
+++ b/charts/README.md
@@ -49,8 +49,23 @@ To disable Orbit, use:
 If a custom Orbit endpoint is required, use:
 
 ```shell
---set cloudserver-front.orbit.endpoint=https://dev.zenko.io
+--set cloudserver-front.orbit.endpoint=https://custom.orbit.endpoint
+--set cloudserver-front.orbit.pushEndpoint=https://custom.push.orbit.endpoint
 ```
+
+and if `poll` mode is needed, use:
+```shell
+--set cloudserver-front.orbit.mode=poll
+```
+
+**Note:** `poll` mode should only be used when long-lived websocket connections
+are not available. In `poll` mode:
+
++ `orbit.pushEndpoint` setting will be ignored
++ Interactive features such as Multicloud Browser, Metadata Search, CRR object
+  status and CRR object retry will not be available
++ Configuration updates may take up to a minute to propagate
++ Statistics reports are sent less frequently
 
 Autoscaling
 -----------
--- a/charts/cloudserver-front/templates/deployment.yaml
+++ b/charts/cloudserver-front/templates/deployment.yaml
@@ -77,6 +77,10 @@ spec:
 {{- else }}
             - name: MANAGEMENT_ENDPOINT
               value: "{{- .Values.orbit.endpoint -}}"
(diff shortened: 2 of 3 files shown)


charts/cloudserver-front/templates/deployment.yaml
```diff
@@ -77,6 +77,10 @@ spec:
 {{- else }}
             - name: MANAGEMENT_ENDPOINT
               value: "{{- .Values.orbit.endpoint -}}"
+            - name: MANAGEMENT_MODE
```
ssalaues: Likely these need to be added to the backbeat chart as well but otherwise LGTM

charts/README.md
```diff
@@ -49,7 +49,7 @@ To disable Orbit, use:
 If a custom Orbit endpoint is required, use:
 
 ```shell
---set cloudserver-front.orbit.endpoint=https://dev.zenko.io
+--set cloudserver-front.orbit.endpoint=https://custom.orbit.endpoint
```
RachedBenMustapha: If a custom endpoint is provided then a custom push endpoint must be used as well so the example should reflect that. Chances are that the push endpoint will not have access to the same data as the custom endpoint.

charts/cloudserver-front/templates/deployment.yaml
RachedBenMustapha: @ssalaues you are right about `PUSH_ENDPOINT`. `MANAGEMENT_MODE` is not used by backbeat (and in the near future will not be used by cloudserver either)

alexanderchan-scality: I believe the recent changes to backbeat has it get the configurations from mongo, so it doesn't require the management/push endpoints.

charts/README.md
```diff
@@ -49,9 +49,19 @@ To disable Orbit, use:
...
 ```
 
+and if `poll` mode is needed, use:
+```shell
+--set cloudserver-front.orbit.mode=poll
+```
+
+**Note:** In `poll` mode, Cloudserver will request changes from Orbit instead
+of receiving updates when one become available, so the `orbit.pushEndpoint`
+value will be ignored.
```
RachedBenMustapha: I wasn't going to expose this setting in the docs but since we are, a couple of things are worth mentioning as well. In poll mode:
- interactive features such as Multicloud Browser, Metadata Search, CRR object status and CRR object retry are not available
- configuration updates may take up to a minute to propagate
- statistics reports are sent less frequently

Poll mode should be used when long-lived websocket connections are not available.

charts/README.md
```diff
@@ -49,9 +49,24 @@ To disable Orbit, use:
...
+```shell
+--set cloudserver-front.orbit.mode=poll
+```
+
+**Note:** `poll` mode should only be used when long-lived websocket connections
+are not available. In `poll` mode:
+
++ `orbit.pushEndpoint` setting will be ignored
++ Interactive features such as Multicloud Browser, Metadata Search, CRR object
+  status and CRR object retry will not available
```
jonathan-gramain: minor typo "will not be available"

PR #173, opened 2018-07-05
feat: add support for https proxy certificate

giacomoguiulfo:


charts/cloudserver-front/templates/deployment.yaml
```diff
@@ -57,6 +57,8 @@ spec:
               value: "{{ .Values.proxy.https }}"
             - name: HTTPS_PROXY
               value: "{{ .Values.proxy.https }}"
+            - name: HTTPS_PROXY_CERTIFICATE
+              value: "/ssl/ca.crt"
```
rahulreddy: Can this be `{{ .Values.proxy.https_cert }}` instead of hardcoding the path?

ssalaues: This is where the cert gets mounted inside of the container. The file is generated from the secret that always gets mounted into /ssl so it needs to be hard coded.

PR #174, opened 2018-07-06
Feature/zenko 596 add redis ha option

ssalaues:


charts/backbeat/templates/replication/status_deployment.yaml
```diff
@@ -41,7 +41,7 @@ spec:
             - name: EXTENSIONS_REPLICATION_SOURCE_S3_PORT
               value: "80"
             - name: REDIS_HOST
-              value: "{{- printf "%s-%s" .Release.Name "redis" | trunc 63 | trimSuffix "-" -}}"
+              value: "{{- printf "%s-%s" .Release.Name "redis-ha-master-svc" | trunc 63 | trimSuffix "-" -}}"
```
rahulreddy: Please don't change Backbeat, this breaks compatibility. Backbeat only writes to LocalCache atm.

ssalaues: I added this because all of the other backbeat services are configured to use redis-ha.  I added this to make it uniform and prevent issues. If they all should be on localcache then we should make that change as well.

ssalaues: I just saw that #160 switched all the backbeat redis to use the local cache (my branch was out of date). I'll drop this commit.

PR #176, opened 2018-07-09
docs: Add info on upgrading Zenko

ssalaues:


charts/README.md
```diff
@@ -121,6 +121,33 @@ $ helm test $RELEASE_NAME --cleanup
...
+-------------------------
+
+To upgrade an existing Zenko installation to the latest version or simply change deployment
+configuration, you will need to run the following commands. This assumes that your release
+name is `zenko` and that you have previously cloned the repository. 
+
+```shell
+$ cd Zenko/charts
+$ git pull
+$ helm dep build ./zenko
```
giacomoguiulfo: @ssalaues Maybe use `dependency` instead of `dep` to be more explicit of what the command does?

ssalaues:
Made it more explicit by changing to `dependency` over the original `dep`

PR #179, opened 2018-07-10
ft: add options for storage limits

ssalaues:


giacomoguiulfo:
It's probably not necessary, but should we bump `S3-data` image as well to keep them both up to date?

PR #180, opened 2018-07-11
bug(zenko-696): Fix CRR Tests

tmacro:


tests/create_buckets.py
```diff
@@ -0,0 +1,96 @@
...
+        raise exp
+
+
+TIMEOUT = 120  # 2 Minutes
+
+K8S_NAMESPACE = os.getenv('ZENKO_K8S_NAMESPACE')
+ZENKO_HELM_RELEASE = os.getenv('ZENKO_HELM_RELEASE')
+ZENKO_ENDPOINT = get_env('CLOUDSERVER_FRONT_ENDPOINT',
+                         'http://%s-cloudserver-front:80' % ZENKO_HELM_RELEASE)
+VERIFY_CERTIFICATES = get_env('VERIFY_CERTIFICATES', False)
```
rahulreddy: Why do we need to set this to false and especially since s3 client is talking to zenko over http?

tests/create_buckets.py
```diff
@@ -0,0 +1,96 @@
...
+created = False
+while time.time() < timeout:
+    try:
+        for bucket in buckets:
+            bucket_safe_create(s3client.Bucket(bucket))
+        created = True
+        break
+    except Exception:
+        print('Failed to create buckets, backing off and trying again')
+        time.sleep(1)
```
jonathan-gramain: I guess you meant `time.sleep(backoff)`?

tests/create_buckets.py
```diff
@@ -0,0 +1,96 @@
...
+import os
+
+
+def get_env(key, default=None, error=False):
+    if not error:
+        return os.environ.get(key, default)
+    return os.environ[key]
+
+
+def bucket_safe_create(bucket):
```
JianqinWang: IIRC, this function is defined here already: https://github.com/scality/Zenko/blob/5a3318738a22752b1453fab0e38e8458d52883f1/tests/zenko_e2e/util/bucket.py#L11

My python is rusty, is there a reason why this is being redefined instead of being imported?

tmacro: create_buckets is just a one off script, while zenko_e2e is a separate project. I'd prefer no to entangle the two over saving a few lines.

tests/create_buckets.py
tmacro: For configuration Right now we're using http, however that could change in the future. Mostly the code for loading config was copy/pasted from the e2e tests, kinda a "if it ain't broke, don't fix it' situation

jonathan-gramain:
@bert-e wait

tests/create_buckets.py
rahulreddy: I mean ideally we should not have unused code

eve/main.yml
```diff
@@ -37,6 +37,8 @@ models:
       GCP_CRR_SRC_BUCKET_NAME: ci-zenko-gcp-crr-src-bucket
       GCP_CRR_MPU_BUCKET_NAME: ci-zenko-gcp-crr-mpu-bucket
       MULTI_CRR_SRC_BUCKET: ci-zenko-multi-crr-src-bucket
+      ZENKO_ACCESS_KEY: HEYIMAACCESSKEY
```
giacomoguiulfo: `HEYIMANACCESSKEY`

tmacro: incorrect grammar yes, typo no

tests/create_buckets.py
```diff
@@ -0,0 +1,96 @@
...
+            bucket_safe_create(s3client.Bucket(bucket))
+        created = True
+        break
+    except Exception:
+        print('Failed to create buckets, backing off and trying again')
+        time.sleep(backoff)
+        backoff = backoff ** 2
+
+if not created:
+    print('Failed to create buckets and TIMEOUT has elasped!')
```
bennettbuchanan: Could be nice to use a template string with the timeout value.

tests/create_buckets.py
```diff
@@ -0,0 +1,96 @@
...
+        for container in i.status.container_statuses:
+            if not container.ready:
+                print('%s is not ready' % container.name)
+                passed = False
+                break
+        if not passed:
+            break
+    if not passed:
+        time.sleep(1)
+        passed = True
```
bennettbuchanan: Won't this be reset similarly above?

tests/create_buckets.py
```diff
@@ -0,0 +1,96 @@
...
+    for i in ret.items:
+        for container in i.status.container_statuses:
+            if not container.ready:
+                print('%s is not ready' % container.name)
+                passed = False
+                break
+        if not passed:
+            break
+    if not passed:
+        time.sleep(1)
```
bennettbuchanan: No exponential backoff for this one?

tmacro: I see this more as a poll than something that needs to be retried and backed off

PR #182, opened 2018-07-11
bug(zenko-699): Fix Lifecycle Test

tmacro:


tests/local-values.yml
```diff
@@ -0,0 +1,75 @@
+nodeCount: &nodeCount 1
```
ssalaues: Is this for local testing?

tests/local-values.yml
```diff
@@ -0,0 +1,75 @@
...
+    name: default
+  replicas:
+    servers: *nodeCount
+    sentinels: *nodeCount
+
+cloudserver-front:
+  image:
+    pullPolicy: Always
+  orbit:
+    endpoint: "http://ciutil-orbit-simulator:4222"
```
ssalaues: If this is for local testing do we want orbit-simulator references?

tests/local-values.yml
tmacro: Yes

tests/local-values.yml
tmacro: orbit sim still runs for local test runs

tests/local-values.yml
ssalaues: Just posting here the discussion Taylor and I had offline:
The concern was fundamentally changing an existing cluster when running this make command but it seems like it already previously had this intended affect. It's out of the scope of this PR but we should look into changing up the Makefile to better suit our needs.

PR #185, opened 2018-07-12
feat(zenko-702) Add MPU to CRR

tmacro:


rahulreddy:
Have you established that put_object does mpus automatically beyond some certain threshold of object size?

PR #186, opened 2018-07-12
bf: allow optional proxy certs

ssalaues:
This allows for the proxy certificates to be passed if necessary.

### DIFF
--- a/charts/cloudserver-front/values.yaml
+++ b/charts/cloudserver-front/values.yaml
@@ -39,10 +39,12 @@ proxy:
   # If you want to use an HTTP proxy, add the respective endpoints after
   # 'http:' and/or 'https:'. If the HTTP proxy endpoint is set but the HTTPS
   # one isn't, the HTTP proxy will be used for HTTPS traffic as well.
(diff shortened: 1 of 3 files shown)


charts/cloudserver-front/values.yaml
```diff
@@ -39,10 +39,12 @@ proxy:
   # If you want to use an HTTP proxy, add the respective endpoints after
   # 'http:' and/or 'https:'. If the HTTP proxy endpoint is set but the HTTPS
   # one isn't, the HTTP proxy will be used for HTTPS traffic as well.
-  # Additionally, if the HTTPS proxy is set, you will need to provide an
-  # ssl certificate in a file named 'ca.crt' at the root path of this chart.
+  # Additionally you can pass a CA certifcate that will be added to the trusted
+  # certs. If the HTTPS proxy is configured and caCert is true, helm
```
rahulreddy: Let's remove `HTTPS proxy` in favor of `proxy` as CA cert would have to be specified HTTP->HTTPS setups.

PR #188, opened 2018-07-16
chore: rename zenko-filer to zenko-nfs

giacomoguiulfo:


ssalaues:
Merging as the CI doesn't test for Zenko-nfs and the install stage was successful. (Tests were failing due to CI issues that have been merged)

PR #190, opened 2018-07-17
ZENKO-667 feat: side band protocol ports

ploki:


ploki:
Dear reviewer, 4077034 is part of another PR #188

ssalaues:
Merged in #188 could you rebase please

charts/zenko-nfs/values.yaml
```diff
@@ -5,6 +5,10 @@ replicaCount: 1
 
 service:
   type: NodePort
+  extra:
+    rquotad: 875
```
giacope: @ploki Should these ports be configurable?

ploki: Not particularly. Only in case of unexpected port conflicts

ssalaues:
@ploki I noticed there are two [squashme] commits. Can you squash these before we merge?

charts/zenko-nfs/values.yaml
giacope: Then they shouldn't be configurable.

ploki: They should be configurable.
Can you tell me If you see a reason for these properties to not be configurable?

giacope: There is no need to make them configurable, and it helps to avoid possible configuration errors. The only scenario in which there could be port conflicts is at the pod level (containers within a pod with the same port), but that will be a conflict caused by us.

Besides that, this is the way we have being doing it in Zenko charts. PR's in the past have been requested changes like this one.

ploki:
@bert-e reset

ploki:
@ssalaues I squash the branch and rebased it on `development/0.9`

@giacomoguiulfo I started to remove the template parameters but was not really happy to replace the parameters by literal numbers in several places in the yaml files. Is it okay to consider these values as constants in the `values.yaml`?

giacope:
@ploki Quoting NicolasT when we last discussed this issue:
`I still fail to see why we'd want to make the port of our Services 'configurable', unless we retrieve them from the environment or DNS in other software connecting to them.`

My main concern is that we should stay with one way of doing it. It's doesn't look nice to have some ports configurable and some don't for no reason.

giacope:
I won't block for the above though 🙂. However, has this been tested and is it possible to do a NFSv3 mount from outside the Kubernetes Cluster by exposing these ports?

ploki:
@giacomoguiulfo Where do we define constants in the bunch files?

giacope:
@ploki IMHO, they should go in the templates or as partials (`_hepler.tpl`). Ideally, `values.yaml` should only expose values that you expect the user to change for x and y reasons.

ploki:
@bert-e reset wait

PR #194, opened 2018-07-18
rf: ZENKO-758 rename replication pods

jonathan-gramain:
Add "replication" to pod names to avoid confusions, and change
"producer" and "consumer" to more meaningful names.

New pod names:

- backbeat-producer -> backbeat-replication-populator
- backbeat-consumer -> backbeat-replication-data-processor
- backbeat-status -> backbeat-replication-status-processor

If there's a risk it does not fit anymore in DNS name limit (63 char?)
we can consider shortening names, but ideally (to me) they would be as
self-descriptive as they can get.

### DIFF
--- a/charts/backbeat/templates/replication/data_processor_deployment.yaml
+++ b/charts/backbeat/templates/replication/data_processor_deployment.yaml
@@ -1,14 +1,14 @@
 apiVersion: extensions/v1beta1
 kind: Deployment
 metadata:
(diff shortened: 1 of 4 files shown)


jonathan-gramain:
Maybe the pod instance hash also counts in the name length? But anyway it should still leave some legroom.

ssalaues:
@jonathan-gramain Typically just the service name is used for communication which doesn't have the hashes but yeah even still gives us some room.

PR #195, opened 2018-07-18
feature: ZENKO-367 CRR to AWS end-to-end tests

bennettbuchanan:
This PR adds CRR to AWS replication end-to-end tests. For now, I've skipped tests which were previously passing but are not currently in Zenko—these skipped tests require further investigation.

### DIFF
--- a/tests/zenko_e2e/backbeat/s3SDK.js
+++ b/tests/zenko_e2e/backbeat/s3SDK.js
@@ -1,0 +1,36 @@
+const S3 = require('aws-sdk').S3;
+
+const scalityS3Client = new S3({
+    accessKeyId: process.env.ZENKO_BACKBEAT_ACCESS_KEY,
+    secretAccessKey: process.env.ZENKO_BACKBEAT_SECRET_KEY,
+    sslEnabled: false,
+    endpoint: 'http://zenko.local',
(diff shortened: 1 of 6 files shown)


bennettbuchanan:
@bert-e wait

ssalaues:
Can you rebase off master? We merged in some fixes to the CI this morning

bennettbuchanan:
@ssalaues I've rebased to master, thanks!

tests/zenko_e2e/backbeat/s3SDK.js
```diff
@@ -0,0 +1,36 @@
+const S3 = require('aws-sdk').S3;
+
+const scalityS3Client = new S3({
+    accessKeyId: process.env.ZENKO_BACKBEAT_ACCESS_KEY,
+    secretAccessKey: process.env.ZENKO_BACKBEAT_SECRET_KEY,
+    sslEnabled: false,
+    endpoint: 'http://127.0.0.1:8000',
```
ssalaues: Do we assume that, when the tests are ran, it's expected to have 8000 port-forwarded to localhost? Why not use `zenko.local` ?

ssalaues: Either way is okay really but might be good to have a note of it in the doc.

PR #196, opened 2018-07-19
feat(zenko-739): Add Transient source test

tmacro:


tests/zenko_e2e/cloudserver/test_transient_src.py
```diff
@@ -0,0 +1,12 @@
...
+from ..fixtures import *
+
+
+def test_transient_src(transient_src_bucket, transient_target_bucket,
+                       testfile, objkey):
+    util.mark_test('TRANSIENT SOURCE')
+    util.upload_object(transient_src_bucket, objkey, testfile)
+    assert util.check_object(
+        objkey, testfile, transient_target_bucket, timeout=30)
+    assert not util.get_from_preferred_read(conf.TRANSIENT_SRC_BUCKET,
```
jonathan-gramain: Not 100% sure I get the logic right, but it seems we should instead loop over this call until it fails, because it shall succeed until the garbage collector actually removes the original data. It should be quick after the target replication is complete, but not immediate, so not doing this loop and expecting an immediate failure may introduce flakiness I think.

PR #199, opened 2018-07-19
feature: ZENKO-367 CRR to Azure end-to-end tests

bennettbuchanan:
This PR adds CRR to Azure replication end-to-end tests. I've skipped tests which were previously passing but are not currently in Zenko—these skipped tests require further investigation.


tests/zenko_e2e/backbeat/Using.md
```diff
@@ -32,9 +32,14 @@ export ZENKO_BACKBEAT_SECRET_KEY=<secret-key>
 ```
 export AWS_S3_BACKBEAT_ACCESS_KEY=<access-key>
 export AWS_S3_BACKBEAT_SECRET_KEY=<secret-key>
-export AWS_S3_BACKBEAT_BUCKET_NAME=<destination-bucket-name>
-export AWS_DESTINATION_LOCATION=<destination-location-name>
-export AWS_SOURCE_LOCATION=<source-location-name>
+export AWS_S3_BACKBEAT_BUCKET_NAME=<destination-aws-bucket-name>
```
ssalaues: instead of having all the exports examples here, the step can be to edit the example file (which can be updated with the examples) and `source .secrets.env`

https://github.com/scality/Zenko/blob/master/tests/.secrets.env.example
https://github.com/scality/Zenko/blob/master/tests/.env

bennettbuchanan: Good idea—this will help us align with pre-existing tests and help eventually with automation. I've updated to use any pre-existing env variables and updated those files accordingly.

tests/zenko_e2e/backbeat/Using.md
```diff
@@ -10,34 +10,44 @@
...
+export AZURE_BACKEND_ACCOUNT_NAME=<azure-account-name>
+export AZURE_BACKEND_ACCESS_KEY=<access-key>
+export AZURE_BACKEND_ENDPOINT=<endpoint>
+export AZURE_BACKBEAT_CONTAINER_NAME=<destination-azure-container-name>
+export AZURE_BACKEND_DESTINATION_LOCATION=<destination-azure-location>
 ```
 
 5. Run the test suite: `npm run test_crr`.
+7. If using `*.env` files, source the files:
+   `source .secrets.env && source .secrets.env`.
```
ssalaues: Why do we need to source the secrets twice?

bennettbuchanan: You don't need to, of course. Since the guide we update the file in two different steps in the guide, I figured it would help ensure that it's not forgotten.

tests/zenko_e2e/backbeat/Using.md
```diff
@@ -10,34 +10,44 @@
    ```
 
 2. Create an account using Orbit.
-3. Export the access key and secret key of that account:
+3. Export the access key and secret key of that account (for example, in
```
ssalaues: I think 3 and 7 can easily be combined into one step.  Something like `Secrets can be added to the secret.env file and sourced into your test environment`

tests/zenko_e2e/backbeat/Using.md
```diff
@@ -10,34 +10,44 @@
...
+7. If using `.secrets.env`, source the file: `source .secrets.env`.
 
 ### Tests for CRR to AWS:
 
 1. Create a bucket on AWS `<destination-bucket-name>` with versioning enabled.
 2. In Orbit, create an AWS location `<destination-location-name>` with an AWS
    `<destination-bucket-name>`.
 3. In Orbit, create an AWS location `<source-location-name>`.
-4. Export the access key, secret key, AWS bucket name, and AWS location:
+4. Export the access key, secret key, bucket and container names, and storage
```
ssalaues: Same here with 4 and 6  can be easily combined.

tests/zenko_e2e/backbeat/Using.md
ssalaues: I meant more so why `source .secrets.env && source .secrets.env` vs `source .secrets.env`

tests/zenko_e2e/backbeat/Using.md
```diff
@@ -47,10 +57,10 @@ export AWS_SOURCE_LOCATION=<source-location-name>
 3. Export the access key, secret key, AWS bucket name, and AWS location:
```
ssalaues: For consistency, I would suggest adding the `source` instructions in this section as well. I agree with you, it would help ensure that it's not forgotten.

tests/zenko_e2e/backbeat/Using.md
bennettbuchanan: Okay, I've moved them as two sequential steps. I've kept them as distinct numbers though to maintain the indentation.

tests/zenko_e2e/backbeat/Using.md
bennettbuchanan: Thanks! Updated.

PR #201, opened 2018-07-19
feat: add functional test stage

giacomoguiulfo:


charts/cloudserver/values.yaml
```diff
@@ -53,6 +53,8 @@ proxy:
   https: ""
   caCert: false
 
+env: {}
```
ssalaues: See line 28

charts/cloudserver/templates/manager-deployment.yaml
```diff
@@ -39,10 +39,10 @@ spec:
               value: {{ $value | quote }}
             {{- end }}
           args: ['npm', 'run', 'start_s3server']
-          livenessProbe:
```
tcarmet, line 42: You guys are still having the same issue (pod getting killed by kubernetes) to make it work I disabled the liveness probe on the deployment:
https://github.com/scality/Zenko/blob/13c3e4adbc7ce0db367e7e5f05b9985f66b784c5/charts/cloudserver/templates/deployment.yaml#L45-L48
and not on the `manager-deployment`

Don't know if you guys tried it yet, just dropping the info here :)

eve/main.yml
```diff
@@ -166,6 +166,29 @@ stages:
...
+        env:
+          <<: *global_env
+    - ShellCommand:
+        name: Wait for cloudserver
+        command: sleep 30
+    - ShellCommandWithSecrets:
+        name: Zenko Functional Tests
+        command: |-
+          CLOUDSERVER_POD=$(kubectl get pods --no-headers=true -o custom-columns=:metadata.name | grep cloudserver-manager)
+          kubectl exec $CLOUDSERVER_POD -- bash -c "npm run install_ft_deps"
```
tcarmet: There's no way you guys can apply the same "workflow" that you use for other tests?
```
building a test image -> push to registry -> run with proper commands to test stuff. 
```
I'm curious to know if this is what is causing the liveness probe to fail. (Note that I might be super wrong here as I don't know much about cloudserver :) )

eve/main.yml
```diff
@@ -166,6 +166,29 @@ stages:
         env:
           <<: *global_env
     - ShellCommandWithSecrets: *test_zenko
+    - ShellCommandWithSecrets:
```
tcarmet: When rebasing your branch, note that `ShellCommandWithSecrets` will be deprecated in future eve releases. Just use `ShellCommand` instead.

eve/main.yml
giacope: Yes, that'll be better 🙂

charts/cloudserver/templates/manager-deployment.yaml
giacope: Interesting... I don't see why the non managers will be involved since they are not getting requests at all AFAIK. But these setup needs to be completely refactored. Thanks 👍

giacope:
Closing due to inactivity and because this is very outdated.

PR #189, opened 2018-07-16
rf: ZENKO-744 rename lifecycle producer and consumer

jonathan-gramain:
Rename:

- producer -> bucketProcessor
- consumer -> objectProcessor

Note: waiting for backbeat image update to bump version number.

PR #202, opened 2018-07-20
ft: ZENKO-781 scale out backbeat services

jonathan-gramain:
- Modify number of kafka partitions created to nodeCount

- Set all scalable backbeat components replica count to nodeCount

### DIFF
--- a/charts/single-node-values.yml
+++ b/charts/single-node-values.yml
@@ -10,6 +10,19 @@ cloudserver:
     replicas: *nodeCount
 
 backbeat:
+  replication:
+    dataProcessor:
+      replicaCount: *nodeCount
+    statusProcessor:
+      replicaCount: *nodeCount
+  lifecycle:
+    bucketProcessor:
+      replicaCount: *nodeCount
+    objectProcessor:
+      replicaCount: *nodeCount
+  garbageCollector:
+    consumer:
+      replicaCount: *nodeCount
   mongodb:
     replicas: *nodeCount
 
--- a/charts/zenko/values.yaml
+++ b/charts/zenko/values.yaml
@@ -37,6 +37,19 @@ cloudserver:
     secretKey: deployment-specific-secret-key
 
 backbeat:
+  replication:
+    dataProcessor:
+      replicaCount: *nodeCount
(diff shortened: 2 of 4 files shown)


charts/backbeat/templates/gc/deployment.yaml
```diff
@@ -40,10 +40,12 @@ spec:
             httpGet:
               path: {{ .Values.health.path.liveness}}
               port: {{ .Values.health.port }}
+            initialDelaySeconds: 10
```
giacope: Since most of the backbeat deployments share this value, maybe it should be specified in the `values.yaml` file. Then it could be templated with something like:
```yaml
        	livenessProbe:
{{ toYaml .Values.livenessProbe | indent 12 }}
        	readinessProbe:
{{ toYaml .Values.readinessProbe | indent 12 }}
```

jonathan-gramain: For now I keep those changes as an experiment to see how it goes on the CI, but if it helps I'll do this change!

charts/backbeat/templates/gc/deployment.yaml
```diff
@@ -40,10 +40,12 @@ spec:
             httpGet:
               path: {{ .Values.health.path.liveness}}
               port: {{ .Values.health.port }}
+            initialDelaySeconds: 30
```
ssalaues: I think the `periodDelay` of something like 10-15 would be very useful in this PR as well since the dependency services (kafka and zookeeper) usually take the longest to come up. This way the exponential `CrashLoopBackoff` doesn't cause the pods to quickly build up a long retry period.

jonathan-gramain:
depends on https://github.com/scality/Zenko/pull/189 to be merged first, because of the lack of default values for the CI otherwise.

charts/backbeat/templates/gc/deployment.yaml
jonathan-gramain: Since the problem in this PR does not seem related to this initial delay but to conflicts with other pending PRs, let's keep those useful hints aside for a follow-up PR.

PR #204, opened 2018-07-20
ZENKO-737: CRR Pause Resume e2e tests - AWS

JianqinWang:


tests/zenko_e2e/backbeat/ReplicationUtility.js
```diff
@@ -561,6 +563,57 @@ class ReplicationUtility {
             return cb();
         });
     }
+    
+    assertNoBlob(containerName, key, cb) {
+        this.azure.getBlobProperties(containerName, key, err => {
+            assert(err.code);
```
bennettbuchanan: Could be worthwhile to assert the error code `BlobNotFound`.

tests/zenko_e2e/backbeat/ReplicationUtility.js
```diff
@@ -561,6 +563,57 @@ class ReplicationUtility {
...
+            return cb();
+        });
+    }
+
+    checkReplicationStatus(locationName, cb) {
```
bennettbuchanan: I like this approach of using helper methods for the test suite. I think this warrants its own utility class since it's querying the backbeat API routes. Perhaps introduce something like `BackbeatAPIUtility`.

tests/zenko_e2e/backbeat/ReplicationUtility.js
```diff
@@ -561,6 +563,57 @@ class ReplicationUtility {
...
+        const hostname = 'zenko.local';
+        let path = '/_/backbeat/api/crr/status';
+        if (locationName) {
+            path = `${path}/${locationName}`;
+        }
+        makeGETRequest(path, (err, res) => {
+            getResponseBody(res, (err, body) => {
```
bennettbuchanan: How come no error checking for the use of `makeGETRequest` and `getResponseBody`?

tests/zenko_e2e/backbeat/ReplicationUtility.js
```diff
@@ -561,6 +563,57 @@ class ReplicationUtility {
...
+    pauseReplication(locationName, cb) {
+        let path = '/_/backbeat/api/crr/pause';
+        if (locationName) {
+            path = `${path}/${locationName}`;
+        }
+        makePOSTRequest(path, '{}', cb);
+    }
+
+    resumeReplication(locationName, schedule, hoursScheduled, cb) {
+        const host = 'zenko.local';
```
dora-korpar: this is used a couple of times in different functions, could be pulled out and made global like `src` & `destLocation`

bennettbuchanan: `host` isn't used. We need to introduce a linter soon! 👌

JianqinWang: this actually is not longer needed here, ty for noticing

tests/zenko_e2e/backbeat/ReplicationUtility.js
```diff
@@ -561,6 +563,57 @@ class ReplicationUtility {
...
+    
+    assertNoBlob(containerName, key, cb) {
+        this.azure.getBlobProperties(containerName, key, err => {
+            assert(err.code);
+            return cb();
+        });
+    }
+
+    checkReplicationStatus(locationName, cb) {
+        const hostname = 'zenko.local';
```
bennettbuchanan: `hostname` isn't used. We need to introduce a linter soon! 👌

dora-korpar: oh, not needed here either then :)

tests/zenko_e2e/backbeat/tests/crr-pause-resume/awsBackend.js
```diff
@@ -0,0 +1,107 @@
...
+const awsUtils = new ReplicationUtility(awsS3Client);
+const srcBucket = `source-bucket-${Date.now()}`;
+const destBucket = process.env.AWS_S3_BACKBEAT_BUCKET_NAME;
+const destLocation = process.env.AWS_DESTINATION_LOCATION;
+const hex = crypto.createHash('md5')
+    .update(Math.random().toString())
+    .digest('hex');
+const keyPrefix = `${srcBucket}/${hex}`;
+const key = `${keyPrefix}/object-to-replicate-${Date.now()}`;
+const copyKey = `${key}-copy`;
```
bennettbuchanan: No need for the `copyKey` or `keyutf8`.

tests/zenko_e2e/backbeat/tests/crr-pause-resume/awsBackend.js
```diff
@@ -0,0 +1,107 @@
...
+        next => scalityUtils.deleteVersionedBucket(srcBucket, next),
+        next => awsUtils.deleteAllVersions(destBucket,
+            `${srcBucket}/${keyPrefix}`, next),
+    ], done));
+
+    after(done => scalityUtils.resumeReplication(null, null, null, done));
+
+    it('should pause and resume replication', done => series([
+        next => scalityUtils.putObject(srcBucket, key, Buffer.alloc(1), next),
+        next => scalityUtils.compareObjectsAWS(srcBucket, destBucket, key,
```
bennettbuchanan: I think we can just wait for replication to complete—i.e. no need to get and compare the objects since that is being tested elsewhere. There is a helper method for this `waitUntilReplicated`. Probably the same applies for tests below that are comparing the objects.

tests/zenko_e2e/backbeat/tests/crr-pause-resume/awsBackend.js
```diff
@@ -0,0 +1,107 @@
...
+        next => scalityUtils.checkReplicationStatus(null, (err, data) => {
+            assert(data[destLocation]);
+            assert.strictEqual(data[destLocation], 'disabled');
+            return next();
+        }),
+    ], done));
+
+    it('should retrieve resume status', done => series([
+        next => scalityUtils.pauseReplication(null, next),
+        next => scalityUtils.checkReplicationStatus(null, (err, data) => {
```
bennettbuchanan: I think it's worthwhile to check for the error to help determine a failure that's unrelated to the assertions.

tests/zenko_e2e/backbeat/tests/crr-pause-resume/awsBackend.js
JianqinWang: keyutf8 removed, copyKey is used

tests/zenko_e2e/backbeat/ReplicationUtility.js
JianqinWang: Removed this function as not relevant to AWS

tests/zenko_e2e/backbeat/ReplicationUtility.js
```diff
@@ -561,6 +563,48 @@ class ReplicationUtility {
             return cb();
         });
     }
+
+    checkReplicationStatus(locationName, cb) {
```
bennettbuchanan: Maybe `check` methods should be renamed to something like `getReplicationStatus` since no assertions are happening in the methods. Alternatively, I think it would be good to add the assertions in the helper methods—that way it condenses the test cases. 😸

tests/zenko_e2e/backbeat/ReplicationUtility.js
JianqinWang: fixed

tests/zenko_e2e/backbeat/BackbeatAPIUtility.js
```diff
@@ -0,0 +1,55 @@
+const assert = require('assert');
+const { makeGETRequest, makePOSTRequest, getResponseBody } =
+    require('./utils/request');
+
+class BackbeatAPIUtility {
+    constructor() {}
```
bennettbuchanan: If I'm not mistaken, we can omit the constructor if it's not used.

tests/zenko_e2e/backbeat/BackbeatAPIUtility.js
```diff
@@ -0,0 +1,55 @@
...
+
+    getReplicationStatus(locationName, cb) {
+        let path = '/_/backbeat/api/crr/status';
+        if (locationName) {
+            path = `${path}/${locationName}`;
+        }
+        makeGETRequest(path, (err, res) => {
+            assert.ifError(err);
```
bennettbuchanan: Since we are asserting the error in the `it` statement, maybe it's cleaner to return the errors from the helper. When asserting in both places, the test suite assertion isn't needed. Also, then you can just pass in the `cb` to `getResponseBody`. 😄

tests/zenko_e2e/backbeat/BackbeatAPIUtility.js
```diff
@@ -0,0 +1,55 @@
...
+            getResponseBody(res, (err, body) => {
+                assert.ifError(err);
+                return cb(null, body);
+            });
+        });
+    }
+
+    pauseReplication(locationName, cb) {
```
bennettbuchanan: Maybe we want to add a test case which uses the path `${path}/${locationName}`. I don't see a test case in which `pauseReplication` is being called with a particular location name.

tests/zenko_e2e/backbeat/BackbeatAPIUtility.js
JianqinWang: Will remove this and test it x)

tests/zenko_e2e/backbeat/BackbeatAPIUtility.js
JianqinWang: I see, makes sense, will fix asap

tests/zenko_e2e/backbeat/BackbeatAPIUtility.js
JianqinWang: I was thinking I could do that part when I setup 1-multiple  CRR testing :')

tests/zenko_e2e/backbeat/tests/crr-pause-resume/awsBackend.js
```diff
@@ -0,0 +1,116 @@
...
+        next => backbeatAPIUtils.pauseReplication(null, next),
+        next => backbeatAPIUtils.getReplicationStatus(null, (err, data) => {
+            assert.ifError(err);
+            assert(data[destLocation]);
+            assert.strictEqual(data[destLocation], 'disabled');
+            return next();
+        }),
+        next => scalityUtils.putObject(srcBucket, key2, Buffer.alloc(1),
+            next),
+        next => awsUtils.assertNoObject(destBucket, key2, next),
```
jonathan-gramain: Do we wait some amount of time between the PUT on source and this call? I think it would improve the chances to fail if replication is not paused correctly. Probably 15s or less is enough as we don't want to increase the test run time too much.

tests/zenko_e2e/backbeat/tests/crr-pause-resume/awsBackend.js
```diff
@@ -0,0 +1,116 @@
...
+            assert.ifError(err);
+            assert(data[destLocation]);
+            assert.strictEqual(data[destLocation], 'enabled');
+            return next();
+        }),
+        next => scalityUtils.waitUntilReplicated(srcBucket, key2,
+            undefined, next),
+    ], done));
+
+    it('should retrieve pause status', done => series([
```
jonathan-gramain: Not sure this test has value as we do the same check in the above test, and also in the following test, better remove IMHO.

JianqinWang:
need to do some quick fixes :') sorry for the messy commits

rahulreddy:
Is this a bad rebase?

tests/zenko_e2e/backbeat/tests/crr-pause-resume/awsBackend.js
```diff
@@ -0,0 +1,116 @@
...
+        next => backbeatAPIUtils.pauseReplication(null, next),
+        next => backbeatAPIUtils.getReplicationStatus(null, (err, data) => {
+            assert.ifError(err);
+            assert(data[destLocation]);
+            assert.strictEqual(data[destLocation], 'disabled');
+            return next();
+        }),
+    ], done));
+
+    it('should retrieve resume status', done => series([
```
jonathan-gramain: Same, it looks like it's duplicating the first test which is more complete. I would opt to remove it.

JianqinWang: I was hoping that life would be easier if I had to remove tests instead of make up tests :sweat_smile: , so I wrote them even though I was not 100% about necessity. @philipyoo @bennettbuchanan opinions on removal of the pause + resume status checks?

tests/zenko_e2e/backbeat/tests/crr-pause-resume/awsBackend.js
JianqinWang: mm, if we want to make sure replication is paused correctly, the pause would probably go somewhere between `backbeatAPIUtils.pauseReplication` and  `scalityUtils.putObject`?

tests/zenko_e2e/backbeat/tests/crr-pause-resume/awsBackend.js
```diff
@@ -44,6 +44,7 @@ describe('Replication Pause-Resume with AWS backend', function() {
         next => scalityUtils.waitUntilReplicated(srcBucket, key,
             undefined, next),
         next => backbeatAPIUtils.pauseReplication(null, next),
+        next => setTimeout(next, 5000),
```
jonathan-gramain: I think this one should be removed, as we expect the status to be `disabled` right away.

tests/zenko_e2e/backbeat/tests/crr-pause-resume/awsBackend.js
```diff
@@ -52,8 +53,10 @@ describe('Replication Pause-Resume with AWS backend', function() {
         }),
         next => scalityUtils.putObject(srcBucket, key2, Buffer.alloc(1),
             next),
+        next => setTimeout(next, 15000),
         next => awsUtils.assertNoObject(destBucket, key2, next),
         next => backbeatAPIUtils.resumeReplication(null, null, null, next),
+        next => setTimeout(next, 5000),
```
jonathan-gramain: same thing here, should be removed

tests/zenko_e2e/backbeat/tests/crr-pause-resume/awsBackend.js
jonathan-gramain: after discussion, let's keep them (since publish/subscribe of redis may introduce delay)

PR #211, opened 2018-07-23
ft: cloudserver soft anti-affinity

ssalaues:
Sets up a default soft anti affinity for the cloudserver chart


ssalaues:
Added a note explaining the anti affinity settings in the values.yaml

giacope:
@ssalaues Actually, wouldn't it be better if the anti-affinity is on the Zenko chart instead Cloudserver one, to be more modular?

PR #213, opened 2018-07-24
ci: update method to fetch secrets

sylvain-killian:
ShellCommandWithSecrets is deprecated in Eve 1.3.41. Use
secrets from vault instead.


sylvain-killian:
I forgot a bit, will be back soon!

PR #218, opened 2018-07-24
chore: add note about http(s) proxies

giacope:
Maybe it's not the best way to phrase it, but the idea is that only 1 proxy should be specified, either the HTTP or the HTTPS one. Open to suggestions 🙂

### DIFF
--- a/charts/cloudserver/values.yaml
+++ b/charts/cloudserver/values.yaml
@@ -34,7 +34,7 @@ mongodb:
 # When Orbit is enabled, there will be an additional fully stateless replica
 # added which serves as the "manager" for reporting metrics. For example,
 # with a "replicaCount: 2" and Orbit management enabled, there would be a
-# total of 3 pods deployed (2 worker replicas + 1 manager). 
+# total of 3 pods deployed (2 worker replicas + 1 manager).
 replicaCount: 2
 
 image:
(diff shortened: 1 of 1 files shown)


charts/cloudserver/values.yaml
```diff
@@ -49,6 +49,7 @@ proxy:
   # Additionally you can pass a CA certifcate that will be added to the trusted
   # certs. If the proxy URL is configured and caCert is true, helm
   # will look for a file named 'ca.crt' at the root path of this chart.
+  # Note: To avoid unexpected behavior, only specify an HTTP or an HTTPS proxy.
```
jonathan-gramain: `only specify one of "http" or "https" proxy options`?

giacope: Updated!

charts/cloudserver/values.yaml
```diff
@@ -49,6 +49,7 @@ proxy:
   # Additionally you can pass a CA certifcate that will be added to the trusted
   # certs. If the proxy URL is configured and caCert is true, helm
   # will look for a file named 'ca.crt' at the root path of this chart.
+  # Note: To avoid unexpected behavior, only specify one of "http" or "https" proxy options.
```
jonathan-gramain: minor but lets-do-this-while-its-hot: can you wrap the line so that it fits into 80 columns?

jonathan-gramain (approved):

philipyoo (approved):

tmacro (approved):

charts/cloudserver/values.yaml
giacope: Done!

PR #220, opened 2018-07-25
rf: remove wait flags and programatically wait

ssalaues:
Reduces flakiness with the installation step of the CI where sometime pods would be in indefinitate Crashloop state that would cause the CI potentially wait 800 seconds for an install to finish. The Crashloop state is normal especially in the CI environment where the services are expecting the buckets that are predefined environment but don't yet exist until test initialization. The CI still explicitly expects the stack to stabilize but this is now part of the test initialization thus saving time on sleep commands and unneccessary waits. Additionally, this allows devs to collect logs on a stack that was unable to stabilize for debugging purposes.


jonathan-gramain (approved):

ssalaues:
Tested with the `latest` stage for compatibility
https://eve.devsca.com/github/scality/zenko/#/builders/4/builds/1669

ssalaues:
Should also fix CI issues with PR #202

tmacro (approved):

PR #224, opened 2018-07-26
bf(ZENKO-852): Prevent CI object name conflicts

tmacro:
This adds a per run prefix to objects and then checks for the prefix before deleting objects. This is to prevent concurrent test runs from deleting each others objects and causing a incorrect test failure.

### DIFF
--- a/tests/zenko_e2e/util/bucket.py
+++ b/tests/zenko_e2e/util/bucket.py
@@ -107,7 +107,9 @@ def cleanup_bucket(bucket, replicated=False, delete_bucket=True): # noqa pylint:
             objects = []
             versions = version_list['Versions']
             for v in versions:  # pylint: disable=invalid-name
-                objects.append({'VersionId': v['VersionId'], 'Key': v['Key']})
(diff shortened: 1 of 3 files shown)


ssalaues (approved):

JianqinWang (approved):

tests/zenko_e2e/util/bucket.py
```diff
@@ -107,7 +107,9 @@ def cleanup_bucket(bucket, replicated=False, delete_bucket=True): # noqa pylint:
             objects = []
             versions = version_list['Versions']
             for v in versions:  # pylint: disable=invalid-name
-                objects.append({'VersionId': v['VersionId'], 'Key': v['Key']})
+                if v['Key'].startswith(conf.OBJ_PREFIX):
```
jonathan-gramain, line 110: I think we could just specify `prefix = conf.OBJ_PREFIX` before calling `list_object_versions()` for more efficiency?

PS: I know this is merged already :rofl:

PR #225, opened 2018-07-26
Rewrote Installation Guide

wabernat:
Two new files (composed from a couple or three others, informed by my bitter experience trying to follow them). 

The installation guide doc has settings for a 3-node Metal K8s cluster. I would like to get settings in for a five-node cluster before GA (if time permits).

Deleted the original MetalK8s-CentOS74 install guide. The original had a "Useful kubectl commands," section, which I moved to Operations.

Got rid of empty Introduction.rst file. I have no idea what that was about. 

We still need upgrade info.

The current file follows the earlier document's conventions for :file:`filename` and other undefined code and document type snippets. I am assuming that these are known to Sphinx, but if not, I recommend taking out all program-type definitions unless they are actually pulling some freight.


ssalaues:
reopening original pr #135

PR #230, opened 2018-07-27
ZENKO-865 - clarify setup for e2e CRR tests

JianqinWang:


tests/backbeat/Using.md
```diff
@@ -9,18 +9,29 @@
    127.0.0.1 zenko.local
    ```
 
-2. Create an account using Orbit.
-3. Export the access key and secret key of that account (for example, in
+2. Find the Zenko service endpoint for S3 calls to Cloudserver by running:
+    ```
+    kubectl get services
+    ```
+    Look for the service named `zenko-cloudserver`, the value under `CLUSTER-IP`
```
bennettbuchanan: Minor: I see `ClusterIP` instead of `CLUSTER-IP`. Maybe different versions of kubectl?

tests/backbeat/Using.md
```diff
@@ -9,18 +9,29 @@
...
+    should be set as the Zenko endpoint for testing:
+
+```
+export CLOUDSERVER_ENDPOINT=<cluster-ip of zenko-cloudserver>
```
bennettbuchanan: Thinking out loud here 💭. Maybe to follow the pattern of avoiding spaces we could use `<zenko-cloudserver-cluster-ip>`.

bennettbuchanan (approved):
Thanks for maintaining this! 😸

tests/backbeat/Using.md
JianqinWang: updated and squashed!

tmacro (approved):

philipyoo (approved):

tests/backbeat/Using.md
ssalaues: Instead of the cluster IP it would be preferable to simply use the value under `NAME` which resolves in the DNS to the `CLUSTER-IP` but at least with the `NAME` we'll always resolve to the correct location (in case the IP ever changes).

tests/backbeat/Using.md
```diff
@@ -9,18 +9,29 @@
    127.0.0.1 zenko.local
```
ssalaues: Not your fault but would also help minimize confusion is if this changed to something like`<insert-kube-node-ip-here>` zenko.local. This way it's more universal. Otherwise this and line 5 are assuming that:
- that you are ssh-ed to one of the kube servers and doing testing from there
or
- doing a port-forward from a remote kube cluster to your laptop

But doesn't make sense for the many people that use `kubectl` locally on a laptop to connect to the remote servers or running on something like minikube

bennettbuchanan (approved):
🚢

ssalaues (approved):

PR #221, opened 2018-07-25
feature: ZENKO-845 Object monitor throughput test

bennettbuchanan:

PR #233, opened 2018-07-28
feature: ZENKO-846 Add retry test

bennettbuchanan:
Adds the following test cases:
* Tests for failed CRR retries (a single retry and up to 128 concurrent retries). 
* Test checking that the CRR metrics are properly updated when a retry occurs.

Automating these tests is proving especially challenging since we need to update the backbeat pod such that exponential back-off is very short (i.e. one second), so requires a unique backbeat pod with its configuration updated as such. Also, the test relies on deleting and recreating the destination bucket to create failures. This use of buckets is unreliable when multiple tests are using that same bucket. Ideas for strategies to automate these tests are welcome!

### DIFF
--- a/tests/backbeat/package.json
+++ b/tests/backbeat/package.json
@@ -21,6 +21,7 @@
   "scripts": {
     "test_crr": "mocha --exit -t 10000 --recursive tests/crr",
     "test_api": "mocha --exit -t 10000 --recursive tests/api",
+    "test_retry": "mocha --exit -t 10000 --recursive tests/retry",
     "test_crr_pause_resume": "mocha --exit -t 10000 --recursive tests/crr-pause-resume"
   },
   "author": ""
--- a/tests/backbeat/tests/retry/retry.js
+++ b/tests/backbeat/tests/retry/retry.js
@@ -1,0 +1,166 @@
+const assert = require('assert');
+const crypto = require('crypto');
+const { series, parallel, times, timesSeries, doWhilst } = require('async');
+
+const { scalityS3Client, awsS3Client } = require('../../s3SDK');
+const ReplicationUtility = require('../../ReplicationUtility');
+const { makeGETRequest, makePOSTRequest, getResponseBody } =
+    require('../../utils/request');
+
+const scalityUtils = new ReplicationUtility(scalityS3Client);
+const awsUtils = new ReplicationUtility(awsS3Client);
+const srcBucket = `source-bucket-${Date.now()}`;
+const destFailBucket = process.env.AWS_S3_FAIL_BACKBEAT_BUCKET_NAME;
+const destFailLocation = process.env.AWS_S3_FAIL_BACKEND_DESTINATION_LOCATION;
+
+const hex = crypto.createHash('md5')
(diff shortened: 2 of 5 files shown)


tests/backbeat/package.json
```diff
@@ -15,12 +15,13 @@
     "aws-sdk": "2.28.0",
     "azure-storage": "^2.10.0",
     "@google-cloud/storage": "^1.6.0",
-    "mocha": "2.3.4",
+    "mocha": "^5.2.0",
```
bennettbuchanan: Need to update mocha version so that we can use `this.retries`.

ssalaues (approved):

philipyoo (approved):

philipyoo:
I think test is failing because the get request is made before data is available in redis. We can set a timeout before making the get request

bennettbuchanan:
I agree @philipyoo, there is definitely a race condition there. I've refactored that test to continually make the GET request in an `async.doWhilst` until the metrics have been updated (in this PR https://github.com/scality/Zenko/pull/221). 😸

tests/backbeat/tests/retry/retry.js
```diff
@@ -0,0 +1,164 @@
...
+                return doWhilst(callback =>
+                    makeGETRequest(path, (err, res) => {
+                        assert.ifError(err);
+                        assert.strictEqual(res.statusCode, 200);
+                        getResponseBody(res, (err, body) => {
+                            assert.ifError(err);
+                            const { results } = body.completions;
+                            // We have reached an expiration of metrics.
+                            if (results.count < prevCompletions.count) {
+                                return callback(new Error('Invalid test case'));
```
jonathan-gramain: I'm not sure but a bug in the code may well trigger this case (bug in metrics returning less metrics than expected), so instead maybe use a regular `assert`?

tests/backbeat/tests/retry/retry.js
```diff
@@ -0,0 +1,164 @@
...
+    const { backlog, completions, failures } = body;
+    assert.strictEqual((backlog.results.count - prevBacklog.count), 0);
+    assert.strictEqual((backlog.results.size - prevBacklog.size), 0);
+    assert.strictEqual((completions.results.count - prevCompletions.count), 1);
+    assert.strictEqual((completions.results.size - prevCompletions.size), 1);
+    assert.strictEqual((failures.results.count - prevFailures.count), 0);
+    assert.strictEqual((failures.results.size - prevFailures.size), 0);
+}
+
+function performRetries(N, keys, done) {
```
jonathan-gramain: minor: I think you could get rid of parameter `N` and use `keys.length` instead

tests/backbeat/tests/retry/retry.js
bennettbuchanan: Very good point. Updated.

tests/backbeat/tests/retry/retry.js
bennettbuchanan: Nice! Updated to remove that param.

PR #235, opened 2018-07-30
bf(ZENKO-881): Reduce Azure Test Flakiness

tmacro:
add a prefix checking to cleanup_azure_bucket, similar to cleanup_bucket

### DIFF
--- a/tests/zenko_e2e/util/azure.py
+++ b/tests/zenko_e2e/util/azure.py
@@ -1,3 +1,4 @@
+from .. import conf
 
 
 class ObjectStub:  # pylint: disable=too-few-public-methods
@@ -67,6 +68,7 @@ class AzureBucket:
 def cleanup_azure_bucket(bucket, delete_bucket=True):
     blobs = list(bucket.objects.all())
     for blob in blobs:
-        bucket.delete_blob(blob.name)
+        if blob.name.startswith(conf.OBJ_PREFIX):
+            bucket.delete_blob(blob.name)
     if delete_bucket:
         bucket.delete()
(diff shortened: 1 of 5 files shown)


JianqinWang (approved):

ssalaues (approved):

jonathan-gramain (approved):

tests/zenko_e2e/util/azure.py
```diff
@@ -67,6 +68,7 @@ def delete_blob(self, name):
 def cleanup_azure_bucket(bucket, delete_bucket=True):
     blobs = list(bucket.objects.all())
     for blob in blobs:
-        bucket.delete_blob(blob.name)
+        if blob.name.startswith(conf.OBJ_PREFIX):
```
bennettbuchanan, line 71: Does this account for the bucket_match false addition?

bennettbuchanan: I.e. is the source bucket name prepended to the key name on the destination?

tmacro: At the moment none of the python e2e tests run with bucket_match=false, and I think at this point other changes would have to be made to support it

bennettbuchanan (approved):

bennettbuchanan:
It may be worthwhile also to add a couple retries to the Azure tests since we seem to be getting `ESOCKETTIMEDOUT`. Also, feel free to add a `.skip` to the backbeat object monitor tests as there is known flakiness that should be addressed by https://github.com/scality/Zenko/pull/221.

tmacro:
Added the additions proposed by @bennettbuchanan

PR #236, opened 2018-07-30
ZENKO-882 improvement: Add ESLint for E2E tests

bennettbuchanan:


ssalaues (approved):

philipyoo (approved):

tcarmet:
@bert-e wait

bennettbuchanan:
I'd like to not leave this one behind (i.e. "un-stale" it). I think linting is worth it if we continue adding to or modifying these test. 

I think we're missing an eslint config file, and probably there will be some work to rebase. Hopefully I can get to this before it becomes stale again.

PR #239, opened 2018-07-31
ft: update crr metrics e2e tests

philipyoo:


tests/backbeat/Using.md
```diff
@@ -100,10 +100,17 @@
    and `.secrets.env`):
 
     ```
+    export CLOUDSERVER_HOST=<zenko-cloudserver-name>
```
bennettbuchanan, line 107: 👍

tests/backbeat/Using.md
```diff
@@ -100,10 +100,17 @@
...
     export AWS_S3_BACKEND_ACCESS_KEY=<aws-access-key>
     export AWS_S3_BACKEND_SECRET_KEY=<aws-secret-key>
     export AWS_S3_BACKBEAT_BUCKET_NAME=<destination-aws-bucket-name>
     export AWS_S3_BACKEND_DESTINATION_LOCATION=<destination-aws-location-name>
+    export AWS_S3_BACKEND_SOURCE_LOCATION=<source-aws-location-name>
```
bennettbuchanan: Where are we using an AWS source bucket in the API tests that we need to define this?

tests/backbeat/Using.md
```diff
@@ -100,10 +100,17 @@
    and `.secrets.env`):
```
bennettbuchanan, line 104: I'd also add the steps needed to create the Azure destination location. I think these steps would be the same as those for creating Azure resources for CRR tests—probably just copying over those steps will do. 😸

tests/backbeat/tests/api/metrics.js
```diff
@@ -27,49 +31,90 @@ function makeDelayedGETRequest(endpoint, cb) {
...
-                    assert(body[key].description);
-                    assert(body[key].results);
-                    const resultKeys = Object.keys(body[key].results);
+                assert.equal(res.statusCode, 200);
+                getResponseBody(res, (err, body) => {
+                    assert.ifError(err);
+                    const metricTypes = ['backlog', 'completions',
+                        'throughput', 'failures'];
+                    const keys = Object.keys(body);
+                    keys.forEach(key => {
```
bennettbuchanan: Seems like maybe we would want to iterate through `metricTypes` and comparing them to the object body so that we can catch a case where a metric is missing from the response.

tests/backbeat/tests/api/metrics.js
```diff
@@ -27,49 +31,90 @@ function makeDelayedGETRequest(endpoint, cb) {
...
 
     beforeEach(done => series([
         next => scalityUtils.createVersionedBucket(srcBucket, next),
         next => scalityUtils.putBucketReplicationMultipleBackend(srcBucket,
-            destBucket, roleArn, destLocation, next),
+            'placeholder', roleArn, storageClass, next),
     ], done));
 
-    afterEach(done => series([
+    afterEach(done => parallel([
```
bennettbuchanan: Looks like it could be cleaner to use two describe statements because these `beforeEach` and `afterEach` hooks are not needed for the tests that are checking the API route properties (e.g. the test `should return all metrics for path ${path}`). That is, those tests should pass without having created a bucket and setup replication on it.

tests/backbeat/tests/api/metrics.js
```diff
@@ -83,8 +128,8 @@ describe.skip('Backbeat replication metrics', function dF() {
                 }),
             next => scalityUtils.putObject(srcBucket, key, Buffer.alloc(100),
                 next),
-            next => scalityUtils.compareObjectsAWS(srcBucket, destBucket, key,
-                undefined, next),
+            next => scalityUtils.compareObjectsAWS(srcBucket, awsDestBucket,
+                key, undefined, next),
             next => makeDelayedGETRequest('/_/backbeat/api/metrics/crr/all',
```
bennettbuchanan: Let's use the `doWhilst` here to wait for the expected response. I'm comfortable saying the test failed if the metrics aren't there by the time the `REPLICATION_TIMEOUT` is exceeded.

tests/backbeat/tests/api/metrics.js
```diff
@@ -99,66 +144,48 @@ describe.skip('Backbeat replication metrics', function dF() {
                         assert(byteResult - prevDataBytes === 100);
 
                         const throughputSize = body.throughput.results.size;
-                        assert(throughputSize > 0);
+                        assert(throughputSize >= 0);
```
bennettbuchanan: Maybe you can use a unique key name for this test (assuming it maintains the common key prefix) such that you can guarantee the expected throughput size.

tests/backbeat/tests/api/metrics.js
```diff
@@ -99,66 +144,48 @@ describe.skip('Backbeat replication metrics', function dF() {
...
-        next => makeGETRequest('/_/backbeat/api/metrics/crr/all/backlog',
-            (err, res) => {
-                assert.ifError(err);
-                assert.equal(res.statusCode, 200);
-                getResponseBody(res, (err, body) => {
-                    assert.ifError(err);
-                    const keys = Object.keys(body);
-                    assert.strictEqual(keys.length, 1);
-                    assert.strictEqual(keys[0], 'backlog');
+    it('should get object-level progress metrics', done => {
```
bennettbuchanan: Looks like this test case is already handled in https://github.com/scality/Zenko/blob/development/0.9/tests/backbeat/tests/api/objectMonitor.js#L39.

tests/backbeat/tests/api/metrics.js
```diff
@@ -99,66 +144,48 @@ describe.skip('Backbeat replication metrics', function dF() {
...
+                        assert(keys.includes('progress'));
+
+                        next();
+                    });
+                }),
+        ], done);
+    });
 
-    it('should get bucket-level throughput metrics', done => {
+    it('should get object-level throughput metrics', done => {
```
bennettbuchanan: This PR https://github.com/scality/Zenko/pull/221 should handle this test case as well. Probably a refactor is in order to move all the backbeat API tests into a single file.

bennettbuchanan (commented):
Nice! ❤️these new test cases.

tests/backbeat/Using.md
philipyoo: ya i noticed we use both ENDPOINT and HOST in setup :O

tests/backbeat/tests/api/metrics.js
philipyoo: for now, i'll remove the object-level tests from this file

tests/backbeat/tests/api/metrics.js
bennettbuchanan: As per our conversation, the object-level metrics don't apply here, so the above comment is `void` 😂.

tests/backbeat/tests/api/metrics.js
```diff
@@ -1,191 +1,202 @@
...
+            'placeholder', roleArn, storageClass, next),
     ], done));
 
-    it('should get bucket-level throughput metrics', done => {
-        let versionId;
-        const destinationKey = `${srcBucket}/${key}`;
+    afterEach(done => parallel([
+        next => scalityUtils.deleteVersionedBucket(srcBucket, next),
+        next => awsUtils.deleteAllVersions(awsDestBucket, destKeyPrefix, next),
+        next => scalityUtils.deleteAllBlobs(destContainer, destKeyPrefix, next),
```
bennettbuchanan: I noticed a race condition here in another test run: since Azure does not have versioning, if `scalityUtils.deleteVersionedBucket` manages to replicate the deletion of the object version before `scalityUtils.deleteAllBlobs` tries to clean up the destination container, then a `BlobNotFound` error is thrown and `afterEach` fails. We can avoid this case by doing the cleanup in series, where we delete the blobs before deleting the source bucket.

tests/backbeat/tests/api/metrics.js
```diff
@@ -1,191 +1,202 @@
...
-
-                        next();
-                    });
-                }),
+                });
+            }),
+            // NOTE: metrics are doubled because 2 destination locations
+            next => scalityUtils.putObject(srcBucket, key, Buffer.alloc(100),
+                next),
+            next => scalityUtils.compareObjectsAWS(srcBucket, awsDestBucket,
```
bennettbuchanan: Since this is a one-to-many setup we can just use the util `waitUntilReplicated`—i.e. no need to compare the destination object.

tests/backbeat/tests/api/metrics.js
```diff
@@ -1,191 +1,202 @@
...
+            },
+            next => {
+                const expectedBody = {
+                    throughput: {
+                        description: 'Current throughput for replication ' +
+                        'operations in ops/sec (count) and bytes/sec (size) ' +
+                        'in the last 900 seconds',
+                        results: {
+                            count: (parseFloat(prevThroughput.count) +
+                                (2 / 900)).toFixed(2),
```
bennettbuchanan, line 189: Minor: looks like this will generally be `'0.00'`, maybe we could consider putting N objects to the point that the value is increased in the API response.

tests/backbeat/tests/api/metrics.js
```diff
@@ -1,191 +1,202 @@
...
-                    const resultKeys = Object.keys(body.backlog.results);
+                    const type = Object.keys(body)[0];
+                    const data = body[type];
+                    assert(data.description);
+                    assert(data.results);
+                    const resultKeys = Object.keys(data.results);
                     assert(resultKeys.includes('count'));
                     assert(resultKeys.includes('size'));
-                    next();
+                    return done()
```
JianqinWang, line 108: missing ';'

tests/backbeat/tests/api/metrics.js
```diff
@@ -1,191 +1,202 @@
...
-    it('should get bucket-level throughput metrics', done => {
-        let versionId;
-        const destinationKey = `${srcBucket}/${key}`;
+    afterEach(done => series([
+        next => scalityUtils.deleteAllBlobs(destContainer, destKeyPrefix, next),
+        next => awsUtils.deleteAllVersions(awsDestBucket, destKeyPrefix, next),
+        next => scalityUtils.deleteVersionedBucket(srcBucket, next),
+    ], done));
 
+    it.skip('should report metrics when replication occurs', done => {
```
bennettbuchanan, line 133: We still want to skip this one? LGTM 😸

bennettbuchanan (approved):

tests/backbeat/tests/api/metrics.js
philipyoo: i was experiencing some issues still where metrics data from other tests were still included. For now going to commit it with the plans to continue testing

bennettbuchanan: Okay, still good as a manual test for now. 👍

tests/backbeat/tests/api/metrics.js
philipyoo: do you mean replicate N number of objects and each time check that throughput increases?

bennettbuchanan: Was thinking that replicating >= 5 objects would result in the API responding with a `count` floating point > `0.00`. It's possibly overkill since having the `size` property increase probably suffices.

philipyoo: Ah I see, I'll look into adding it with the fix when re-enabling this test case 👍

PR #240, opened 2018-07-31
ZENKO-887 change to Installation/Install_guide.rst

wabernat:
This makes a one-line change to Install_guide.rst. Note, however, that I had to git add this file and directory to and paste in a copy of the file because the previous two versions are hung up in builds, so the history may look odd.

### DIFF
--- a/docs/Installation/Installation_Guide.rst
+++ b/docs/Installation/Installation_Guide.rst
@@ -50,8 +50,6 @@ Sizing
 
 Sizing for Metal K8s
 --------------------
-To perform functional tests using a Scality cloud lab, divide these
-requirements by two.
 
 -  Cores
 
@@ -64,7 +62,7 @@ requirements by two.
    -  4 GB for etcd
    -  8 GB for master
    -  8 GB for node
-      
+
 -  Storage (for the system)
 
    -  20 GB for the root filesystem
@@ -138,96 +136,103 @@ and their basic configuration, including masters, etcds, and nodes.
 
 To specify the machines on which the Ansible_-based deployment system shall
 install MetalK8s, you must provide an *inventory*. This inventory contains a
-*hosts* file that lists all hosts in the cluster and *kube-node.yml*, a script
-that contains configuration information.
+*hosts* file that lists all hosts in the cluster and *kube-node.yml*, a
+configuration file.
 
 .. _Ansible: https://www.ansible.com
 
 To create an inventory:
 
-  1. Log in to the master machine and navigate to the metal-k8s repo.
+1. Log in to the master machine and navigate to the metal-k8s repo.
 
-  2. Create a directory inside the metal-k8s directory (for example,
-     :file:`inventory/quickstart-cluster`) in which the inventory will
-     be stored. Change to that directory.
+2. Create a directory inside the metal-k8s directory (for example,
+   :file:`inventory/quickstart-cluster`) in which the inventory will
+   be stored. Change to that directory.
 
-     .. code-block:: shell
+   .. code-block:: shell
 
-       $ cd metal-k8s
-       $ mkdir -p inventory/quickstart-cluster
-       $ cd inventory/quickstart-cluster/
+     $ cd metal-k8s
+     $ mkdir -p inventory/quickstart-cluster
+     $ cd inventory/quickstart-cluster/
 
-  3. Create the :file:`hosts` file, which lists all hosts.
+3. Create the :file:`hosts` file, which lists all hosts.
 
-     .. code-block:: ini
+   .. code-block:: ini
 
-        node-01 ansible_host=10.0.0.1 ansible_user=centos
-        node-02 ansible_host=10.0.0.2 ansible_user=centos
-        node-03 ansible_host=10.0.0.3 ansible_user=centos
+     node-01 ansible_host=10.0.0.1 ansible_user=centos
+     node-02 ansible_host=10.0.0.2 ansible_user=centos
+     node-03 ansible_host=10.0.0.3 ansible_user=centos
+     node-04 ansible_host=10.0.0.4 ansible_user=centos
+     node-05 ansible_host=10.0.0.5 ansible_user=centos
 
-        [kube-master]
-        node-01
-        node-02
-        node-03
+     [kube-master]
+     node-01
+     node-02
+     node-03
+     node-04
+     node-05
 
-        [etcd]
-        node-01
-        node-02
-        node-03
+     [etcd]
+     node-01
+     node-02
+     node-03
+     node-04
+     node-05
+   
+     [kube-node]
+     node-01
+     node-02
+     node-03
+     node-04
+     node-05
+   
+     [k8s-cluster:children]
+     kube-node
+     kube-master
 
-        [kube-node]
-        node-01
-        node-02
-        node-03
+   Change the host names, IP addresses, and user names to conform to your infrastructure.
+   For example, if your servers are named "server1", "server2", and "server3", copy the code block
+   above and replace ALL instances of "node-0" with "server".
 
-        [k8s-cluster:children]
-        kube-node
-        kube-master
+4. Create a :file:`group_vars` subdirectory in the directory you created in
+   step 2 (the same directory as the :file:`hosts` file).
 
-     Change the host names, IP addresses, and user names to conform to
-     your infrastructure. For example, if your servers are named "server1",
-     "server2", and "server3", copy the code block above and replace ALL
-     instances of "node-0" with "server".
+   .. code-block:: shell
 
-  4. Create a :file:`group_vars` subdirectory in the directory you created in
-     step 2 (the same directory as the :file:`hosts` file).
+    $ mkdir group_vars ; cd group_vars
 
-     .. code-block:: shell
+5. Create a file, :file:`kube-node.yml`, in the :file:`group_vars`
+   subdirectory of the inventory. This file declares how to set up storage
+   (in the default configuration) on hosts in the *kube-node* group; that is,
+   hosts on which pods shall be scheduled:
 
-      $ mkdir group_vars ; cd group_vars
+   .. code-block:: yaml
 
-  5. Create a file, :file:`kube-node.yml`, in the :file:`group_vars`
-     subdirectory of the inventory. This file declares how to set up storage
-     (in the default configuration) on hosts in the *kube-node* group; that is,
-     hosts on which Pods shall be scheduled:
+     metalk8s_lvm_default_vg: False
+     metalk8s_lvm_vgs: ['vg_metalk8s']
+     metalk8s_lvm_drives_vg_metalk8s: ['/dev/vdb']
+     metalk8s_lvm_lvs_vg_metalk8s:
+      lv01:
+        size: 52G
+      lv02:
+        size: 52G
+      lv03:
+        size: 52G
+      lv04:
+        size: 11G
+      lv05:
+        size: 11G
+      lv06:
+        size: 11G
+      lv07:
+        size: 5G
+      lv08:
+        size: 5G
 
-     .. code-block:: yaml
-
-      metalk8s_lvm_default_vg: False
-      metalk8s_lvm_vgs: ['kubevg']
-      metalk8s_lvm_drives_kubevg: ['/dev/vdb']
-      metalk8s_lvm_lvs_kubevg:
-       lv01:
-         size: 52G
-       lv02:
-         size: 52G
-       lv03:
-         size: 52G
-       lv04:
-         size: 11G
-       lv05:
-         size: 11G
-       lv06:
-         size: 11G
-       lv07:
-         size: 5G
-       lv08:
-         size: 5G
-
-     In this example, every *kube-node* host is assumed to have a disk
-     available as :file:`/dev/vdb` that can be used to set up Kubernetes
-     *PersistentVolumes*. For more information about storage, see
-     :doc:`../architecture/storage`.
+   In this example, every *kube-node* host is assumed to have a disk
+   available as :file:`/dev/vdb` that can be used to set up Kubernetes
+   *PersistentVolumes*. For more information about storage, see
+   :doc:`../architecture/storage`.
 
 
 Enter the MetalK8s Virtual Environment Shell
@@ -237,11 +242,11 @@ To install MetalK8s, you must issue commands from within a virtual shell.
 The following steps ensure you can access the virtual environment.
 
 1. Install python-virtualenv:
-   ::
 
-   $ yum install python-virtualenv
+ ::
 
-   Your CentOS image may already have this virtualenv preinstalled.
+  $ yum install python-virtualenv
+  Your CentOS image may already have this virtualenv preinstalled.
 
 2. To install a supported version of Ansible and its dependencies, along with
    some Kubernetes tools (:program:`kubectl` and :program:`helm`), MetalK8s
@@ -249,23 +254,24 @@ The following steps ensure you can access the virtual environment.
    To enter this environment, run ``make shell`` (this takes a few seconds
    when first run).
 
-  .. code::
+ ::
 
-    $ cd metal-k8s
-    $ make shell
-    Creating virtualenv...
-    Installing Python dependencies...
-    Downloading kubectl...
-    Downloading Helm...
-    Launching MetalK8s shell environment. Run 'exit' to quit.
-    (metal-k8s) $
+  $ cd metal-k8s
+  $ make shell
+  Creating virtualenv...
+  Installing Python dependencies...
+  Downloading kubectl...
+  Downloading Helm...
+  Launching MetalK8s shell environment. Run 'exit' to quit.
+  (metal-k8s) $
+
 
 Deploy the Cluster
 ==================
 
 Run the following command to deploy the cluster::
 
-    (metal-k8s) $ ansible-playbook -i inventory/quickstart-cluster -b playbooks/deploy.yml
+  (metal-k8s) $ ansible-playbook -i inventory/quickstart-cluster -b playbooks/deploy.yml
 
 Deployment takes about a half hour.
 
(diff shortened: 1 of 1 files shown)


docs/Installation/Installation_Guide.rst
```diff
@@ -0,0 +1,610 @@
...
+    Adding local repo with URL: http://127.0.0.1:8879/charts
+    $HELM_HOME has been configured at /home/centos/.helm.
+    Warning: Tiller is already installed in the cluster.
+    (Use --client-only to suppress this message, or --upgrade to upgrade Tiller to the current version.)
+    Happy Helming!
+    (metal-k8s) [centos@node01 ~]$
+
+   Helm can now install applications on the Kubernetes cluster.
+
+3. Declare the ZooKeeper repository:
```
ssalaues: This section is no longer needed

docs/Installation/Installation_Guide.rst
```diff
@@ -0,0 +1,610 @@
...
+
+   ::
+
+      $ helm repo add zenko-zookeeper https://scality.github.io/zenko-zookeeper/charts
+      "zenko-zookeeper" has been added to your repositories
+
+      $ helm repo add incubator http://storage.googleapis.com/kubernetes-charts-incubator
```
ssalaues: While these are no longer needed we need the following step `helm repo add scality https://scality.github.io/charts/`

ssalaues (changes requested):

giacope (changes requested):
Not all the comments are requests.

ssalaues (approved):

giacope (approved):

docs/Installation/Installation_Guide.rst
```diff
@@ -1,579 +1,581 @@
...
+While it is possible to run Zenko on a single machine, it's designed for
+cluster operation. If you can set up a Kubernetes cluster on your own, review
+the **General Cluster Requirements** and skip to **Installing Zenko**, below.
+
+Otherwise, you can set up a cluster quickly using MetalK8s_, Scality's
+open-source Kubernetes cluster project, as described in "Setting Up a Metal K8s
+Cluster," below.
+
+The following section describes general cluster requirements, which are tested
+on Metal K8s. Because MetalK8s is designed to operate without support from
```
NicolasT, line 19: `MetalK8s`, not `Metal K8s`.

docs/Installation/Installation_Guide.rst
```diff
@@ -1,579 +1,581 @@
...
+
+****************************
+General Cluster Requirements
+****************************
+
+Setting up a testing cluster requires at least three machines (these can be
+VMs) running CentOS_ 7.4 (The recommended mimimum for Zenko production service
+is five server nodes with three masters/etcds, but for testing and
+familiarization, three masters and three nodes is fine). You must have SSH
+access to these machines and they must have SSH access to each other. (You
```
NicolasT, line 36: No need for SSH access between nodes?

docs/Installation/Installation_Guide.rst
```diff
@@ -1,579 +1,581 @@
...
+can copy SSH credentials from one machine to the next and log in once to
```
NicolasT, line 37: Never copy SSH private keys to other machines... There's no need to do it when deploying MetalK8s, and it's bad practice. Worst case you use agent forwarding.

docs/Installation/Installation_Guide.rst
```diff
@@ -1,579 +1,581 @@
...
+
+.. _MetalK8s: https://github.com/scality/metal-k8s/
+.. _CentOS: https://www.centos.org
+.. _Kubernetes: https://kubernetes.io
+
+
+Sizing
+======
+
+Sizing for Metal K8s
```
NicolasT, line 51: `MetalK8s`

docs/Installation/Installation_Guide.rst
```diff
@@ -1,579 +1,581 @@
...
+-  RAM
+
+   -  4 GB for etcd
+   -  8 GB for master
+   -  8 GB for node
+
+-  Storage (for the system)
+
+   -  20 GB for the root filesystem
+   -  16 GB for etcd
```
NicolasT, line 69: If we size `etcd` storage, we should document how to set up the system(s) to have `etcd` storage not on `/` (which is good practice in any case).

docs/Installation/Installation_Guide.rst
```diff
@@ -1,579 +1,581 @@
...
+
+   This requirement is not for the system device. You may have to attach a
+   separate storage volume to your cloud server instance.
+
+All servers must be in CentOS 7.4, and accessible with ssh.
+
+Proxies
+=======
+
+If you are behind a proxy, add the following lines to your local machine’s
```
NicolasT, line 96: What is 'your local machine'? It may need to be set on all machines in the cluster, no?

docs/Installation/Installation_Guide.rst
```diff
@@ -1,579 +1,581 @@
...
+/etc/environment file:
+
+::
+
+    http_proxy=http://user;pass@<my-ip>:<my-port>
+    https_proxy=http://user;pass@<my-ip>:<my-port>
+    no_proxy=localhost,127.0.0.1,10.*
+
+******************************
+Setting Up a Metal K8s Cluster
```
NicolasT, line 106: `MetalK8s`

docs/Installation/Installation_Guide.rst
```diff
@@ -1,579 +1,581 @@
...
+
+Clone or Copy the MetalK8s Git Repo
+===================================
+
+Either log in to your primary machine (any of the nodes can be the primary)
+and clone the metal-k8s repo directly from GitHub:
+
+.. code-block:: shell
+
+   $ git clone https://github.com/scality/metal-k8s
```
NicolasT, line 121: Using this, users would be installing some random in-development version of MetalK8s, rather than the one we want to have them deploy, so this is not the right approach to 'download' MetalK8s.

docs/Installation/Installation_Guide.rst
```diff
@@ -1,579 +1,581 @@
...
+
+MetalK8s provides a stable, easy-to-deploy base for you to test Zenko in a
+live Kubernetes cluster environment.
+
+
+Clone or Copy the MetalK8s Git Repo
+===================================
+
+Either log in to your primary machine (any of the nodes can be the primary)
+and clone the metal-k8s repo directly from GitHub:
```
NicolasT, line 117: `MetalK8s`

docs/Installation/Installation_Guide.rst
```diff
@@ -1,579 +1,581 @@
...
+
+.. code-block:: shell
+
+   $ git clone https://github.com/scality/metal-k8s
+
+or put the metal-k8s project onto the machine using sftp:
```
NicolasT, line 123: `MetalK8s`

docs/Installation/Installation_Guide.rst
```diff
@@ -1,579 +1,581 @@
...
+.. code-block:: shell
+
+  $ sftp centos@10.0.0.1
+  Connected to 10.0.0.1
+  sftp> put -r metal-k8s
+
+Define an Inventory
+===================
+
+Each server must be configured in an inventory file that identifies servers
```
NicolasT, line 134: An inventory is *not* a *file*. It's a directory with various files, potentially in subdirectories.

docs/Installation/Installation_Guide.rst
```diff
@@ -1,579 +1,581 @@
...
+and their basic configuration, including masters, etcds, and nodes.
+
+To specify the machines on which the Ansible_-based deployment system shall
+install MetalK8s, you must provide an *inventory*. This inventory contains a
+*hosts* file that lists all hosts in the cluster and *kube-node.yml*, a
```
NicolasT, line 139: `kube-node.yml` doesn't go next to the `hosts` file, but in a `group_vars` subdirectory of the inventory.

docs/Installation/Installation_Guide.rst
```diff
@@ -1,579 +1,581 @@
...
+    $ mkdir group_vars ; cd group_vars
+
+5. Create a file, :file:`kube-node.yml`, in the :file:`group_vars`
+   subdirectory of the inventory. This file declares how to set up storage
+   (in the default configuration) on hosts in the *kube-node* group; that is,
+   hosts on which pods shall be scheduled:
+
+   .. code-block:: yaml
+
+     metalk8s_lvm_default_vg: False
```
NicolasT, line 211: Why?

docs/Installation/Installation_Guide.rst
```diff
@@ -1,579 +1,581 @@
...
+     metalk8s_lvm_vgs: ['vg_metalk8s']
```
NicolasT, line 212: Shouldn't be required if `metalk8s_lvm_default_vg` were not set to `False`.

docs/Installation/Installation_Guide.rst
```diff
@@ -1,579 +1,581 @@
...
+   :doc:`../architecture/storage`.
+
+
+Enter the MetalK8s Virtual Environment Shell
+============================================
+
+To install MetalK8s, you must issue commands from within a virtual shell.
+The following steps ensure you can access the virtual environment.
+
+1. Install python-virtualenv:
```
NicolasT, line 244: No need to install `python-virtualenv`?

docs/Installation/Installation_Guide.rst
```diff
@@ -1,579 +1,581 @@
...
+
+Cluster Services
+----------------
+
+Services to operate and monitor your MetalK8s cluster are provided. To access
+these dashboards:
+
+1. Make sure kubectl is installed on your local machine::
+
+   $ yum install kubectl
```
NicolasT, line 351: You get this in a `make shell` environment, which also ensures you get a version which works with the deployed cluster, which is definitely not guaranteed when installing from CentOS repositories.

docs/Installation/Installation_Guide.rst
```diff
@@ -1,579 +1,581 @@
...
+   |                         |                                                         |                                                                                                 | in the Kubernetes host.               |
+   +-------------------------+---------------------------------------------------------+-------------------------------------------------------------------------------------------------+---------------------------------------+
+   | `Grafana`_              | Monitoring dashboards for cluster services              | http://localhost:8001/api/v1/namespaces/kube-ops/services/kube-prometheus-grafana:http/proxy/   |                                       |
+   +-------------------------+---------------------------------------------------------+-------------------------------------------------------------------------------------------------+---------------------------------------+
+   | `Cerebro`_              | An administration and monitoring console for            | http://localhost:8001/api/v1/namespaces/kube-ops/services/cerebro:http/proxy/                   | When accessing Cerebro, connect it to |
```
NicolasT, line 378: We pre-provision a config.

docs/Installation/Installation_Guide.rst
```diff
@@ -1,579 +1,581 @@
...
+   |                         | Elasticsearch clusters                                  |                                                                                                 | http://elasticsearch:9200 to operate  |
+   |                         |                                                         |                                                                                                 | the MetalK8s Elasticsearch cluster.   |
+   +-------------------------+---------------------------------------------------------+-------------------------------------------------------------------------------------------------+---------------------------------------+
+   | `Kibana`_               | A search console for logs indexed in Elasticsearch      | http://localhost:8001/api/v1/namespaces/kube-ops/services/http:kibana:/proxy/                   | When accessing Kibana for the first   |
```
NicolasT, line 382: We pre-provision a config.

docs/Installation/Installation_Guide.rst
```diff
@@ -1,579 +1,581 @@
...
+   |                         |                                                         |                                                                                                 | the ``logstash-*`` index, using the   |
+   |                         |                                                         |                                                                                                 | ``@timestamp`` field as *Time Filter  |
+   |                         |                                                         |                                                                                                 | field name*.                          |
+   +-------------------------+---------------------------------------------------------+-------------------------------------------------------------------------------------------------+---------------------------------------+
+
+See :doc:`../architecture/cluster-services` for more about these services
```
NicolasT, line 389: Do these docs build?

docs/Installation/Installation_Guide.rst
wabernat: I don't understand the distinction you're making.

NicolasT: > they must have SSH access to each other

That's not a requirement, and if it were it shouldn't be one.

docs/Installation/Installation_Guide.rst
wabernat: done.

docs/Installation/Installation_Guide.rst
wabernat: To whose expertise shall I appeal for this information?

docs/Installation/Installation_Guide.rst
wabernat: done

docs/Installation/Installation_Guide.rst
wabernat: done.

docs/Installation/Installation_Guide.rst
wabernat: fixed.

docs/Installation/Installation_Guide.rst
wabernat: Where is the latest stable branch for clones to target?

docs/Installation/Installation_Guide.rst
wabernat: Done.

docs/Installation/Installation_Guide.rst
wabernat: Done

docs/Installation/Installation_Guide.rst
wabernat: fixed.

docs/Installation/Installation_Guide.rst
wabernat: Salim suggested it. My build was not building. This seemed to move me a little closer to built.

docs/Installation/Installation_Guide.rst
wabernat: Removed bad docs link. 

Can't comment about build other than to note that the merge request went through, and the linter for RST is quite permissive/nonexistent.

docs/Installation/Installation_Guide.rst
wabernat: OK, I talked through this. Fixed.

docs/Installation/Installation_Guide.rst
wabernat: fixed

docs/Installation/Installation_Guide.rst
wabernat: fixed.

docs/Installation/Installation_Guide.rst
wabernat: deleted.

docs/Installation/Installation_Guide.rst
wabernat: How would you have this reflected in the documentation? Is Cerebro preconfigured to elasticsearch:9200?

docs/Installation/Installation_Guide.rst
wabernat: Fixed in Zenko-921

docs/Installation/Installation_Guide.rst
wabernat: Following precedent in ZENKO-921 (Kibana, next comment) and deleting setup note.

docs/Installation/Installation_Guide.rst
wabernat: removed.

docs/Installation/Installation_Guide.rst
wabernat: removed.

docs/Installation/Installation_Guide.rst
wabernat: removed.

docs/Installation/Installation_Guide.rst
wabernat: Is this a show-stopper? I need to know either: a) that I can omit the etcd sizing requirement or b) how to configure the system to put etcd somewhere other than / .

If I don't hear back, I'm filing this under "nice to have" and moving the other changes along.

smaffulli:
Folks, this commit (although merged) contains information that is not pertinent to Zenko IMO. How to install a MetalK8s cluster should be a link to MetalK8s docs. The instructions to install Zenko are also too MetalK8s-centric which is limiting.

I think we should revert this change. 

And before going any further, I would suggest to focus on agreeing on a common architecture for the whole Zenko documentation, setup the infrastructure to build it properly (take MetalK8s docs as an example to follow) and put the content we have in various formats/sources into one coherent information architecture.  Such conversation already started on https://github.com/scality/Zenko/pull/149

docs/Installation/Installation_Guide.rst
NicolasT: I think this ought to be discussed with out TS folks. Right now, we don't have any recommendations for putting various types of data on different volumes, but maybe that should be the case.

docs/Installation/Installation_Guide.rst
NicolasT: I'd recommend downloading the release files rather than pulling things from Git, for 'stable' installations, so https://github.com/scality/metalk8s/releases

docs/Installation/Installation_Guide.rst
NicolasT: Yeah, there's an entry in a pre-populated list of 'known clusters' when you open Cerebro.

docs/Installation/Installation_Guide.rst
NicolasT: This would only be found by actually building the docs. For MetalK8s, we're working on a step in CI which performs this, and as such makes sure any proposed change to the docs doesn't break its build.

ssalaues:
@smaffulli I think the reason why we were okay with having this MetalK8s info on Zenko was because so many devs were having issues with specifically installing Zenko on MetalK8s. The docs for MetalK8s is great for just installing MetalK8s but we need a bit more configurations for Zenko on MetalK8s (mainly volume provisioning)

And if our devs were having pretty much the same issues with installs, likely others will run into the same issues as well.

docs/Installation/Installation_Guide.rst
wabernat: Done.

docs/Installation/Installation_Guide.rst
wabernat: Done.

PR #241, opened 2018-07-31
bf(ZENKO-888): Fix Prometheus test

tmacro:


ssalaues (approved):

tests/zenko_e2e/util/bucket.py
```diff
@@ -129,9 +129,9 @@ def cleanup_bucket(bucket, replicated=False, delete_bucket=True): # noqa pylint:
         except BaseException:
             pass
 
-        is_truncated = version_list['IsTruncated']
+        is_truncated = version_list.get('IsTruncated')
         if is_truncated:
-            key_marker = version_list['NextKeyMarker']
+            key_marker = version_list.get('NextKeyMarker')
```
bennettbuchanan, line 134: I think we'll want to include the `NextVersionIdMarker` as well in the subsequent request.

PR #244, opened 2018-07-31
disable prometheus test

tmacro:


bennettbuchanan (approved):

tests/zenko_e2e/prometheus/test_basic.py
```diff
@@ -13,6 +13,7 @@ def test_prometheus_healthy(prometheus_client):
 POD = 'kubernetes-pods'
 
 
+@pytest.skip("Broken")
```
bennettbuchanan: Minor: maybe worth adding a comment with reasoning on this.

PR #245, opened 2018-07-31
bf(ZENKO-888): Fixd GCP object deletion bug

tmacro:


tests/zenko_e2e/util/bucket.py
```diff
@@ -136,6 +136,14 @@ def cleanup_bucket(bucket, replicated=False, delete_bucket=True): # noqa pylint:
         bucket.delete()
 
 
+def cleanup_gcp_bucket(bucket, delete_bucket=False):
```
jonathan-gramain, line 139: could it make sense to generalize as `cleanup_nonversioned_bucket`? (thinking of azure as well)

jonathan-gramain (approved):

tests/zenko_e2e/util/bucket.py
tmacro: The tests actually use the Azure sdk on the backside so they're a tad different. We can probably create a wrapper class for the GCP and Azure SDKs to make their buckets appear to be AWS buckets objects to cleanup the cleanup function a little

tmacro: also the cleanup_buckets function should probably refactored as well to use the high level OOP API instead of the low level one. then we should be able to merge all three

ssalaues (approved):

PR #250, opened 2018-08-03
feature: ZENKO 799 location quota tests

dora-korpar:
First iteration of tests for location quota feature :D
Depends on https://github.com/scality/orbit-simulator/pull/9
For reviewing ease: First commit is the creation of a node_tests folder within the tests dir and moving backbeat tests. Second commit is the tests

PR #251, opened 2018-08-03
bugfix: ZENKO-915 One-to-many CRR test flakiness

bennettbuchanan:


ssalaues (approved):

giacope (approved):

ssalaues:
Since #250 just merged, there are some conflicts so the queue should fail. Please fix the conflicts once the Berte reports back. (It looks like the folders got moved around)

jonathan-gramain (approved):

tests/node_tests/backbeat/tests/crr/awsBackend.js
```diff
@@ -23,6 +23,7 @@ const REPLICATION_TIMEOUT = 300000;
 
 describe('Replication with AWS backend', function() {
     this.timeout(REPLICATION_TIMEOUT);
+    this.retries(3);
```
bennettbuchanan, line 26: I've noticed HTTP 503 status code (Service Unavailable) when Cloudserver tries to put objects to AWS during CRR. It seems like a retry of the test is warranted for that case rather than a CI failure, though I'd ideally like to minimize reliance on retries.

PR #252, opened 2018-08-07
bf: ZENKO-916 modify ACL replication test case

jonathan-gramain:
Instead of waiting for replication and expecting no change, expect the
replication status to stay COMPLETED. This because we don't want ACLs
to trigger a replication at all as it would break transient source by
reading the data from the local source which has been deleted already.

### DIFF
--- a/charts/cloudserver/values.yaml
+++ b/charts/cloudserver/values.yaml
@@ -79,7 +79,7 @@ replicaFactor: 1
 
 image:
   repository: zenko/cloudserver
-  tag: 8.0.8-RC4
+  tag: 8.0.9
   pullPolicy: IfNotPresent
 
 proxy:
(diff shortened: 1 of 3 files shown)


jonathan-gramain:
@bert-e wait

ssalaues (approved):

dora-korpar (approved):

tmacro (approved):

giacope:
@bert-e wait

giacope (changes requested):
Changes to update a docker image tag should be on a separate commit, and the commit message should contain the entire changelog of that release (and the image should be created with the proper script to update the wiki and generate the changelog). Also, whenever Cloudserver is updated, it is preferred that s3-data image tag is updated too (in the same commit).

PR #257, opened 2018-08-08
ZENKO-927

wabernat:
This implements Nicolas's requested changes to ZENKO-887 Installation_Guide.rst

### DIFF
--- a/docs/Installation/Installation_Guide.rst
+++ b/docs/Installation/Installation_Guide.rst
@@ -283,6 +283,8 @@ If your system can reach port *6443* on the first *kube-master* node, you can
    node-01     Ready     master,node      1m        v1.10.4
    node-02     Ready     master,node      1m        v1.10.4
    node-03     Ready     master,node      1m        v1.10.4
+   node-04     Ready     master,node      1m        v1.10.4
+   node-05     Ready     master,node      1m        v1.10.4
 
 * List all pods
 
@@ -330,15 +332,15 @@ Cluster Services
 ----------------
 
 Services to operate and monitor your MetalK8s cluster are provided. To access
-these dashboards:
+these remote services from your local machine:
 
 1. Copy the credentials in
    :file:`inventory/quickstart-cluster/artifacts/admin.conf` to your local
-   machine. Export this path locally with
+   machine at ~/.kube . Export this path locally with
 
    ::
 
-   $ export KUBECONFIG=`pwd`/inventory/quickstart-cluster/artifacts/admin.conf
+    $ export KUBECONFIG="~/.kube/admin.conf"
 
 2. On your cluster, open port 6443 for remote access to cluster services.
 
@@ -481,7 +483,7 @@ parameters as appropriate.)
       -  zenko.local
 
     cloudserver:
-       endpoint: "zenko.local"
+      endpoint: "zenko.local"
 
    You can edit these parameters, using each component’s values.yaml file
    as your guide. Save this file.
--- a/docs/Installation/Installation_Guide.rst
+++ b/docs/Installation/Installation_Guide.rst
@@ -12,11 +12,11 @@ cluster operation. If you can set up a Kubernetes cluster on your own, review
 the **General Cluster Requirements** and skip to **Installing Zenko**, below.
 
 Otherwise, you can set up a cluster quickly using MetalK8s_, Scality's
-open-source Kubernetes cluster project, as described in "Setting Up a Metal K8s
+open-source Kubernetes cluster project, as described in "Setting Up a MetalK8s
 Cluster," below.
 
 The following section describes general cluster requirements, which are tested
-on Metal K8s. Because MetalK8s is designed to operate without support from
+on MetalK8s. Because MetalK8s is designed to operate without support from
 public cloud resources, the following sizing requirements are assumed good for
 other cloud Kubernetes deployments, where such resources are preinstalled and
 available on demand.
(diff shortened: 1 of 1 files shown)


docs/Installation/Installation_Guide.rst
```diff
@@ -284,15 +270,15 @@ credentials to access the cluster. Export this location in the shell to give
...
 If your system can reach port *6443* on the first *kube-master* node, you can
 
 * List the nodes
 
-::
+  ::
 
-   (metal-k8s) $ kubectl get nodes
+   (metalk8s) $ kubectl get nodes
    NAME        STATUS    ROLES            AGE       VERSION
```
JianqinWang, line 282: Since the example `hosts` file has 5 nodes, this example should also list 5 nodes for consistency.

docs/Installation/Installation_Guide.rst
```diff
@@ -346,21 +332,17 @@ Cluster Services
 Services to operate and monitor your MetalK8s cluster are provided. To access
 these dashboards:
 
-1. Make sure kubectl is installed on your local machine::
-
-   $ yum install kubectl
-
-2. Copy the credentials in
+1. Copy the credentials in
```
JianqinWang: Would be helpful to specify where to save this file in local machine (`~/.kube/config` for ubuntu as well as macos, I believe)

docs/Installation/Installation_Guide.rst
```diff
@@ -500,7 +481,7 @@ parameters as appropriate.)
       -  zenko.local
 
     cloudserver:
-     endpoint: "zenko.local"
+       endpoint: "zenko.local"
```
bennettbuchanan: Noticed this change. I'm not sure, but want to ensure that the yaml files are indented to the right level.

bennettbuchanan (approved):

docs/Installation/Installation_Guide.rst
```diff
@@ -500,7 +481,7 @@ parameters as appropriate.)
...
```
JianqinWang: The guide I've been following to install Zenko is here https://github.com/scality/Zenko/tree/development/0.9/charts

It gives the option of installing Zenko using flags in a copy-paste command, instead of editing `options.yml`

I feel like it would be good to at least link this guide somewhere (if not include it fully) so others can find it, since the guide there is very complete and discusses other flags/values for initial installation as well as upgrading + validating your deployment.

JianqinWang:
Another note - creating 200 GB Volumes for deployment has caused issues with the example values in `kube-node.yml`, where the metalk8s deployment runs out of space. Most of us have been using 400 GB volumes instead so that the ansible deployment can finish.

docs/Installation/Installation_Guide.rst
wabernat: done.

docs/Installation/Installation_Guide.rst
wabernat: fixed.

docs/Installation/Installation_Guide.rst
JianqinWang: After discussing with William - asking to copy the file to 'your local machine' may be misleading, as 'local machine' is often associated with your computer device. This step can be simplified to just specifying that we export `KUBECONFIG` in our deployment node, or something similar.

docs/Installation/Installation_Guide.rst
```diff
@@ -178,21 +173,22 @@ To create an inventory:
...
      [k8s-cluster:children]
      kube-node
      kube-master
 
-   Change the host names, IP addresses, and user names to conform to your infrastructure.
-   For example, if your servers are named "server1", "server2", and "server3", copy the code block
-   above and replace ALL instances of "node-0" with "server".
+   Change the host names, IP addresses, and user names to conform to your
+   infrastructure. For example, if your servers are named "server1", "server2",
+   and "server3", copy the code block above and replace ALL instances of
```
jonathan-gramain, line 190: Since the code block above seems to be made for a 5 server deployment, maybe the example should also mention up to "server5", or the example should be reduced to 3 servers.

docs/Installation/Installation_Guide.rst
wabernat: done.

docs/Installation/Installation_Guide.rst
wabernat: Today's discourse has suggested a profusion of alternate installation methods. My intention with this guide is to put out something out that is simple and proven, that we can support. I don't dismiss your suggestion out of hand, nor irrevocably, but I don't think this failing to provide alternative/advanced techniques should be a blocker. Can this slide, or do you feel strongly otherwise?

JianqinWang: Could you add something at the end like,

For more installation options and info, please see *link to the other guide*.

If not for this PR, then one in the close future so we can start consolidating docs (or at least chaining them together so finding one let's you know of the others)

docs/Installation/Installation_Guide.rst
wabernat: fixed.

giacope (changes requested):
Please update commit messages so that they don't include people's names.

docs/Installation/Installation_Guide.rst
JianqinWang: Now that the example `hosts` file has been changed to 3 servers, please revert the change here  so the document is consistent.

wabernat:
This branch was totally lost in the shuffle and is not worth trying to merge. 

Abandoning.

PR #258, opened 2018-08-08
Improvement/ZENKO-934 refactor e2e Makefile

tmacro:


tests/Makefile
```diff
@@ -1,132 +1,239 @@
...
-DOCKER_IMAGE := $(DOCKER_IMAGE_NAME):$(DOCKER_IMAGE_TAG)
 
 DOCKER_ARGS :=
 PYTEST_ARGS :=
 DOTENV := $(PWD)/.env
-HELM_TEST_PODS := {range .items[?(@.metadata.annotations.helm\.sh\/hook)]}{@.metadata.name}{"\n"}{end}
-E2E_POD := zenko-e2e
+
+# These are for dumping logs, but they're really long so store them as variables
+E2E_PODS := {range .items[?(@.metadata.annotations.helm\.sh\/hook)]}{@.metadata.name}{"\n"}{end}
```
jonathan-gramain: Could this var be confused with `E2E_POD` below as they don't seem to serve the same purpose?

jonathan-gramain (approved):

anurag4DSB (approved):

eve/main.yml
```diff
@@ -16,6 +16,22 @@ models:
       source: tests/artifacts
       urls:
         - "*"
+  - ShellCommand: &fix_curl # Workaround, to be removed
+      name: Install curl
+      command: |-
+        apk --update upgrade
+        apk --update add curl
```
ssalaues: We don't need this now with the updated image right?

tmacro: removed

tests/Makefile
```diff
@@ -1,132 +1,239 @@
--include .local.make
-DOCKER_IMAGE_NAME := docker.io/zenko/zenko-e2e
-DOCKER_IMAGE_TAG := latest
-
-DOCKER := docker
-TOX := tox
+# Some bin aliases
+DOCKER := @docker
+TOX := @tox
+HELM := @helm
```
giacope: Maybe it's not a good idea to suppress all echoing by default with `@`s. Since most stuff is being suppressed anyway, why not just run by default with the Makefile `-s` or `--silent` flag, which is easier to override or change.

tmacro: we could also use the same format as metalk8s's Makefile with a `$(V)` variable being defined, to toggle verboseness

giacope: Yep, or something like:
```bash
ifndef VERBOSE
MAKEFLAGS	+= --no-print-directory --silent
endif
```

tmacro: ooo didn't know you could edit the Makefile flags on the fly, this seems like the cleanest way to me

tests/Makefile
```diff
@@ -1,132 +1,252 @@
...
+	$(KUBECTL) run $(E2E_POD) \
+		--rm \
+		--restart=Never \
+		--attach=True \
+		--image=$(E2E_IMAGE)  \
+		--namespace $(HELM_NAMESPACE) \
+		$(shell ../eve/workers/ci_env.sh env) || (echo "Tests have failed" ; kubectl get pods -L redis-role ; exit 1) >&2
+.PHONY: test-e2e
+
+test-backbeat: | build-bb-e2e
```
ssalaues: We should change this to something like `test-node` or `run-node-tests`. I know these tests were changed to up to be more generic. Because like `test_location_quota` is an Cloudserver test.

ssalaues (changes requested):
Easy change but otherwise LGTM and should help keep the yaml length down

ssalaues (approved):

giacope (approved):

PR #259, opened 2018-08-08
ZENKO-949 bugfix: increase default prometheus replica count

giacope:
Increase the default number of Prometheus replicas to 2 for failover purposes.

### DIFF
--- a/charts/zenko/values.yaml
+++ b/charts/zenko/values.yaml
@@ -80,6 +80,8 @@ prometheus:
     enabled: false
   pushgateway:
     enabled: false
+  server:
+    replicaCount: 2
 
 mongodb-replicaset:
   replicaSet: rs0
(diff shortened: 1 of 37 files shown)


ssalaues (approved):

charts/zenko/values.yaml
```diff
@@ -80,6 +80,8 @@ prometheus:
     enabled: false
   pushgateway:
     enabled: false
+  server:
+    replicaCount: 2
```
ssalaues, line 84: Does this work without any chart changes?

giacope:
@bert-e wait

charts/zenko/values.yaml
giacope: No 😥

ssalaues:
Duplicate of #267

PR #262, opened 2018-08-10
bf: ZENKO-968 expose NO_PROXY in values.yml

jonathan-gramain:
Add a value in cloudserver chart to expose the NO_PROXY exceptions.

### DIFF
--- a/charts/cloudserver/templates/_env.tpl
+++ b/charts/cloudserver/templates/_env.tpl
@@ -43,6 +43,12 @@ env:
   - name: NODE_EXTRA_CA_CERTS
     value: "/ssl/ca.crt"
   {{- end }}
+  {{- if .Values.proxy.no_proxy }}
+  - name: no_proxy
+    value: "{{ .Values.proxy.no_proxy }}"
+  - name: NO_PROXY
+    value: "{{ .Values.proxy.no_proxy }}"
+  {{- end }}
   - name: S3METADATA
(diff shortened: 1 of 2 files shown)


charts/cloudserver/templates/_env.tpl
```diff
@@ -43,6 +43,12 @@ env:
   - name: NODE_EXTRA_CA_CERTS
     value: "/ssl/ca.crt"
   {{- end }}
+  {{- if .Values.proxy.no_proxy }}
```
giacomoguiulfo: Should we set `no_PROXY` as well?

charts/cloudserver/values.yaml
```diff
@@ -91,9 +91,11 @@ proxy:
   # will look for a file named 'ca.crt' at the root path of this chart.
   # Note: To avoid unexpected behavior, only specify one of "http" or "https"
   # proxy options.
+  # If you have exceptions to proxification you can use the no_proxy variable.
```
giacope, line 94: Minor: I think it looks cleaner if the `Note:` goes in the end.

giacope (approved):

JianqinWang (approved):

charts/cloudserver/values.yaml
jonathan-gramain: Saw too late, sorry will skip this

PR #264, opened 2018-08-10
ZENKO-933 bugfix: redis chart

ssalaues:
Review only for now since we should thoroughly test this before merging.

Refactor of the redis-ha chart. This fixes issues that would cause redis not to fail over at all or too late.


ssalaues:
@bert-e wait

charts/redis-ha/templates/redis-ha-configmap.yaml
```diff
@@ -0,0 +1,29 @@
...
+    min-slaves-to-write {{ .Values.redis.minSlavesToWrite }}
+    repl-diskless-sync {{ .Values.redis.disklessReplication }}
+    rdbcompression yes
+    rdbchecksum yes
+    {{ .Values.redis.rdbBackups }}
+    {{ .Values.redis.addionalConfig }}
+
+  sentinel.conf: |
+    dir /data
+    sentinel monitor {{ .Values.redis.masterGroupName }} {{ template "redis-ha.fullname" . }}-0.{{ template "redis-ha.fullname" . }} 6379 {{ .Values.sentinel.quorum }}
```
rahulreddy: I can't tell if we are getting the bootstraplist here or just the first host.

charts/redis-ha/templates/redis-ha-statefulset.yaml
```diff
@@ -0,0 +1,223 @@
...
+            S_PATCH=/data/conf/sentinel.patch
+
+            set -ex
+            function sentinel_update(){
+                if [[ "$OLD_IP" != "$POD_IP" ]]; then
+                    echo "Removing references to previous IP"
+                    sed -i "s/*.sentinel known.*$OLD_IP//" $SENTINEL_CONF
+                fi
+                if [[ `grep "sentinel monitor" $SENTINEL_CONF` ]]; then
+                    sed -i "s/^.*sentinel monitor.*/sentinel monitor {{ .Values.redis.masterGroupName }} $1 6379 {{ .Values.sentinel.quorum }}/" $SENTINEL_CONF
```
rahulreddy: Use the var `.Values.redis.port` instead of hardcoding the port.

charts/redis-ha/templates/redis-ha-statefulset.yaml
```diff
@@ -0,0 +1,223 @@
...
+            MASTER=`redis-cli -h {{ template "redis-ha.fullname" . }} -p {{ .Values.sentinel.port }} sentinel get-master-addr-by-name {{ .Values.redis.masterGroupName }} | grep -E '[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}'`
+            OLD_IP=`cat /data/conf/ip`
+            REDIS_CONF=/data/conf/redis.conf
+            SENTINEL_CONF=/data/conf/sentinel.conf
+            R_PATCH=/data/conf/redis.patch
+            S_PATCH=/data/conf/sentinel.patch
+
+            set -ex
+            function sentinel_update(){
+                if [[ "$OLD_IP" != "$POD_IP" ]]; then
```
rahulreddy: Why do you have to sed the old ip with the new ones? It should be able to pickup new valid ips and just blacklist old invalid ips?

charts/redis-ha/templates/redis-ha-configmap.yaml
ssalaues: I copied Federation which initializes the conf with the first host as the master and then configures the rest as slaved by default.

charts/redis-ha/values.yaml
```diff
@@ -0,0 +1,102 @@
...
+
+image: redis
+tag: 4.0.11-stretch
+## replicas number for each component
+replicas: 3
+
+## Redis specific configuration options
+redis:
+  port: 6379
+  masterGroupName: mymaster
```
rahulreddy, line 14: I think it's ok if we switch master name to `zenko`. Sentinel should allow other slaves to update their configuration and interpret this as the new master name. This should be tested obviously.

rahulreddy (commented):
LGTM in general. I added a few comments where I had concerns. There are a few references where port numbers like 6379 are hardcoded, can they be amended to use values?

charts/redis-ha/templates/redis-ha-configmap.yaml
rahulreddy: Yeah actually that's true. We are good then.

charts/redis-ha/templates/redis-ha-statefulset.yaml
ssalaues: You're right they should be able to blacklist. I'll remove it and test.

charts/redis-ha/values.yaml
ssalaues: I tested this and didn't work (seamlessly at least) because kubernetes updates the pods in reverse order. So when the redis-ha-2 goes down and comes back up, it tries to contact the master in the `zenko` group that does not exist and goes into crashloop state preventing the remaining pods from updating. However deleting all the pods manually does result in the intended group change.

The default that we would be upgrading from on previous chart is `mymaster` so it might be safest to leave it like this.

rahulreddy: Ah damn, thanks for testing. Let's leave it as-is.

giacope (approved):
/lgtm

rahulreddy (approved):

charts/redis-ha/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+[Redis](http://redis.io/) is an advanced key-value cache and store. It is often referred to as a data structure server since keys can contain strings, hashes, lists, sets, sorted sets, bitmaps and hyperloglogs.
+
+## TL;DR;
+
+```bash
+$ helm install stable/redis-ha
+```
+
+By default this chart install one master pod containing redis master container and sentinel container, 2 sentinels and 1 redis slave.
```
dora-korpar, line 11: installs

charts/redis-ha/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+## Installing the Chart
+
+To install the chart
+
+```bash
+$ helm install stable/redis-ha
+```
+
+The command deploys Redis on the Kubernetes cluster in the default configuration. By default this chart install one master pod containing redis master container and sentinel container along with 2 redis slave pods each containing their own sentinel sidecars. The [configuration](#configuration) section lists the parameters that can be configured during installation.
```
dora-korpar, line 30: installs 😬

ssalaues: Welp looks like @wabernatScality is gonna have his work cut out for him on this one

dora-korpar (approved):

PR #265, opened 2018-08-13
feat(ZENKO-472): Setup nightly cosbench run

tmacro:
REVIEW ONLY
This PR needs to be rebased against development/0.9 when #258 is merged to iron out any inconsistencies between their respective tests/Makefile, as I have been developing them in parallel. Other than that this PR should be ready to go!

### DIFF
--- a/eve/workers/cosbench/workload.short.xml.tmpl
+++ b/eve/workers/cosbench/workload.short.xml.tmpl
@@ -13,7 +13,7 @@
         </workstage>
 
         <!-- Write 0 Byte -->
-        <workstage name="write_0B_1T">
+        <!-- <workstage name="write_0B_1T">
             <work name="write_0B_1T" workers="1" runtime="300" division="object">
                 <operation type="write" ratio="100" config="sizes=c(0)MB;containers=r(1,1);objects=r(100001,200000);"/>
             </work>
@@ -22,10 +22,10 @@
             <work name="write_0B_256T" workers="256" runtime="300" division="object">
                 <operation type="write" ratio="100" config="sizes=c(0)MB;containers=r(1,1);objects=r(100001,200000);"/>
             </work>
-        </workstage>
+        </workstage> -->
 
         <!-- Read 0 Bytes -->
-        <workstage name="read_0B_1T">
+        <!-- <workstage name="read_0B_1T">
             <work name="read_0B_1T" workers="1" runtime="300" division="object">
                 <operation type="read" ratio="100" config="containers=r(1,1);objects=r(1,100000);"/>
             </work>
@@ -34,10 +34,10 @@
             <work name="read_0B_256T" workers="256" runtime="300" division="object">
                 <operation type="read" ratio="100" config="containers=r(1,1);objects=r(1,100000);"/>
             </work>
-        </workstage>
+        </workstage> -->
 
         <!-- Mixed 0B -->
-        <workstage name="mixed_0B_1T">
+        <!-- <workstage name="mixed_0B_1T">
             <work name="mixed_0B_1T" workers="1" runtime="300" division="object">
                 <operation type="read" ratio="50" config="containers=r(1,1);objects=r(1,100000);"/>
                 <operation type="write" ratio="50" config="sizes=c(0)MB;containers=r(1,1);objects=r(100001,200000);"/>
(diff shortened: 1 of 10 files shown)


tmacro:
@bert-e wait

eve/workers/cosbench/workload.short.xml.tmpl
```diff
@@ -0,0 +1,63 @@
...
+
+        <!-- Create 0 Byte Objects -->
+        <workstage name="prepare">
+            <work type="prepare" workers="512" config="containers=r(1,1);objects=r(1,100000);sizes=c(0)MB;" />
+        </workstage>
+
+        <!-- Write 0 Byte -->
+        <workstage name="write_0B_1T">
+            <work name="write_0B_1T" workers="16" runtime="300" division="object">
+                <operation type="write" ratio="100" config="sizes=c(0)MB;containers=r(1,1);objects=r(1,100000);"/>
```
anurag4DSB: I would recommend using different ranges for reads and writes.
For Example: prepare and read can be: (1,100000) and write can be (100001,200000)

eve/workers/cosbench/workload.short.xml.tmpl
```diff
@@ -0,0 +1,63 @@
...
+            </work>
+        </workstage>
+        <workstage name="write_0B_256T">
+            <work name="write_0B_256T" workers="256" runtime="300" division="object">
+                <operation type="write" ratio="100" config="sizes=c(0)MB;containers=r(1,1);objects=r(1,100000);"/>
+            </work>
+        </workstage>
+
+        <!-- Read 0 Bytes -->
+        <workstage name="read_0B_1T">
```
anurag4DSB: I would also suggest changing the workers to 64 and 512.
@vrancurel  What do you think?

anurag4DSB (changes requested):

eve/workers/cosbench/workload.short.xml.tmpl
vrancurel: Can we do 1, 8, 32, 64, 128, 256, 512 ?

eve/workers/cosbench/workload.short.xml.tmpl
anurag4DSB: and eventually, you can have 2 work stages for cleanup
```
<work type="cleanup" workers="512" config="containers=r(1,1);objects=r(1,100000)" />
<work type="cleanup" workers="512" config="containers=r(1,1);objects=r(100001,200000)" />
```

eve/workers/cosbench/workload.short.xml.tmpl
giacope: @vrancurel As of right now, `Eve` can't run long enough to test all those scenarios. Max is 2 I think.

ssalaues: @vrancurel There's a time limit of 1 hour on the CI so with the setup, tests, and 2 stages we already get very close to this limit which is why we decided for only 2 stages to start with. It's currently a limitation of Eve's access to the GKE cluster that I believe releng is working on.

tmacro: @vrancurel The long workload I made is in `workload.xml.tmpl`. it uses 1,2,4,8,16,32,64,128,256,512,1024. When we get it setup with a proper metalk8s and Zenko deploy, we'll be able to switch.

eve/workers/cosbench/workload.short.xml.tmpl
```diff
@@ -0,0 +1,64 @@
...
+
+        <!-- Create 0 Byte Objects -->
+        <workstage name="prepare">
+            <work type="prepare" workers="1024" config="containers=r(1,1);objects=r(1,100000);sizes=c(0)MB;" />
+        </workstage>
+
+        <!-- Write 0 Byte -->
+        <workstage name="write_0B_1T">
+            <work name="write_0B_1T" workers="1" runtime="300" division="object">
+                <operation type="write" ratio="100" config="sizes=c(0)MB;containers=r(1,1);objects=r(100000,200000);"/>
```
anurag4DSB: Ideally, it should be (100001,200000)

anurag4DSB:
I understand we can add only 2 types of threads given the contraints of 1 hour on the CI. I would still suggest confirming with @vrancurel which numbers he wants for the current 2 workload setup. Currently its 1 thread and 256 thread.

anurag4DSB (approved):

PR #274, opened 2018-08-15
ZENKO-1003 bugfix: Move dump-logs to seperate ShellCommand

tmacro:


rahulreddy (approved):

giacope (approved):

ssalaues (approved):

ssalaues:
@bert-e wait
just so it doesn't auto merge with test commit but feel free to delete the comment when ready

eve/ci-values.yml
```diff
@@ -45,7 +45,7 @@ global:
 cloudserver:
   image:
     pullPolicy: Always
-  replicaCount: 0
+  replicaCount: 3
```
ssalaues: Maybe should add `replicaFactor: 1` while we're at it otherwise it's 30 cloud servers in CI which is 30 more logs to sift through

ssalaues:
Also I'm pretty sure the build failed because `global.orbit.workerMode` is set to push mode. So all the 30 cloud servers that spun up in the previous build were trying to push to orbit simulator.

PR #278, opened 2018-08-16
ZENKO-1010 refactor: zenko charts structure

giacope:
**This change is upgradable.**

So basically the `charts` directory has been renamed to `kubernetes` and it will now look like this:

 └──     kubernetes/
 │  ├────     README.md
 │  ├────     setup-no-orbit.md
 │  ├────     single-node-values.yml
 │  └────     zenko/
 │  │  ├────     Chart.yaml
 │  │  └────     charts/
 │  │  │  └────     backbeat/
 │  │  │  └────     cloudserver/
 │  │  │  └────     grafana/
 │  │  │  └────     mongodb-replicaset/
 │  │  │  └────     prometheus/
 │  │  │  └────     redis-ha/
 │  │  │  └────     s3-data/
 │  │  │  └────     zenko-nfs/
 │  │  │  └────     zenko-queue/
 │  │  │  └────     zenko-quorum/
 │  │  └────     templates/
 │  │  │  ├────     NOTES.txt
 │  │  │  ├────     _helpers.tpl
 │  │  │  ├────     datasource.yaml
 │  │  │  ├────     ingress.yaml
 │  │  │  └────     secrets.yaml
 │  │  └────     values.yaml

With this new format, the command `helm dependency build zenko` will **NEVER** have to be used again. There is no need to build dependencies since they are all present.
It will now be possible to install Zenko with only **ONE** command: `helm install zenko`.


ssalaues:
@bert-e wait

ssalaues (approved):
Overall I like this be we should be sure to thoroughly test this before merging. Like have several people test upgrades, fresh installs, update documentation, etc.

kubernetes/zenko/charts/zenko-nfs/values.yaml
```diff
@@ -1,6 +1,9 @@
 # Default values for zenko-nfs.
 # This is a YAML-formatted file.
 # Declare variables to be passed into your templates.
+
+enabled: false
```
ssalaues: Don't need at this level

ssalaues (changes requested):
One little change

kubernetes/zenko/charts/zenko-nfs/values.yaml
giacope: It does in order to pass `helm lint --strict`. Changed default value to `true`.

ssalaues (approved):

tmacro (approved):

PR #279, opened 2018-08-17
ZENKO-1004: fix flaky nodejs E2E tests

JianqinWang:
Here are the steps in this PR that will hopefully fix some of the known flakiness in Zenko tests:

- use different  `sourceBucket` names for each test, to avoid conflicts during the tear down in afterEach
- use `uuid` instead of `Date.now()` to generate unique bucket names
- raise wait timer between checking replication statuses for the crr pause resume tests
- Skipping MPU replication by range tests in 1-M scenarios, ticket for that test is opened here: https://scality.atlassian.net/browse/ZENKO-1015

### DIFF
--- a/tests/node_tests/backbeat/ReplicationUtility.js
+++ b/tests/node_tests/backbeat/ReplicationUtility.js
@@ -627,21 +627,12 @@ class ReplicationUtility {
 
     compareObjectsAWS(srcBucket, destBucket, key, optionalField, cb) {
         return async.series([
-            next => this.waitUntilReplicated(srcBucket, key, undefined, err => {
-                console.log('compareObjectsAWS:waitUntilReplicated', err);
-                return next(err);
-            }),
-            next => this.getObject(srcBucket, key, (err, data) => {
-                console.log('compareObjectsAWS:getObject', err, data);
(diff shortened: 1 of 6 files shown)


ssalaues (approved):

JianqinWang:
@bert-e wait

bennettbuchanan:
Is it proven that a source bucket name conflict is the cause for flakiness? For example, this is still failing on the flaky test we were tracking. There is, of course, the case that the source bucket is created at the exact same millisecond across tests, but I would guess it to be so rare as to not show up as noticeable flakiness. The beforeEach and afterEach should not have race conditions within the same describe statement.

As you've done here, using a `uuid` is probably indeed a more reliable option than using a timestamp, but again I'm not sure it is the root cause.

JianqinWang:
You're right, the test is still failing. Other than bucket name conflicts, is there anything that comes to your mind as to what can cause this issue?

bennettbuchanan:
My initial thought is that the `NoSuchBucket` error is coming from the public clouds (i.e. for these tests either AWS or GCP). It's unclear to me why that would be the case though. It will require further investigation.

PR #280, opened 2018-08-17
Bugfix/zenko 1017 redis fixes

ssalaues:
Fixes:

- Ability to upgrade redis in general (at the helm level)
- When persistence is disabled, the init script would not copy over the config when needed
    - Also adds more stability to the chart

Feature:

- Switched to parallel pod management for quicker startup and cleanups.
    - Does not affect upgrades which are still rolling upgrades
    - Allows us to change the redis master group to `zenko` instead of `mymaster`

### DIFF
--- a/charts/redis-ha/Chart.yaml
+++ b/charts/redis-ha/Chart.yaml
@@ -5,7 +5,7 @@ keywords:
 - redis
 - keyvalue
 - database
-version: 3.0.0-Z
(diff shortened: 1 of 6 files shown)


ssalaues:
@bert-e wait

giacope (approved):

tmacro (approved):

dora-korpar (approved):

PR #281, opened 2018-08-20
ZENKO-1016 Improvement: Add tagging to flaky node e2e tests

tmacro:
To help improve the CI, add tags to flaky test. These test will be skipped during normal test runs, to help with the endless rebuilding, and will be run during the nightly build to prevent loss of test coverage.

### DIFF
--- a/tests/node_tests/backbeat/tests/crr/oneToMany.js
+++ b/tests/node_tests/backbeat/tests/crr/oneToMany.js
@@ -27,6 +27,7 @@ const copyKey = `${key}-copy`;
 const copySource = `/${srcBucket}/${key}`;
 const REPLICATION_TIMEOUT = 300000;
 
+
 describe('Replication with AWS, Azure, and GCP backends (one-to-many)',
 function() {
     this.timeout(REPLICATION_TIMEOUT);
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -228,7 +228,7 @@ test-node-flaky:
 	$(call test-node)
 
 test: | install-common install-zenko test-e2e test-node dump-logs
-test-latest: | install-common get-latest-commit install-latest-zenko test-e2e test-node-flaky dump-logs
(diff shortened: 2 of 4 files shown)


anurag4DSB (approved):

ssalaues (approved):

tests/node_tests/backbeat/tests/crr/oneToMany.js
```diff
@@ -27,6 +27,7 @@ const copyKey = `${key}-copy`;
 const copySource = `/${srcBucket}/${key}`;
 const REPLICATION_TIMEOUT = 300000;
 
+
```
bennettbuchanan, line 30: Seeing only an added line in this file, so wondering if we meant to add a tag to these tests as well.

bennettbuchanan (commented):
I agree that we should not be slowed by test flakiness, so it makes sense that we take measures to address it. For example, previously we had been adding retries as a strategy to address this problem. However, my concern would be that running only once may conceal a larger problem with the functionality. That is, if something is merged during the day that actually breaks this test entirely (or increases it's flakiness), then when this test fails in the evening it would be easy to dismiss it as test flakiness. Partly the question is: do we know why tagged test are flaky, and if not, how will we follow up to make the test suite reliable?

ssalaues:
@bert-e wait

ssalaues:
@bennettbuchanan you're right we should be able to track these flaky tests instead of losing them to the ether. Also the idea is that we're going to run the nightly tests several times in attempt to measure the flakiness.

@tmacro can you create a ticket for the flaky tests and add a comment to the tagged test pointing to the ticket?

tmacro:
@bennettbuchanan the envisioned workflow is, after a test is tagged flaky, a matching ticket is created to either fix the flakiness, or discover the bug causing it. This way the tests won't stay "flaky" for very long, nor be swept under the rug as someone is actively investigating it.

bennettbuchanan (approved):

PR #283, opened 2018-08-20
bf: ZENKO-1013 use backbeat api metrics routes for stats

alexanderchan-scality:


alexanderchan-scality:
https://eve.devsca.com/github/scality/zenko/#/builders/4/builds/2804

PR #287, opened 2018-08-21
ZENKO-1026 move manager out of cloudserver service

ssalaues:
This takes out the manager pod from the cloudserver service so it doesn't have to handle requests along side manager duties.

Note: As pointed out, the manager pod will still service requests that originate from secure channel (browser, search, CRR ops, etc)

### DIFF
--- a/kubernetes/zenko/charts/cloudserver/templates/manager-deployment.yaml
+++ b/kubernetes/zenko/charts/cloudserver/templates/manager-deployment.yaml
@@ -4,7 +4,7 @@ kind: Deployment
 metadata:
   name: {{ template "cloudserver.fullname" . }}-manager
   labels:
-    app: {{ template "cloudserver.name" . }}
+    app: {{ template "cloudserver.name" . }}-manager
     chart: {{ template "cloudserver.chart" . }}
     release: {{ .Release.Name }}
     heritage: {{ .Release.Service }}
@@ -12,7 +12,7 @@ spec:
   replicas: 1
   selector:
     matchLabels:
-      app: {{ template "cloudserver.name" . }}
+      app: {{ template "cloudserver.name" . }}-manager
       release: {{ .Release.Name }}
   template:
     metadata:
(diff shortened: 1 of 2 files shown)


kubernetes/zenko/charts/cloudserver/templates/manager-deployment.yaml
```diff
@@ -4,15 +4,15 @@ kind: Deployment
 metadata:
   name: {{ template "cloudserver.fullname" . }}-manager
   labels:
-    app: {{ template "cloudserver.name" . }}
```
giacomoguiulfo: I think we should start using partials for the different component names that a chart can have.

kubernetes/zenko/charts/cloudserver/templates/manager-deployment.yaml
```diff
@@ -4,15 +4,15 @@ kind: Deployment
 metadata:
   name: {{ template "cloudserver.fullname" . }}-manager
```
giacope, line 5: Maybe we should start using partials for the different component names that a chart can have. Something like: `{{ template "cloudserver.manager.fullname" . }}

rahulreddy (approved):

giacomoguiulfo:
@bert-e wait

kubernetes/zenko/charts/cloudserver/templates/manager-deployment.yaml
ssalaues: Yeah I think that's a good idea but it maybe overkill for the cloudserver chart. I think it makes more sense for something like the backbeat chart which is expected to have many required components and references to these components. Whereas in this case this is simply an optional deployment.

anurag4DSB (approved):

kubernetes/zenko/charts/cloudserver/templates/manager-deployment.yaml
giacope: We can add it to the `TODO` list then, for when the charts are refactored.

giacope (approved):

ssalaues:
Updated the ci-values with the `workerMode: poll` to work with the CI

rachedbenmustapha:
@ssalaues for completeness, note that the manager pod will still service requests that originate from secure channel (browser, search, CRR ops, etc), but I think that's perfectly fine.

ssalaues:
@bert-e reset

PR #291, opened 2018-08-22
ft: retain s3-data pvc on helm delete

ssalaues:
Prevents helm from deleting the s3-data pvc automatically on `helm delete`


tmacro (approved):
less is more :)

JianqinWang (approved):

PR #294, opened 2018-08-23
ZENKO-965 bugfix: incorrect Grafana password

giacope:
The issue of Grafana having the wrong password is most likely due to https://github.com/helm/charts/issues/2627. These changes add a checksum annotation to the deployment, which increases the chances of the password stored in the secret been the same one stored on the Grafana server. However, it's not a guarantee.
The only way to guarantee it is by specifying a password on the values.yaml file.


tmacro (approved):

giacope:
@tmacro closing this in favor of #297

PR #298, opened 2018-08-24
bugfix: ZENKO-1024 Add pending counters tests

bennettbuchanan:


ssalaues:
@bert-e wait

ssalaues:
@bennettbuchanan this branch is 89 commits behind can you fetch the latest and rebase please

bennettbuchanan:
@ssalaues I've rebased this 😸

tests/node_tests/backbeat/tests/retry/pendingMetrics.js
```diff
@@ -0,0 +1,162 @@
...
+
+function getPendingPath(location) {
+    const pathPrefix = '/_/backbeat/api/metrics/crr';
+    return `${pathPrefix}/${location}/pending`;
+}
+
+describe('Backbeat API pending metrics', function() {
+    this.timeout(REPLICATION_TIMEOUT);
+    const roleArn = 'arn:aws:iam::root:role/s3-replication-role';
+    const expectedBody = {
```
dora-korpar: Probably not a big issue since `expectedBody.pending.results` is set for every test, but it would be better practice to use `Object.assign` to copy `expectedBody` in each of the tests, so they aren't all accessing a global variable that could get mutated

tests/node_tests/backbeat/tests/retry/retry.js
```diff
@@ -20,14 +20,17 @@ const keyPrefix = `${srcBucket}/${hex}`;
 const key = `object-to-replicate-${Date.now()}`;
 const REPLICATION_TIMEOUT = 1200000;
```
dora-korpar: 20-min timeout O.O

alexanderchan-scality (approved):

tests/node_tests/backbeat/tests/retry/pendingMetrics.js
bennettbuchanan: Okay, I agree. To avoid this I've moved it to the helper method where I pass in the expected results. This also has the advantage of reducing some code.

dora-korpar (approved):

tests/node_tests/backbeat/tests/retry/retry.js
bennettbuchanan: I believe this is leftover from running the test suite without updating the exponential backoff in backbeat's config. The goal is to not create tests that take so long (and instead use a custom backbeat image with an updated config) so I've reset this to 5 minutes like the other replication timeout constants.

ssalaues (approved):

rahulreddy (approved):

bennettbuchanan:
I've now validated that these tests are passing with RC6.

PR #303, opened 2018-08-27
ZENKO-1065 feat: add CRR parallelism factor

giacope:


vrancurel:
I would default to a rep factor of 5 but let's see what happens with 2 first

vrancurel (approved):

alexanderchan-scality (approved):

ssalaues (approved):

rahulreddy (approved):

rahulreddy:
So 2 is what we will have in par with S3C. We will come up with a better value after spending reasonable effort on optimizing performance.

giacope:
@bert-e help

PR #304, opened 2018-08-27
improvement: ZENKO-1004 Write file on MD5 mismatch

bennettbuchanan:
When there is an MD5 content mismatch for replicated objects, write the file contents to the build artifacts for further investigation.

### DIFF
--- a/tests/node_tests/backbeat/ReplicationUtility.js
+++ b/tests/node_tests/backbeat/ReplicationUtility.js
@@ -21,15 +21,15 @@ class ReplicationUtility {
     _compareObjectBody(body1, body2) {
         const digest1 = crypto.createHash('md5').update(body1).digest('hex');
         const digest2 = crypto.createHash('md5').update(body2).digest('hex');
-        // if (digest1 !== digest2) {
-        //     // dump data for later investigation
-        //     const filePrefix = `${process.env.CIRCLE_ARTIFACTS}/` +
-        //               `genericStaas_backbeat_md5_mismatch_body`;
-        //     fs.writeFileSync(`${filePrefix}1.bin`, body1);
-        //     fs.writeFileSync(`${filePrefix}2.bin`, body2);
-        //     console.error('md5 mismatch: data dumped in ' +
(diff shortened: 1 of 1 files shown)


tmacro (approved):

ssalaues (approved):

ssalaues:
@bert-e wait

Lets just double check the files are uploaded to the expected place. If it does, feel free to delete this comment.

bennettbuchanan:
@bert-e wait

tests/node_tests/backbeat/ReplicationUtility.js
```diff
@@ -21,15 +21,15 @@ class ReplicationUtility {
...
-        //     // dump data for later investigation
-        //     const filePrefix = `${process.env.CIRCLE_ARTIFACTS}/` +
-        //               `genericStaas_backbeat_md5_mismatch_body`;
-        //     fs.writeFileSync(`${filePrefix}1.bin`, body1);
-        //     fs.writeFileSync(`${filePrefix}2.bin`, body2);
-        //     console.error('md5 mismatch: data dumped in ' +
-        //                   `${filePrefix}{1,2}.bin`);
-        // }
+        if (digest1 !== digest2) {
+            // dump data for later investigation
```
rahulreddy, line 25: It will be helpful to log the reqUids as well.

rahulreddy (commented):
Btw this is not a bugfix, more like an investigative effort...improvement then?

tests/node_tests/backbeat/ReplicationUtility.js
bennettbuchanan: Isn't the reqUid only available in the codebase through werelogs?

rahulreddy: It's also written as a response headers `x-amz-request-id` and `x-amz-id-2`.

bennettbuchanan:
Closing this for now since I don't think we've determined that the files are being written to the correct place. This kind of change would be helpful for debugging, in particular for the intermittent socket timeout errors we had been seeing. We can always restore, if needed.

PR #308, opened 2018-08-29
improvement: create bootstrap lists from global or replica count

ssalaues:
Allows for the backbeat chart and cloudserver chart to create bootstrap lists of all statefulsets based off global `nodeCount` or `replica` count if custom configuration is needed but default to the global `nodeCount`

### DIFF
--- a/kubernetes/zenko/charts/backbeat/templates/_helpers.tpl
+++ b/kubernetes/zenko/charts/backbeat/templates/_helpers.tpl
@@ -35,16 +35,34 @@ Create chart name and version as used by the chart label.
 Create the default mongodb replicaset hosts string
 */}}
 {{- define "backbeat.mongodb-hosts" -}}
-{{- $count := (atoi (printf "%d" (int64 .Values.mongodb.replicas))) -}}
+{{- $count := (int ( default .Values.global.nodeCount .Values.mongodb.replicas)) -}}
 {{- $release := .Release.Name -}}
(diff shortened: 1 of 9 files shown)


kubernetes/zenko/charts/backbeat/templates/_helpers.tpl
```diff
@@ -35,16 +35,34 @@ Create chart name and version as used by the chart label.
...
-{{- $count := (atoi (printf "%d" (int64 .Values.mongodb.replicas))) -}}
+{{- $count := (int ( default .Values.global.nodeCount .Values.mongodb.replicas)) -}}
 {{- $release := .Release.Name -}}
 {{- range $v := until $count }}{{ $release }}-mongodb-replicaset-{{ $v }}.{{ $release }}-mongodb-replicaset:27017{{ if ne $v (sub $count 1) }},{{- end -}}{{- end -}}
 {{- end -}}
 
+{{/*
+Create default queue hosts string
+*/}}
+{{- define "backbeat.queue" -}}
```
jonathan-gramain: What about `backbeat.queue-hosts`, `backbeat.quorum-hosts` to precise things?

kubernetes/zenko/charts/backbeat/values.yaml
```diff
@@ -19,18 +20,24 @@ logging:
...
-  replicas: 3
+  replicas:
 
 health:
   port: 4042
   path:
     liveness: /_/health/readiness
 
+queue:
+  replicas:
```
giacope, line 36: Shouldn't this have a default value?

giacope (changes requested):
Needs chart version bump

ssalaues:
Closing this PR. A better solution would be to modify the entry point scripts of the services that require bootstrap lists.

PR #311, opened 2018-08-30
ZENKO-1098 improvement: update zenko-queue chart

giacope:
### This PR
- Updates `zenko-queue` chart from version `0.7.2` to `0.9.1` (latest upstream), which includes:
	- Topic management through the values.yaml.
	- Upgrades metrics exporter from `v1.0.1` to `v1.2.0`.  
	- Other small features/bugfixes (i.e. init-container failures, headless service annotations, secret management).
- Disables `auto-creation-of-topics`, and configures the necessary ones (`backbeat-replication`, `backbeat-metrics`, etc.).
- Increases Kafka's global (for all topics) max message size from 1MB to 5MB.


giacomoguiulfo:
@bert-e wait

anurag4DSB:
pre-merge failed, trying to rebuild.

tmacro (approved):

giacomoguiulfo:
@bert-e wait

kubernetes/zenko/values.yaml
```diff
@@ -117,17 +117,42 @@ zenko-queue:
...
   prometheus:
     jmx:
       enabled: true
     kafka:
       enabled: true
-zookeeper:
-  enabled: false # Disables zenko-queue from deploying its own quorum
+  topics:
+    - name: backbeat-gc
+      partitions: *crrParallelism
```
jonathan-gramain: If I'm not mistaken we're changing the number of partitions here, shall be reverted IMHO to the original value `partitionCount * partitionFactor`

jonathan-gramain (changes requested):

kubernetes/zenko/values.yaml
giacope: I don't really see the point of `partitionCount * partitionFactor`. Before we were doing it in a sense of abstraction because we were setting partitions in a more global scope. But this gives you more freedom and the ability to do it per topic. I kinda only makes sense to have the `factor` for the data-processor topic.

anurag4DSB (approved):

vrancurel (commented):
please keep previous kafka version

kubernetes/zenko/values.yaml
giacope: Also, I think we should prevent the use of `Yaml` anchors and helm's templating facilities to have "factor" variables as much as we can. Both things are not a "thing" in the Helm community.

giacope:
@vrancurel Done

ssalaues (approved):

kubernetes/zenko/values.yaml
```diff
@@ -117,17 +114,42 @@ zenko-queue:
...
   prometheus:
     jmx:
       enabled: true
     kafka:
       enabled: true
-zookeeper:
-  enabled: false # Disables zenko-queue from deploying its own quorum
+  topics:
+    - name: backbeat-gc
+      partitions: 2
```
jonathan-gramain: What is the rationale for setting 2 as the default value? Why not `*nodeCount` as it used to be or `*replicaCount * *replicaFactor` to have one partition per running process, which seems a more reasonable default to me?

giacope: Good point! Yes I meant `*nodeCount`.
That said, `*replicaCount * *replicaFactor` (or any sort of calculation that requires templating) will lead us into `yaml/helm-hell` and will require complicated, zenko-specific changes to the Zenko-Queue chart, so I don't consider it a possibilty.

jonathan-gramain: OK, still not great as the number of partitions should be *at least* the number of processes, but I get your concerns on manageability so we can start with this, but I think we will have to iterate again at one point.

jonathan-gramain (approved):

kubernetes/zenko/values.yaml
```diff
@@ -117,17 +114,42 @@ zenko-queue:
...
+      partitions: *nodeCount
+      replicationFactor: 3
+    - name: backbeat-lifecycle-object-tasks
+      partitions: *nodeCount
+      replicationFactor: 3
+    - name: backbeat-lifecycle-bucket-tasks
+      partitions: *nodeCount
+      replicationFactor: 3
+    - name: backbeat-metrics
+      partitions: 1
```
jonathan-gramain: why not `*nodeCount` here?

giacope: @jonathan-gramain `BackbeatAPI` does not depend on `nodeCount` since it's preferred for it to only have 1 instance. So it is like this to follow the `"to have one partition per running process"` statement.

giacope: Unless some other process uses this topic besides API?

PR #325, opened 2018-09-11
Revert "ZENKO-1134 [zenko-queue]: cleanup job when it succeeds"

ssalaues:
This reverts commit 9e3d3ee5d0d58655ace7e64ab61f85a7b18c774f.

Bug was introduced with the above commit that can cause helm installs to fail as a result of helm waiting for queue job to finish (exceeds helm timeout).

fixes ZENKO-1134

We will revert the commit and the original issue addressed can be handled in a separate PR.

### DIFF
--- a/kubernetes/zenko/charts/zenko-queue/templates/job-config.yaml
+++ b/kubernetes/zenko/charts/zenko-queue/templates/job-config.yaml
@@ -9,10 +9,6 @@ metadata:
     chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
     heritage: "{{ .Release.Service }}"
     release: "{{ .Release.Name }}"
-  annotations:
-    "helm.sh/hook": post-install,post-upgrade,post-rollback
-    "helm.sh/hook-weight": "-5"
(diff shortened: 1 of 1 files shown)


giacope (approved):

rahulreddy (approved):
I think it's ok for rc7

rahulreddy:
I tried rebuilding the q branch it seems to go on forever and there are errors reported in the logs https://eve.devsca.com/github/scality/zenko/#/builders/4/builds/3348

ssalaues:
I think these took longer because the retries on the python test suite have a backoff retry mechanism. So likely there were several reruns but eventually passed. 

From the error, it looks like if failed because of an Eve timeout. Specifically the time limit on how long Eve can have access to a kubernetes cluster. We need this PR too because the CI wont helm install properly either so any other PRs will fail. We're trying to figure out how it passed in the first place (maybe some kind of caching?)

philipyoo (approved):

rahulreddy:
@ssalaues I think it was caching the previous build? I did force it though (not rebuild), so there should be no caching in play.

PR #329, opened 2018-09-13
bugfix: redis chart corner cases

ssalaues:
This fixes issues where redis servers can potentially get partitioned after failover and recovery or in certain cases after upgrades. These changes end up making the redis service more robust, better at failovers, and further simplifies implementation.

fixes # ZENKO-1145

- The relevant fixes themselves are within the configmap
https://github.com/scality/Zenko/pull/329/files#diff-0dd094d379f8ed1f968171377c597949
- There's some fluff being imported here as well (like the README changes) because of the way the upstream chart need to be setup.
- Changes made with input from the helm chart community.
- Lots of reviews and testing along with a user that installed the previous version of this and seemed to have encountered the specific issue this now fixes https://github.com/helm/charts/pull/7323
- All this code already exists in the scality/charts fork as well https://github.com/scality/charts/tree/redis-ha-refactor
- The custom values we had originally in the redis-ha/values.yaml have been moved to the zenko/values.yaml context

I believe these changes bring the redis-ha chart into a much more production ready state.

### DIFF
--- a/kubernetes/zenko/charts/redis-ha/values.yaml
+++ b/kubernetes/zenko/charts/redis-ha/values.yaml
@@ -4,18 +4,20 @@
 
 enabled: true
 
-image: redis
-tag: 4.0.11-stretch
+image:
+  repository: redis
+  tag: 4.0.11-stretch
+  pullPolicy: IfNotPresent
 ## replicas number for each component
 replicas: 3
 
 ## Redis specific configuration options
 redis:
   port: 6379
-  masterGroupName: zenko
-  auth: false          # Not yet implemented - Configures redis with AUTH (requirepass & masterauth conf params)
-  # Any additional redis conf settings can be set as key value pairs as seen below
+  masterGroupName: mymaster
   config:
+    ## Additional redis conf options can be added below
+    ## For all available options see http://download.redis.io/redis-stable/redis.conf
     min-slaves-to-write: 1
     min-slaves-max-lag: 5   # Value in seconds
     maxmemory: "0"       # Max memory to use for each redis instance. Default is unlimited.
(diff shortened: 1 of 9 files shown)


kubernetes/zenko/charts/redis-ha/values.yaml
```diff
@@ -6,16 +6,17 @@ enabled: true
 
 image: redis
 tag: 4.0.11-stretch
+pullPolicy: IfNotPresent
```
giacope: This would look better for this set of values:
```yaml
image:
  repository: redis
  tag: 4.0.11-stretch
  pullPolicy: IfNotPresent
```
For reference, that's how image values look when creating a chart with `helm create` and therefore I would recommend doing so in order to keep a common format among the charts.

kubernetes/zenko/charts/redis-ha/values.yaml
```diff
@@ -59,30 +63,17 @@ securityContext:
...
-## Defaults to a random 10-character alphanumeric string if not set and auth is true
-## ref: https://github.com/kubernetes/charts/blob/master/stable/redis-ha/templates/redis-auth-secret.yaml
-##
-## redisPassword:
+# redisPassword:
+auth: false
 
 persistentVolume:
-  enabled: true
+  enabled: false
```
giacope: As per the review guidelines https://github.com/helm/charts/blob/master/REVIEW_GUIDELINES.md
"Persistence should be enabled by default"

ssalaues:
Tested an upgrade from RC7 and it was able to correct a partitioned redis. (Along with seamless upgrade)

kubernetes/zenko/charts/redis-ha/values.yaml
ssalaues: The two issues have been resolved. Thank you!

giacope (approved):
/lgtm

tmacro (approved):

PR #340, opened 2018-09-19
ci: Build documentation in pipeline

NicolasT:
Making sure the documentation can be built at any point in time is useful. Let's enforce this in CI.

The infrastructure was borrowed from the MetalK8s project (ref d9c18547f3e2240bada58db5126ecce8d0d58fa5).

Also, enable the `sphinxcontrib-spelling` extension.

See: https://github.com/scality/metalk8s


NicolasT:
Hmh, odd CI failures... Maybe because the version of Eve running on this repository is too old. Any insights, @tcarmet ?

eve/main.yml
```diff
@@ -104,11 +104,31 @@ stages:
...
         stage_names:
         - test-zenko
 
+  build-doc:
+    worker:
+      type: kube_pod
+      path: eve/workers/doc-builder.yaml
+      images:
+        doc-builder:
+          context: '.'
```
ssalaues, line 123: For some reason it looks like Eve is having issues with reading the `context`

tcarmet: yeah my bad, there's an old version of eve on zenko. I'm updating it!

tcarmet: update done! Working as expected now :)

NicolasT: Thanks @tcarmet!

NicolasT:
I made some changes to the actual docs to make the tree actually build.

These may not be the changes @wabernatScality actually wants, so may certainly need to be fixed up later.

However! Despite this, I'd argue getting this integrated and ensuring any later edits actually build is more important than keeping non-building docs around.

wabernat (approved):

smaffulli:
Absolutely, we have to have this. I just pulled the latest development/1.0 branch and docs don't build cleanly.

ssalaues (approved):

PR #341, opened 2018-09-20
[redis-ha]: remove memory/cpu limits for redis

ssalaues:
Removes hard memory limit on redis

fixes ZENKO-1189

### DIFF
--- a/kubernetes/zenko/resources.yaml
+++ b/kubernetes/zenko/resources.yaml
@@ -64,13 +64,11 @@ prometheus:
    resources: {}
 
 redis-ha:
-  server:
+  redis:
     resources:
       requests:
         memory: 200Mi
         cpu: 100m
-      limits:
-        memory: 700Mi
   sentinel:
     resources:
       requests:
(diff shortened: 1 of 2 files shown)


anurag4DSB (approved):

kubernetes/zenko/resources.yaml
```diff
@@ -64,13 +64,12 @@ prometheus:
    resources: {}
 
 redis-ha:
-  server:
```
rahulreddy, line 67: Why is this changing from server => redis?

JianqinWang (approved):

kubernetes/zenko/resources.yaml
ssalaues: The file was out of date. `server` was how the old redis chart organized these values

kubernetes/zenko/values.yaml
```diff
@@ -164,6 +164,8 @@ redis-ha:
   replicas: *nodeCount
   redis:
     masterGroupName: zenko
+    resources:
+      limits: {}
```
giacope: This doesn't work. Try `""` on the children level.

giacope (changes requested):

kubernetes/zenko/values.yaml
ssalaues: Fixed with `null`

giacope (approved):

PR #347, opened 2018-09-21
bf: ZENKO-1168 add initial delay before liveness probes

jonathan-gramain:

Add a delay (default 2mn) before starting liveness probes on backbeat
and cloudserver pods. This prevents premature kill & restart when the
container is a bit slow to start.

fixes ZENKO-1168

### DIFF
--- a/kubernetes/zenko/charts/backbeat/templates/api/deployment.yaml
+++ b/kubernetes/zenko/charts/backbeat/templates/api/deployment.yaml
@@ -52,6 +52,7 @@ spec:
             httpGet:
               path: /_/healthcheck
               port: api
+            initialDelaySeconds: {{ .Values.health.initialDelaySeconds }}
           # TODO readinessProbe:
           resources:
 {{ toYaml .Values.api.resources | indent 12 }}
--- a/kubernetes/zenko/charts/cloudserver/values.yaml
+++ b/kubernetes/zenko/charts/cloudserver/values.yaml
@@ -64,6 +64,12 @@ logging:
 
 allowHealthchecksFrom: '0.0.0.0/0'
 
+health:
+  port: http
+  path:
+    liveness: /_/healthcheck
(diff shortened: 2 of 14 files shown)


kubernetes/zenko/charts/backbeat/templates/api/deployment.yaml
```diff
@@ -52,6 +52,7 @@ spec:
             httpGet:
               path: /_/healthcheck
```
ssalaues, line 53: Is it safer for this one to use a hard coded value vs. `.Values.health.path.liveness`?

ssalaues: Oh never mind, I see they are different now.

rahulreddy (commented):
Would adding readiness probes would be the right course of action instead of setting an initial delay?

giacope:
@rahulreddy I don't think so (at least for most of backbeat). Readiness probe is to prevent requests from reaching the pod until it's ready. The backbeat pods, except for the API, don't receive requests (except for health check ones).

kubernetes/zenko/charts/cloudserver/values.yaml
```diff
@@ -64,6 +64,12 @@ logging:
 
 allowHealthchecksFrom: '0.0.0.0/0'
 
+health:
+  port: http
```
ssalaues, line 68: I think just having the `initialDelaySeconds` configurable would suffice. Typically we don't want user configurable values that could break the deployment and being able to change the container port (which isn't configurable either). Same pretty much goes for the health check route because we want to ensure that the health route we want is always used.

I know the backbeat chart does it this way however this is will likely be removed soon as we are refactoring the backbeat chart.

ssalaues (changes requested):
Don't think we need a configurable port/path for the cloudserver liveness. Otherwise LGTM

PR #348, opened 2018-09-21
ZENKO 1187 sizing defaults

ssalaues:
Updates sizing defaults for 1.0 to more reasonable values that total to a 200GB per node minimum.

fixes ZENKO-1187

This breaks any upgrades to Zenko from previous versions. (Upgrades would require modifying the values.yaml to previous settings). It's best to do this prior to any production install as it will give us more peace of mind even on 'default' installs where a user may not diligently update the storage values to production ready settings.

### DIFF
--- a/kubernetes/single-node-values.yaml
+++ b/kubernetes/single-node-values.yaml
@@ -11,6 +11,10 @@ cloudserver:
   mongodb:
     replicas: *nodeCount
 
+s3-data:
+  persistentVolume:
+    size: 10Gi
+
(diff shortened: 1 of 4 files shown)


giacope (changes requested):
single node values?

giacope (approved):

anurag4DSB (approved):

alexanderchan-scality (approved):

rahulreddy (approved):

PR #375, opened 2018-10-03
improvement: ZENKO-1217 add tuneable to retry timeout

rahulreddy:
This tunable controls for how long (in seconds) a replication entry will
be retried before setting the object as failed

### DIFF
--- a/kubernetes/zenko/charts/backbeat/templates/replication/data_processor_deployment.yaml
+++ b/kubernetes/zenko/charts/backbeat/templates/replication/data_processor_deployment.yaml
@@ -98,6 +98,8 @@ spec:
               value: "{{ template "backbeat.redis-hosts" . }}"
             - name: REDIS_HA_NAME
               value: "{{ .Values.redis.sentinel.name }}"
+            - name: EXTENSIONS_REPLICATION_QUEUE_PROCESSOR_RETRY_TIMEOUT_S
+              value: "{{ .Values.replication.dataProcessor.retryTimeoutS }}"
           livenessProbe:
             httpGet:
               path: {{ .Values.health.path.liveness}}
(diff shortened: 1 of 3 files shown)


rahulreddy:
@bert-e wait

philipyoo (approved):

alexanderchan-scality (approved):

ssalaues:
Do we want this as a patch in 1.0?

kubernetes/zenko/charts/backbeat/templates/replication/data_processor_deployment.yaml
```diff
@@ -98,6 +98,8 @@ spec:
               value: "{{ template "backbeat.redis-hosts" . }}"
             - name: REDIS_HA_NAME
               value: "{{ .Values.redis.sentinel.name }}"
+            - name: EXTENSIONS_REPLICATION_QUEUE_PROCESSOR_RETRY_TIMEOUT_S
```
bennettbuchanan, line 101: Depending on the user's needs, it seems like we might as well introduce `EXTENSIONS_REPLICATION_REPLICATION_STATUS_PROCESSOR_RETRY_TIMEOUT_S` as another tunable because we have that option exposed now.

bennettbuchanan (approved):

rahulreddy:
@ssalaues I just realized that, changed the base to 1.0

kubernetes/zenko/charts/backbeat/templates/replication/data_processor_deployment.yaml
rahulreddy: Ideally I don't want to expose that yet. I made it available to avoid any future new versions of Backbeat's docker images.

ssalaues (approved):

giacomoguiulfo:
@bert-e 	wait

PR #401, opened 2018-10-15
improvement: add retry on failure cron

ssalaues:
Adds a separate cron for retrying failed CRR objects added to s3utils https://github.com/scality/s3utils/pull/11

ZENKO-1258

### DIFF
--- a/kubernetes/zenko/templates/maintenance/retry-failed-cron.yaml
+++ b/kubernetes/zenko/templates/maintenance/retry-failed-cron.yaml
@@ -1,0 +1,47 @@
+{{- if .Values.maintenance.enabled -}}
+apiVersion: batch/v1beta1
+kind: CronJob
+metadata:
+  name: {{ template "zenko.fullname" . }}-retry-failed
+  labels:
+    app: {{ template "zenko.name" . }}
+    chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+spec:
+  schedule: "{{ .Values.maintenance.retryFailed.schedule }}"
+  concurrencyPolicy: Forbid
+  successfulJobsHistoryLimit: {{ .Values.maintenance.successfulJobsHistory }}
+  failedJobsHistoryLimit: 3
+  jobTemplate:
+    spec:
+      backoffLimit: 1
+      template:
+        spec:
+          restartPolicy: OnFailure
+          containers:
+          - name: failed
+            image: {{ .Values.maintenance.image.repository }}:{{ .Values.maintenance.image.tag }}
+            imagePullPolicy: {{ .Values.maintenance.image.pullPolicy }}
+            command: ["/bin/bash"]
+            args:
+            - -c
+            - node autoRetryFailedCRR.js 
+            env:
+            - name: ENDPOINT
+              value: "{{ .Release.Name }}-cloudserver"
+            - name: ACCESS_KEY
+              valueFrom:
+                secretKeyRef:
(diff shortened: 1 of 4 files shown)


ssalaues:
@bert-e reset

rahulreddy (approved):

JianqinWang (approved):

kubernetes/zenko/templates/maintenance/retry-failed-cron.yaml
```diff
@@ -0,0 +1,47 @@
+{{- if .Values.maintenance.enabled -}}
+apiVersion: batch/v1beta1
+kind: CronJob
+metadata:
+  name: {{ template "zenko.fullname" . }}-retry-failed
+  labels:
+    app: {{ template "zenko.name" . }}
+    chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
```
giacope, line 8: @ssalaues 
We should probably make a template for this:
```go
{{/*
Create chart name and version as used by the chart label.
*/}}
{{- define "zenko.chart" -}}
{{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" -}}
{{- end -}}
```

kubernetes/zenko/values.yaml
```diff
@@ -48,6 +33,25 @@ global:
...
   replicationEndpoints: []
 
+maintenance:
+  # Enables a retry CronJobs that will attempt to retry CRR objects stuck in
+  # pending or failed state based off the specified schedule. For Orbit enabled
+  # installs, Zenko will need to be registered and account credentials
+  # (secretKey and accessKey) will need to be provisioned prior to enabling this
+  # maintenance feature.
+  enabled: false
+  accessKey:
```
giacope, line 43: `""`?

kubernetes/zenko/templates/maintenance/secrets.yaml
```diff
@@ -1,15 +1,15 @@
-{{- if .Values.maintenance.stalled.enabled -}}
+{{- if .Values.maintenance.enabled -}}
 apiVersion: v1
 kind: Secret
 metadata:
-  name: {{ template "zenko.fullname" . }}-retry-stalled-secrets
+  name: {{ template "zenko.fullname" . }}-retry-secrets
```
giacope, line 5: We should probably template full names.

kubernetes/zenko/templates/maintenance/retry-pending-cron.yaml
```diff
@@ -1,17 +1,17 @@
-{{- if .Values.maintenance.stalled.enabled -}}
+{{- if .Values.maintenance.enabled -}}
 apiVersion: batch/v1beta1
 kind: CronJob
 metadata:
-  name: {{ .Release.Name }}-retry-stalled-objects
+  name: {{ template "zenko.fullname" . }}-retry-pending
```
giacope, line 5: We should probably template full names.

giacope:
@ssalaues This required a chart version bump.

kubernetes/zenko/values.yaml
ssalaues: It doesn't require the empty quotes. Also it will fail silently but install if there are empty quotes and the maintenance is enabled. Where as it will fail to install if the values are empty and is attempted to be enabled.

kubernetes/zenko/templates/maintenance/secrets.yaml
ssalaues: What do you mean template full names? Isn't that what is going on here? Or you do you mean to include the "retry-secrets" ?

ssalaues:
@giacomoguiulfo this is included in the hotfix.1 version bump.

PR #404, opened 2018-10-16
improvement: add pods with debug tools

ssalaues:
Adds an optional kafkaclient and s3utils pods if `--set maintenance.debug=true` at install or upgrade.

### DIFF
--- a/kubernetes/zenko/templates/_helpers.tpl
+++ b/kubernetes/zenko/templates/_helpers.tpl
@@ -15,6 +15,42 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this 
 {{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" -}}
 {{- end -}}
 
+{{/*
+Create chart name and version as used by the chart label.
+*/}}
+{{- define "zenko.chart" -}}
+{{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" -}}
+{{- end -}}
+
+{{- /*
+Print the standard Helm labels.
+*/ -}}
+{{- define "zenko.labels.standard" -}}
+app: {{ template "zenko.name" . }}
+chart: {{ template "zenko.chart" . }}
+heritage: {{ .Release.Service | quote }}
+release: {{ .Release.Name | quote }}
+{{- end -}}
+
+{{/*
+Create a common name for retry resources
+*/}}
+{{- define "zenko.retry" -}}
+{{- $name := default .Chart.Name .Values.nameOverride -}}
+{{- printf "%s-%s-retry" .Release.Name $name | trunc 63 | trimSuffix "-" -}}
+{{- end -}}
+
+{{/*
+Create a common name for debug resources
+*/}}
+{{- define "zenko.debug" -}}
+{{- $name := default .Chart.Name .Values.nameOverride -}}
+{{- printf "%s-%s-debug" .Release.Name $name | trunc 63 | trimSuffix "-" -}}
+{{- end -}}
+
+{{/*
+Disables kafka from deploying it's own zookeeper
+*/}}
 {{- define ".Values.zookeeper.enabled" -}}
 {{- printf "false" -}}
 {{- end -}}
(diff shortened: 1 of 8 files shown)


anurag4DSB (approved):

rahulreddy (approved):

kubernetes/zenko/templates/maintenance/debug/kafkaclient.yaml
```diff
@@ -0,0 +1,14 @@
+{{- if .Values.maintenance.debug -}}
+apiVersion: v1
+kind: Pod
+metadata:
+  name: {{ template "zenko.fullname" . }}-debug-kafka-client
+spec:
+  containers:
+  - name: kafka
+    image: solsson/kafka:0.11.0.0
```
bennettbuchanan: Thinking out loud: could there be a case where we need an image pull policy for this pod? Wondering if, as kafka releases/updates CLI tools, we could take advantage of them by using the latest image rather than keeping it at v `0.11.0.0`.

ssalaues: That's a really fair point. I'll make it configurable.

ssalaues:
@bert-e wait

kubernetes/zenko/templates/maintenance/debug/kafkaclient.yaml
```diff
@@ -0,0 +1,14 @@
+{{- if .Values.maintenance.debug -}}
+apiVersion: v1
+kind: Pod
+metadata:
+  name: {{ template "zenko.fullname" . }}-debug-kafka-client
```
giacope: What about the rest of the `standard` labels?

kubernetes/zenko/templates/maintenance/debug/s3utils.yaml
```diff
@@ -0,0 +1,15 @@
+{{- if .Values.maintenance.debug -}}
+apiVersion: v1
+kind: Pod
+metadata:
+  name: {{ template "zenko.fullname" . }}-debug-s3utils
```
giacope: Same here

giacope (changes requested):

kubernetes/zenko/values.yaml
```diff
@@ -51,6 +51,7 @@ maintenance:
     repository: zenko/s3utils
     tag: 0.3
     pullPolicy: IfNotPresent
+  debug: false
```
giacomoguiulfo: I would prefer, `enabled` instead of `debug`.

kubernetes/zenko/templates/maintenance/debug/kafkaclient.yaml
giacope: Also, I would prefer the name to be fully templated.

ssalaues:
@giacomoguiulfo I templated common names but I didn't create a template for each different name. It seemed the most appropriate use of templating the names.

kubernetes/zenko/templates/_helpers.tpl
```diff
@@ -15,6 +15,42 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this
...
+app: {{ template "zenko.name" . }}
+chart: {{ template "zenko.chart" . }}
+heritage: {{ .Release.Service | quote }}
+release: {{ .Release.Name | quote }}
+{{- end -}}
+
+{{/*
+Create a common name for retry resources
+*/}}
+{{- define "zenko.retry" -}}
```
giacomoguiulfo: why not, zenko.retry.fullname?

kubernetes/zenko/values.yaml
```diff
@@ -40,18 +40,29 @@ maintenance:
...
+    # instanciate the pod.
+    enabled: false
+    accessKey: ""
+    secretKey: ""
+  s3UtilsImage:
     repository: zenko/s3utils
     tag: 0.3
     pullPolicy: IfNotPresent
-  debug: false
+  kafkaClientImage:
```
giacope: Minor. This doesn't look very `helm`. Why not:
```yaml
debug:
  kafkaClient:
    image:
      repository: solsson/kafka
      tag: 0.11.0.0
      pullPolicy: IfNotPresent
  s3Utils:
    image:
    ...
```

giacope (approved):

kubernetes/zenko/values.yaml
ssalaues: Fixed

ssalaues:
@giacomoguiulfo @rahulreddy There is a new commit where I changed the debug tools to full on deployments (previously simple pod definitions)

PR #410, opened 2018-10-24
feat: add zenko-queue-manager chart

giacope:

This PR adds the Zenko-Queue-Manager chart ([Kafka-Manager](https://github.com/yahoo/kafka-manager)) to the Zenko stack. It is enabled by default and the zenko-queue cluster is configured automagically.

If possible, try this chart out and see if it works fine 😀

This more or less how it looks on with the Zenko-Queue cluster configured:
screenshot 2018-10-24 at 12 22 41 pm

### DIFF
--- a/images/kafka-manager/Dockerfile
+++ b/images/kafka-manager/Dockerfile
@@ -1,0 +1,20 @@
+FROM hseeberger/scala-sbt
+
+ENV KM_VERSION=1.3.3.18
+
+WORKDIR /tmp
+
+RUN wget https://github.com/yahoo/kafka-manager/archive/${KM_VERSION}.tar.gz && \
+    tar xvf ${KM_VERSION}.tar.gz && \
+    cd kafka-manager-${KM_VERSION} && \
+    sbt clean dist
+
+WORKDIR /opt
+
+RUN unzip -d . /tmp/kafka-manager-${KM_VERSION}/target/universal/kafka-manager-${KM_VERSION}.zip && \
+    rm -fr /tmp/${KM_VERSION}.tar.gz /tmp/kafka-manager-${KM_VERSION}
+
+WORKDIR /opt/kafka-manager-${KM_VERSION}
+
+EXPOSE 9000
+ENTRYPOINT ["./bin/kafka-manager","-Dconfig.file=conf/application.conf"]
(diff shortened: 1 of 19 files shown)


giacomoguiulfo:
@bert-e wait

kubernetes/zenko/charts/zenko-queue-manager/requirements.yaml
```diff
@@ -0,0 +1,5 @@
+dependencies:
+- name: zookeeper
```
ssalaues, line 2: Kafka doesn't need to be a dependency right? So if you just deploy the manager and ZK then you can manually add clusters as necessary?

ssalaues (approved):

kubernetes/zenko/charts/zenko-queue-manager/requirements.yaml
giacope: 1. Yeah, Kafka manager only needs Zookeeper to initialize properly (to manage state). This can either be its personal quorum or a shared one with a Kafka cluster.

2. Correct.

The reason for this dependency is only to match the chart I'm pushing upstream since that one won't pass CI without stabilizing and it needs Zookeeper for that (readiness probe).

images/kafka-manager/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+WORKDIR /tmp
+
+RUN wget https://github.com/yahoo/kafka-manager/archive/${KM_VERSION}.tar.gz && \
+    tar xvf ${KM_VERSION}.tar.gz && \
+    cd kafka-manager-${KM_VERSION} && \
+    sbt clean dist
+
+WORKDIR /opt
+
+RUN unzip -d . /tmp/kafka-manager-${KM_VERSION}/target/universal/kafka-manager-${KM_VERSION}.zip && \
```
tmacro, line 14: This can be combined with the RUN on L#7 to avoid having the .tar and .zip in the final image

tmacro (commented):
LGTM

images/kafka-manager/Dockerfile
giacope: Mmm I see. Well it's a Scala image. It will take several hours to rebuild 😅. I would prefer to leave it as-is for now.

rahulreddy (approved):

PR #425, opened 2018-10-31
Improvement/doc zenko 1224

wabernat:
This PR add needed information concerning deleting objects from versioned buckets using the s3utils scripts.

Fixes Zenko 1224,

### DIFF
--- a/docs/Operation-Architecture/Orbit_UI/File_Operations/Deleting_Files.rst
+++ b/docs/Operation-Architecture/Orbit_UI/File_Operations/Deleting_Files.rst
@@ -1,20 +1,23 @@
-Deleting Files
-==============
+Deleting Objects
+================
 
-To delete files from a selected bucket:
+To delete objects from a selected bucket:
 
-#. Click the check box next to each file to be deleted. The number of
-   files to be deleted is indicated in the top bar of the file list.
+#. Click the check box next to each object to be deleted. The number 
+   of objects to be deleted is indicated in the top bar of the file 
+   list.
 
    |image0|
 
 #. Click the **Delete** button.
+
    |image1|
+
 #. Orbit requests confirmation of the deletion.
 
    |image2|
 
-#. The file is deleted from the bucket.
+#. The object is deleted from the bucket.
 
    .. important::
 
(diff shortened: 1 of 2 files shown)


wabernat:
@bert-e reset

wabernat:
This issue went stale last time around. I have simplified it and will make other changes in other tickets/builds. If there is a reason this can't be reviewed/approved, please advise.

wabernat:
@bert-e reset

wabernat:
@bert-e force_reset

wabernat:
@bert-e approve

jonathan-gramain (approved):

docs/Operation-Architecture/Orbit_UI/File_Operations/Deleting_Files.rst
```diff
@@ -30,67 +33,39 @@ To delete files from a selected bucket:
...
+completely (both removing the object data and its object ID from the namespace).
 
-       $ aws s3api delete-object --bucket <bucket-name> --version-id <versionID> --key <key> --endpoint-url http://<zenko.endpoint.url> --profile <aws-profile>
+The cleanupBuckets.js script is available in the s3utils pod. 
 
-    which in the present example returns:
+Run it as follows:
 
-    ::
+  # Exec into the pod with ``kubectl exec -it s3utils bash``
```
ssalaues: We don't enable this pod by default so before we exec into the pod, it has to be enable in either the options file or at run time.

This is what it would look like in the options file.
```
maintenance:
  debug:
    enabled: true
```

The other option is to have it something like below which doesn't require options file configuration and will simply instantiate and launch you into an s3utils pod.
`kubectl run s3utils --image=zenko/s3utils:0.5 -it bash`

ssalaues (changes requested):
A note about the exec line otherwise looks good to me.

docs/Operation-Architecture/Orbit_UI/File_Operations/Deleting_Files.rst
```diff
@@ -30,67 +33,57 @@ To delete files from a selected bucket:
...
-          }
-       ]
-     }
+	The s3utils pod is disabled by default. You can also enable it 
+	by adding the following to the options file::
 
-#. To delete object versions, use the command:
+	  maintenance:
+	    debug:
+	      enabled: true
```
ssalaues, line 65: I forgot to add, doing it this way would also require applying an access/secret key pair.
```suggestion
	      enabled: true
	      # An access/secret key to access Zenko that will be used to configure the s3utils pod
	      accessKey: <access-key>
	      secretKey: <secret-key>
```

docs/Operation-Architecture/Orbit_UI/File_Operations/Deleting_Files.rst
```diff
@@ -30,67 +33,57 @@ To delete files from a selected bucket:
...
-             },
-             "IsLatest": true,
-             "VersionId": "39383436323730383139313839343939393939395247303031202032302e30",
-             "Key": "file.txt",
-             "LastModified": "2018-09-18T17:30:08.103Z"
-          }
-       ]
-     }
+	The s3utils pod is disabled by default. You can also enable it 
+	by adding the following to the options file::
```
ssalaues: Any change to the options file requires a `helm upgrade`
```suggestion
	by adding the following to the options file and upgrading your Zenko deployment::
```

docs/Operation-Architecture/Orbit_UI/File_Operations/Deleting_Files.rst
```diff
@@ -30,67 +33,57 @@ To delete files from a selected bucket:
...
-#. To delete object versions, use the command:
+	  maintenance:
+	    debug:
+	      enabled: true
 
-    ::
+  #. Exec into the pod with::
 
-       $ aws s3api delete-object --bucket <bucket-name> --version-id <versionID> --key <key> --endpoint-url http://<zenko.endpoint.url> --profile <aws-profile>
+     $ kubectl exec -it s3utils bash
```
ssalaues: Adding it via the options file will require the name to be searched for since the name can/will change (similar to looking for the cloud server manager for the InstanceID):
```suggestion
     # First grep for your s3utils pod
     $ kubectl get pods | grep s3utils
     myzenko-zenko-debug-s3utils-7f77f9b5b9-627gz   1/1  Running   0   31m
     # Then exec into the pod
     $ kubectl exec -it myzenko-zenko-debug-s3utils-7f77f9b5b9-627gz bash
```

ssalaues (commented):
With these last few clarifications, the rest makes sense and should function when followed line by line.

ssalaues (approved):

wabernat:
@bert-e reset

wabernat:
@bert-e force_reset

PR #438, opened 2018-11-09
chore: bump backbeat to 8.0.19

ssalaues:
bump backbeat chart for the ZENKO-1175 oplog fix

### DIFF
--- a/kubernetes/zenko/charts/backbeat/Chart.yaml
+++ b/kubernetes/zenko/charts/backbeat/Chart.yaml
@@ -1,5 +1,5 @@
 apiVersion: v1
-appVersion: "8.0.18"
+appVersion: "8.0.19"
 description: A Helm chart for Zenko Backbeat
 name: backbeat
-version: 1.0.0
+version: 1.0.2
--- a/kubernetes/zenko/charts/backbeat/values.yaml
+++ b/kubernetes/zenko/charts/backbeat/values.yaml
@@ -10,7 +10,7 @@ global:
 
 image:
   repository: zenko/backbeat
-  tag: 8.0.18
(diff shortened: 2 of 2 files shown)


alexanderchan-scality (approved):

anurag4DSB (approved):

philipyoo (approved):

kubernetes/zenko/charts/backbeat/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: v1
-appVersion: "8.0.18"
+appVersion: "8.0.19"
 description: A Helm chart for Zenko Backbeat
 name: backbeat
-version: 1.0.0
+version: 1.0.2
```
JianqinWang, line 5: we're skipping 1.0.1?

JianqinWang (approved):

tmacro (approved):

kubernetes/zenko/charts/backbeat/Chart.yaml
ssalaues: It's open to debate. Me and @giacomoguiulfo were discussing this and couldn't decide whether keeping the Chart versions as their own version or mirroring the Zenko release version. No changes had been made to the backbeat chart which is why it hasn't changed until now.

Any thoughts? Would it be more confusing for the chart versions to have their own diverging versions or follow the Zenko release version?

JianqinWang: From my pov, having multiple diverging versions would be harder to follow, especially since our documentation doesn't get edited to keep up sometimes.

PR #448, opened 2018-11-16
feat(ZENKO-1336): add rclone submodule

giacope:

This PR adds rclone as a submodule. The Dockerfile contains the necessary steps to   patch rclone to our needs.

### DIFF
--- a/images/rclone/Dockerfile
+++ b/images/rclone/Dockerfile
@@ -1,0 +1,23 @@
+FROM golang:1.10-alpine3.7 AS build
+
+COPY rclone /go/src/github.com/ncw/rclone/
+COPY patches /go/src/github.com/ncw/rclone/patches
+WORKDIR /go/src/github.com/ncw/rclone/
+
+RUN for file in patches/*.diff ; do \
+        patch -p 1 < "$file" || exit 1 ; \
+    done && \
+    CGO_ENABLED=0 go build -a -installsuffix cgo && \
+    mv rclone /bin/rclone
+
+FROM golang:alpine AS dev
+ENV MOUNT_PATH /data
+COPY --from=build /bin/rclone /bin
+WORKDIR ${MOUNT_PATH}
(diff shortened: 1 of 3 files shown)


ssalaues:
@giacomoguiulfo I think the difference is negligible especially with compression. Plus with the current Dockerfile, we could keep it on Docker automated builds.

giacope:
@ssalaues Updated to use busybox instead of scratch.

ssalaues (approved):

tmacro (approved):

giacope:
@bert-e approve

NicolasT:
If you do patches, why use a submodule?

Either use a subtree, patch in-tree and periodically sync, or simply have a Dockerfile which downloads the source tree, applies patches, builds the binary, no submodule required.

ssalaues:
@NicolasT I like your Dockerfile idea the best. Keeps it simple and removes need to deal with submodules.

images/rclone/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+COPY patches /go/src/github.com/ncw/rclone/patches
+WORKDIR /go/src/github.com/ncw/rclone/
+
+RUN for file in patches/*.diff ; do \
+        patch -p 1 < "$file" || exit 1 ; \
+    done && \
+    CGO_ENABLED=0 go build -a -installsuffix cgo && \
+    mv rclone /bin/rclone
+
+FROM golang:alpine AS dev
```
NicolasT, line 13: What's the use of this intermediate containers?

giacope: Build provides an environment to build the binary.

Dev's purpose is to have a lightweight version of build for testing and debugging purposes. It's not intermediate. It's a different target.

NicolasT: Ah, I wasn't aware multi-stage builds got each stage tagged and kept for later execution.

PR #449, opened 2018-11-20
bugfix: ZENKO-1312 reduce replicaFactor of backbeat processor

jonathan-gramain:
Set the default replicaFactor for backbeat processor pods to 1 instead
of 2, to match the default number of partitions for
backbeat-replication topic which is the number of nodes.

### DIFF
--- a/kubernetes/zenko/values.yaml
+++ b/kubernetes/zenko/values.yaml
@@ -81,7 +81,10 @@ backbeat:
   replication:
     dataProcessor:
       replicaCount: *nodeCount
-      replicaFactor: 2
+      # If replicaFactor is increased, make sure to increase the
+      # number of partitions for "backbeat-replication" topic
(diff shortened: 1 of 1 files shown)


ssalaues (commented):
Ahh so it should only be `replicaFactor: 2` if the `backbeat-replication` topic has 6 partitions instead of the default `nodeCount`?

jonathan-gramain:
@ssalaues Indeed, otherwise, while still okay functionally, the extra pods will stay idle not being able to consume from any partition.

ssalaues (approved):

giacope (approved):

ssalaues:
@jonathan-gramain whenever you're ready, this just needs the `berte approve` from you

jonathan-gramain:
@bert-e approve

PR #464, opened 2018-11-30
feat(rclone): add system attribute patch

giacope:

- Removes rclone as a submodule and downloads it instead (in the Dockerfile).
- Adds a new patch to ingest system attributes metadata.
- Uses the latest rclone release as the base (v1.45 / 6 days ago)


images/rclone/Dockerfile
```diff
@@ -1,11 +1,16 @@
 FROM golang:1.10-alpine3.7 AS build
 
-COPY rclone /go/src/github.com/ncw/rclone/
+WORKDIR "/go/src/github.com/ncw/"
+RUN wget https://github.com/ncw/rclone/releases/download/v1.45/rclone-v1.45.tar.gz && \
```
ssalaues: Might be good to have the version as an `ENV` this way it's easier to change the version when needed.

ssalaues (changes requested):

images/rclone/Dockerfile
giacope: Done

ssalaues (approved):

images/rclone/Dockerfile
```diff
@@ -1,11 +1,18 @@
 FROM golang:1.10-alpine3.7 AS build
 
-COPY rclone /go/src/github.com/ncw/rclone/
+ENV RCLONE_VERSION "v1.45"
+
+WORKDIR "/go/src/github.com/ncw/"
+RUN wget https://github.com/ncw/rclone/releases/download/${RCLONE_VERSION}/rclone-${RCLONE_VERSION}.tar.gz && \
```
tmacro: I'd move the wget to an `ADD` layer. Since you're using a build image anyway, you don't need to clean it up, also `ADD` compares hashes of the previous files that have been downloaded to determine whether a rebuild needs to happen, while `RUN` will just do a lexicographic check of the command which can lead to weird issues when building new images.

tmacro (changes requested):

images/rclone/Dockerfile
giacope: Done!

tmacro (approved):

giacope:
@bert-e approve

giacope:
@bert-e help

giacope:
@bert-e help

PR #484, opened 2018-12-28
feat(ZENKO-1364): cosmos delete and orbit support

giacope:

- Fix some typos/names
- Add delete support for Cosmos
- Update Cloudserver image to support Cosmos with Orbit enabled.

### DIFF
--- a/kubernetes/cosmos/templates/pfsd-deployment.yaml
+++ b/kubernetes/cosmos/templates/pfsd-deployment.yaml
@@ -32,6 +32,8 @@ spec:
               containerPort: 9992
               protocol: TCP
           env:
+            - name: PFSD_READONLY
+              value: {{ .Values.persistentVolume.readOnly | quote }}
             - name: PFSD_MOUNT_PATH
               value: /data
             - name: LISTEN_ADDR
--- a/kubernetes/cosmos/values.yaml
+++ b/kubernetes/cosmos/values.yaml
@@ -5,7 +5,7 @@ pfsd:
 
   image:
     repository: zenko/cloudserver
-    tag: 8.1.3
+    tag: 8.1.4
     pullPolicy: IfNotPresent
 
   service:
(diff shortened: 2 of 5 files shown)


giacope:
@bert-e help

kubernetes/cosmos/values.yaml
```diff
@@ -46,3 +46,4 @@ persistentVolume:
   accessModes:
     - ReadWriteMany
   size: 1Gi
+  readOnly: false
```
ssalaues, line 49: @vrancurel do we want `readOnly: true` by default?

vrancurel: what is stored there ?

ssalaues: This is for access to the NFS. I think `readOnly: false` by default is okay so that lifecycle policies can be applied but wanted to double check.

jonathan-gramain (approved):

kubernetes/cosmos/values.yaml
vrancurel: Okay

ssalaues (approved):

kubernetes/cosmos/templates/pfsd-deployment.yaml
```diff
@@ -32,6 +32,8 @@ spec:
               containerPort: 9992
               protocol: TCP
           env:
+            - name: PFSD_READONLY
+              value: {{ .Values.persistentVolume.readOnly }}
```
ssalaues: This should be quoted or `| quote`. I just tried it with true and got a templating error.

ssalaues (changes requested):

kubernetes/cosmos/templates/pfsd-deployment.yaml
giacope: Done!

giacope:
@bert-e approve

ssalaues (approved):

PR #487, opened 2018-12-29
docs(README.md): add cosmos installation process with Orbit

giacope:

Adds steps for configuring Cosmos with an Orbit-enabled Zenko instance.


ssalaues (approved):

wabernat (approved):
Line 58 has a typo: 'backendj"

I have made all the changes I thought reasonable in the RST file. I think it would be best to include a link or reference to the RST documentation from this location, rather than maintaining two parallel documents. I am approving this, but would like to revisit it on your return.

giacope:
@bert-e approve

PR #517, opened 2019-02-15
feat: add cosmos helm operator

giacope:

### DIFF
--- a/kubernetes/cosmos/values.yaml
+++ b/kubernetes/cosmos/values.yaml
@@ -10,7 +10,9 @@ pfsd:
     pullPolicy: IfNotPresent
 
   service:
-    name: ""
+    ## Override the generated service name.
+    # name: ""
+
     type: ClusterIP
     port: 80
 
@@ -44,10 +46,13 @@ rclone:
 
   remote:
     credentials:
-      existingSercret: ""
+      ## Specify an existing secret to get credentials from.
+      # existingSercret: ""
 
+      ## Otherwise use the following credentials.
       accessKey: my-access-key
       secretKey: my-secret-key
+
     endpoint: http://zenko-cloudserver
     region: mynfs
     bucket: newbucket
--- a/kubernetes/cosmos/values.yaml
+++ b/kubernetes/cosmos/values.yaml
@@ -10,6 +10,7 @@ pfsd:
     pullPolicy: IfNotPresent
 
   service:
+    name: ""
     type: ClusterIP
     port: 80
 
@@ -37,32 +38,28 @@ rclone:
     tag: 1.45
     pullPolicy: IfNotPresent
 
+  suspend: true
   schedule: "0 */12 * * *"
   successfulJobsHistory: 1
 
   remote:
-    accessKey: my-access-key
-    secretKey: my-secret-key
-    endpoint: http://cloudserver
-    region: pfsd-1
-    bucket: pfsd-1
+    credentials:
+      existingSercret: ""
+
+      accessKey: my-access-key
+      secretKey: my-secret-key
+    endpoint: http://zenko-cloudserver
+    region: mynfs
+    bucket: newbucket
 
   resources: {}
   nodeSelector: {}
   tolerations: []
   affinity: {}
 
-persistentVolume:
-  enabled: false
-  volumeConfig:
-    ## Example NFS config
-    # nfs:
-    #   server: 10.100.1.42
-    #   path: /data
-    #   readOnly: false
-    # mountOptions:
-    # - hard
-    # - nfsvers=4.1
-  accessModes:
-    - ReadWriteMany
-  size: 1Gi
+backend:
+  ## One of [nfs, smb, ftp]
+  type: nfs
+  server: 10.100.1.28
+  path: /data
+  options: "hard,nfsvers=4.1"
--- a/kubernetes/zenko/templates/cosmos-operator/clusterrole.yaml
+++ b/kubernetes/zenko/templates/cosmos-operator/clusterrole.yaml
@@ -1,0 +1,22 @@
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: {{ template "cosmos-operator.fullname" . }}
+  labels:
+    app: {{ template "cosmos-operator.name" . }}
+    chart: {{ template "zenko.chart" . }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+rules:
+- apiGroups: [""]
+  resources:
+  - persistentvolumes
+  verbs:
+  - "*"
+- apiGroups: ["storage.k8s.io"]
+  resources: ["storageclasses"]
+  verbs:
+  - "*"
+- apiGroups: [""]
+  resources: ["events"]
+  verbs: ["list", "watch", "create", "update", "patch"]
--- a/kubernetes/zenko/templates/cosmos-operator/deployment.yaml
+++ b/kubernetes/zenko/templates/cosmos-operator/deployment.yaml
@@ -1,0 +1,41 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ template "cosmos-operator.fullname" . }}
+  labels:
+    app: {{ template "cosmos-operator.name" . }}
+    chart: {{ template "zenko.chart" . }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: {{ template "cosmos-operator.name" . }}
+      name: {{ template "cosmos-operator.fullname" . }}
+      release: {{ .Release.Name }}
+  template:
+    metadata:
+      labels:
+        app: {{ template "cosmos-operator.name" . }}
+        name: {{ template "cosmos-operator.fullname" . }}
+        release: {{ .Release.Name }}
+    spec:
+      serviceAccountName: {{ template "cosmos-operator.fullname" . }}
+      containers:
+        - name: {{ template "cosmos-operator.fullname" . }}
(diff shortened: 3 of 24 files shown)


tmacro (approved):

kubernetes/cosmos/values.yaml
```diff
@@ -37,32 +38,28 @@ rclone:
...
   successfulJobsHistory: 1
 
   remote:
-    accessKey: my-access-key
-    secretKey: my-secret-key
-    endpoint: http://cloudserver
-    region: pfsd-1
-    bucket: pfsd-1
+    credentials:
+      existingSercret: ""
```
ssalaues: Can prob just leave this commented out with a small note of what it can be used for

kubernetes/cosmos/values.yaml
```diff
@@ -37,32 +38,28 @@ rclone:
...
-    #   server: 10.100.1.42
-    #   path: /data
-    #   readOnly: false
-    # mountOptions:
-    # - hard
-    # - nfsvers=4.1
-  accessModes:
-    - ReadWriteMany
-  size: 1Gi
+backend:
```
ssalaues, line 65: I'm not a huge fan of obscuring the PV options like this because it directly blocks any other out of the box use-case (localFS, block, iSCSI, even SMB). The reason is because each type of PV will have completely different types of config values and can't account for every single one. For example, we have SMB listed however the parameters are in a completely different structure and will even vary based off the SMB FlexVolume plugin that is being used.

Also the way it is prior to these changes already allows for the future FTP use-case since PV is disabled by default and wouldn't need to add the `if not (eq .Values.backend.type "ftp"` all over.

kubernetes/cosmos/values.yaml
```diff
@@ -10,6 +10,7 @@ pfsd:
     pullPolicy: IfNotPresent
 
   service:
+    name: ""
```
ssalaues: Prob can be commented out too with a note saying it's a override of whatever default

kubernetes/zenko/templates/cosmos-operator/clusterrole.yaml
```diff
@@ -0,0 +1,23 @@
...
+    chart: {{ template "zenko.chart" . }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+rules:
+- apiGroups: [""]
+  resources:
+  - persistentvolumeclaims
+  - persistentvolumes
+  verbs:
+  - "*"
```
ssalaues, line 15: I'm not sure but do we need all the verbs? For PVs don't we just need create/delete?

kubernetes/zenko/templates/cosmos-operator/operator.yaml
```diff
@@ -0,0 +1,49 @@
...
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.namespace
+            - name: POD_NAME
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.name
+            - name: OPERATOR_NAME
+              value: {{ template "cosmos-operator.fullname" . }}
+          resources:
```
ssalaues: Do we want these under zenko/values.yaml and passed along?

kubernetes/zenko/templates/cosmos-operator/role.yaml
```diff
@@ -0,0 +1,65 @@
...
+  - '*'
+- apiGroups:
+  - ""
+  resources:
+  - namespaces
+  verbs:
+  - get
+- apiGroups:
+  - apps
+  resources:
```
ssalaues, line 32: Don't we just need deployments here?

kubernetes/zenko/templates/cosmos-operator/role.yaml
```diff
@@ -0,0 +1,65 @@
...
+  - servicemonitors
+  verbs:
+  - get
+  - create
+- apiGroups:
+  - zenko.io
+  resources:
+  - '*'
+  verbs:
+  - '*'
```
ssalaues, line 41: We should double check the verbs and resources everywhere

ssalaues (changes requested):
Have some concerns about specific changes but overall looking good

kubernetes/zenko/templates/cosmos-operator/clusterrole.yaml
giacope: We probably need `watch` and `list` verbs too. Perhaps `update` and `patch` are used internally by the operator for meta-stuff. I don't see a problem in giving more permissions if we already need `create` and `delete` since those are the most "dangerous" ones.

kubernetes/zenko/templates/cosmos-operator/role.yaml
giacope: Probably. I just used the defaults for helm chart operators since they are generic enough to be used by most charts since the operator can operate multiple charts. But I'll remove them for now.

kubernetes/zenko/templates/cosmos-operator/role.yaml
giacope: Any suggestions? 😅
I deleted this element particularly because it only has to be listed in the `ClusterRole`.

kubernetes/zenko/templates/cosmos-operator/clusterrole.yaml
```diff
@@ -0,0 +1,23 @@
...
+- apiGroups: ["storage.k8s.io"]
+  resources: ["storageclasses"]
+  verbs:
+  - "*"
```
NicolasT, line 19: This 'operator' creates StorageClasses and such?!

kubernetes/zenko/templates/cosmos-operator/deployment.yaml
```diff
@@ -0,0 +1,44 @@
...
+      serviceAccountName: {{ template "cosmos-operator.fullname" . }}
+      containers:
+        - name: {{ template "cosmos-operator.fullname" . }}
+          image: "{{ .Values.cosmos.operator.image.repository }}:{{ .Values.cosmos.operator.image.tag }}"
+          imagePullPolicy: {{ .Values.cosmos.operator.image.pullPolicy }}
+          ports:
+          - containerPort: 60000
+            name: metrics
+          env:
+            - name: WATCH_NAMESPACE
```
NicolasT, line 30: If the operator runs in a specific namespace (which is a good initial approach!), why does it get PVC access rights through a ClusterRole(Binding)?

kubernetes/zenko/templates/cosmos-operator/role.yaml
```diff
@@ -0,0 +1,56 @@
...
+    heritage: {{ .Release.Service }}
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - pods
+  - services
+  - endpoints
+  - persistentvolumeclaims
+  - persistentvolumes
```
NicolasT: PVs are not namespaced so not sure how that'd work with a Role.

kubernetes/zenko/templates/cosmos-operator/role.yaml
```diff
@@ -0,0 +1,56 @@
...
+- apiGroups:
+  - apps
+  resources:
+  - deployments
+  verbs:
+  - '*'
+- apiGroups:
+  - monitoring.coreos.com
+  resources:
+  - servicemonitors
```
NicolasT: You're using this?

kubernetes/zenko/templates/cosmos-operator/deployment.yaml
giacope: It doesn't need to. Removing it...

kubernetes/zenko/templates/cosmos-operator/clusterrole.yaml
ssalaues: Yeah but the whole point is to only give access where explicitly needed. Just because `create` and `delete` are required doesn't mean it's okay to needlessly give out permissions by default. Especially since these specific roles aren't namespaced and are in fact cluster wide

kubernetes/zenko/templates/cosmos-operator/clusterrole.yaml
giacope: Seems like the only way... Any suggestions?

kubernetes/zenko/templates/cosmos-operator/role.yaml
giacope: Yes, removing it...

kubernetes/zenko/templates/cosmos-operator/role.yaml
giacope: It was there in the generated `role.yaml`. Probably because the operator has support for a Prometheus metrics exporter. But it doesn't seem to be active on the helm operator so I'm removing it...

kubernetes/zenko/templates/cosmos-operator/clusterrole.yaml
giacope: The Scheduler (coming soon) just needs to create/delete. But the operator itself should need all these privileges so that CRs can be fully modified (manually for example) if needed and operations on these CRs shouldn't be limited.

kubernetes/zenko/templates/cosmos-operator/clusterrole.yaml
giacope: Or well I guess we could ask for the user to provide a StorageClass name to use.

giacope: or install a `StorageClass` as part of Zenko (thanks @ssalaues)

giacope: What do you think? @NicolasT

giacope:
@tmacro, closing this in favor of #529.

PR #528, opened 2019-02-22
bugfix: ZENKO-1513 remove flaky sanity check

jonathan-gramain:
A call to assertNoObject() was done on the source in CRR tests after
putting a delete marker, it's suspected this is flaky because of
eventual consistency (but local zenko should be strongly consistent with commits to MongoDB?)

Removed it as it is not useful for the test itself, will see how it goes.


ssalaues (approved):

jonathan-gramain:
@bert-e approve

philipyoo (approved):

PR #529, opened 2019-02-22
feat(ZENKO-1358): add cosmos helm operator

ssalaues:
Proposed changes to #517

### DIFF
--- a/kubernetes/cosmos/values.yaml
+++ b/kubernetes/cosmos/values.yaml
@@ -40,7 +40,7 @@ rclone:
     tag: 1.45
     pullPolicy: IfNotPresent
 
-  suspend: true
+  suspend: false
   schedule: "0 */12 * * *"
   successfulJobsHistory: 1
 
@@ -62,9 +62,19 @@ rclone:
   tolerations: []
   affinity: {}
 
-backend:
-  ## One of [nfs, smb, ftp]
-  type: nfs
-  server: 10.100.1.28
-  path: /data
-  options: "hard,nfsvers=4.1"
+persistentVolume:
+  enabled: false
+  # Specifiy a existing storage class to use
+  # storageClass:
+  volumeConfig:
+    ## Example NFS config
+    # nfs:
+    #   server: 10.100.1.42
+    #   path: /data
+    #   readOnly: false
+    # mountOptions:
+    # - hard
+    # - nfsvers=4.1
+  accessModes:
+    - ReadWriteMany
+  size: 1Gi
--- a/kubernetes/cosmos/values.yaml
+++ b/kubernetes/cosmos/values.yaml
@@ -45,13 +45,11 @@ rclone:
   successfulJobsHistory: 1
 
   remote:
-    credentials:
-      ## Specify an existing secret to get credentials from.
-      # existingSercret: ""
-
-      ## Otherwise use the following credentials.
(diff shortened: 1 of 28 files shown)


kubernetes/cosmos/values.yaml
```diff
@@ -62,9 +60,19 @@ rclone:
...
+    # nfs:
+    #   server: 10.100.1.42
+    #   path: /data
+    #   readOnly: false
+    # mountOptions:
+    # - hard
+    # - nfsvers=4.1
+  accessModes:
+    - ReadWriteMany
+  size: 1Gi
```
giacope: Maybe we can remove this?

kubernetes/cosmos/values.yaml
```diff
@@ -62,9 +60,19 @@ rclone:
...
+  volumeConfig:
+    ## Example NFS config
+    # nfs:
+    #   server: 10.100.1.42
+    #   path: /data
+    #   readOnly: false
+    # mountOptions:
+    # - hard
+    # - nfsvers=4.1
+  accessModes:
```
giacope: Maybe we can remove this?

kubernetes/cosmos/operator/examples/cosmos-cr.yaml
```diff
@@ -14,6 +14,7 @@ spec:
       bucket: "s3bucket"
   backend:
```
giacope: This example needs to be updated to match the current values.

kubernetes/zenko/templates/cosmos-operator/sc.yaml
```diff
@@ -0,0 +1,14 @@
+{{- if .Values.cosmos.enabled }}
```
giacope: Shouldn't we create it only if they don't specify one in the values?

kubernetes/zenko/templates/cosmos-operator/crd.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if .Values.cosmos.enabled }}
```
giacope: Should we really make cosmos optional? I think it should just be like backbeat... (deployed even if unused)

kubernetes/cosmos/values.yaml
ssalaues: For what reason?

kubernetes/cosmos/values.yaml
ssalaues: For what reason?

kubernetes/zenko/templates/cosmos-operator/crd.yaml
ssalaues: Mostly for the reason that it creates privileged accounts. If someone didn't want a service running that has/creates this kind of access, it could be turned off

kubernetes/cosmos/values.yaml
giacope: because its value is useless :)

kubernetes/cosmos/values.yaml
giacope: because its value is useless :)

giacope (approved):

tmacro (approved):

ssalaues:
@bert-e approve

PR #534, opened 2019-02-27
feat: add cosmos scheduler

giacope:

This PR adds the Cosmos Scheduler, a new component that will create and delete Cosmos CRs based on the creation and deletion of NFS Ingestion Buckets.

Vendor dependencies will come in another PR to simplify the review process.


cosmos/scheduler/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM alpine
+RUN apk add gettext
+COPY ./cosmos /bin/
+ENTRYPOINT ["cosmos"]
```
giacope: Probably need a better Dockerfile (that doesn't depend on a locally built binary)

cosmos/scheduler/cmd/scheduler.go
```diff
@@ -0,0 +1,247 @@
...
+			s.handler(msgCh)
+		}
+	}()
+	for {
+		if err := s.waitForOverlayUpdate(); err != nil {
+			log.Println(err)
+		} else {
+			msgCh <- "overlay update"
+		}
+	}
```
giacope, line 125: Not sure if this the best way to tackle the scheduler logic. Open to suggestions...

cosmos/scheduler/cmd/scheduler.go
```diff
@@ -0,0 +1,247 @@
...
+	})
+	if err != nil {
+		return err
+	}
+	log.Println("ingestion credentials secret created succesfully")
+	return nil
+}
+
+func (s *Scheduler) getCosmosLocationBson() (bson.A, error) {
+	locations, err := s.Pensieve.GetLocationsWithTypes([]string{"nfs"})
```
giacope, line 190: Be more specific about location types instead of relying on `strstr`.

cosmos/scheduler/cmd/scheduler.go
```diff
@@ -0,0 +1,247 @@
...
+				if err := cur.Decode(&elem); err != nil {
+					log.Println(err)
+					return
+				}
+				ch <- elem
+			}
+			if err := cur.Err(); err != nil {
+				log.Println(err)
+			}
+		}
```
giacope, line 222: Opinions for a way for tackling this notification system are welcome :)

cosmos/scheduler/cmd/scheduler.go
```diff
@@ -0,0 +1,247 @@
...
+	}
+}
+
+func (s *Scheduler) handler(ch <-chan interface{}) {
```
giacope: Need a better method name

cosmos/scheduler/cmd/scheduler.go
```diff
@@ -0,0 +1,247 @@
...
+	bucketUpdates := make(chan BucketTransaction, 1)
+	quitWatch := make(chan bool, 1)
+	go s.watchBucketEvents(bucketUpdates, quitWatch)
+	for {
+		select {
+		case bucket := <-bucketUpdates:
+			if bucket.OperationType == "insert" {
```
giacope: This could be wrapped in another method

cosmos/scheduler/cmd/scheduler.go
```diff
@@ -0,0 +1,247 @@
...
+				log.Println("creating cosmos for bucket", bucket.FullDocument.Value.Name)
+				location, err := s.Pensieve.GetLocationWithName(bucket.FullDocument.Value.LocationConstraint)
+				if err != nil {
+					log.Println(err)
+					return
+				}
+				s.CreateCosmosFromLocation(location, bucket.FullDocument.Value.Name)
+			} else if bucket.OperationType == "replace" && bucket.FullDocument.Value.Deleted == true {
+				log.Println("deleting cosmos for bucket", bucket.FullDocument.Value.Name)
+				err := s.KubeAlpha.Cosmoses(s.Namespace).Delete(bucket.FullDocument.Value.LocationConstraint+"-"+bucket.FullDocument.Value.Name, &metav1.DeleteOptions{})
```
giacope: Downto here

cosmos/scheduler/cmd/scheduler.go
```diff
@@ -0,0 +1,247 @@
...
+			log.Println("update on overlay")
+			return
+		}
+	}
+}
+
+// CreateCosmosFromLocation creates a new Cosmos CR using data from a
+// *MongodbURL.Location. It assumes the location to be of type "NFS".
+func (s *Scheduler) CreateCosmosFromLocation(location *pensieve.Location, bucket string) error {
+	nfs := pensieve.NewNFSLocation(location.Details.Endpoint)
```
giacope, line 237: Be more generic about the location for future use cases

cosmos/scheduler/main.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+	kubeClient, err := kubernetes.NewForConfig(config)
+	(&scheduler.Scheduler{
+		KubeAlpha:     kubeAlpha,
+		KubeClientset: kubeClient,
+		Pensieve:      pensieve.NewHelper(mongoClient.Database(metadataDatabase).Collection(pensieveCollection)),
+		Namespace:     viper.GetString("namespace"),
+		Cloudserver:   viper.GetString("cloudserver_endpoint"),
+		MongodbClient: mongoClient,
+	}).Run()
```
giacope, line 82: Suggestions for a better way of dealing with the configuration of the scheduler are welcome :)

cosmos/scheduler/pkg/pensieve.go
```diff
@@ -0,0 +1,235 @@
...
+// have one of the types passed as argument or an error.
+func (pensieve *Helper) GetLocationsWithTypes(types []string) ([]Location, error) {
+	locations, err := pensieve.GetLocations()
+	if err != nil {
+		return nil, err
+	}
+	ret := make([]Location, 0)
+	for _, loc := range locations {
+		for _, t := range types {
+			if strings.Contains(loc.LocationType, t) {
```
giacope, line 112: change for ==

cosmos/scheduler/pkg/pensieve.go
```diff
@@ -0,0 +1,235 @@
...
+	Proto   string
+	Version string
+	IPAddr  string
+	Path    string
+	Options string
+}
+
+// NewNFSLocation returns a *NFSLocation the corresponds to the given endpoint.
+// It assumes that the endpoint is properly formed.
+func NewNFSLocation(endpoint string) *NFSLocation {
```
giacope, line 221: Make this function error out properly when the endpoint is malformed.

kubernetes/zenko/templates/cosmos-scheduler/deployment.yaml
```diff
@@ -0,0 +1,34 @@
+apiVersion: apps/v1
```
giacope, line 2: Add proper templates and meta stuff to this deployment.

cosmos/scheduler/cmd/scheduler.go
```diff
@@ -0,0 +1,247 @@
...
+	if err != nil {
+		log.Println(err)
+		return
+	}
+	if len(locationBson) == 0 {
+		return
+	}
+	collection := s.MongodbClient.Database("metadata").Collection("__metastore")
+	ctx := context.Background()
+	cur, err := collection.Watch(ctx, mongo.Pipeline{
```
alexanderchan-scality: it might be safer to set `batchSize` for the watch to only return one document at a time.

cosmos/scheduler/cmd/scheduler.go
```diff
@@ -0,0 +1,247 @@
...
+	}
+	if !exists {
+		if err = s.createIngestionSecret(); err != nil {
+			panic(err.Error())
+		}
+	}
+	msgCh := make(chan interface{}, 1)
+	go func() {
+		for {
+			s.handler(msgCh)
```
alexanderchan-scality: for clarification, why does the `handler` need to quit for each "overlay update"? I may be missing something, but it doesn't look like the `handler` does anything with that event. For each `__metadata` changes, a check for the latest overlay configuration is always performed, so the `handler` routine doesn't need to "restart"

If that quit doesn't matter, I believe the `handler` can be run without the loop?

cosmos/scheduler/cmd/scheduler.go
```diff
@@ -0,0 +1,247 @@
...
+	for {
+		if err := s.waitForOverlayUpdate(); err != nil {
+			log.Println(err)
+		} else {
+			msgCh <- "overlay update"
+		}
+	}
+}
+
+func (s *Scheduler) waitForOverlayUpdate() error {
```
alexanderchan-scality: I think `waitForOverlayUpdate` is creating a new `changeStream` for each  it's called.
maybe move the loop logic into the method?

an example using `sync.WaitGroup`:
(not sure how this should exit gracefully)

```
func (s *Scheduler) Run() {
          ...
          wg := sync.WaitGroup
          defer wg.Wait()
          wg.Add(1)
          ... handler(msgCh, &wg) goroutine
          wg.Add(1)
          go waitForOverlayUpdate(msgCh, &wg)
          ...
}

func (s *Schedule) waitForOverlayUpdate(ch chan<- string, wg *sync.WaitGroup) {
        defer wg.Done()
        collection := s.MongodbClient.Database("metadata").Collection("PENSIEVE")
	ctx := context.Background()
	cur, err := collection.Watch(ctx, mongo.Pipeline{})
        defer cur.Close(ctx)
	if err != nil {
		log.Println(err)
	}
        for {
                if cur.Next(ctx) {
                        ch <- "update overlay"
                } else {
                        if err := cur.Err(); err != nil {
                                log.Println(err)
                        }
                }
        }
}
```

cosmos/scheduler/cmd/scheduler.go
giacope: The reason for that is because every time the handler restarts, it starts a new `s.watchBucketEvents()` which will create a new watch stream with a new set of buckets.

cosmos/scheduler/cmd/scheduler.go
```diff
@@ -0,0 +1,247 @@
...
+		log.Println(err)
+		return
+	}
+	defer cur.Close(ctx)
+	for {
+		select {
+		case <-quit:
+			return
+		default:
+			if cur.Next(ctx) {
```
alexanderchan-scality: from our discussion, the idea of making this section a goroutine and closing the cursor on quit does seem to work. on cursor close, if `cursor.Next(ctx)` is block for a new entry, `false` will be returned and it will unblock. the  `cursor.Err()` will also be set to `ended session was used`.

cosmos/scheduler/cmd/scheduler.go
```diff
@@ -0,0 +1,265 @@
...
+}
+
+func (s *Scheduler) createIngestionSecret() (string, error) {
+	accessKey, secretKey, err := s.Pensieve.GetServiceAccountCredentials("md-ingestion")
+	if err != nil {
+		return "", err
+	}
+	secret, err := s.KubeClientset.CoreV1().Secrets(s.Namespace).Create(&v1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "cosmos-metadata-ingestion",
```
ssalaues, line 171: Add a todo here as it would conflict if there are multiple Zenkos within one namespace.

ssalaues (approved):

cosmos/scheduler/cmd/scheduler.go
giacope: Done

cosmos/scheduler/cmd/scheduler.go
```diff
@@ -0,0 +1,266 @@
...
+				var elem BucketTransaction
+				if err := cur.Decode(&elem); err != nil {
+					log.Println(err)
+					return
+				}
+				ch <- elem
+			} else {
+				return
+			}
+			if err := cur.Err(); err != nil {
```
alexanderchan-scality: should this be in the `else` statement? correct me if I am wrong, this case doesn't seem reachable if `cur.Next() == false`

cosmos/scheduler/cmd/scheduler.go
```diff
@@ -0,0 +1,266 @@
...
+	collection := s.MongodbClient.Database("metadata").Collection("PENSIEVE")
+	ctx := context.Background()
+	cur, err := collection.Watch(ctx, mongo.Pipeline{})
+	if err != nil {
+		log.Println(err)
+		close(ch)
+		return nil
+	}
+	go func() {
+		defer cur.Close(ctx)
```
alexanderchan-scality: the function doesn't end for the defer to kick in

cosmos/scheduler/cmd/scheduler.go
```diff
@@ -0,0 +1,266 @@
...
+		for {
+			if cur.Next(ctx) {
+				ch <- "overlay update"
+			}
+			if err := cur.Err(); err != nil {
```
alexanderchan-scality, line 142: should `return` here?

cosmos/scheduler/cmd/scheduler.go
```diff
@@ -0,0 +1,266 @@
...
+
+// Run starts the Cosmos Scheduler
+func (s *Scheduler) Run() {
+	log.Println("starting cosmos scheduler")
+	err := s.configureIngestionSecret()
+	if err != nil {
+		log.Panicln(err.Error())
+	}
+	quit := make(chan bool)
+	overlayUpdates := s.watchOverlayUpdates()
```
alexanderchan-scality: should check for `overlayUpdates != nil`?

alexanderchan-scality (commented):
first pass comments

cosmos/scheduler/cmd/scheduler.go
giacope: That is correct

cosmos/scheduler/cmd/scheduler.go
giacope: no, removing the `defer` instead...

cosmos/README.md
```diff
@@ -0,0 +1,22 @@
+# Cosmos
+
+This directory holds most the code the corresponds to the Cosmos ingestion framework.
```
alexanderchan-scality: minor
```suggestion
This directory holds most the code that corresponds to the Cosmos ingestion framework.
```

cosmos/README.md
```diff
@@ -0,0 +1,22 @@
...
+The subdirecory layout is as follows:
```
alexanderchan-scality: minor
```suggestion
The subdirectory layout is as follows:
```

cosmos/README.md
```diff
@@ -0,0 +1,22 @@
...
+    Cosmos CRs in the Kubernetes API.
+
+- `/operator`
+
+    This directory contains the necessary files for building the Cosmos Helm Operator
+    Docker image.
+
+- `/scheduler`
+
+    This directory holds the code for the Orechestration Daemon.
```
alexanderchan-scality: minor
```suggestion
    This directory holds the code for the Orchestration Daemon.
```

cosmos/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+- `/vendor`
+
+    This directory contains the dependecies for all the Cosmos Go packages.
```
alexanderchan-scality: minor
```suggestion
    This directory contains the dependencies for all the Cosmos Go packages.
```

alexanderchan-scality (approved):

giacope:
@bert-e approve

tmacro (approved):

PR #537, opened 2019-03-05
bf: fix crd install on upgrades

ssalaues:
Zenko upgrades doesn't install the Cosmos CRD. This is a result of the `crd-install` helm hook which isn't absolutely required in our use-case since we do not template any Custom Resources along with the initial install/upgrades

ZENKO-1548

Known helm issue:
https://github.com/helm/helm/issues/4697


giacope (approved):

JianqinWang (approved):

ssalaues:
@bert-e approve

PR #543, opened 2019-03-07
ZENKO-1556 PV guarantee

ssalaues:
- Updates the nfs mount options to use a comma separated string
- Updates the PVC to explicitly look for a PV of the corresponding name
- Updates examples to reflect changes

ZENKO-1556


ssalaues:
wrong branch name

PR #547, opened 2019-03-11
Fix s3-fuzzer

tmacro:
Refactor and fix nightly s3 listing fuzzer run

### DIFF
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -242,16 +242,19 @@ run-listing-fuzzer:
 		--restart=Never \
 		--attach=True \
 		--image=zenko/s3fuzz:latest \
+		--image-pull-policy="Always" \
 		--namespace $(HELM_NAMESPACE) \
-		--env ZENKO_HOST="http://$(ZENKO_HELM_RELEASE)-cloudserver:80" \
-		--env COUNT=1 \
-		--env SEED=$(shell date '+%s') \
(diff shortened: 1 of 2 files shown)


tests/Makefile
```diff
@@ -241,17 +241,20 @@ run-listing-fuzzer:
 		--rm \
 		--restart=Never \
 		--attach=True \
-		--image=zenko/s3fuzz:latest \
+		--image=tmacro/s3fuzz:latest \
```
bennettbuchanan: Do we want to pull from your personal docker hub image?

bennettbuchanan (approved):

tests/Makefile
rahulreddy: That is of concern, we should be using zenko's registry

tmacro: From the slack message: `custom docker image left in purposefully until done with testing`

bennettbuchanan: Would it be possible to push a custom test image to the zenko account in docker hub instead? If we do that, the test won't be bound to your account. 

Or maybe I'm misunderstanding that you'll push another commit to change it back before this PR is merged.

tmacro: I meant the second, I'm doing some testing to make sure its deterministic and then I'll swap it over

ssalaues (approved):

tmacro:
@bert-e approve

PR #590, opened 2019-03-20
ft: ZENKO-1238 E2E s3c ingestion tests

JianqinWang:
- tests for ingestion and oob updates from RING S3C

ingestion tests on their own passed here:
https://eve.devsca.com/github/scality/zenko/#/builders/2/builds/2372


JianqinWang:
reopening: https://github.com/scality/Zenko/pull/591

PR #612, opened 2019-03-25
documentation: ZENKO-1519_document_lifecycle_transition REVISED

wabernat:
This PR adds changes entered after the PR was approved. 

This PR fixes ZENKO-1519

ZENKO-1519 already has been merged. Some errors were spotted. This PR addresses those errors.


wabernat:
@bert-e reset

wabernat:
@bert-e approve

docs/docsource/Operation-Architecture/Orbit_UI/Bucket_Management/Lifecycle_Management-Transition.rst
```diff
@@ -1,6 +1,21 @@
...
+Object lifecycle transition policies enable you to change the location of an
+object or object type based on its age.
+
+Transition policies do not fully transition non-versioned sources. For such 
+sources, the object is copied to the destination, but it is not removed from 
+the source. For example, if a user sets a policy to transition an object from
+Azure (source) to AWS S3, the object is copied from Azure to AWS S3, but not
+removed from Azure.
+
+This affects the following location constraints: Azure, GCP, and--if the
```
jonathan-gramain: Once we have this mechanism in place, GCP will be able to be versioned as well as non-versioned. Zenko does not yet support versioning on top of GCP, but the GCP bucket itself can be versioned, and that is what matters for supporting deletion in transition policies.

So I would simply move "GCP" to the "if the source is non-versioned" list.

It might be worth adding that sproxyd-RING is supported, regardless of the zenko bucket being versioned or nonversioned (same for S3 local storage aka. s3-data, but AFAIK we don't officially support it so not sure it should be documented).

wabernat: done

docs/docsource/Operation-Architecture/Orbit_UI/Bucket_Management/Lifecycle_Management-Transition.rst
```diff
@@ -28,7 +44,7 @@ To establish a lifecycle transition rule:
       :scale: 75 %
       :align: center
 
-   You may name a directory or subdirectory to which the rule applies. Enter
+   You may specify an prefix to identify objects to which the rule applies. Enter
```
JianqinWang, line 47: 'an prefix' -> 'a prefix'

tmacro (approved):

ssalaues (approved):

ssalaues:
@bert-e bypass_build_status

docs/docsource/Operation-Architecture/Orbit_UI/Bucket_Management/Lifecycle_Management-Transition.rst
```diff
@@ -1,8 +1,24 @@
 Object Lifecycle Management: Transition
 =======================================
 
+Object lifecycle transition policies enable you to change the location of an
+object or object type based on its age.
+
+Transition policies do not fully transition non-versioned sources. For such
```
jonathan-gramain, line 7: The latest news is that for nonversioned source, instead of deleting the object, we will add a tag telling that the object is deleted. (there will still be a tiny chance that we might tag an updated version of the object instead of the one already transitioned, but at least we would not inadvertently delete a more recent object in such case).

So this is still true, but could be extended when we have the details of what we will do exactly.

jonathan-gramain (approved):

PR #629, opened 2019-03-30
bugfix: ZENKO-1677 increase tests lifecycle cron delay to 1mn

jonathan-gramain:
Revert tests cron job delay to every minute, so that transitions do not
overlap on the same object. In theory nothing very bad should happen
in such case (no data loss) but it's possible that transitions cannot
occur timely and need to be retried.

### DIFF
--- a/eve/ci-values.yaml
+++ b/eve/ci-values.yaml
@@ -128,7 +128,7 @@ backbeat:
     replicaFactor: 2
   lifecycle:
     conductor:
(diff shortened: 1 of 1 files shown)


bennettbuchanan (approved):

ssalaues (approved):

ssalaues:
@bert-e bypass_build_status

jonathan-gramain:
@bert-e approve

PR #661, opened 2019-05-02
ft(ZENKO-1769): Lower backbeat replication retry timeouts for CI

tmacro:
Lowers the range of time backbeat waits before retrying a failed replication to 2-10 seconds. Transient errors such as socket hangups can cause replication to fail, which coupled with a 1 minute initial timeout would cause tests to timeout before backbeat has a chance to retry. This should reduce flakiness by allowing backbeat to retry objects before a test timeouts and fails.

### DIFF
--- a/eve/ci-values.yaml
+++ b/eve/ci-values.yaml
@@ -96,6 +96,31 @@ backbeat:
     pullPolicy: Always
   replication:
     replicaFactor: 2
+    retry:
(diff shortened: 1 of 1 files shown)


ssalaues (approved):

jonathan-gramain (approved):

tmacro:
@bert-e approve

PR #664, opened 2019-05-03
bf: safely handle multiple ingestion buckets

ssalaues:
This safely handles when an additional ingestion bucket is created. It does this
by using a label with the bucket value that is referenced when attempting to
`ApplyChanges`.

ZENKO-1756

### DIFF
--- a/cosmos/scheduler/Makefile
+++ b/cosmos/scheduler/Makefile
@@ -1,5 +1,5 @@
 COSMOS_IMAGE := zenko/cosmos-scheduler
-COSMOS_TAG := 0.2.1
+COSMOS_TAG := 0.2.2
 COSMOS_BINARY := ./scheduler
 
 .PHONY: setup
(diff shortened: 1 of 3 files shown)


tmacro (approved):

dora-korpar (approved):

rachedbenmustapha:
@ssalaues about the PascalCase adjustment, it's only needed if you want to make the functions/methods exported to other packages - uppercase denotes public visibility. It's generally discouraged to export more than the strict minimum as a "public interface", to make maintainer's lives easier when trying not to break API/ABIs. My suggestion would be to leave camelCase in place unless we want to publicly support these APIs.

ssalaues:
@RachedBenMustapha yeah those are good points. It was originally a mix of the two and I wasn't entirely sure of the significance of either. I'll revert it camelCase then.

ssalaues:
@bert-e approve

ssalaues:
@bert-e status

PR #666, opened 2019-05-07
bugfix: zenko 1715 mongo upgrade

ssalaues:
Adds additional check to prevent a false positive upgrade. Also adds more info on the upgrade procedure.

ZENKO-1715

### DIFF
--- a/kubernetes/zenko/templates/maintenance/mongo-capabilites.yaml
+++ b/kubernetes/zenko/templates/maintenance/mongo-capabilites.yaml
@@ -3,17 +3,16 @@
 apiVersion: batch/v1
 kind: Job
 metadata:
-  name: {{ template "zenko.fullname" . }}-mongo-capabilities
+  name: {{ template "zenko.fullname" . }}-mongodb-upgrade
   labels:
-    component: maintenance
+    component: maintenance-job
 {{ include "zenko.labels.standard" . | indent 4 }}
 spec:
   template:
     metadata:
       labels:
-        component: maintenance
-        app: {{ template "zenko.name" . }}
-        release: {{ .Release.Name | quote }}
+        component: mongodb-upgrade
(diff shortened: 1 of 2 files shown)


kubernetes/zenko/templates/maintenance/mongo-capabilites.yaml
```diff
@@ -58,6 +58,12 @@ spec:
                    sleep(10000);
                }
                db.adminCommand({ setFeatureCompatibilityVersion: "3.6" });
+               sleep(10000);
+               version = db.adminCommand({ getParameter: 1, featureCompatibilityVersion: 1 }).featureCompatibilityVersion.version;
+               if (version != '3.6'){
+                   print("Unable to upgrade capabilities, manual upgrade man be neccessary")
```
nicolas2bert: typo ? `may be`

nicolas2bert (approved):

kubernetes/zenko/templates/maintenance/mongo-capabilites.yaml
```diff
@@ -58,6 +58,12 @@ spec:
...
+                   quit(1)
```
JianqinWang: do we need `;` here

kubernetes/zenko/templates/maintenance/mongo-capabilites.yaml
ssalaues: Oops! You're right, thank you

kubernetes/zenko/templates/maintenance/mongo-capabilites.yaml
ssalaues: My bad, I'll fix that.

JianqinWang (approved):

ssalaues:
@bert-e approve

PR #667, opened 2019-05-07
ZENKO-1776: package logs as .tar.gz files for easier download

JianqinWang:

Package the logs so it's easier to download.

Also fixes issue where queue and quorum pod logs are not dumped properly. Currently there are 3 tests run on 3 parallel zenko instances, but for each zenko instance, the queue and quorum pods do not have unique names. When the logs are being dumped, those logs will be overwritten. 

fixes ZENKO-1776

### DIFF
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -35,7 +35,7 @@ models:
   - Upload: &upload_artifacts
       source: tests/artifacts
       urls:
-        - "*"
+        - "*.tar.gz"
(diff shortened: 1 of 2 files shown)


ssalaues (approved):

ssalaues:
@bert-e bypass_build_status

tmacro (approved):

JianqinWang:
@bert-e approve

PR #675, opened 2019-05-08
bugfix: use unique src bucket name for py tests for each build

JianqinWang:
Now that we're running multiple builds simultaneous for E2E CI, using the same src bucket name against GCP will cause issues (multiple buckets on Zenko with same name conducts CRR to the actual GCP backend).

We append the name with the build number which should make sure that when multiple Zenko tests are CRR'ing to the same destination bucket, we will not have conflicts.

Also fixing `SRC_BUCKET_NAME`s for various other backends, even if they use mocks. If we run multiple Nightly tests in parallel, we could see issues.

These changes only affect python tests


rahulreddy (approved):

ssalaues (approved):

ssalaues:
@bert-e bypass_build_status

JianqinWang:
@bert-e approve

PR #678, opened 2019-05-09
bf: ZENKO-1744 ingestion populator max reads

philipyoo:
Add max reads variable for ingestion populator to limit entries read per batch from s3c

### DIFF
--- a/kubernetes/zenko/charts/backbeat/templates/ingestion/producer_deployment.yaml
+++ b/kubernetes/zenko/charts/backbeat/templates/ingestion/producer_deployment.yaml
@@ -46,6 +46,8 @@ spec:
               value: "{{ template "backbeat.redis-hosts" . }}"
             - name: REDIS_HA_NAME
               value: "{{ .Values.redis.sentinel.name }}"
+            - name: QUEUE_POPULATOR_BATCH_MAX_READ
+              value: "1000"
           livenessProbe:
             httpGet:
               path: {{ .Values.health.path.liveness }}


philipyoo:
@bert-e reset

JianqinWang (approved):

ssalaues (approved):

kubernetes/zenko/charts/backbeat/templates/ingestion/producer_deployment.yaml
```diff
@@ -46,6 +46,8 @@ spec:
               value: "{{ template "backbeat.redis-hosts" . }}"
             - name: REDIS_HA_NAME
               value: "{{ .Values.redis.sentinel.name }}"
+            - name: QUEUE_POPULATOR_BATCH_MAX_READ
+              value: "1000"
```
JianqinWang, line 50: https://github.com/scality/Zenko/blob/eac712e601bbf9fbe6c78e1706724f14f2e78ee4/kubernetes/zenko/charts/backbeat/values.yaml#L60 Maybe it's preferred that this value be added there, so it's easier for tuning (all values and variables in one file)

philipyoo: I'm sort of conflicted with this change itself since I feel like it shouldn't be tuneable by a user. I'd rather try and avoid more flexibility for adjusting this value

JianqinWang: If that's the case, shouldn't we leave it out of the deployment chart in general, and only reference/set this value within the backbeat code? So it gets built into the image

philipyoo:
rather than a potential tuneable, planning on fixing this value on backbeat configs itself

philipyoo:
Sorry 🙏 re-opening

kubernetes/zenko/charts/backbeat/templates/ingestion/producer_deployment.yaml
rahulreddy: ~~Does it actually affect CRR too?~~ Nevermind, I see that this change only applies to ingestion chart

philipyoo: only if adding to the replication pod charts

philipyoo:
@bert-e approve

PR #682, opened 2019-05-10
ci: fixes false positives on stability checks

ssalaues:
This fixes an issue where a CI build was allowed to proceed with
tests before being fully stabilized. This is due to the short window where a
container restarts and is momentarily ready before crashing for whatever
reason. The fix is to add a stabilization period of 10 second where everything
is expected to be ready.

ZENKO-1796

### DIFF
--- a/tests/zenko_tests/create_buckets.py
+++ b/tests/zenko_tests/create_buckets.py
@@ -22,6 +22,8 @@ FAIL_STATUS = [
     'Error',
 ]
 
+PASSING_COUNT = 10
+
 def is_ignored(name):
     for ignored in IGNORED_PODS:
         if ignored.search(name):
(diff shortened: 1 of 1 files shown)


JianqinWang (approved):

dora-korpar (approved):

tests/zenko_tests/create_buckets.py
```diff
@@ -68,13 +69,15 @@ def wait_for_pods(delete, timeout):
...
+                        passed_count = 0
+        if passed:
+            if passed_count >= 10:
+                break
+            passed_count = passed_count + 1
+        time.sleep(1)
 
     if not passed:
-      _log.info('Containers have not become ready and %d has elasped' % timeout)
+      _log.info('Containers have not become ready and %d has elasped' % timeout - time.time())
```
jonathan-gramain: minor: could add unit: `%d seconds have elapsed`

jonathan-gramain (approved):

tests/zenko_tests/create_buckets.py
ssalaues: I decided to omit that change completely as I was just using that change for debugging but thank you!

ssalaues:
@bert-e approve

PR #699, opened 2019-05-17
Documentation/zenko 1817 standardize directory names titles

wabernat:
This PR standardizes and simplifies the "book" titles and file hierarchies for the Zenko documentation. We need it because: 
A) it makes no sense to call our web help a "Guide" or "Manual"
B) if all operation documentation resides in a directory called "operation", then we can automate our references to the documentation more practically. Subject titles are Operation, Installation, and Reference. Directories are named operation, installation, and reference.  

This fixes #ZENKO-1817


docs/docsource/index.rst
```diff
@@ -9,6 +9,6 @@ Zenko
 .. toctree::
    :maxdepth: 2
 
-   Installation/index
-   Operation-Architecture/index
-   Reference/index
+   installation/index
```
ssalaues, line 12: Looks like tox is failing here 
```
/build/docs/docsource/index.rst:9:toctree contains reference to nonexisting document 'installation/index'
```

wabernat: Okay, I'm going to sound the alarm. I have a clean build on my local machine. Lucie was describing issues that sounded very similar. Can you help me by telling me if you can see any delta between the requirements and settings embodied in the local Makefile and what BitBucket is running? 

My suspicion is that RelEng has changed BitBucket's build process but not updated our Makefiles.

ssalaues: It looks like it's just a case-sensitive issue with the capitalization of `installation`. The folder being referenced in this file is `installation` but the actual folder with this file is named `Installation`. At least according to the diff shown in this PR.

wabernat: This is confusing. The whole point of this was to get rid of the "Operation-Architecture"/"Installation/"Reference" (uppercase) naming scheme, and the branch I (thought I) pushed has the lowercase directories in the source, which are thus named in the index.rst file. 

Are you looking at: 2137e615fcfda577f68321281688714e28bc2183 ?

ssalaues: Yes, look at the in browser diff. If you search for `Installation/index` you will see the file and full path `docs/docsource/Installation/index.rst`
https://github.com/scality/Zenko/commit/2137e615fcfda577f68321281688714e28bc2183#diff-bcddfd19dd179dc84102c570e6c1a9b0R1

This means that the folder `Installation` is still capitalized while this index.rst file is expecting `installation`

wabernat: This is baffling. 

My local branch shows lowercase in both the directory names and in index.html. I see NO uppercase directory names at this level. Git log shows the same hash as the one on GitHub. If I do a git diff, there is a null response. If I pull, it says I'm up to date. If I push, it says the same. Yet the link you sent indicates GitHub has not accepted my changes and is retaining the old capitalized directory names.

From the local side, this was a non-issue. I pushed it to GitHub, and now we see this problem that I explicitly fixed before I pushed. 

I don't understand A) how this is happening, or B) how I can fix it.

wabernat: Satori: 

Git is not case-sensitive concerning directory names. Operation-Architecture changed (because I deleted "-Architecture" in addition to the case change). Not so for installation and reference.

hacking...

wabernat:
@bert-e approve

ssalaues (approved):

tmacro (approved):

PR #706, opened 2019-05-22
ci: fix log dump

ssalaues:
Always exit 0 on log generation. This resolves the flaky logs dump step that occasionally fails because it is trying to access logs to a container that no longer exists. It also adds `warnOnFailure` to non-critical steps along with adding missing `haltOnFailure` to critical steps.

ZENKO-1833


JianqinWang (approved):

rahulreddy (approved):

ssalaues:
@bert-e approve

PR #731, opened 2019-06-07
ci: fixes OOM issues on data processor pods

ssalaues:
Increases the upper memory limit for the  data processor pods. The memory usage has increased significantly possibly due to https://github.com/scality/Zenko/pull/708 leading to the pods being killed by K8s when it quickly reaches the threshold.

fixes ZENKO-1868

### DIFF
--- a/eve/ci-values.yaml
+++ b/eve/ci-values.yaml
@@ -163,15 +163,22 @@ backbeat:
 
   replication:
     dataProcessor:
+      resources:
+        limits:
+          cpu: 250m
+          memory: 1Gi
+        requests:
+          cpu: 100m
+          memory: 512Mi
+    populator:
       resources: &backbeat
         limits:
           cpu: 250m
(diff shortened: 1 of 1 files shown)


dora-korpar (approved):

JianqinWang (approved):

rahulreddy (approved):

ssalaues:
@bert-e wait

tmacro (approved):

ssalaues:
I just force pushed an update to the PR. The original changes updated the memory requests/limits for all of the backbeat pods instead of just the data processor. This would be very bad since we can have 8 builds each running 3 Zenko installs with 20 backbeat containers per Zenko (480 containers just for backbeat) leading to far higher memory requests than we can support in our CI.

ssalaues:
@bert-e approve

tcarmet (approved):

PR #738, opened 2019-06-11
ZENKO-1879 improve ci stability

ghost:

Add two main changes to the E2E tests:
 - Use limits-only for resources to ensure availability of the resources and improve stability of the service
 - Use a script to get the output from Zenko-e2e pod,  which is resilient to disconnection between GKE (launcher environment of tests) and the dedicated Scality.cloud cluster (hosting the kubernetes cluster where the service is run for the tests).

### DIFF
--- a/eve/ci-values.yaml
+++ b/eve/ci-values.yaml
@@ -16,8 +16,11 @@ s3-data:
     enabled: false
   resources:
     requests:
-      cpu: 50m
-      memory: 128Mi
+      cpu: 200m
+      memory: 512Mi
+    limits:
+      cpu: 200m
+      memory: 512Mi
 
 
 mongodb-replicaset:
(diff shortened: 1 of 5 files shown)


jonathan-gramain (approved):

ssalaues (changes requested):
The `limits` can be provisioned far beyond the capabilities of a node. However the `requests` themselves require the node to absolutely have the resources unassigned. You can test this out easily by running several concurrent builds and run a `kubectl describe node` you will see that the limits themselves can be over 100% of node capacity but the requests cannot. https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/#how-pods-with-resource-requests-are-scheduled

I think having both would be appropriate but perhaps it might be best to simply have the requests/limits as the same to prevent spikes in usage?

ghost:
@bert-e force_reset

ssalaues:
@DavidPineauScality The changes look good to me. However can you target `development/1.0`?

ghost:
@bert-e force_reset

ghost:
.

ssalaues (approved):

tmacro (approved):

ghost:
Do you think it's necessary to update my branch before going forward ?
Every branch is now passing tests.

ssalaues:
@DavidPineauScality it should be fine. The changes were minimal and shouldn't conflict. Also I typically keep watch on the queues for any necessary rebuilds.

ghost:
@bert-e approve

PR #758, opened 2019-06-17
bf: semVer check for pre-releases

ssalaues:
Currently, helm install will fail on most managed K8s installs. This is because of the custom versions used such as `v1.11.7-gke.6` is considered a pre-release in the library used by helm and thus doesn't satisfy the requirements. This fix allows for compatibility with such versions by evaluating them within the requirement range.

fixes # ZENKO-1904

Related to:
https://github.com/helm/helm/issues/3810

### DIFF
--- a/kubernetes/zenko/Chart.yaml
+++ b/kubernetes/zenko/Chart.yaml
@@ -1,7 +1,7 @@
 name: zenko
 version: 1.1.0-rc.3
 appVersion: 1.1.0-rc.3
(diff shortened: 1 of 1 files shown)


nicolas2bert (approved):

dora-korpar (approved):

tmacro (approved):

ssalaues:
@bert-e approve

rahulreddy:
@bert-e bypass_build_status

PR #786, opened 2019-07-03
improvements: add health checks

ssalaues:
- Adds a health check to the scheduler for the mongo connection
- Adds some comments and better logging
- Adds startup check for credentials instead of solely relying on overlay updates

### DIFF
--- a/cosmos/scheduler/cmd/scheduler.go
+++ b/cosmos/scheduler/cmd/scheduler.go
@@ -41,6 +41,10 @@ func (s *Scheduler) Run() {
 		return
 	}
 	go s.healthCheckServer()
+	err = s.configureIngestionSecret(true)
+	if err != nil{
+		log.Fatal("error configuring secret:", err)
+	}
 	for {
 		go s.watchBucketUpdates(quit)
 		<-overlayUpdates
@@ -65,7 +69,10 @@ func (s *Scheduler) healthCheckServer() {
 	log.Fatal(http.ListenAndServe(":8080", nil))
 }
 
-func (s *Scheduler) configureIngestionSecret() error {
+// Checks for existing secret and creates one if non existant. If the a configmap
+// exists but does not match the credentials in MongoDB then the secret will be updated.
+// The init bool is used for start up sequences otherwise logging becomes too verbose.
+func (s *Scheduler) configureIngestionSecret(init bool) error {
 	accessKey, secretKey, err := s.Pensieve.GetServiceAccountCredentials("md-ingestion")
 	if err != nil {
 		return err
@@ -85,7 +92,7 @@ func (s *Scheduler) configureIngestionSecret() error {
 			return err
 		}
 		log.Println("ingestion credentials successfully patched")
-	}  else {
+	}  else if init {
 		log.Println("found existing up-to-date ingestion secret")
 	}
 	return nil
@@ -165,7 +172,7 @@ func (s *Scheduler) watchOverlayUpdates() (chan interface{}, error) {
 			if err := cur.Err(); err != nil {
 				log.Println(err)
 			}
-			err := s.configureIngestionSecret()
+			err := s.configureIngestionSecret(false)
 			if err != nil {
 				close(ch)
 				cur.Close(ctx)
@@ -248,7 +255,7 @@ type BucketTransaction struct {
 func (s *Scheduler) applyChanges(bucket *BucketTransaction) {
 	location, err := s.Pensieve.GetLocationWithName(bucket.FullDocument.Value.LocationConstraint)
     if err != nil {
-		log.Println("error getting location", err)
+		log.Println("error getting location:", err)
 	} 
 	switch bucket.OperationType {
 	case "insert" :
@@ -264,7 +271,7 @@ func (s *Scheduler) applyChanges(bucket *BucketTransaction) {
 			if err != nil {
 				log.Println(err)
 			}
-		} else {
+		} else if bucket.FullDocument.Value.Deleted == true {
 			log.Println("received delete request but no cosmos created for bucket:", bucket.FullDocument.Value.Name)
 		}
 	}
--- a/cosmos/scheduler/cmd/scheduler.go
+++ b/cosmos/scheduler/cmd/scheduler.go
@@ -3,9 +3,12 @@ package scheduler
 import (
 	"context"
 	"log"
+	"time"
+	"net/http"
 
 	"github.com/mongodb/mongo-go-driver/bson"
 	"github.com/mongodb/mongo-go-driver/mongo"
+	"github.com/mongodb/mongo-go-driver/mongo/readpref"
 	"github.com/scality/zenko/cosmos/api/types/v1alpha1"
 	clientV1alpha1 "github.com/scality/zenko/cosmos/clientset/v1alpha1"
 	"github.com/scality/zenko/cosmos/scheduler/pkg"
@@ -37,12 +40,29 @@ func (s *Scheduler) Run() {
 		log.Println(err)
 		return
 	}
+	go s.healthCheckServer()
 	for {
 		go s.watchBucketUpdates(quit)
 		<-overlayUpdates
 		log.Println("received an overlay update")
 		quit <- true
 	}
+}
+
+func (s *Scheduler) healthCheckServer() {
+	health := func(w http.ResponseWriter, req *http.Request){
+		ctx, cancel := context.WithTimeout(req.Context(), 5*time.Second)
+		defer cancel()
+		err := s.MongodbClient.Ping(ctx, readpref.Nearest())
+		if err != nil {
+			log.Println("MongoDB healthcheck failed:", err)
+			http.Error(w, "Internal error, could not contact MongoDB", 500)
+		} else {
+			w.WriteHeader(http.StatusOK)
+		}
+	}
+	http.HandleFunc("/healthcheck", health)
+	log.Fatal(http.ListenAndServe(":8080", nil))
 }
 
 func (s *Scheduler) configureIngestionSecret() error {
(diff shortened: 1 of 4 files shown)


anurag4DSB (approved):

rahulreddy (approved):

cosmos/scheduler/cmd/scheduler.go
```diff
@@ -31,47 +33,61 @@ type Scheduler struct {
...
-	quit := make(chan bool)
 	overlayUpdates, err := s.watchOverlayUpdates()
 	if err != nil {
 		log.Println(err)
 		return
 	}
+	// safely create or update secret on startup
+	go s.configureIngestionSecret(true)
+	go s.checkMongoConnection()
+	go s.watchBucketUpdates()
```
alexanderchan-scality: correct me if I am wrong, with `watchBucketUpdates` being moved out of the for loop, the  `locationBson` on line 91 doesn't seem to be updating on overlay updates

ssalaues: Yeah you're totally right, I'll fix that. Thank you!

cosmos/scheduler/cmd/scheduler.go
```diff
@@ -31,47 +33,61 @@ type Scheduler struct {
...
+			log.Panicln("ping healthcheck failed:", err)
+		}
+	}
+}
+
+func (s *Scheduler) configureIngestionSecret(init bool) {
 	accessKey, secretKey, err := s.Pensieve.GetServiceAccountCredentials("md-ingestion")
 	if err != nil {
-		return err
+		log.Panicln("error getting service account credentials")
```
rachedbenmustapha: Panics are better suited for programming errors. This is more Fatal material.

cosmos/scheduler/cmd/scheduler.go
```diff
@@ -138,19 +145,18 @@ func (s *Scheduler) watchOverlayUpdates() (chan interface{}, error) {
...
-		for {
-			if cur.Next(ctx) {
-				ch <- "overlay update"
-			}
-			if err := cur.Err(); err != nil {
-				log.Println(err)
-			}
-			err := s.configureIngestionSecret()
-			if err != nil {
+		defer cur.Close(ctx)
```
rachedbenmustapha: I'm not sure if the race detector would be happy about accessing `cur` here in a different goroutine than where it's created and without locking. I suggest instanciating as many of the concurrency-unsafe objects as possible in the goroutine (the channel is fine though). It would also fix the quirk of deferring a close on something opened in another scope.

cosmos/scheduler/cmd/scheduler.go
```diff
@@ -95,36 +111,27 @@ func (s *Scheduler) watchBucketUpdates(quit chan bool) error {
 		}}},
 	})
 	if err != nil {
-		log.Println(err)
+		log.Println("error on collection watch",err)
 		return err
 	}
 	ch := make(chan BucketTransaction)
 	go func() {
 		defer cur.Close(ctx)
```
rachedbenmustapha: Same comment here as for `watchOverlayUpdates`.

cosmos/scheduler/cmd/scheduler.go
```diff
@@ -31,47 +33,61 @@ type Scheduler struct {
...
 func (s *Scheduler) Run() {
 	log.Println("starting cosmos scheduler")
-	quit := make(chan bool)
 	overlayUpdates, err := s.watchOverlayUpdates()
 	if err != nil {
 		log.Println(err)
 		return
 	}
+	// safely create or update secret on startup
+	go s.configureIngestionSecret(true)
```
rachedbenmustapha: I there a reason this is done asynchrously?

ssalaues: I was going back and forth on this but ultimately I didn't think it needed to be a blocking operation in this context.

cosmos/scheduler/cmd/scheduler.go
ssalaues: Yeah I was just reading about that. I think I'll need to change a few of these.

cosmos/scheduler/cmd/scheduler.go
ssalaues: Yeah that makes sense. There's definitely lots more that need to be fixed/updated. It's clear to me that I need to spend more time on these changes so I think I'll drop this commit from this PR and open a separate PR with your suggestions/considerations.

cosmos/scheduler/cmd/scheduler.go
rachedbenmustapha: Rule of thumb is that things should be synchronous unless there's a real need for them to happen later and out of our sight. Here anything that needs to use the new secret may run too early, creating a race or deadlock (depending), because we don't have a way of knowing when (if) the operation is successful.

cosmos/scheduler/cmd/scheduler.go
```diff
@@ -37,6 +39,7 @@ func (s *Scheduler) Run() {
 		log.Println(err)
 		return
 	}
+	go s.checkMongoConnection()
```
rachedbenmustapha: Instead of manually managing timers and crashing, have you thought about implementing a tiny `/healthcheck` route that does the ping, and can be called by kube directly? This makes it easy to have "flexible" behavior (such as interval, initial delay, thresholds, ...) at no cost.

cosmos/scheduler/cmd/scheduler.go
ssalaues: Ah yeah those are definitely good reasons. Thank you!

cosmos/scheduler/cmd/scheduler.go
ssalaues: I have considered that as part of the larger refactor that I'm working on. But I'll add this route instead of having temporary solutions.

cosmos/scheduler/cmd/scheduler.go
```diff
@@ -45,7 +52,23 @@ func (s *Scheduler) Run() {
 	}
 }
 
-func (s *Scheduler) configureIngestionSecret() error {
+func (s *Scheduler) healthCheck() {
```
ssalaues: @RachedBenMustapha I just pushed this section as the updated health check.

cosmos/scheduler/cmd/scheduler.go
```diff
@@ -45,7 +52,23 @@ func (s *Scheduler) Run() {
...
+	health := func(w http.ResponseWriter, req *http.Request){
+		ctx := context.Background()
+		err := s.MongodbClient.Ping(ctx, readpref.Nearest())
+		if err != nil{
+			log.Println("MongoDB healthcheck failed:", err)
+			http.Error(w, "Internal error, could not contact MongoDB", 500)
+		}
```
rachedbenmustapha, line 66: If err is nil, an OK response should be written, the request might hang otherwise

cosmos/scheduler/cmd/scheduler.go
```diff
@@ -45,7 +52,23 @@ func (s *Scheduler) Run() {
 	}
 }
 
-func (s *Scheduler) configureIngestionSecret() error {
+func (s *Scheduler) healthCheck() {
+	health := func(w http.ResponseWriter, req *http.Request){
+		ctx := context.Background()
```
rachedbenmustapha: The context from the request can be reused, this way cancellation (if the connection is severed or the server is shutting down for example) can be cleanly propagated.

cosmos/scheduler/cmd/scheduler.go
```diff
@@ -37,6 +39,11 @@ func (s *Scheduler) Run() {
 		log.Println(err)
 		return
 	}
+	err = s.configureIngestionSecret(true)
+	if err != nil{
+		log.Fatal("error configuring secret:", err)
+	}
+	s.healthCheck()
```
rachedbenmustapha: `healthCheck` is probably blocking so a goroutine would be warranted here

cosmos/scheduler/cmd/scheduler.go
rachedbenmustapha: nitpick: it's actually a health check server, not a health check :)

cosmos/scheduler/cmd/scheduler.go
rachedbenmustapha: (or even better a WithTimeout on the request's context, to avoid a hanging mongodb from having health checks take longer than the check interval)

cosmos/scheduler/cmd/scheduler.go
ssalaues: Oops yeah looks like my rebase got messed up. Thank you.

cosmos/scheduler/cmd/scheduler.go
ssalaues: When I tested this with curl, I got a `HTTP/1.1 200 OK` response which is why I left it out. However if you think it's better to be explicit then I can add it?

rachedbenmustapha: Explicit 100% guaranteed better :)

cosmos/scheduler/cmd/scheduler.go
ssalaues: Reusing the context makes sense, however If it is WithTimeout, then can it be reused?

rachedbenmustapha: Contexts are organised in a tree, so WithTimeout creates a child context of the request context which will get canceled all the same, and also have its own time constraint

cosmos/scheduler/cmd/scheduler.go
```diff
@@ -45,6 +49,22 @@ func (s *Scheduler) Run() {
 	}
 }
 
+func (s *Scheduler) healthCheckServer() {
+	health := func(w http.ResponseWriter, req *http.Request){
+		ctx, cancel := context.WithTimeout(req.Context(), 5*time.Second)
```
rachedbenmustapha, line 58: minor: `defer cancel()` as soon as it's created allows to just forget about it, no matter how many branches can return. 

Can be cleaned up in another PR later, no rush.

ssalaues:
@bert-e approve

PR #790, opened 2019-07-12
ci: fix clusterrole for nightlies

ssalaues:
Fixes the nightly failures on the cosmos tests by using a common release name. This is done by removing the `ZENKO_HELM_RELEASE` and relying on the default release name of `zenko-test` so that creation of the cluster role binding is accurate across all build stages. Additionally it explicitly  cleans up the clusterrolebinding as part of the cleanup.

### DIFF
--- a/eve/scripts/cluster-registry-secret.sh
+++ b/eve/scripts/cluster-registry-secret.sh
@@ -1,5 +1,5 @@
 #!/bin/sh
-set -ex
+set -e
 
 DOCKERCONFIG="$(cat ~/.docker/config.json | base64 | tr -d \\n)"
 
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -24,15 +24,15 @@ models:
   - ShellCommand: &k8s_cleanup
       name: Cleanup
       command: |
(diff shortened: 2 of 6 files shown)


eve/scripts/cluster-registry-secret.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/sh
-set -ex
+set -x
```
ssalaues: This is so the secrets don't get exposed in the stdout.

eve/scripts/cluster-registry-secret.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/sh
-set -ex
+set -e
```
ssalaues, line 2: This is so the secrets don't get exposed in the stdout.

JianqinWang (approved):

dora-korpar (approved):

tmacro (approved):

jonathan-gramain (approved):

ssalaues:
@bert-e approve

PR #792, opened 2019-07-17
ZENKO-1999 - cosmos scheduler refactor

ssalaues:
Refactors the cosmos scheduler code making it more legible, extensible, and improves logging. 

Additionally:
- it is more deterministic with reading the configuration overlays
- Fixes remaining possible racy and deadlock scenarios
- More easily allows for future locations to be added

### DIFF
--- a/cosmos/scheduler/cmd/scheduler.go
+++ b/cosmos/scheduler/cmd/scheduler.go
@@ -127,6 +127,9 @@ func (s *Scheduler) configureIngestionSecret(init bool) error {
 	return nil
 }
 
+// watchBucketUpdates will watch for any new "mirror mode" buckets created in the "__metastore".
+// A new bucket created in a supported BucketTransaction.LocationType will send the transaction
+// on the returned channel.
 func (s *Scheduler) watchBucketUpdates(ctx context.Context) (chan BucketTransaction) {
 	ch := make(chan BucketTransaction)
 	go func() {
@@ -167,6 +170,9 @@ func (s *Scheduler) watchBucketUpdates(ctx context.Context) (chan BucketTransact
 	return ch
 }
 
+// watchOverlayUpdates watches for any new configuration overlays inserted into the
+// 'PENSIEVE' collection. It returns a channel interface that will send on every
+// new overlay update recieved.
 func (s *Scheduler) watchOverlayUpdates(ctx context.Context) (chan interface{}) {
 	ch := make(chan interface{})
 	go func() {
@@ -199,6 +205,8 @@ func (s *Scheduler) watchOverlayUpdates(ctx context.Context) (chan interface{}) 
 	return ch
 }
 
+// getIngestionSecret returns a key value map with 'accessKey' and 'secretKey'.
+// The keys are strings and the values are byte arrays.
 func (s *Scheduler) getIngestionSecret() (map[string][]byte) {
 	secrets, err := s.KubeClientset.CoreV1().
 		Secrets(s.Namespace).Get(s.SecretName, metav1.GetOptions{})
@@ -211,6 +219,8 @@ func (s *Scheduler) getIngestionSecret() (map[string][]byte) {
 	return nil
 }
 
+// setIngestionSecret creates or updates a Kubernetes secret from arguments accessKey and secretKey.
+// The bool argument can be used to patch an existing secret.
 func (s *Scheduler) setIngestionSecret(accessKey string, secretKey string, patch bool) error {
 	kubeSecret := &v1.Secret{
 		ObjectMeta: metav1.ObjectMeta{
@@ -295,6 +305,9 @@ func (s *Scheduler) getCosmosLocationBson(locationType string) (bson.A, error) {
 	return ret, nil
 }
 
+// applyChanges will create a new cosmos custom resource based off the given *BucketTransaction. If
+// the *BucketTransaction is in response to a bucket deletion in Zenko, then any existing custom
+// resource labeled with the bucket will be removed.
 func (s *Scheduler) applyChanges(bucket *BucketTransaction) {
 	location, err := s.Pensieve.GetLocationWithName(bucket.FullDocument.Value.LocationConstraint)
     if err != nil {
@@ -321,8 +334,9 @@ func (s *Scheduler) applyChanges(bucket *BucketTransaction) {
 	return
 }
 
-// Check for existing Cosmos on specified location labeled with the given bucket.
-// Currently only 1 ingestion bucket is supported per location
+// checkForCosmos looks for existing Cosmos custom resources on specified
+// *pensieve.Location labeled with the given bucket string. Currently only 1 ingestion
+// bucket is supported per location.
 func (s *Scheduler) checkForCosmos(location *pensieve.Location, bucket string) bool {
 	cosmos, err := s.KubeAlpha.Cosmoses(s.Namespace).List(metav1.ListOptions{
 		LabelSelector: "bucket="+bucket,
@@ -337,7 +351,7 @@ func (s *Scheduler) checkForCosmos(location *pensieve.Location, bucket string) b
 	return false
 }
 
-// createCosmosFromLocation creates a new Cosmos CR using data from a
+// createCosmosFromLocation creates a new Cosmos custom resource using data from a
 // *MongodbURL.Location. It assumes the location to be of type "NFS".
 func (s *Scheduler) createCosmosFromLocation(location *pensieve.Location, bucket string) error {
 	nfs := pensieve.NewNFSLocation(location.Details.Endpoint)
--- a/cosmos/scheduler/cmd/scheduler.go
+++ b/cosmos/scheduler/cmd/scheduler.go
@@ -60,18 +60,14 @@ func (s *Scheduler) Run() {
 		log.Fatal("run failed configuring secret: ", err)
 	}
 
-	quit := make(chan bool)
-	overlayUpdates, err := s.watchOverlayUpdates()
-	if err != nil {
-		log.Fatal(err)
-	}
-
+	overlayUpdates := s.watchOverlayUpdates(context.Background())
 	for {
-		go s.watchBucketUpdates("nfs", quit)
+		ctx, cancel := context.WithCancel(context.Background())
+		go s.watchBucketUpdates(ctx, "location-nfs-mount-v1")
 		// blocks until an overlay update is recieved on the overlayUpdates channel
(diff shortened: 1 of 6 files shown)


cosmos/scheduler/cmd/scheduler.go
```diff
@@ -32,33 +34,50 @@ type Scheduler struct {
...
+// at :8080/healthcheck
 func (s *Scheduler) healthCheckServer() {
 	health := func(w http.ResponseWriter, req *http.Request){
 		ctx, cancel := context.WithTimeout(req.Context(), 5*time.Second)
 		defer cancel()
-		err := s.MongodbClient.Ping(ctx, readpref.Secondary())
+		err := s.MongodbClient.Ping(ctx, readpref.Nearest())
 		if err != nil {
 			log.Println("MongoDB healthcheck failed:", err)
 			http.Error(w, "Internal error, could not contact MongoDB", 500)
```
rachedbenmustapha, line 89: In idiomatic go, there would be a `return` here, leaving the normal code path un-indented.

cosmos/scheduler/cmd/scheduler.go
```diff
@@ -76,65 +95,73 @@ func (s *Scheduler) healthCheckServer() {
 func (s *Scheduler) configureIngestionSecret(init bool) error {
 	accessKey, secretKey, err := s.Pensieve.GetServiceAccountCredentials("md-ingestion")
 	if err != nil {
+		log.Println("error getting service account credentials")
 		return err
 	}
 	secret := s.getIngestionSecret()
 	if secret == nil {
 		log.Println("creating ingestion secret")
 		err = s.ingestionSecret(accessKey, secretKey, false)
```
rachedbenmustapha: The name `ingestionSecret` is a bit confusing, it looks like a getter. Something with write or save or set in the name would be less surprising.

cosmos/scheduler/cmd/scheduler.go
```diff
@@ -76,65 +95,73 @@ func (s *Scheduler) healthCheckServer() {
...
 		if err != nil {
-			return err
+			log.Println("error creating ingestion secret", err)
+		} else {
+			log.Println("ingestion secret created successfully")
```
rachedbenmustapha: The non-error flow is usually un-indented

cosmos/scheduler/main.go
```diff
@@ -53,8 +53,13 @@ func init() {
 func main() {
 	mongoOptions := options.Client()
 	mongoOptions.SetAppName("cosmos-scheduler")
-	mongoOptions.SetReadPreference(readpref.Secondary())
 	mongoOptions.ApplyURI("mongodb://"+viper.GetString("mongodb_hosts"))
+	// incase of a singleton where no secondary is available
+	if len(mongoOptions.Hosts) == 1 {
+		mongoOptions.SetReadPreference(readpref.Nearest())
+	} else {
+		mongoOptions.SetReadPreference(readpref.Secondary())
```
rachedbenmustapha: Wouldn't we always want the primary?

cosmos/scheduler/cmd/scheduler.go
```diff
@@ -76,65 +95,73 @@ func (s *Scheduler) healthCheckServer() {
...
-		}}},
-	})
-	if err != nil {
-		log.Println(err)
-		return err
-	}
 	ch := make(chan BucketTransaction)
+	defer close(ch)
 	go func() {
+		collection := s.MongodbClient.Database("metadata").Collection("__metastore")
```
rachedbenmustapha: I know this is not new in this PR, but something to take care of later: the database name should be configurable (sandbox deployment uses `MONGODB_DATABASE`), for cases where zenko instances are co-hosted.

cosmos/scheduler/cmd/scheduler.go
```diff
@@ -76,65 +95,73 @@ func (s *Scheduler) healthCheckServer() {
...
+	return err
 }
 
-func (s *Scheduler) watchBucketUpdates(quit chan bool) error {
-	locationBson, err := s.getCosmosLocationBson()
+// watchBucketUpdates will take the locationType and watch for any new buckets created
+// for the locationType. A new bucket created in the locationType will trigger an applyChanges
+// for the bucket. If a true is sent on the bool channel then the function will return closing all
+// open connections.
+func (s *Scheduler) watchBucketUpdates(locationType string, quit chan bool) error {
```
rachedbenmustapha: Instead of having a special quit channel, a context can be passed as the first argument, and in the `select` loop, use `case <-ctx.Done():`.

The argument context can then be used as-is (with or without a cancellation wrapper, depending on the need) in nested context-aware function calls.

This allows a top-level `Context` cancellation to trickle down, and makes standard easy to read function signatures (once one gets used to them).

cosmos/scheduler/cmd/scheduler.go
```diff
@@ -32,33 +34,50 @@ type Scheduler struct {
...
 }
 
+// healthCheckServer will ping the MongoDB connection on every request recieved
+// at :8080/healthcheck
 func (s *Scheduler) healthCheckServer() {
 	health := func(w http.ResponseWriter, req *http.Request){
 		ctx, cancel := context.WithTimeout(req.Context(), 5*time.Second)
 		defer cancel()
-		err := s.MongodbClient.Ping(ctx, readpref.Secondary())
+		err := s.MongodbClient.Ping(ctx, readpref.Nearest())
```
rachedbenmustapha: The health check should probably use the same read preference as the standard code path, to catch the same kind of problems. For example the health check can be happy reading from a secondary, while the application itself is failing reads on the primary.

cosmos/scheduler/cmd/scheduler.go
```diff
@@ -32,33 +34,50 @@ type Scheduler struct {
...
+
+	quit := make(chan bool)
 	for {
-		go s.watchBucketUpdates(quit)
-		<-overlayUpdates
+		go s.watchBucketUpdates("nfs", quit)
+		// blocks until an overlay update is recieved on the overlayUpdates channel
+		<- overlayUpdates
 		log.Println("received an overlay update")
 		quit <- true
```
rachedbenmustapha: Why quit here? It seems that this is doing "watch, receive, quit" in a loop, which has the potential to miss events because of the asynchronicity involved.

rachedbenmustapha (commented):
first pass

cosmos/scheduler/main.go
ssalaues: My rationale behind this was because this program will never need to write to MongoDB hence the preference on the secondary (although secondaries will be unavailable singleton deployments). This way there is more available connections to the primary for services that do require it.

cosmos/scheduler/cmd/scheduler.go
ssalaues: Yeah that's a good point. I'll add this in.

cosmos/scheduler/cmd/scheduler.go
ssalaues: Yeah this is one I kept changing to try to get rid of the quit channel but couldn't come up with a good solution. I think this is great suggestion and will make these changes.

cosmos/scheduler/cmd/scheduler.go
ssalaues: Agreed, I think I can more easily rework this now with your previous comments.

cosmos/scheduler/cmd/scheduler.go
```diff
@@ -32,39 +35,57 @@ type Scheduler struct {
...
 	for {
-		go s.watchBucketUpdates(quit)
-		<-overlayUpdates
+		ctxNfs, cancelNfs := context.WithCancel(ctx)
+		go s.watchBucketUpdates(ctxNfs, "location-nfs-mount-v1")
+		// blocks until an overlay update is recieved on the overlayUpdates channel
+		<- overlayUpdates
+		cancelNfs()
 		log.Println("received an overlay update")
-		quit <- true
```
ssalaues, line 53: TODO: the loop is still essentially watch, update, quit. Will still be reworking this.

cosmos/scheduler/main.go
ssalaues: Switched this to the Primary.

cosmos/scheduler/cmd/scheduler.go
```diff
@@ -32,39 +39,58 @@ type Scheduler struct {
...
 	}
+
+	overlayUpdates := s.watchOverlayUpdates(context.Background())
+	bucketCh := s.watchBucketUpdates(context.Background())
 	for {
-		go s.watchBucketUpdates(quit)
-		<-overlayUpdates
-		log.Println("received an overlay update")
-		quit <- true
+		select {
```
ssalaues, line 69: @RachedBenMustapha I replaced the "watch, receive, quit" loop here with a select loop where the goroutines are no longer interdependent on each other or kill/recreated on every update. 

It's added as a separate commit to make reviewing a bit easier:
https://github.com/scality/Zenko/pull/792/commits/554002e92739824489f8725fee78a4e0df8fd926

cosmos/scheduler/cmd/scheduler.go
```diff
@@ -17,12 +19,17 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 )
+const (
+	nfsLocation = "location-nfs-mount-v1"
+	scalityRing = "location-ring-s3-v1"
```
ssalaues: Was temporarily using `scalityRing` as an explicit check to return early but no longer needed. This can be deleted.

alexanderchan-scality (approved):

rachedbenmustapha (approved):
Looks more solid than it was, good improvement.

Suggestions for next steps:
- integrate `go vet` and `go fmt` checks in the CI (or even better, something like https://golangci.com/), there are anomalies that I didn't mention, that these tools would bring up automatically
- Use a single top-level context that can be passed down to all goroutines, and that can be used as global cancelation for `SIGTERM` for example
- Unless mongodb's watches always return a first result when set up, this only seems to be catching mongodb changes happening while the process is running, not while it's being restarted, upgraded, crashed, etc.

ssalaues:
@RachedBenMustapha thanks for the feedback and great suggestions for steps moving forward. The last one especially has been annoying not only for me but for the user experience as well.

ssalaues:
Rebased and added the version bump to Zenko.

ssalaues:
@bert-e approve

PR #810, opened 2019-08-14
[ZENKOIO-72] Cosmos Chart and Scheduler Improvements 

ssalaues:
- Makes the scheduler create function more easily extensible with a format for future locations 
- Allows the cosmos chart to template both source and destination credentials into Kubernetes Secrets. 

fixes ZENKOIO-72


kubernetes/cosmos/templates/rclone-cronjob.yaml
```diff
@@ -71,7 +83,7 @@ spec:
             emptyDir: {}
 {{- end }}
 {{- end -}}
-{{- if .Values.rclone.initialIngestion }}
+{{- if and .Values.rclone.initialIngestion .Release.IsInstall }}
```
ssalaues, line 86: `IsInstall` doesn't seem to work E2E with the operator. Will have to revert this part back.

bennettbuchanan (approved):

JianqinWang (approved):

tmacro (approved):

kubernetes/cosmos/templates/rclone-cronjob.yaml
ssalaues: Added the `IsInstall` back but turned the job into a post-install hook. This way the job isn't immutable. Also will only be ran at install time and not at every upgrade.

ssalaues:
@bert-e approve

PR #827, opened 2019-08-23
improvement: increase kafka queue max heap size

ssalaues:
Increases the Max Java heap size to 6Gb which is commonly used as a default optimal value. The default is max/min 1Gb. These changes set max of 6Gb and min 1Gb.

https://docs.confluent.io/current/kafka/deployment.html
https://www.infoq.com/articles/apache-kafka-best-practices-to-optimize-your-deployment/

### DIFF
--- a/kubernetes/zenko/values.yaml
+++ b/kubernetes/zenko/values.yaml
@@ -197,6 +197,7 @@ zenko-queue:
 ## Extensive list of configurables can be found here:
 ## https://github.com/kubernetes/charts/blob/master/incubator/kafka/values.yaml
   replicas: *nodeCount
+  kafkaHeapOptions: "-Xmx6G -Xms1G"
   rbac:
     enabled: true
   configurationOverrides:


ssalaues:
@bert-e wait

tmacro (approved):

tcarmet:
I don't know how I got myself into this place, but I'm curious.

I've tweaked the same settings one day but with Elasticsearch (as it's java behind I believe it works the same, or maybe not at all) and I had the following issue, it's seems like either ES or the JVM does some pre memory allocation or something and if I didn't had enough memory, in this case it would be a 6Gb of memory requested on the pod resources, well the pod was getting killed all the time by the kubernetes master  (even if the program wasn't doing anything) due to OOM errors.

Can you confirm that you're not hitting this issue if, per example you don't set anything into the `resources` of the pod and kafka can schedule himself just fine? Otherwise you will have to be careful to always match the `resources` pod spec and this value.

hope you don't have to deal with that.

ssalaues:
> I don't know how I got myself into this place, but I'm curious.
> 
> I've tweaked the same settings one day but with Elasticsearch (as it's java behind I believe it works the same, or maybe not at all) and I had the following issue, it's seems like either ES or the JVM does some pre memory allocation or something and if I didn't had enough memory, in this case it would be a 6Gb of memory requested on the pod resources, well the pod was getting killed all the time by the kubernetes master (even if the program wasn't doing anything) due to OOM errors.
> 
> Can you confirm that you're not hitting this issue if, per example you don't set anything into the `resources` of the pod and kafka can schedule himself just fine? Otherwise you will have to be careful to always match the `resources` pod spec and this value.
> 
> hope you don't have to deal with that.

We don't specify resource limits/requests on Zenko so in our case they shouldn't be. But I put the hold on it to test it more thoroughly regardless. However the behavior would likely be essentially the same. The reason for that OOM behavior is because there is a lot of page caching (this can grow to 2-3x the heap size) which in Kubernetes counts towards the total memory allocation of a Pod even though it's not technically heap in use. So if you set the max heap to 6Gb then you will want to set the pod memory limit to a much higher value to prevent the OOM kills by K8s.

https://stackoverflow.com/questions/48609979/how-to-manage-page-cache-resources-when-running-kafka-in-kubernetes

rahulreddy (approved):

tcarmet (approved):
Alright! thank you for your answer @ssalaues I'm learning stuff :)

ssalaues:
@bert-e approve

PR #830, opened 2019-08-30
Feature: zenko 1456 mongodb sys backup

anurag4DSB:
Add backup and restore feature for zenko

fixes #[Zenko-1456](https://scality.atlassian.net/browse/ZENKO-1456)

Current features:
- schedule backups
- local backups retention
- upload to S3 Object Storage (Minio, AWS, Google Cloud, Azure)
- upload to gcloud storage
- upload to SFTP (under testing)
- notifications (Email, Slack)

### DIFF
--- a/kubernetes/zenko/charts/mgob/Chart.yaml
+++ b/kubernetes/zenko/charts/mgob/Chart.yaml
@@ -1,0 +1,7 @@
+apiVersion: v1
+appVersion: "1.0"
+description: MongoDB dockerized backup agent. Runs schedule backups with retention, S3 & SFTP upload, notifications, instrumentation with Prometheus and more.
+name: mgob
+version: 0.1.0
+sources:
+- https://github.com/stefanprodan/mgob
--- a/kubernetes/zenko/charts/mgob/README.md
+++ b/kubernetes/zenko/charts/mgob/README.md
@@ -1,0 +1,243 @@
+# Usage
+
+# mgob
+
+MGOB is a MongoDB backup automation tool built with golang.
+
+#### Features
+
+- schedule backups
+- local backups retention
+- upload to S3 Object Storage (Minio, AWS, Google Cloud, Azure)
+- upload to gcloud storage
+- upload to SFTP (under testing)
+- notifications (Email, Slack)
+
+#### Install
+
+MGOB is available on Docker Hub at [stefanprodan/mgob](https://hub.docker.com/r/stefanprodan/mgob/).
+
+#### Configure
+
+Define a backup plan (yaml format) in the values file under configMap.
+This can be done either in the values.yaml file for mgob or vaules.yaml
+file of Zenko.
+
+_Backup plan_
+
+```yaml
+scheduler:
+  # run every day at 6:00 and 18:00 UTC
+  cron: "0 6,18 */1 * *"
+  # number of backups to keep locally
+  retention: 14
+  # backup operation timeout in minutes
+  timeout: 60
+target:
+  # mongod IP or host name
+  host: "<mongoDB replica set>"
+  # mongodb port
+  port: 27017
+  # mongodb database name, leave blank to backup all databases
+  database: "test"
+  # leave blank or comment out if auth is not enabled
+  username: "admin"
+  password: "secret"
+  # add custom params to mongodump (eg. Auth or SSL support), leave blank if not needed
+  params: "--ssl --authenticationDatabase admin"
+# S3 upload (optional)
+s3:
+  url: "https://s3.amazonaws.com"
+  bucket: "backup"
+  accessKey: "Q3AM3UQ867SPQQA43P2F"
+  secretKey: "zuf+tfteSlswRu7BJ86wekitnifILbZam1KYY3TG"
+  # For Minio and AWS use S3v4 for GCP use S3v2
+  api: "S3v4"
+# GCloud upload (optional)
+gcloud:
+  bucket: "backup"
+  keyFilePath: /path/to/service-account.json
+# Azure blob storage upload (optional)
+azure:
+  containerName: "backup"
+  connectionString: "DefaultEndpointsProtocol=https;AccountName=...;AccountKey=...;EndpointSuffix=core.windows.net"
+# SFTP upload (optional - under testing)
+sftp:
+  host: sftp.company.com
+  port: 2022
+  username: user
+  password: secret
+  # dir must exist on the SFTP server
+  dir: backup
+# Email notifications (optional - under testing)
+smtp:
+  server: smtp.company.com
+  port: 465
+  username: user
+  password: secret
+  from: mgob@company.com
+  to:
+    - devops@company.com
+    - alerts@company.com
+# Slack notifications (optional  - under testing)
+slack:
+  url: https://hooks.slack.com/services/xxxx/xxx/xx
+  channel: devops-alerts
+  username: mgob
+  # 'true' to notify only on failures
+  warnOnly: false
+```
+
+ReplicaSet example:
+
+```yaml
+target:
+  host: "zenko-mongodb-replicaset-0.zenko-mongodb-replicaset,zenko-mongodb-replicaset-1.zenko-mongodb-replicaset,zenko-mongodb-replicaset-2.zenko-mongodb-replicaset"
+  port: 27017
+  database: "test"
+```
+
+Sharded cluster with authentication and SSL example:
+
+```yaml
+target:
+  host: "zenko-mongodb-replicaset-0.zenko-mongodb-replicaset,zenko-mongodb-replicaset-1.zenko-mongodb-replicaset,zenko-mongodb-replicaset-2.zenko-mongodb-replicaset"
+  port: 27017
+  database: "test"
+  username: "admin"
+  password: "secret"
+  params: "--ssl --authenticationDatabase admin"
+```
+
+#### Web API
+
+NOTE: Considering `zenko` is the release name, the service name will be `zenko-mgob`.
+
+```sh
+kubectl get svc zenko-mgob
+NAME         TYPE        CLUSTER-IP   EXTERNAL-IP   PORT(S)    AGE
+zenko-mgob   ClusterIP   None         <none>        8090/TCP   3d
+```
+
+- `zenko-mgob:8090/storage` file server
+- `zenko-mgob:8090/status` backup jobs status
+- `zenko-mgob:8090/metrics` Prometheus endpoint
+- `zenko-mgob:8090/version` mgob version and runtime info
+- `zenko-mgob:8090/debug` pprof endpoint
+
+On demand backup:
+
+- HTTP POST `zenko-mgob:8090/backup/:planID`
+
+```bash
+curl -X POST http://zenko-mgob:8090/backup/metadata
+```
+
+```json
+{
+  "plan": "metadata",
+  "file": "metadata-1567721706.gz",
+  "duration": "1.264375987s",
+  "size": "4.9 kB",
+  "timestamp": "2019-09-05T22:15:06.529181117Z"
+}
+```
+
+Scheduler status:
+
+- HTTP GET `zenko-mgob:8090/status`
+- HTTP GET `zenko-mgob:8090/status/:planID`
+
+```bash
+curl -X GET http://zenko-mgob:8090/status/metadata
+```
+
+```json
+{
+  "plan": "metadata",
+  "next_run": "2019-09-06T05:00:00Z"
+  "last_run": "2017-09-05T21:49:00.000622589Z",
+  "last_run_status": "200",
+  "last_run_log": "Backup finished in 2.339055539s archive metadata-1567720190.gz size 4.7 kB"
+}
+```
+
+#### Logs
+
+View scheduler logs with `kubectl logs {{ release-name }}-mgob-0`:
+
+```bash
+time="2019-09-05T21:49:38Z" level=info msg="Next tmp cleanup run at 2019-09-05 22:00:00 +0000 UTC"
+time="2019-09-05T21:49:38Z" level=info msg="Next run at 2019-09-06 05:00:00 +0000 UTC" plan=metadata
+time="2019-09-05T21:49:38Z" level=info msg="starting http server on port 8090"
+time="2019-09-05T21:49:50Z" level=info msg="On demand backup started" plan=metadata
+time="2019-09-05T21:49:50Z" level=info msg="new dump" archive="/tmp/metadata-1567720190.gz" err=<nil> mlog="/tmp/metadata-1567720190.log" planDir="/storage/metadata"
+time="2019-09-05T21:49:51Z" level=info msg="S3 upload finished `/storage/metadata/metadata-1567720190.gz` -> `metadata/anurag-west-1/metadata-1567720190.gz` Total: 4.74 KiB, Transferred: 4.74 KiB, Speed: 30.07 KiB/s " plan=metadata
+time="2019-09-05T21:49:51Z" level=info msg="On demand backup finished in 1.222549426s archive metadata-1567720190.gz size 4.9 kB" plan=metadata
+```
+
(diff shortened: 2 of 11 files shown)


kubernetes/zenko/charts/mgob/Chart.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
+appVersion: "1.0"
+description: MongoDB dockerized backup agent. Runs schedule backups with retention, S3 & SFTP upload, notifications, instrumentation with Prometheus and more.
+name: mgob
+version: 0.1.0
```
ssalaues, line 5: Should have the sources listed here to point to https://github.com/stefanprodan/mgob. Similar to:
https://github.com/helm/charts/blob/master/stable/redis-ha/Chart.yaml#L21

kubernetes/zenko/charts/mgob/notes.md
```diff
@@ -0,0 +1,235 @@
+# Usage
```
ssalaues: This file should be a `README.md` so that it loads automatically when we browse the chart.

kubernetes/zenko/charts/mgob/notes.md
```diff
@@ -0,0 +1,235 @@
...
+  port: 27017
+  database: "test"
+  username: "admin"
+  password: "secret"
+  params: "--ssl --authenticationDatabase admin"
+```
+
+#### Web API
+
+- `zenko-mgob-0:8090/storage` file server
```
ssalaues: Is there a reason why it should be pointing to the specific pod and not the service name? Seems like most other references point to the service name.

kubernetes/zenko/charts/mgob/templates/_helpers.tpl
```diff
@@ -0,0 +1,46 @@
...
+{{- $count := (atoi (printf "%d" (int64 .Values.mongodb.replicas))) -}}
+{{- $release := .Release.Name -}}
+{{- range $v := until $count }}{{ $release }}-mongodb-replicaset-{{ $v }}.{{ $release }}-mongodb-replicaset{{ if ne $v (sub $count 1) }},{{- end -}}{{- end -}}
+{{- end -}}
+
+
+{{- define "mgob.configMap-rendered" -}}
+{{- $defaults := .Values.mgob.configMap.defaults -}}
+{{- if not .Values.mgob.configMap.defaults.target.host -}}
+{{ $_ := set $defaults.target "host" (include "mgob.mongodb-hosts" .) }}
```
ssalaues, line 41: Just to make sure I understand, if no hosts are defined then it defaults to the "mgob.mongodb-hosts" template?

kubernetes/zenko/charts/mgob/values.yaml
```diff
@@ -0,0 +1,83 @@
+# Default values for zenko.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+mgob:
```
ssalaues: Since this chart **IS** `mgob` you don't need to specify a `mgob` section in the values.yaml. This will simplify many of the references to the values as well.

kubernetes/zenko/charts/mgob/templates/statefulset.yaml
```diff
@@ -0,0 +1,57 @@
+apiVersion: apps/v1 # Change to "apps/v1beta1" for Kubernetes < 1.10.
```
ssalaues: Instead of having a comment this should be a templating condition like `if .Capabilities.APIVersions` or `KubeVersion`
https://github.com/helm/helm/blob/master/docs/chart_template_guide/builtin_objects.md

kubernetes/zenko/charts/mgob/values.yaml
```diff
@@ -0,0 +1,83 @@
...
+  enabled: true
```
ssalaues, line 12: There should be no need to disable it. Also it seems that this only disables the PVC?

kubernetes/zenko/values.yaml
```diff
@@ -283,3 +283,39 @@ zenko-queue-manager:
   clusters:
     - name: "zenko-queue-cluster"
       tuning: {}
+
+mgob:
+  mgob:
+   # Enables the backup feature for Zenko. Metadata is backed up for MongoDB
+   # Per default schedule metadata backup takes place as full
+    enabled: false
```
ssalaues: Instead of having the full chart disabled (which is what it seems like is the intent here). You could simply have the cron disabled.

kubernetes/zenko/charts/mgob/templates/secret.yaml
```diff
@@ -0,0 +1,17 @@
+{{- range $secret := .Values.secret }}
+{{- if $secret.data -}}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ $secret.name }}
+  labels:
+    release: "{{ $.Release.Name }}"
```
ssalaues: Labels should really be a template function. I know many of the original Zenko charts do it this way but it's a bad practice that we should move away from.

kubernetes/zenko/charts/mgob/values.yaml
```diff
@@ -0,0 +1,83 @@
...
+        #   username: mgob
+        #   # 'true' to notify only on failures
+        #   warnOnly: false
+
+service:
+  name: mgob
+  externalPort: 8090
+  internalPort: 8090
+selector:
+  role: mongo-backup
```
ssalaues: Having the selector as configurable is overkill and unnecessary. Also since it's not based off the release name it is no longer unique so you will have multiple installs cross referencing each other.

kubernetes/zenko/charts/mgob/values.yaml
```diff
@@ -0,0 +1,83 @@
...
+          bucket: "my-db-backups"
+          accessKey: "ABCDEFGHIJKLMNOPQRST"
+          secretKey: "be0da49b599b24115c7f53c92c729cbb2c8a17e5"
+        # slack:
+        #   username: mgob
+        #   # 'true' to notify only on failures
+        #   warnOnly: false
+
+service:
+  name: mgob
```
ssalaues: Having a static service name different from the release is confusing and could lead to conflicts especially since this is the default. Typically if the names are to be overridden then `.Values.nameOverride` is to be used which is already integrated into your `_helpers.tpl` file. This will (should) properly template the override to all resources.

kubernetes/zenko/charts/mgob/values.yaml
```diff
@@ -0,0 +1,83 @@
...
+  externalPort: 8090
+  internalPort: 8090
```
ssalaues: You really only need to specify one port listed as just `port` which will be how anything external communicates. The `internalPort` can be static since it's only Service -> Pod communication which never really needs to change.

kubernetes/zenko/charts/mgob/values.yaml
```diff
@@ -0,0 +1,83 @@
...
+selector:
+  role: mongo-backup
+resources:
+  limits:
```
ssalaues: As long as this performs adequately within these resource restrains then this should be fine. Otherwise typically these are left blank.

kubernetes/zenko/charts/mgob/values.yaml
```diff
@@ -0,0 +1,83 @@
...
+  persistentVolume:
+    enabled: true
+    accessModes:
+      - ReadWriteOnce
+    annotations: {}
+    existingClaim: ""
+    size: 10Gi
+
+  configMap:
+    name: "mgob-config"
```
ssalaues: The ConfigMap name should also be using the templated name. It's okay to have an optional value to point to a custom ConfigMap but this shouldn't be the default. (See other comments about static names)

kubernetes/zenko/charts/mgob/templates/statefulset.yaml
```diff
@@ -0,0 +1,57 @@
...
+kind: StatefulSet
+metadata:
+  name: {{ template "mgob.fullname" . }}
+spec:
+  serviceName: {{ template "mgob.fullname" . }}
+  replicas: 1
+  selector:
+    matchLabels:
+      role: {{ .Values.selector.role }}
+  template:
```
ssalaues, line 11: Should have an `annotations` section with the checksum of the ConfigMap. This will force an update to the pods if the ConfigMap ever changes.
https://github.com/helm/charts/blob/master/stable/redis-ha/templates/redis-ha-statefulset.yaml#L20

ssalaues (changes requested):
Lots of good work in here but still needs refinement.

kubernetes/zenko/charts/mgob/Chart.yaml
anurag4DSB: great catch, done

kubernetes/zenko/charts/mgob/templates/statefulset.yaml
anurag4DSB: I was really missing this. Thank you, done  :)

kubernetes/zenko/charts/mgob/values.yaml
anurag4DSB: got it

kubernetes/zenko/charts/mgob/notes.md
anurag4DSB: Changed

kubernetes/zenko/charts/mgob/notes.md
anurag4DSB: done

kubernetes/zenko/charts/mgob/values.yaml
anurag4DSB: done

kubernetes/zenko/charts/mgob/templates/statefulset.yaml
anurag4DSB: makes sense, removed the comment

kubernetes/zenko/charts/mgob/templates/secret.yaml
anurag4DSB: done

kubernetes/zenko/charts/mgob/values.yaml
anurag4DSB: done

kubernetes/zenko/charts/mgob/values.yaml
anurag4DSB: done

kubernetes/zenko/charts/mgob/values.yaml
anurag4DSB: done

kubernetes/zenko/charts/mgob/values.yaml
anurag4DSB: changed.

kubernetes/zenko/charts/mgob/templates/statefulset.yaml
anurag4DSB: Great catch, I was missing this.

kubernetes/zenko/charts/mgob/templates/_helpers.tpl
```diff
@@ -36,11 +36,20 @@ Create the default mongodb replicaset hosts string
...
+{{- range $name, $value := .Values.configMap.plans -}}
 {{ $name }}: {{ merge $value $defaults | toYaml | quote }}
 {{- end -}}
 {{- end -}}
+
+{{- define "mgob.role" -}}
+role: {{ template "mgob.fullname" . }}-backup
+{{- end -}}
+
+{{- define "mgob.secret.lables" -}}
```
ssalaues: `lables` -> `labels`

kubernetes/zenko/charts/mgob/templates/statefulset.yaml
```diff
@@ -0,0 +1,59 @@
+apiVersion: apps/v1
+kind: StatefulSet
```
ssalaues, line 2: Why is this a StatefulSet but with a defined PVC? A StatefulSet generates its own PVC and I'm not entirely sure why there is a custom one.
https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#stable-storage

anurag4DSB: done in the latest commit.

tmacro (approved):

kubernetes/zenko/charts/mgob/templates/statefulset.yaml
```diff
@@ -0,0 +1,61 @@
...
+{{- range $name, $value := .Values.configMap.plans }}
+            - key: {{ $name }}
+              path: {{ $name }}.yml
+{{ end }}
+{{- range $secret := .Values.secret }}
+        - name: {{ $secret.name }}
+          secret:
+            secretName: {{ $secret.name }}
+{{ end }}
+        {{ if .Values.existingClaim }}
```
ssalaues: You should be using `volumeClaimTemplates` for the StatefulSet. This go along with what we discussed previously about the claims on Deployments vs StatefulSet.

kubernetes/zenko/values.yaml
```diff
@@ -294,3 +294,37 @@ zenko-queue-manager:
   clusters:
     - name: "zenko-queue-cluster"
       tuning: {}
+
+mgob:
+ # Enables the backup feature for Zenko. Metadata is backed up for MongoDB
+ # Per default schedule metadata backup takes place as full
+  enabled: false
```
ssalaues: Is this supposed to do something?

kubernetes/zenko/values.yaml
```diff
@@ -294,3 +294,37 @@ zenko-queue-manager:
...
+  persistentVolume:
```
ssalaues: These should just be the defaults in the mgob chart values.

kubernetes/zenko/values.yaml
```diff
@@ -294,3 +294,37 @@ zenko-queue-manager:
...
+    enabled: true
+    accessModes:
+      - ReadWriteOnce
+    annotations: {}
+    existingClaim: ""
+    size: 10Gi
+  configMap:
+    defaults:
+      target:
+        database: "metadata"
```
ssalaues: Looks like a lot of these are already defined in the mgob chart values. Why are they being redefined here?

ssalaues (changes requested):

kubernetes/zenko/charts/mgob/templates/statefulset.yaml
anurag4DSB: Made the change, thanks for the explanation :)

kubernetes/zenko/values.yaml
anurag4DSB: Understood how I was using this wrong.

kubernetes/zenko/values.yaml
anurag4DSB: done

kubernetes/zenko/values.yaml
anurag4DSB: makes sense, removed. Will work with the documentation team for details.

kubernetes/zenko/charts/mgob/README.md
```diff
@@ -0,0 +1,243 @@
...
+
+- schedule backups
+- local backups retention
+- upload to S3 Object Storage (Minio, AWS, Google Cloud, Azure)
+- upload to gcloud storage
+- upload to SFTP (under testing)
+- notifications (Email, Slack)
+
+#### Install
+
```
ssalaues, line 17: Minor and not blocking for this but on your next revision, you should add an example of how to install the chart.

ssalaues (approved):
Looks solid. Good job!

kubernetes/zenko/charts/mgob/README.md
anurag4DSB: cool, sounds good :)

anurag4DSB:
@bert-e approve

PR #839, opened 2019-09-06
ft: add backbeat api hpa

ssalaues:
Adds a horizontal pod autoscaler to the Backbeat API deployment. This will allow the API pods to scale up to `nodeCount` as necessary in situations when a single pod starts to use more than 70% of CPU usage. This serves not only to prevent a bottleneck in the API service but as well as an indicator that the service is under heavy load.

### DIFF
--- a/kubernetes/zenko/values.yaml
+++ b/kubernetes/zenko/values.yaml
@@ -86,6 +86,16 @@ cloudserver:
       # secret:
 
 backbeat:
+  api:
+    autoscaling:
+      enabled: true
+      config:
+        maxReplicas: *nodeCount
+    resources:
+      requests:
+        cpu: 250m
+      limits:
+        cpu: 1
   replication:
     dataProcessor:
       replicaCount: *nodeCount
--- a/kubernetes/single-node-values.yaml
+++ b/kubernetes/single-node-values.yaml
@@ -16,6 +16,10 @@ s3-data:
     size: 10Gi
 
 backbeat:
+  api:
+    autoscaler:
+      enabled: false
+    resources: {}
   replication:
     dataProcessor:
       replicaCount: *nodeCount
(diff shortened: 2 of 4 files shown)


bennettbuchanan (approved):

JianqinWang (approved):

kubernetes/zenko/values.yaml
```diff
@@ -86,6 +86,16 @@ cloudserver:
       # secret:
 
 backbeat:
+  api:
+    autoscaling:
+      enabled: true
+      config:
+        maxReplicas: *nodeCount
```
NicolasT, line 93: I don't really get why this should be `nodeCount`. We use single-process NodeJS containers, right? So the number of processes you could in theory run is the number of CPUs in the cluster, which is likely (much) more than the number of nodes in the cluster.

I'm not sure what the  'correct' number here should be, or how to calculate it. However, `nodeCount` seems at least as arbitrary as 3, 5, 10 or 21...

ssalaues: Since we are basing all of our default replicas in Zenko off `nodeCount` it isn't arbitrary at all to continue that nomenclature here. It's the safest default we can assume without bottlenecking at a single pod/process while at the same time allowing anyone to change the value based off their core count if they so desire.

Additionally, this is an internal-only facing service at the moment so all the load that it receives is from our other processes. This is how we know that typically 1 pod/process is enough, however under heavy CRR operations, for example, we also know that 1 pod/process can become exhausted therefore tripling the capacity (for a 3 node) can be a safe assumption.

ssalaues: This is also why I think the hpa is a better solution than having a static replica count here. Because if the cluster sees that a single pod is near the designated capacity, then it can add pods as necessary to fill the demand.

ssalaues:
@bert-e approve

PR #846, opened 2019-09-12
bugfix: ZENKO-2104 Disable flaky test

bennettbuchanan:
Disable a flaky CRR object monitor test while a cause/fix is being investigated.

### DIFF
--- a/tests/zenko_tests/node_tests/backbeat/tests/api/objectMonitor.js
+++ b/tests/zenko_tests/node_tests/backbeat/tests/api/objectMonitor.js
@@ -91,7 +91,8 @@ describe('Backbeat object monitor CRR metrics', function() {
         },
     ], done));
 
-    it('should monitor part uploads of an MPU object', done => waterfall([
+    // Disabled because the test is flaky. See ZENKO-2104.
+    it.skip('should monitor part uploads of an MPU object', done => waterfall([
         next => scalityUtils.completeMPUAWS(srcBucket, key, 50, next),
         (data, next) => {
             const path = `/_/backbeat/api/metrics/crr/${destLocation}` +


bennettbuchanan:
@bert-e reset

tests/zenko_tests/node_tests/backbeat/tests/api/objectMonitor.js
```diff
@@ -91,7 +91,8 @@ describe('Backbeat object monitor CRR metrics', function() {
         },
     ], done));
 
-    it('should monitor part uploads of an MPU object', done => waterfall([
+    // Disabled because the test is flaky. See ZENKO-2104.
+    it.skip('should monitor part uploads of an MPU object', done => waterfall([
```
ssalaues, line 95: While I'm all for having less flaky tests, do you think this is the best course of action? Just mainly for regression testing. 

However with that being said, considering how flaky the test is, it's likely not doing much to help us at this point either.

tmacro: Echoing salim, but don't want to block the build

tmacro (approved):

tests/zenko_tests/node_tests/backbeat/tests/api/objectMonitor.js
bennettbuchanan: Indeed, my thought was to disable the test while it is being investigated. That way it won't send false negatives during regression testing and slow down development cycles that aren't related to testing. Once things are stabilized, we could reenable it. The strategy is definitely debatable and I could be convinced otherwise. 😄

bennettbuchanan: (Strategy is based on reading from Google [here](https://testing.googleblog.com/2016/05/flaky-tests-at-google-and-how-we.html) re: "quarantining" tests.)

ssalaues: Seems sound 👍

ssalaues (approved):

bennettbuchanan:
@bert-e approve

PR #852, opened 2019-09-17
documentation: ZENKOIO-89_import_template_improvements

wabernat:
This PR updates the documentation settings to bring Zenko up to the same look as the 8.0 S3C and RING builds. These settings build minimum required outputs; however, html/pdf trickery is not activated, so there's not much difference between html and pdf (i.e., no glossary or special pdf-only content).

This fixes #ZENKOIO-89


wabernat:
@bert-e approve

wabernat:
@bert-e reset

lucieleonard:
- HTML: it is looking as expected. The template works, we should allow more configuration but that would be a PR in the template itself. TBD.
- PDF: I don't know if I do it right, but I can run the three pubs into 1 PDF, or run 3 separate commands to have 3 PDFs. I think we prefer three separated PDFs, am I right?
When built, the PDFs have no cover. You can fix that in /scaldoc/preamble.tex by changing \maketitle into \sphinxmaketitle (comes from the sphinx upgrade). This applies the right cover for each PDF.
If you'd like to have 1 command = 3 PDFs, we should do that in another PR.
But if it's ok like this, just change the command I mentioned and your PDFs and HTML look fine and you can merge.
If you need something else, please ask

wabernat:
It needs to be three separate PDFs. I don't care how many commands it takes. 

I tried the \sphinxmaketitle trick. It worked great on my local build, but Eve would have none of it, so I had to undo it. 

I will try deleting .tox again and we'll see if that fixes it.

wabernat:
Re-reading, I see we are talking past each other. The problem with PDFs is that there is not a way to use any index_pdf.rst files or "only" tags. When I try to add such, the settings do not transfer cleanly from the RING/S3C source, and I do not get a clean build. When I specify an html build, Sphinx does not detect that I'm asking for  html and uses index_pdf instead of index.rst. In other words, the "conditional text" feature does not work. The PDF builds do not pick up pdf-only output. 

This build is fine as far as it goes, but does not go far enough. I agree that we should resolve these matters in one or more subsequent tickets (assuming I can get /sphinxmaketitle to work!)

smaffulli:
Is the new skin affecting building the PDFs or is the issue with the PDF a separate one?

wabernat:
On reflection, I would score the PDF issue as a separate one. I don't believe I integrated the PDF text condition successfully into Zenko prior to this upgrade. It's just that importing the RING/S3C conf.py required that I stare straight at the PDF conditional logic, and it's not working for me. 

There is also a trick involving the covers of the books (\sphinxmaketitle) which works great for me, but breaks the build when I push the source to Eve. I have been dealing with other issues today, so I've made no progress on this.

wabernat:
@bert-e reset

tmacro (approved):

ssalaues (approved):

wabernat:
@bert-e reset

wabernat:
Batter my heart, O three-person'd God!

anurag4DSB:
@bert-e ping

PR #874, opened 2019-10-07
Bugfixes: MD5 and VersionID for AWS OOB

ssalaues:
- fixes MD only of MPU objects
This fixes an bug that causes MD only MPUs transfers to hang due to failure. This was a result of using the etag as an MD5 (which in the case of MPUs are not necessarily MD5 compatible). It resulted in not sending the expected headers to Cloudserver which would in turn reply with an error.

- Fix Meta map and add s3 source version ID
This commit fixes some issues in the RClone patches that produce the map of the metadata headers of an S3 request. This map of the metadata should have been in a map[string]*string data structure (array of string pointers). In fixing this, I was able to add the versionID in the readMetadata method to the *Object.meta parameter and pass it along with the request to Cloudserver. This also correctly carries over any additional metadata from the original object.

### DIFF
--- a/images/rclone/patches/001-md-only.diff
+++ b/images/rclone/patches/001-md-only.diff
@@ -1,5 +1,5 @@
 diff --git a/backend/s3/s3.go b/backend/s3/s3.go
-index 0da73a15..e7e8df8a 100644
+index fdd5ba6fe..246db518f 100644
 --- a/backend/s3/s3.go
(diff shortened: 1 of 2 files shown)


dora-korpar (approved):

tmacro (approved):

ssalaues:
@bert-e approve

PR #875, opened 2019-10-09
ZENKO-2167: Add regression tests and fix lingering bugs

ssalaues:
Regression tests to validate that version IDs from amazon get ingested into Zenko properly. Additionally, bumping the Rclone images revealed more bugs related to NFS ingestion and the way the filesystem data was translated into the object metadata which was fixed in a separate PR.


JianqinWang (approved):

alexanderchan-scality (approved):

ssalaues:
@bert-e approve

PR #877, opened 2019-10-10
ft: Watch for location credential updates from Orbit

ssalaues:
What does this PR do, and why do we need it?
Refactors some of the credential management into a reconciler to reduce duplicated code. In doing this I was able to more easily add a new function that will validate that all the location credentials are up-to-date or update them if they are not on every new overlay pushed from Orbit (e.g. any location changes). This function can be later extended to validate the location state against CRD state.


tmacro (approved):

alexanderchan-scality (approved):

ssalaues:
@bert-e approve

PR #881, opened 2019-10-16
ft: chart release script

ssalaues:
- The first commit adds a script that takes the intended release version as an argument and optional path to the zenko folder (otherwise assumes it is being ran relative to `eve/scripts`). It will then look for the latest CloudServer and Backbeat images based off the corresponding minor version (1.2.x release will look for latest CloudServer/Backbeat 8.2.x).
Usage example: `./release 1.2.0-rc.1 [path to zenko]`

- The second commit makes the `values.yaml` and `Chart.yaml` follow a uniform spec. While there is a visual diff, functionally it is the same.

### DIFF
--- a/kubernetes/zenko/charts/backbeat/values.yaml
+++ b/kubernetes/zenko/charts/backbeat/values.yaml
@@ -15,7 +15,7 @@ image:
 
 logging:
    # Options: info, debug, trace
-   level: info
+  level: info
 
 mongodb:
   replicaSet: rs0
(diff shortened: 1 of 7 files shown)


tmacro (approved):

kubernetes/zenko/charts/backbeat/values.yaml
```diff
@@ -96,7 +96,7 @@ ingestion:
 
 lifecycle:
   conductor:
-    cronRule: "0 */6 * * *"
+    cronRule: 0 */6 * * *
```
JianqinWang, line 99: yaml can correctly read this? damn (probably a poorly phrased question, but this format is also acceptable for yaml?)

JianqinWang (approved):

anurag4DSB (approved):

kubernetes/zenko/charts/backbeat/values.yaml
ssalaues: Typically yaml parsers read almost everything as a string. The exceptions are Ints and Bools.

kubernetes/zenko/charts/cloudserver/values.yaml
```diff
@@ -57,12 +57,12 @@ externalBackends:
     keepAlive: false
     keepAliveMsecs: 1000
     maxFreeSockets: 256
-    maxSockets: null
+    maxSockets:
```
ssalaues, line 60: Double checked to validate this still gets templated as `null` by the entry point into the `config.json`

tmacro:
@bert-e bypass_build_status

ssalaues:
@bert-e approve

PR #893, opened 2019-10-30
improvement/ZENKOIO-129-test-refs-and-only

lucieleonard:
…nt publications

Testing purposes: only directive and cross-reference

none


lucieleonard:
Checked the builds: it worked.

PR #898, opened 2019-11-05
feat(ZENKO-1987): Add opal chart

tmacro:

Adds opal deployment
fixes #ZENKO-1987

Pay special attention to `kubernetes/zenko/templates/ingress.yaml` and the addition of the cert-manager charts as well as `kubernetes/zenko/values.yaml` changes

### DIFF
--- a/kubernetes/zenko/charts/opal/values.yaml
+++ b/kubernetes/zenko/charts/opal/values.yaml
@@ -1,0 +1,161 @@
+global:
+  orbit:
+    enabled: true
+    endpoint: https://api.zenko.io
+    pushEndpoint: https://push.api.zenko.io
+    workerMode: poll
+
+  locationConstraints: {}
+
+users: {}
+logging:
+  level: info
+
+mongodb:
+  replicaSet: rs0
+  replicas: 3
+
+kmip:
+  enabled: false
+  port: 5696
+  hosts: []
+  compoundCreate: false
+  bucketAttributeName: ''
+  pipelineDepth: 8
+
+proxy:
+  http: ''
+  https: ''
+  caCert: false
+  no_proxy: ''
+
+redis:
+  sentinel:
+    name: zenko
+  replicas: 3
+
+allowHealthchecksFrom: '0.0.0.0/0'
+
+blobserver:
+  enabled: False
+  endpoint: blob.zenko.local
+  enable_dns_storage_accounts: True
+  replicaCount: 3
+  replicaFactor: 1
+
+  asyncDeletes:
+    successfulJobsHistory: 1
+    failedJobsHistory: 3
+    jobs:
+      - name: container
+        command: ['yarn', 'cron:delete_container']
+        maxRetry: 1
+        maxConcurrentProcs: 10
+        maxConcurrentDeletes: 100
+        schedule: '*/5 * * * *'
+      - name: bucket
+        command: ['yarn', 'cron:delete_bucket']
+        maxRetry: 1
+        maxConcurrentProcs: 10
+        maxConcurrentDeletes: 100
+        schedule: '*/5 * * * *'
+      - name: overview
+        command: ['yarn', 'cron:delete_overview']
+        maxRetry: 1
+        maxConcurrentProcs: 10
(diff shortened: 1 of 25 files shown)


anurag4DSB (approved):

kubernetes/zenko/values.yaml
```diff
@@ -12,18 +12,42 @@ ingress:
   # Used to create an Ingress record.
   # This must match the 'cloudserver' 'endpoint', unless your client
   # supports different hostnames.
-  hosts:
-    - zenko.local
```
rahulreddy, line 16: I would not recommend removing this as it would break existing deployments when they upgrade. We can deprecate it, honor both and suggest using the new attribute. Also we need to document this, or you can use `hosts`

kubernetes/zenko/values.yaml
```diff
@@ -12,18 +12,42 @@ ingress:
...
+  # The root domain for endpoints
+  root_domain: zenko.local
+
+  # Subdomain to serve cloudserver traffic
+  # Will be added as a prefix to root_domain
+  # eg s3.example.com
+  s3_subdomain: s3
```
rahulreddy, line 29: If you leave `s3_subdomain` as an empty string would that still work and `zenko.local` remain as the default endpoint for S3? I am not talking about when both Blobserver and Cloudserver are deployed, but only when Blob is deployed.

kubernetes/zenko/values.yaml
```diff
@@ -84,6 +108,17 @@ cloudserver:
     # accountName:
       # access:
       # secret:
+      #
+      #
+s3-data:
```
rahulreddy: Do we need this or is this an artifact from testing?

rahulreddy (changes requested):

kubernetes/zenko/charts/opal/values.yaml
```diff
@@ -0,0 +1,160 @@
...
+        schedule: '*/5 * * * *'
+      - name: gc
+        command: ['yarn', 'cron:gc']
+        maxRetry: 1
+        maxConcurrentProcs: 10
+        maxConcurrentDeletes: 100
+        schedule: '*/5 * * * *'
+
+  image:
+    repository: registry.scality.com/sf-eng/blobserver
```
tcarmet, line 76: We need to discuss the use of the repo `sf-eng` to be the one that we deliver.

I thought that it was supposed to be a playground for all of us to share and use images hosted on a registry, and now I see it here.

If you like the naming of it and want to use it that way, fine, but we need to define and advocate the fact that it is going to be used not only for dev purposes.

Otherwise I recommend that we create a new repo, in which we use for proper deliver, and continue to use `sf-eng` for CI and dev.

@tmacro @rahulreddy

tcarmet (dismissed):

kubernetes/zenko/charts/opal/values.yaml
tmacro: That makes sense to me as well, not sure if we want to change it for the 1.2 release or not

kubernetes/zenko/values.yaml
tmacro: I've added a fallback where if `ingress.hosts` is defined it will be respected, along with a deprecation notice in `values.yaml` and in the message printed after install/upgrade if it is being used

kubernetes/zenko/charts/opal/values.yaml
tcarmet: I'm fine with this statement, I've created [ZENKO-2336](https://scality.atlassian.net/browse/ZENKO-2336) to create a discussion on the matter.

But still as `sf-eng` is a private repo how do we plan to deliver `blobserver`? If I don't have an answer to this question I'm not sure we can deliver 1.2 (note: I have a couple of ideas in mind but I need to understand what's the usecase in delivering blobserver on this release).

tmacro: So far its been implied that we would provide an `imagePullSecret` for those wanting to use blobserver

rahulreddy (approved):

PR #899, opened 2019-11-07
Cosmos improvements and updates

ssalaues:
- Disables the initial ingestion from starting immediately (basis for pause/resume)
- Define threading options in CR so they're more accessible (they were available before but had to be added in manually)
- Omit unused values from the Cosmos CR (makes it easier to read/debug/etc)
- Updates the tests to manually trigger ingestion by patching the CR and triggering the job

ZENKOIO-132
ZENKOIO-133


JianqinWang (approved):

anurag4DSB:
Tried the changes with AWS OOB, works perfectly as intended.
`kubectl edit cosmos` allows you to set `suspend: false` 
This way we can control when the OOB ingestion starts.

anurag4DSB (approved):

ssalaues:
@anurag4DSB Setting `suspend: false` will indeed enable the cron job. However if you want to manually trigger an ingestion right away you can set `triggerIngestion: true`. Both options work independently from each other.

ssalaues:
@bert-e wait

tmacro:
@bert-e bypass_build_status

anurag4DSB:
@ssalaues Thank you for the clarification. I understand cronjob schedule and once we set `triggerIngestion: true`, it will trigger ingestion immediately.

what if we need to trigger another ingestion manually? Do we set `triggerIngestion: false` and then`true` again?

Also, what would you recommend? Changing these options under `rclone` or individual `jobTemplate`?

ssalaues:
> what if we need to trigger another ingestion manually? Do we set `triggerIngestion: false` and then`true` again?

Yeah that will work. It can be useful for testing. This can be done all in one command as well so you don't have to edit every time.
 
> Also, what would you recommend? Changing these options under `rclone` or individual `jobTemplate`?

I'm not sure I understand. Everything you should change is listed under `spec:`.

ssalaues:
@bert-e approve

PR #901, opened 2019-11-07
doc: give a default value for latex_title

slaperche-scality:
`latex_title` is only initialized when either `install`, `operation` or `reference` tag is given on the command line.
When building for RTD, none of those are given, and thus `latex_title` is not defined and we have an error.

To workaround this issue, we declare `latex_title` with a default value (this is harmless since RTD doesn't build the PDF documentation, only the HTML one).


slaperche-scality:
Wrong branch name again :sweat_smile:

PR #903, opened 2019-11-12
bf: fix templating error causing deployment issues

ssalaues:
When the generated version hash results in a pure int, it confuses the parser expecting string labels. This results in initial ingestion never being triggered and the operator in an endless error loop cycle. Quoting the label fixes this.
 
fixes # ZENKO-2221


alexanderchan-scality (approved):

JianqinWang (approved):

tmacro (approved):

ssalaues:
@bert-e approve

PR #904, opened 2019-11-13
feature: add pause/resume route

ssalaues:
Adds a route that will return the current status of the location cron on GET. The state can be modified on POST. Included is a README update with the routes and example usage with curl.

fixes #<ISSUE>

### DIFF
--- a/cosmos/api/types/v1alpha1/cosmos.go
+++ b/cosmos/api/types/v1alpha1/cosmos.go
@@ -31,7 +31,7 @@ type CosmosPfsdSpec struct {
 type CosmosRcloneSpec struct {
 	Schedule      string                       `json:"schedule"`
 	Suspend       bool                         `json:"suspend"`
-	TriggerIngest bool                         `json:"triggerIngestion"`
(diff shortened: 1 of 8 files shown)


JianqinWang (approved):

tmacro (approved):

rachedbenmustapha (commented):
I can see why exposing a REST endpoint for suspend/resume would be useful for manual operation (testing, troubleshooting, ...), but in the global context of Zenko, wouldn't it make sense to monitor the Zookeeper keys that backbeat-api already uses to manage suspend/resume state, and react to changes?

ssalaues:
Yeah that would make sense to have only one universal source of truth to read from. I'll leave that decision to the powers that be since this was the design choice made at the last minute as an attempt to squeeze it in for 1.2.0 GA.

dora-korpar:
@bert-e bypass_author_approval

vrancurel (approved):

vrancurel:
@bert-e bypass_author_approval

vrancurel:
@bert-e help

PR #908, opened 2019-11-23
documentation: ZENKO-2212_blobserver_API

wabernat:
Added Blobserver API documentation. Reorganized Reference guide. Made changes to conf.py and custom.css.

This fixes ZENKO-2122

I deleted mentions of tiering (hot, cold, archive, and premium), snapshots, and page and append blob types. Look for information that ties into these elements that I might have missed. 

I recommend you three confer and shard this by topic--cover your own area of expertise first, then take another third of the changes, so you only need to wade through 2/3 of this, and I still get two readers on each section. 

I still need to fix the PDF builds for this (I have been ignoring the LatexPDF side so as to cover more ground).


wabernat:
Moved to ZENKO-1633

PR #909, opened 2019-11-25
documentation: ZENKO-2212_2nd_blobserver_API

wabernat:
This is a duplicate push of the original ZENKO-2212. Bert-E wasn't digging on my nomenclature. 

This PR adds blobserver API documentation, reorganizes the Reference Guide, and makes changes to conf.py and custom.css.

This PR fixes #ZENKO-2212 

I deleted mentions of tiering (hot, cold, archive, and premium), snapshots, and page and append blob types. Look for information that ties into these elements that I might have missed.

I recommend you three confer and shard this by topic--cover your own area of expertise first, then take another third of the changes, so you only need to wade through 2/3 of this, and I still get two readers on each section.

I still need to fix the PDF builds for this (I have been ignoring the LatexPDF side so as to cover more ground).


wabernat:
This approach was not correct. The fault, dear Brutus, lies with Bert-E, not with our stars. Closed.

PR #914, opened 2019-12-04
ZENKO-2236 fixing read the docs build

tcarmet:
We broke the read the docs build because the requirements.txt
file needs to be along side the sphinx conf file (conf.py)


tmacro (approved):

dashagurova (commented):
There is another problem (latex_title is not defined), I fixed it for 1.2 but not for 1.1
Since you want to merge to 1.1 I will add the fix for 'latex_title' as well :)

dashagurova:
@bert-e approve

dashagurova (approved):

PR #939, opened 2020-01-02
ci: ZENKO-2254 limit the number of simultaneous builds to 4

tcarmet:
To ease up with the CI infrastructure let's limit the number
of simultaneous builds we can handle for now.

### DIFF
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -189,6 +189,7 @@ models:
 
 stages:
   pre-merge:
+    simultaneous_builds: 4
     worker:
       type: local
     steps:


JianqinWang:
hasn't the limit always been 4 ...? or less than 4

jonathan-gramain (approved):

JianqinWang (approved):

rahulreddy (approved):

tcarmet:
@jianqinWang it has been upgraded to 8 on the server side. So here I'm making sure we control it. I may play with this value in the near future so it's not definitive. If you see setup issue saying that the install failed. We should lower this value.

PR #959, opened 2020-02-13
documentation/ZENKO-2329_fix_bad_link

wabernat:
Fixed bad link in operation/Architecture/Orbit.rst

Fixes ZENKO-2329


wabernat:
bad commit message

PR #998, opened 2020-04-03
bf(ZENKO(2515): Add missing conditionals for opal

tmacro:
Opal is disabled by default and should only be deployed if set in the user's values. Currently blobserver and some of its async jobs are created using the default. This PR adds the missing conditionals to the helm chart to prevent this. It also fixes a minor quoting mistake.

fixes ZENKO-2515


alexanderchan-scality (approved):

anurag4DSB (approved):

tmacro:
@bert-e approve

PR #1012, opened 2020-05-14
Generate config.yaml for ISO solution

miniscruff:
* Simple makefile to get started
* One make target to generate config.yaml
* Empty README.md to fill in later
* Add nested `_build` to gitignore

Generate config.yaml for the solution iso

fixes ZENKO-2600 in jira


miniscruff:
Recreating with the parent task number

PR #1018, opened 2020-05-18
feature: ZENKO-2413 Build ISO Solution

miniscruff:
* Deps.txt to store zenko dependencies
* Copy some operator yaml files into ISO
* Build manifest.yaml into ISO
* Pull and store images in /images
* Generate registry-config.inc from docker image python script
* Generate a solutions ISO similar to examples

fixes ZENKO-2413


solution/zenko-cr.yaml
```diff
@@ -0,0 +1,48 @@
+---
```
rachedbenmustapha: This is not a Zenko CR but a Zenko Version CR

miniscruff: oh, the contents are right though just the name is wrong?

solution/zenko-version-cr.yaml
```diff
@@ -0,0 +1,48 @@
...
+      tag: 2.12-2.3.0
+    shell:
+      image: busybox
+      tag: 1.31.0-glibc
+    mongodb:
+      image: mongodb
+      tag: 4.0
+    redis:
+      image: redis
+      tag: 5.0.3-v1
```
alexanderchan-scality: the `mongodb` and `redis` should have the following structure
```suggestion
    mongodb: 4.0
    redis: 5.0.3-v1
```

solution/deps.txt
```diff
@@ -0,0 +1,9 @@
+banzaicloud/kafka-operator:v0.11.2
+jimmidyson/configmap-reload:v0.3.0
+pravega/zookeeper-operator:latest
+quay.io/jetstack/cert-manager-controller:v0.11.0
+quay.io/jetstack/cert-manager-webhook:v0.11.0
+quay.io/jetstack/cert-manager-cainjector:v0.11.0
+quay.io/coreos/prometheus-operator:v0.38.1
+quay.io/coreos/prometheus-config-reloader:v0.38.1
+registry.scality.com/sf-eng/zenko-operator:10794ca
```
alexanderchan-scality: missing kubedb operator as dep:
* kubedb/operator:v0.13.0-rc.0

solution/zenko-version-cr.yaml
miniscruff: Hmm, are these not equivalent? My bash parsing uses the image: and tag: lines to pull from, these two were special cases I was trying to unspecial. Not a huge deal though.

miniscruff: Actually, regardless I dont seem to be able to pull mongo or redis of these tags.
* `docker pull mongodb:4.0`
* `docker pull mongodb:4.0-v1` ( this was the actual image used )
* `docker pull redis:5.0.3-v1`

Am I missing something here?

alexanderchan-scality: they aren't. the `mongodb` and `redis` services will be deployed by the `kubedb` operator, and they have a different way of defining images via [`mongodbversions`](https://kubedb.com/docs/0.9.0/concepts/catalog/mongodb/) and [`redisverisons` ](https://kubedb.com/docs/0.9.0/concepts/catalog/redis/)custom resources. 
In the `zenkoversions` manifest here, we are just declaring which versions should be used. 

Changes made in https://github.com/scality/zenko-operator/pull/54 will allow for custom versions that pull from the private registry to be defined, which should make the manifest more uniform

alexanderchan-scality: the mongodbversion has the following images:
```
apiVersion: catalog.kubedb.com/v1alpha1
kind: MongoDBVersion
metadata:
...
  name: 4.0-v1
spec:
  db:
    image: kubedb/mongo:4.0-v1
  exporter:
    image: kubedb/mongodb_exporter:v1.0.0
  initContainer:
    image: kubedb/busybox
  podSecurityPolicies:
    databasePolicyName: mongodb-db
    snapshotterPolicyName: mongodb-snapshot
  tools:
    image: kubedb/mongo-tools:4.0-v1
  version: 4.0.5

```

alexanderchan-scality: the redisversion:
```
apiVersion: catalog.kubedb.com/v1alpha1
kind: RedisVersion
metadata:
...
  name: 5.0.3-v1
spec:
  db:
    image: kubedb/redis:5.0.3-v1
  exporter:
    image: kubedb/redis_exporter:v0.21.1
  podSecurityPolicies:
    databasePolicyName: redis-db
  version: 5.0.3

```

miniscruff: oh ok, I can accept your change but won't I still need to pre-pull and save the mongo and redis docker images in the ISO anyway?

alexanderchan-scality: yeah, you will still need to do so.

miniscruff: oh nice, let me get this in there.

solution/deps.txt
miniscruff: This should be resolved now along with the other kubedb dependencies

solution/zenko-cr.yaml
rachedbenmustapha: The content's fine, it's just that the name is misleading regarding what it is.

There is the question of whether we consider this file to be the single source of truth regarding image versions contained in a global Zenko release, or if we want this to be generated from a more canonical source. Both options are open.

miniscruff: I have renamed it, for the time being, it is used as a partial source containing some of the dependencies where the others are placed in dep.txt. Kind of a weird split but since this file was in another repo I thought it was better to use it as much as I could and create another file for what is missing. A single source of truth for all dependencies would be nice though.

solution/build.sh
```diff
@@ -0,0 +1,136 @@
...
+BUILD_TIMESTAMP=${BUILD_TIMESTAMP}
+BUILD_HOST=${BUILD_HOST}
+EOF
+}
+
+function gen_operator_yaml()
+{
+	OPERATOR_IMAGE=$(grep /zenko-operator: deps.txt)
+	# we need to escape / with \/ in our sed command
+	sed "s/REPLACE_IMAGE/${OPERATOR_IMAGE//\//\\/}/" operator.yaml > ${ISO_ROOT}/operator/operator.yaml
```
jonathan-gramain: You can use any character as delimiter for the sed substitution e.g. `s:a:b:`, might reduce the number of backslashes :)

solution/build.sh
```diff
@@ -0,0 +1,136 @@
...
+}
+
+function build_registry_config()
+{
+	docker run \
+    	 	--name static-oci-registry \
+	 	--mount type=bind,source=${ISO_ROOT}/images,destination=/var/lib/images \
+		--mount type=bind,source=${ISO_ROOT},destination=/var/run \
+	 	--rm \
+    		docker.io/nicolast/static-container-registry:latest \
```
jonathan-gramain, line 138: Is `nicolast` the current reference repository for the registry? Might be good to keep a TODO around to think about changing it if it is temporary, and maybe use a shell variable for it.

solution/build.sh
```diff
@@ -0,0 +1,136 @@
...
+HARDLINK=hardlink
+OPERATOR_SDK=operator-sdk
+OPERATOR_SDK_OPTS=
+SKOPEO=skopeo
+SKOPEO_OPTS="--override-os linux --insecure-policy"
+
+# grab our dependencies from our deps.txt file as an array
+readarray -t DEP_IMAGES < deps.txt
+# add the CR dependencies by joining image: and tag: lines
+readarray -t CR_IMAGES < <(grep image: zenko-version-cr.yaml | awk '{print $2}')
```
jonathan-gramain: It might be the best option for now to `grep` in the yaml files, but it would be more robust if some sort of `jq` for yaml could be used instead.

solution/build.sh
miniscruff: Oh... thats probably a good idea... I definitely didn't typo this 10 times or anything...

solution/build.sh
miniscruff: https://github.com/NicolasT/static-container-registry doesn't look temporary, but maybe @NicolasT has a recommendation?

solution/build.sh
miniscruff: Does jq work for yaml or are you saying if something similar did exist? Cause that would be swell.

jonathan-gramain: I don't know if a `jq` for yaml exists, but it would be useful if it does! Otherwise I'm sure some python scripting can do the job too.

rachedbenmustapha: A middle ground would be to write the `zenko-version-cr.yaml` file in json instead and use `jq`, as the kube client tools don't really care if it's json or yaml in general (we humans just tend to do yaml more).

miniscruff: The big reason I am using the yaml in the first place was to copy paste it from zenko-operator. If I need to rewrite it I think just merging it into the deps.txt would be fine.

solution/build.sh
miniscruff: On second thought I am not quite sure this applies, I am not using sed here this is bash search and replace. https://linuxconfig.org/introduction-to-bash-shell-parameter-expansions#h6-search-and-replace-pattern

jonathan-gramain: Ah ok, so changing the sed delimiter would not help, alright

tmacro:
```suggestion
	awk -v img="$OPERATOR_IMAGE" '{ gsub("REPLACE_IMAGE", img); print $0 }' operator.yaml > ${ISO_ROOT}/operator/operator.yaml	
```

solution/build.sh
miniscruff: @RachedBenMustapha I merged the zenko-version-cr.yaml into deps for dependencies and removed the paths from the images in the yaml as, from what I can tell, that is what is needed when using the private registry. Although, really just guessing a little here.

solution/build.sh
miniscruff: I had to tweak this a little as I think the image should just be the final name and not any url or path with it. Still unsure on this though.

solution/zenko-version-cr.yaml
```diff
@@ -0,0 +1,43 @@
...
+    zookeeper:
+      image: zookeeper
+      tag: 0.2.4
+    kafka:
+      image: kafka
+      tag: 2.12-2.3.0
+    shell:
+      image: busybox
+      tag: 1.31.0-glibc
+    mongodb: 4.0
```
alexanderchan-scality: hadn't notice this earlier (needs to match the image tag in the dep list)
```suggestion
    mongodb: 4.0-v1
```

miniscruff: oh yea oops, I had changed it to just 4.0 as it wasnt pulling before. Will accept this. ( and make sure I change the commit message properly this time... )

alexanderchan-scality:
currently, the `zenko-operator` dependencies aren't installed when the solution is added/activated. Was wondering if they should be split into separate solutions or should their respective crds/rbac/operator yamls be bundled with `zenko-operator`'s yamls  (@RachedBenMustapha @miniscruff)

rachedbenmustapha:
@alexanderchan-scality @miniscruff all dependencies should be installed/activated within the same solution ISO.

miniscruff:
I have placed all dependencies in the deps.txt that I found or was told. Any more we find can just be placed in there until we get a more robust multi repo situation organized.

rachedbenmustapha:
The deps.txt file only includes referenced docker images. YAML manifests are needed to install the operators we depend on, as the JIRA ticket mentions.

miniscruff:
@RachedBenMustapha ok I added some of the YAML files that @alexanderchan-scality had shared with me. I feel like I didn't do it quite right, or maybe missed some so let me know what you think.

solution/build.sh
```diff
@@ -0,0 +1,151 @@
...
+	wget -O ${BUILD_ROOT}/zookeeper-operator.tar.gz https://github.com/pravega/zookeeper-operator/archive/v0.2.7.tar.gz
+	mkdir ${BUILD_ROOT}/zookeeper-operator
+	tar -C ${BUILD_ROOT}/zookeeper-operator --strip-components=1 -xf ${BUILD_ROOT}/zookeeper-operator.tar.gz
+	find ${BUILD_ROOT}/zookeeper-operator/deploy/ -type f -exec cat {} + >> ${CRDS_DIR}/zookeeper-operator-bundle.yaml
+
+	# combine all kafka files into a bundle.yaml
+	wget -O ${BUILD_ROOT}/kafka-operator.tar.gz https://github.com/banzaicloud/kafka-operator/archive/v0.11.2.tar.gz
+	mkdir ${BUILD_ROOT}/kafka-operator
+	tar -C ${BUILD_ROOT}/kafka-operator --strip-components=1 -xf ${BUILD_ROOT}/kafka-operator.tar.gz
+	find ${BUILD_ROOT}/kafka-operator/config/base -type f -exec cat {} + >> ${CRDS_DIR}/kafka-operator-bundle.yaml
```
tcarmet: Optional: I see a lot of version from different dependencies being retrieved, what about setting them as variables so that we can identify and bump them easier in the future?

solution/zenko-version-cr.yaml
```diff
@@ -0,0 +1,43 @@
+---
+apiVersion: zenko.io/v1alpha1
+kind: ZenkoVersion
+metadata:
+  name: 1.1-dev
+spec:
+  imagePullPolicy: Always
+  versions:
+    cloudserver:
```
tcarmet: We're using two different version of cloudserver, one mentioned as `cloudserver` and the other mentioned as `localData`.

Wouldn't be better if we use the same one? 

cc: @alexanderchan-scality @RachedBenMustapha

solution/zenko-version-cr.yaml
```diff
@@ -0,0 +1,43 @@
+---
+apiVersion: zenko.io/v1alpha1
+kind: ZenkoVersion
+metadata:
+  name: 1.1-dev
```
tcarmet: shouldn't it be `2.0.0-dev`?

tcarmet (commented):
Mostly LGTM, leaving some open discussion

solution/zenko-version-cr.yaml
rachedbenmustapha: There are 2 entries so that they can be decoupled :) The actual code for the 2 services is separate, and it should really be in 2 repos generating 2 images.

solution/build.sh
```diff
@@ -0,0 +1,151 @@
...
+	cp zenko-version-cr.yaml ${ISO_ROOT}/operator/zenko-version-cr.yaml
+}
+
+function download_dependency_yamls()
+{
+	OPERATOR_DIR=${ISO_ROOT}/operator
+	CRDS_DIR=${OPERATOR_DIR}/crds
+
+	wget --directory=${CRDS_DIR} https://github.com/jetstack/cert-manager/releases/download/v0.11.0/cert-manager.yaml
+	wget -O ${CRDS_DIR}/prom-operator-bundle.yaml https://github.com/coreos/prometheus-operator/blob/v0.38.0/bundle.yaml
```
rachedbenmustapha: The prometheus operator already runs as part of metalk8s (I think) so we don't need to install it.

The manifests listed here look good otherwise, it's missing only kubedb (that manages redis and mongodb). Kubedb's a bit trickier as it's not a clean set of yaml manifests but an installer script: https://raw.githubusercontent.com/kubedb/installer/89fab34cf2f5d9e0bcc3c2d5b0f0599f94ff0dca/deploy/kubedb.sh . It seems that the script can be "flattened" to get only the yaml files we need here, as well as maybe a `Job` resource to execute whatever commands we can't get away without.

solution/build.sh
miniscruff: I was thinking of pulling these versions from the deps.txt file, so yes.

Edit: Versions are now pulled from the dep.txt file which acts as the source of truth for image and versions.

solution/build.sh
miniscruff: oh I seem to have missed the kubedb link Alex sent me yesterday, https://github.com/kubedb/installer/tree/v0.13.0-rc.0/deploy is what he sent. I can flatten that folder into a bundle like the others if it works. Or take a look at the shell script to see what I can pull. Still very new to this so not sure what is more appropriate.

solution/zenko-version-cr.yaml
miniscruff: ooh, I can use the git tag here as well so we don't have to update it for each release. cc @rahulreddy @RachedBenMustapha 

Edit: Replaced with VERSION_FULL from above, let me know if this works.

solution/build.sh
alexanderchan-scality: should go with @RachedBenMustapha suggestion of https://raw.githubusercontent.com/kubedb/installer/89fab34cf2f5d9e0bcc3c2d5b0f0599f94ff0dca/deploy/kubedb.sh  rather than the `v0.13.0-rc.0` version; the `v0.13.0-rc.0` version has some issues installing on kubernetes version 1.16+, which metalk8s seems to be running

rachedbenmustapha: I strongly suggest looking inside the installer script and adapting the bundle for our needs:

- we don't need most of the versions from the catalog (actually none since zenko-operator has a PR in review to generate our own versions)
- we should be mindful that the namespace kubedb will be deployed in is not going to be its default (kube-system) so this may have an impact on actual solution "activation" in its own namespace by metalk8s.
- in the first iteration we could do without things like pod security policies, mutating webhooks, validating webhooks, ...

Alternatively kubedb has a chart here: https://github.com/kubedb/installer/tree/master/charts/kubedb that can be "templated" with helm to generate yaml which could be convenient (or not).

miniscruff: @RachedBenMustapha Mind if I hold off on the kubedb work until the next PR? Get this one merged and work on the kubedb after. Try and keep the PR's relatively small.

solution/build.sh
```diff
@@ -0,0 +1,157 @@
...
+	cp -R -f operator/ ${ISO_ROOT}/operator
+	sed "s/VERSION_FULL/${VERSION_FULL}/" zenko-version-cr.yaml > ${ISO_ROOT}/operator/zenko-version-cr.yaml
+}
+
+function download_dependency_yamls()
+{
+	OPERATOR_DIR=${ISO_ROOT}/operator
+	CRDS_DIR=${OPERATOR_DIR}/crds
+
+	wget --directory=${CRDS_DIR} https://github.com/jetstack/cert-manager/releases/download/v0.11.0/cert-manager.yaml || exit 1
```
miniscruff: @tcarmet I added some `|| exit 1` and file checks to make bad stuff actually exit properly.

rachedbenmustapha: `set -e` at the top of the script will make anything that fails exit. It's generally good practice along with `set -u`.

miniscruff: hmm, gotta look up -u. I tried `-e` and it made some other things fail when they were fine. But will give it another shot. I think it was my extra if checks were causing issues that probably just need to not exist with `-e`.

rachedbenmustapha: `-u` (combined with `-e`) aborts when reading an undefined variable. `-e` can cause problems on scripts that were written without it in the first place, but any well-written script will generally work fine with `-e`.

miniscruff: ^ sweet yea, that works now. I was doing extra if checks which kinda voided the -e before. Removed those ( after checking they still work ) and added the -u.

solution/zenko-version-cr.yaml
tcarmet: alright then! thanks for the answer

solution/build.sh
tcarmet: nice!

solution/build.sh
```diff
@@ -0,0 +1,157 @@
...
+	mkdir ${BUILD_ROOT}/zookeeper-operator
+	tar -C ${BUILD_ROOT}/zookeeper-operator --strip-components=1 -xf ${BUILD_ROOT}/zookeeper-operator.tar.gz
+	find ${BUILD_ROOT}/zookeeper-operator/deploy/ -type f -exec cat {} + >> ${CRDS_DIR}/zookeeper-operator-bundle.yaml
+
+	# combine all kafka files into a bundle.yaml
+	KAFKA_VERSION=$(grep /kafka-operator: deps.txt | awk -F ':' '{print $2}')
+	wget -O ${BUILD_ROOT}/kafka-operator.tar.gz https://github.com/banzaicloud/kafka-operator/archive/${KAFKA_VERSION}.tar.gz
+	mkdir ${BUILD_ROOT}/kafka-operator
+	tar -C ${BUILD_ROOT}/kafka-operator --strip-components=1 -xf ${BUILD_ROOT}/kafka-operator.tar.gz
+	find ${BUILD_ROOT}/kafka-operator/config/base -type f -exec cat {} + >> ${CRDS_DIR}/kafka-operator-bundle.yaml
```
alexanderchan-scality: I don't believe the `*-bundle.yaml` files will be picked up for deployment (https://github.com/scality/metalk8s/blob/development/2.6/salt/metalk8s/orchestrate/solutions/deploy-components.sls#L12-L38)

Order is somewhat important in the bundle file, so I don't the find/append is always safe (also there should be the separator, `---`, between each manfiest)

miniscruff: oh dang ok... should I just build manual yaml files now? Not sure where to place them if I did though. I could create a manual ordering and fix the --- between files. But I have no idea how to resolve the salt stuff, nor the actual ordering required.

miniscruff: It seems that I just need to edit the file names to end with `_crd.yaml` for them to be picked up by salt. Not sure if order between bundles matter though, and not sure how to resolve that if it does.

rachedbenmustapha: Beware that not everything is a CRD.

solution/build.sh
```diff
@@ -0,0 +1,183 @@
...
+}
+
+function mkdirs()
+{
+	mkdir -p ${ISO_ROOT}
+}
+
+function gen_manifest_yaml()
+{
+	cat > ${ISO_ROOT}/manifest.yaml <<EOF
```
miniscruff, line 45: @gdemonet @alexandre-allard-scality I made adjustments according to your PR, let me know if this seems to line up with the new format.

solution/build.sh
```diff
@@ -0,0 +1,183 @@
...
+apiVersion: solutions.metalk8s.scality.com/v1alpha1
+kind: SolutionConfig
```
alexandre-allard: The kind is now `Solution`:
```suggestion
kind: Solution
```

solution/build.sh
```diff
@@ -0,0 +1,183 @@
...
+	echo kind: OperatorConfig >> ${CONFIG_PATH}
+	echo repositories: >> ${CONFIG_PATH}
+  	echo "  ${VERSION_FULL}:" >> ${CONFIG_PATH}
+    	echo "    - endpoint: metalk8s-registry/${PRODUCT_LOWERNAME}-${VERSION_FULL}" >> ${CONFIG_PATH}
+      	echo "    images:" >> ${CONFIG_PATH}
+	for dep in ${DEP_IMAGES[@]}; do
+		SHORT_DEP=${dep##*/}
+        	echo "      - ${SHORT_DEP}" >> ${CONFIG_PATH}
+	done
+}
```
alexandre-allard: We don't need to generate this file, it will be done by MetalK8s and passed to the Operator using the `--config` option.
Plus, the generated YAML is wrong (`images` key is under indented):
```suggestion
function gen_operator_config_yaml()
{
	CONFIG_PATH=${ISO_ROOT}/operator/operator-config.yaml

	echo apiVersion: solutions.metalk8s.scality.com/v1alpha1 > ${CONFIG_PATH}
	echo kind: OperatorConfig >> ${CONFIG_PATH}
	echo repositories: >> ${CONFIG_PATH}
  	echo "  ${VERSION_FULL}:" >> ${CONFIG_PATH}
    	echo "    - endpoint: metalk8s-registry/${PRODUCT_LOWERNAME}-${VERSION_FULL}" >> ${CONFIG_PATH}
      	echo "      images:" >> ${CONFIG_PATH}
	for dep in ${DEP_IMAGES[@]}; do
		SHORT_DEP=${dep##*/}
        	echo "        - ${SHORT_DEP}" >> ${CONFIG_PATH}
	done
}
```

solution/build.sh
```diff
@@ -0,0 +1,183 @@
...
+
+# run everything in order
+clean
+mkdirs
+gen_manifest_yaml
+copy_yamls
+download_dependency_yamls
+gen_operator_yaml
+gen_operator_config_yaml
+for img in ${DEP_IMAGES[@]}; do
```
alexandre-allard: When iterating over an array in Bash, it's a good practice to always protect the variable with double quotes (even if there is no spaces), to avoid unwanted behavior because of white characters:
```suggestion
for img in "${DEP_IMAGES[@]}"; do
```

solution/build.sh
```diff
@@ -0,0 +1,183 @@
...
+	echo --- >> ${ZOOKEEPER_CRD_FILE}
+	cat ${BUILD_ROOT}/zookeeper-operator/deploy/all_ns/operator.yaml >> ${ZOOKEEPER_CRD_FILE}
+}
+
+
+function gen_operator_yaml()
+{
+	# we need to escape / with \/ in our sed command
+	sed "s/REPLACE_IMAGE/zenko-operator:${OPERATOR_TAG}/" operator.yaml > ${ISO_ROOT}/operator/operator.yaml
+}
```
alexandre-allard: Not used by the Solution framework, the Solution will generate the Operator Deployment manifest itself.

solution/build.sh
```diff
@@ -0,0 +1,183 @@
...
+function copy_yamls()
+{
+	cp -R -f operator/ ${ISO_ROOT}/operator
+	sed "s/VERSION_FULL/${VERSION_FULL}/" zenko-version-cr.yaml > ${ISO_ROOT}/operator/zenko-version-cr.yaml
+}
+
+function download_dependency_yamls()
+{
+	OPERATOR_DIR=${ISO_ROOT}/operator
+	CRDS_DIR=${OPERATOR_DIR}/crds
```
alexandre-allard: CRDs must be put under `/operator/deploy/crds/`:
```suggestion
	CRDS_DIR=${OPERATOR_DIR}/deploy/crds
```

solution/build.sh
```diff
@@ -0,0 +1,183 @@
...
+	ZOOKEEPER_VERSION=$(grep /zookeeper-operator: deps.txt | awk -F ':' '{print $2}')
+	ZOOKEEPER_CRD_FILE=${CRDS_DIR}/zookeeper_operator_crd.yaml
+	wget -O ${BUILD_ROOT}/zookeeper-operator.tar.gz https://github.com/pravega/zookeeper-operator/archive/v${ZOOKEEPER_VERSION}.tar.gz
+	mkdir ${BUILD_ROOT}/zookeeper-operator
+	tar -C ${BUILD_ROOT}/zookeeper-operator --strip-components=1 -xf ${BUILD_ROOT}/zookeeper-operator.tar.gz
+	cp ${BUILD_ROOT}/zookeeper-operator/deploy/crds/zookeeper_v1beta1_zookeepercluster_crd.yaml ${ZOOKEEPER_CRD_FILE}
+	echo --- >> ${ZOOKEEPER_CRD_FILE}
+	cat ${BUILD_ROOT}/zookeeper-operator/deploy/all_ns/rbac.yaml >> ${ZOOKEEPER_CRD_FILE}
+	echo --- >> ${ZOOKEEPER_CRD_FILE}
+	cat ${BUILD_ROOT}/zookeeper-operator/deploy/all_ns/operator.yaml >> ${ZOOKEEPER_CRD_FILE}
```
alexandre-allard: It will not work, Solution is designed to handle only one object per CRD manifest (the CRD itself).

solution/build.sh
```diff
@@ -0,0 +1,183 @@
...
+	cp -R -f operator/ ${ISO_ROOT}/operator
+	sed "s/VERSION_FULL/${VERSION_FULL}/" zenko-version-cr.yaml > ${ISO_ROOT}/operator/zenko-version-cr.yaml
+}
+
+function download_dependency_yamls()
+{
+	OPERATOR_DIR=${ISO_ROOT}/operator
+	CRDS_DIR=${OPERATOR_DIR}/crds
+
+	CERT_VERSION=$(grep /cert-manager-controller: deps.txt | awk -F ':' '{print $2}')
```
alexandre-allard: Can be achieved with a single `grep`:
```suggestion
	CERT_VERSION=$(grep -oP '/cert-manager-controller:\K.+$' deps.txt)
```
Or a single `awk`:
```suggestion
	CERT_VERSION=$(awk -F ':' '$1 ~ /\/cert-manager-controller$/{ print $2 }' deps.txt)
```

solution/operator/role_binding.yaml
```diff
@@ -0,0 +1,11 @@
...
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: zenko-operator
+subjects:
+- kind: ServiceAccount
+  name: zenko-operator
+roleRef:
+  kind: Role
+  name: zenko-operator
+  apiGroup: rbac.authorization.k8s.io
```
alexandre-allard: We don't need this, MetalK8s will take care of `RoleBinding`.

solution/operator/service_account.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: zenko-operator
```
alexandre-allard: Not needed either, a `ServiceAccount` named `{{ solution.name }}-operator` will be created by MetalK8s.
In this case, this account would be named `zenko-operator`.

solution/operator/crds/kafka_operator_crd.yaml
```diff
@@ -0,0 +1,2472 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
alexandre-allard: Should be put under `solution/operator/deploy/crds/`.
Moreover, each CRD manifest must contain only one CRD.
Other resources such as `Role`, `RoleBinding`, `Deployment`, `Service`, `ServiceAccount`, ...  should not be there and rather be managed by the Zenko Operator itself.

solution/build.sh
gdemonet: (still good practice to keep it in the ISO, for other contexts or manual tests)

solution/build.sh
gdemonet: Indeed it's not a temporary repository, and is being included as a submodule of scality/metalk8s as well (since it's a rather basic Python script, maybe you don't need the Docker container to run it?).

solution/build.sh
gdemonet: (JSON is valid YAML :wink:)

solution/operator.yaml
```diff
@@ -0,0 +1,32 @@
+apiVersion: apps/v1
```
alexandre-allard: As said above, this file is not needed since Solution will take care of creating the Operator `Deployment`

solution/build.sh
```diff
@@ -0,0 +1,122 @@
...
+DEVELOPMENT_RELEASE=1
+BUILD_TIMESTAMP=${BUILD_TIMESTAMP}
+BUILD_HOST=${BUILD_HOST}
+EOF
+}
+
+function gen_operator_yaml()
+{
+	OPERATOR_IMAGE=$(grep /zenko-operator: deps.txt)
+	# we need to escape / with \/ in our sed command
```
gdemonet: (or use a different separator, e.g. `sed "s|REPLACE_IMAGE|${OPERATOR_IMAGE}|" operator.yaml`)

solution/build.sh
```diff
@@ -0,0 +1,183 @@
...
+    solutions.metalk8s.scality.com/display-name: ${PRODUCT_NAME}
+    solutions.metalk8s.scality.com/git: ${GIT_REVISION}
+    solutions.metalk8s.scality.com/development-release: true
+    solutions.metalk8s.scality.com/build-timestamp: ${BUILD_TIMESTAMP}
+    solutions.metalk8s.scality.com/build-host: ${BUILD_HOST}
+  name: ${PRODUCT_LOWERNAME}
+spec:
+  version: ${VERSION_FULL}
+  images:
+    - zenko-operator:${OPERATOR_TAG}
```
gdemonet: You will need the exhaustive list of all the images you want to embed in the ISO in here. In practice, if you want to achieve fully offline installation, you should ship all the images that can be deployed by the Zenko Operator and its dependencies (e.g. KubeDB Operator).

solution/build.sh
```diff
@@ -0,0 +1,183 @@
...
+EOF
+}
+
+function copy_yamls()
+{
+	cp -R -f operator/ ${ISO_ROOT}/operator
+	sed "s/VERSION_FULL/${VERSION_FULL}/" zenko-version-cr.yaml > ${ISO_ROOT}/operator/zenko-version-cr.yaml
+}
+
+function download_dependency_yamls()
```
gdemonet: Why are we saving extra YAML manifests here? Isn't the knowledge of what's supposed to be deployed part of the Zenko Operator? cc @RachedBenMustapha

solution/build.sh
gdemonet: Actually, it makes a lot of sense to build such a config file for use in online context (but probably just build it as part of the Operator image). This would ease testing outside of MetalK8s.

solution/zenko-version-cr.yaml
```diff
@@ -0,0 +1,43 @@
+---
+apiVersion: zenko.io/v1alpha1
+kind: ZenkoVersion
```
gdemonet: What is this for? Where does it come from?
Is it supposed to describe a Zenko release? If yes, maybe the Solution manifest is sufficient?

solution/zenko-version-cr.yaml
```diff
@@ -0,0 +1,43 @@
+---
```
alexandre-allard: This manifest will not be deployed by the Solution framework, what's the purpose of it ? Providing an example shipped with the Solution ?

gdemonet (commented):
Looks close to being done, good job @miniscruff! (and sorry if our docs aren't super clear yet... it's an iterative process :innocent:)

For the general behaviour, I understand we expect all CRDs to be deployed by MetalK8s. Be aware of the implications of using multiple versions of a CRD (see e.g. [this doc](https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning/)), and how that's [not yet supported by MetalK8s](https://github.com/scality/metalk8s/issues/2372). Will need to be handled before we consider upgrading (thus, to be tested before we release).

However, for everything not CRD-related, I'm not sure the manifests should exist in the ISO. Instead, I would imagine deploying the components (either using standard objects like _Deployment_s or some _CustomResource_s), referencing the right image URLs, is the responsibility of Zenko Operator. If it needs some YAML manifests to know which deployment strategy to adopt, maybe they should be baked into its own container image?

solution/build.sh
miniscruff: Oh wait... this file is passed INTO the operator, I thought it was the operator handling it and passing it to metal.

solution/build.sh
miniscruff: ✨

solution/build.sh
miniscruff: oh my comment doesnt actually apply anymore... but good to know.

solution/build.sh
miniscruff: Taking this change for now and can adjust later

solution/build.sh
miniscruff: The docker container avoids having python as a dependency, and is really just as convenient for this case. ( just in case it adds a dependency and versions for python, i can avoid entirely. )

solution/deps.txt
```diff
@@ -0,0 +1,8 @@
+registry.scality.com/sf-eng/blobserver:1.0.8
```
miniscruff, line 1: @RachedBenMustapha not sure if this is exactly what you meant

solution/operator/crds/kafka_operator_crd.yaml
miniscruff: This file was removed entirely

solution/build.sh
miniscruff: This part was removed

solution/zenko-version-cr.yaml
miniscruff: This file was removed

solution/zenko-version-cr.yaml
miniscruff: This file was removed

solution/build.sh
miniscruff: No more yaml manifests

solution/build.sh
miniscruff: Image was removed

solution/operator.yaml
miniscruff: File was removed

solution/operator/service_account.yaml
miniscruff: Removed the file

solution/operator/role_binding.yaml
miniscruff: Removed the file

solution/build.sh
```diff
@@ -0,0 +1,180 @@
...
+{
+	docker run \
+    	 	--name static-oci-registry \
+	 	--mount type=bind,source=${ISO_ROOT}/images,destination=/var/lib/images \
+		--mount type=bind,source=${ISO_ROOT},destination=/var/run \
+	 	--rm \
+    		docker.io/nicolast/static-container-registry:latest \
+		python3 /static-container-registry.py --omit-constants /var/lib/images
+	CONFIG_FILE=${ISO_ROOT}/static-container-registry.conf
+	mv ${CONFIG_FILE} ${ISO_ROOT}/registry-config.inc
```
miniscruff: This seems to be required by solutions
```suggestion
	mv ${CONFIG_FILE} ${ISO_ROOT}/registry-config.inc.j2
```

miniscruff:
Was able to import, activate, and add the solution to a metal 2.5.1 dev cluster. Got stuck on pulling images but is good enough to merge for now imo.

tmacro (approved):

alexanderchan-scality (approved):

PR #1054, opened 2020-09-10
zenko-operator CI

alexanderchan-scality:
**What does this PR do, and why do we need it?**'

Adds CI for zenko-operator:

- [x] bootstrap kind and deploy a zenko instance
- [x] implement scripts for creating users/workflows

ZENKO-2795


rachedbenmustapha:
About the TODO "implement scripts for creating users/workflows" - you can limit your work to just creating an account to get an accesskey/secretkey pair, then use it to create a bucket, then PUT an object. Other contributors will implement workflows and such.

A note on the ZenkoVersion object: I don't know if all docker images have a floating tag, but that would be useful to always test the tip of the release train for all components, along with an `Always` pull policy.

eve/workers/end2end/scripts/build-load-operator.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+
+set -exu
+
+OPERATOR_IMAGE=${1:-zenko-operator:latest}
+OPERATOR_REPO=${2:-git@github.com:scality/zenko-operator.git}
+OPERATOR_BRANCH=${3:-development/1.0}
+
+git clone ${OPERATOR_REPO} -b ${OPERATOR_BRANCH} operator
```
miniscruff: Because we are cloning the zenko operator directly there may be a time where a zenko operator update breaks zenko but we wouldn't know until someone updated zenko. I wonder if we could run a zenko test after each operator update or something similar. Or maybe this is already handled in some way.

alexanderchan-scality: I believe this can be done by scheduled job like the current nightly runs

eve/main.yml
```diff
@@ -36,6 +40,10 @@ models:
         echo "Could not reach Docker daemon from buildbot worker" >&2
         exit 1'
       haltOnFailure: true
+  - SetProperties: &set_operator_properties
+      properties:
+        OPERATOR_IMAGE_NAME: registry.scality.com/sf-eng/zenko-operator
```
tcarmet, line 49: I don't know what we should do here, but I feel bad in leaving the operator image hosted in the `sf-eng` private space rather than public, in a public repository like Zenko.

eve/main.yml
```diff
@@ -36,6 +40,10 @@ models:
...
+        OPERATOR_IMAGE_TAG: latest
```
tcarmet: This is fine for a start but we'll be breaking things if we leave it like that in the long run.

eve/workers/end2end/Dockerfile
```diff
@@ -15,11 +16,13 @@ RUN apt-get update && apt-get install -y \
...
     gnupg2 \
     python3-buildbot-worker \
-    software-properties-common
+    software-properties-common \
+    m4
 
 # install main dependencies
 RUN curl -Lo /usr/local/bin/kind https://kind.sigs.k8s.io/dl/${KIND_VERSION}/kind-$(uname)-amd64 && \
     chmod +x /usr/local/bin/kind && \
+    curl https://raw.githubusercontent.com/helm/helm/${HELM_VERSION}/scripts/get-helm-3 | bash && \
```
tcarmet: Optional: I would rather avoid doing `curl | bash` commands when possible, suggesting the following changes:
```suggestion
  curl -sSL https://get.helm.sh/helm-v${HELM_VERSION}-linux-amd64.tar.gz | tar -xvz && \
    install linux-amd64/helm /usr/local/bin && \
    rm -rf linux-amd64 && \
```

eve/workers/end2end/crds/zenkoversion.yaml
```diff
@@ -0,0 +1,61 @@
...
+      tag: latest-8.1.2
+    backbeat:
+      image: registry.scality.com/zenko-dev/backbeat
+      tag: 161b9c1
+    blobserver:
+      image: registry.scality.com/sf-eng/blobserver
+      tag: 1.0.8
+    jabba:
+      image: registry.scality.com/sf-eng/jabba
+      tag: 940d
```
tcarmet: For my personal knowledge, do we need to install blobserver and jabba?

eve/workers/end2end/kind-config.yaml
```diff
@@ -3,6 +3,22 @@ kind: Cluster
 apiVersion: kind.x-k8s.io/v1alpha4
```
tcarmet: This file is kind of the same used in `bootstrap-kind.sh` correct?

If so, shouldn't we remove it? feels like we're duplicating code but I may be wrong.

eve/workers/end2end/scripts/build-load-operator.sh
tcarmet: +1 here, same story goes for https://github.com/scality/Zenko/pull/1054/files#diff-73e74177595a435f88a0701e605cf08eR46

I believe we can start like that, but focus on fixing a version ASAP.

tcarmet (commented):
Overall LGTM. I left some open subjects to discuss together to see what we do to move on.

Good stuff!

eve/workers/end2end/kind-config.yaml
alexanderchan-scality: this is duplicate code and should be removed

tcarmet:
> # Unsupported issue type
> 
> The Jira issue ZENKO-2795 is of type Task and cannot be used to reference a branch.
> 
> I can only merge the following issue types into development branches:
> Jira issue type 	corresponding branch prefix
> `Bug` 	`bugfix`
> `Design` 	`design`
> `Documentation` 	`documentation`
> `Feature` 	`feature`
> `Improvement` 	`improvement`
> `Story` 	`feature`
> `Sub-task` 	`any`

let me fix this issue for you

eve/workers/end2end/crds/zenkoversion.yaml
alexanderchan-scality: no. blobserver/jabba will not be needed. These images can be removed

eve/main.yml
rachedbenmustapha: Alex is whipping up something as quickly as possible, but ultimately I think the goal is to deploy what's supposed to be deployed in production - ie build and deploy a metalk8s ISO.

tcarmet: perfect then.

tests/zenko_tests/e2e_config/accounts.py
```diff
@@ -0,0 +1,59 @@
...
+    Creates a user account and save its accessKey and secretKey as k8s secret
+    """
+    try:
+        User_V1 = client.get_model('user-v1')
+        u = User_V1(userName=account_name)
+
+        res = client.ui_facing\
+            .createConfigurationOverlayUser(user=u, uuid=uuid)\
+            .response()\
+            .result
```
miniscruff: Instead of using a backslash, you can wrap in parenthesis.

```suggestion
        res = (
            client.ui_facing
            .createConfigurationOverlayUser(user=u, uuid=uuid)
            .response()
            .result
        )
```

tests/zenko_tests/configuration.py
```diff
@@ -0,0 +1,163 @@
...
+    KEY = os.getenv("KEY")
+    UUID = os.getenv("UUID")
+    CONFIG_FILE = os.getenv("CONFIG_FILE", "./e2e-config.yaml")
+    MANAGEMENT_ENDPOINT = os.getenv("MANAGEMENT_ENDPOINT",
+                                    "http://managementapi.zenko.local")
+    MANAGEMENT_ENDPOINT = MANAGEMENT_ENDPOINT.rstrip('/')
+    NAMESPACE = os.getenv("NAMESPACE", "default")
+
+    client = None
+    e2e_config = None
```
miniscruff: I don't think there is a need to initialize these two here

tests/zenko_tests/configuration.py
```diff
@@ -0,0 +1,163 @@
...
+
+def load_end2end_config(config_file=""):
+    """
+    Loads end-to-end test setup file
+    """
+    try:
+        with open(config_file) as file:
+            return yaml.full_load(file)
+    except:
+        raise Exception("Failed to load end2end config file")
```
miniscruff: Not sure the benefit of catching an exception just to raise a generic one. If the yaml file was not found or is not a proper yaml file, full_load will already raise a very similar exception ( maybe more specific ).

tests/zenko_tests/configuration.py
```diff
@@ -0,0 +1,163 @@
...
+    try:
+        config.load_kube_config(config_file=config_file)
+        return
+    except:
+        _log.info("unable to load k8s config file")
+
+    raise Exception("Failed to load k8s config")
+
+
+def load_end2end_config(config_file=""):
```
miniscruff: Not sure a default empty string here is beneficial since it will just crash

eve/workers/end2end/crds/zenkoversion.yaml
tcarmet: alright, let's do so for now, and add them when they'll be required.

eve/workers/end2end/scripts/build-latest-commits.sh
```diff
@@ -0,0 +1,69 @@
...
+
+build_dir() {
+  mkdir -p  ${BUILD_DIR}
+}
+
+build_latest_cloudserver() {
+  git clone ${CLOUDSERVER_REPO} -b ${CLOUDSERVER_BRANCH} ${BUILD_DIR}/cloudserver
+  docker build -t ${CLOUDSERVER_IMAGE}:${CLOUDSERVER_TAG} ${BUILD_DIR}/cloudserver
+  kind load docker-image ${CLOUDSERVER_IMAGE}:${CLOUDSERVER_TAG}
+}
```
miniscruff: Could probably build this into a generic function

tests/zenko_tests/e2e_config/accounts.py
alexanderchan-scality: wasn't sure about doing that, I had thought that it would be into a tuple, probably changed in python3

tests/zenko_tests/configuration.py
alexanderchan-scality: left over from when using multiple try/catch blocks. will fix

tests/zenko_tests/e2e_config/accounts.py
miniscruff: tuples have to have commas, so `(0,)` is a tuple but `(0)` is just 0. Even without parenthesis, it can be a tuple so `return 0,` returns a single item tuple. Funny errors from that one...

eve/workers/end2end/scripts/build-latest-commits.sh
```diff
@@ -0,0 +1,69 @@
...
+
+build_latest_backbeat() {
+  git clone ${BACKBEAT_REPO} -b ${BACKBEAT_BRANCH} ${BUILD_DIR}/backbeat
+  docker build -t ${BACKBEAT_IMAGE}:${BACKBEAT_TAG} ${BUILD_DIR}/backbeat
+  kind load docker-image ${BACKBEAT_IMAGE}:${BACKBEAT_TAG}
+}
```
tcarmet: I believe that applying this for cloudserver and backbeat is not required. It is already done by their own CI in the `post-merge` stage that they have in place. [you can check it per example here](https://github.com/scality/backbeat/blob/development/8.2/eve/main.yml#L135)

In short, when something get's merged in the `development/*` branches it triggers a new job that build and push an image into  `registry.scality.com/zenko/cloudserver:latest-8.2.1` (example for cloudserver), the same is valid for backbeat and cloudserver. (not yet available for s3utils but can be done quite easily)

I also prefer to keep the build/push of the image in their own project so that the project are able to "release" themselve and other repository like Zenko or Z-O can retrieve the package without having to build it.

Also I may understand that the 3 digits version can be an issue, in that case I can make a PR today that translate that also add a tag like `latest-8.2`.

Let me know what you think

alexanderchan-scality: I had looked at those repos and was unable to find images built from `development/8.2`, so I went for the clone and build step.

** it may be a mistake to pull from `dev/8.2` , will fix

tcarmet: I think backbeat currently have a issue with dev/8.2 (just a matter of changing the version referred on the `package.json`)

And I will add a little CI pipeline to s3utils and zenko-ui if it's not the case as well. so that you don't have to build it yourself.

Let me know what you think.

alexanderchan-scality: that sounds good. it would definitely reduce the CI running time

tcarmet: Created this [PR](https://github.com/scality/backbeat/pull/1055) on backbeat that allow us to have `latest-x.y` tags on our registry.
So you'll have `registry.scality.com/zenko/backbeat:latest-8.2` that will always reflect what's on `development/8.2` and so on for all the other development branches in the future.

I'll go ahead and do the same for the other components.

eve/workers/end2end/scripts/configure-e2e.sh
```diff
@@ -0,0 +1,59 @@
...
+- kind: ServiceAccount
+  name: ${SERVICE_ACCOUNT}
+roleRef:
+  kind: Role
+  name: ${SERVICE_ACCOUNT}
+  apiGroup: rbac.authorization.k8s.io
+EOF
+
+kubectl run ${POD_NAME} \
+  --image "registry.scality.com/zenko-dev/zenko-e2e:latest" \
```
tcarmet: let's avoid using `zenko-dev` as I'm planning to deprecate it (this subject remains to be discussed approved together)
```suggestion
  --image "registry.scality.com/zenko/zenko-e2e:latest" \
```

eve/workers/end2end/scripts/run-e2e-test.sh
```diff
@@ -0,0 +1,35 @@
...
+ZENKO_ACCESS_KEY=$(kubectl get secret ${ZENKO_NAME}-account-zenko -o jsonpath='{.data.accessKey}' | base64 -d)
+ZENKO_SECRET_KEY=$(kubectl get secret ${ZENKO_NAME}-account-zenko -o jsonpath='{.data.secretKey}' | base64 -d)
+
+## TODO use existing entrypoint
+if [ "$STAGE" = "end2end" ]; then
+   ARGS='sh -c cd node_tests && npm run test_operator'
+fi
+
+kubectl run ${POD_NAME} \
+  --image "registry.scality.com/zenko-dev/zenko-e2e:latest" \
```
tcarmet:
```suggestion
  --image "registry.scality.com/zenko/zenko-e2e:latest" \
```

eve/workers/end2end/scripts/deploy-zenko.sh
```diff
@@ -0,0 +1,149 @@
...
+  versions:
+    management:
+      ui:
+        image: registry.scality.com/sf-eng/zenko-ui
+        tag: latest
+    cloudserver:
+      image: registry.scality.com/zenko/cloudserver
+      tag: latest-8.1.2
+    backbeat:
+      image: registry.scality.com/zenko-dev/backbeat
```
tcarmet:
```suggestion
      image: registry.scality.com/zenko/backbeat
```

eve/workers/end2end/scripts/deploy-zenko.sh
```diff
@@ -0,0 +1,149 @@
...
+      tag: latest-8.1.1
+    utapi:
+      image: registry.scality.com/sf-eng/utapi
+      tag: zenko-1.0.0
+    secureChannelProxy:
+      image: registry.scality.com/zenko-dev/cloudserver
```
tcarmet:
```suggestion
      image: registry.scality.com/zenko/cloudserver
```

eve/workers/end2end/scripts/deploy-zenko.sh
```diff
@@ -0,0 +1,149 @@
...
+      tag: latest-8.1.2
+    localData:
+      image: registry.scality.com/zenko-dev/cloudserver
```
tcarmet:
```suggestion
      image: registry.scality.com/zenko/cloudserver
```

eve/workers/end2end/scripts/deploy-zenko.sh
```diff
@@ -0,0 +1,149 @@
...
+      tag: latest
+    metrics:
+      image: registry.scality.com/zenko-dev/cloudserver
```
tcarmet:
```suggestion
      image: registry.scality.com/zenko/cloudserver
```

eve/workers/end2end/scripts/build-latest-images.sh
```diff
@@ -0,0 +1,45 @@
...
+S3UTILS_TAG="latest"
+S3UTILS_REPO="git@github.com:scality/s3utils.git"
+S3UTILS_BRANCH="development/8.1"
+
+UI_IMAGE="registry.scality.com/sf-eng/zenko-ui"
+UI_TAG="latest"
+UI_REPO="git@github.com:scality/zenko-ui.git"
+UI_BRANCH="development/1.0"
+
+E2E_IMAGE="registry.scality.com/zenko-dev/zenko-e2e"
```
tcarmet:
```suggestion
E2E_IMAGE="registry.scality.com/zenko/zenko-e2e"
```

tcarmet (commented):
I apologize for the noise, left some suggestion to remove the usage of `zenko-dev` namespace on registry.scality.com as I planning to deprecate or even remove it.

eve/workers/end2end/scripts/configure-e2e.sh
alexanderchan-scality: at the moment, the end2end image is only built and loaded directly into kind, and it is not pushed to the repo.
question: should there be an image repo for end2end images?

rachedbenmustapha:
@tcarmet I agree that there's some cleanup to do (having both zenko-dev and sf-eng namespaces, and maybe even others), are you suggesting that we only keep the zenko namespace and remove the others?

It seems to me that having one "messy" namespace that's used to store internal/intermediate tags for dev purposes, and one "production" namespace with customer-consumable tags would be beneficial, so that:
- we keep our ability to easily deploy nightlies in our dev cluster (pushed whenever the build succeeds, but may be buggy)
- the official zenko repository looks professional and only includes validated images (pushed after successful e2e tests)

There has been 1 incident before involving an RC being deployed instead of GA because we did not separate final vs non-final builds.

tcarmet:
> @tcarmet I agree that there's some cleanup to do (having both zenko-dev and sf-eng namespaces, and maybe even others), are you suggesting that we only keep the zenko namespace and remove the others?
> 
> It seems to me that having one "messy" namespace that's used to store internal/intermediate tags for dev purposes, and one "production" namespace with customer-consumable tags would be beneficial, so that:
> 
>     * we keep our ability to easily deploy nightlies in our dev cluster (pushed whenever the build succeeds, but may be buggy)
> 
>     * the official zenko repository looks professional and only includes validated images (pushed after successful e2e tests)
> 
> 
> There has been 1 incident before involving an RC being deployed instead of GA because we did not separate final vs non-final builds.

I don't mind at all having two namespaces, one for GAs RCs and another one for dev stuff, and that is exactly why I created those at first. I wanted to avoid people outside of the company to use dev images by mistake. as it's too easy to pull them if they're very close to each other.

But then it felt unclear, I see places where we are using the `zenko` namespace with dev images and also places where `zenko-dev` images are being shared to the outside world. That is why I thought about merging them and keep only `zenko`

Now if we make the effort of making this clear, and do the proper cleaning of images on those, I'm all in.

eve/workers/end2end/scripts/configure-e2e.sh
tcarmet: I don't mind that we add a steps/stage that push this images into `registry.scality.com/zenko/zenko-e2e` could be nice to have this image here if we on our computers or cluster wants to pull it and launch some tests.

But that can be done in a later step I believe.

eve/main.yml
```diff
@@ -112,13 +124,87 @@ stages:
...
+        haltOnFailure: true
+    - ShellCommand: &build-e2e-image
+        name: Build end2end test image
+        env: *e2e-env
+        command: |
+          docker build -t ${E2E_IMAGE_NAME}:${E2E_IMAGE_TAG} tests/zenko_tests
+          docker push ${E2E_IMAGE_NAME}:${E2E_IMAGE_TAG}
+          #kind load docker-image ${E2E_IMAGE_NAME}:${E2E_IMAGE_TAG}
+        workdir: build
+        haltOnFailure: true
```
tcarmet, line 180: Can be done later IMO. But if you want to save some time here, this step can be translated into a stage, and the stage itself can be executed at the same time that  the documentation and iso are built. That way to pipeline looks like something like this:

```
---
   | --- build-doc
   | --- build-iso
   | --- build-e2e
---
|
---
    | --- end2end
---
```

It should make us save around 2 min of build time, not crazy but it's a simple win. Let's not do it now though, just an idea to keep in mind :)

tcarmet (approved):
LGTM

There's some task that still needs to be done, but I would much rather like if we do it step by step, otherwise we'll never get the work done :)

I'm gonna do my best in listing them here with their appropriate tickets if something should not be done we'll just close the ticket as won't do:
* opensource zenko-operator
* stop building component images and only pull them instead (even though this has been fixed for some of our components, a few remains to be done, but as the work is not done yet we need to build them here in the meantime)
* ensure all docker images are public (I see zenko-ui per example being pulled from a private namespace)

I'll be editing this comment to attach the tickets link

tmacro (approved):

tcarmet:
wake up bert-e

alexanderchan-scality:
squashed & rebased

PR #1060, opened 2020-09-30
Bugfix/zenko 2846 missing role and role binding in solution

miniscruff:
Role.yaml is required for metal solution to properly configure zenko operator.
Also the CRD's but they were just in the wrong path.

fixes ZENKO-2846


solution/build.sh
```diff
@@ -170,11 +126,9 @@ clean
 mkdirs
 gen_manifest_yaml
 copy_yamls
-# download_dependency_yamls
-# gen_operator_yaml
-# gen_operator_config_yaml
 for img in "${DEP_IMAGES[@]}"; do
-    ${DOCKER} ${DOCKER_OPTS} pull ${img}
+    # only pull if the image isnt already local
+    ${DOCKER} image inspect ${img} > /dev/null 2>&1 || ${DOCKER} ${DOCKER_OPTS} pull ${img}
```
tcarmet, line 131: doesn't the `docker pull` command already performs this check?

tcarmet (approved):
LGTM

just left a curiosity comment.

Edit: Just realize it's a draft PR, I'll do some re reviews along the way. but for now LGTM :)

miniscruff:
> LGTM
> 
> just left a curiosity comment.
> 
> Edit: Just realize it's a draft PR, I'll do some re reviews along the way. but for now LGTM :)

It is a draft PR cause I made it at like 3am and I didn't let the build finish... lol

solution/build.sh
miniscruff: I think no matter what it will pull in a manifest file to check something? I added this as it will fail to a 404 like error when I was using a local image for zenko operator for tests. Because it couldn't pull an image that doesn't exist, even though I had it local.

solution/deps.txt
```diff
@@ -1,8 +1,11 @@
 registry.scality.com/sf-eng/blobserver:1.0.8
 registry.scality.com/sf-eng/jabba:940d
-registry.scality.com/sf-eng/zenko-operator:10794ca
-registry.scality.com/zenko-dev/cloudserver:latest-8.1.2
+registry.scality.com/sf-eng/utapi:zenko-1.0.0
+registry.scality.com/sf-eng/vault:zenko-2.0.0
+registry.scality.com/sf-eng/zenko-operator:1.0.0-local
```
miniscruff: Edit: The real tag we need is AFTER my `--config` PR otherwise that portion of the solution will fail.

alexanderchan-scality (approved):

tcarmet (approved):

PR #1068, opened 2020-10-28
Feature/zenko 2884 package kube db as solution

miniscruff:
Creates a second solution for kubedb and other dependencies

fixes ZENKO-2884

### DIFF
--- a/solution-base/build.sh
+++ b/solution-base/build.sh
@@ -1,0 +1,174 @@
+#!/bin/bash
+set -eu
+
+PWD=$(pwd)
+BUILD_ROOT=${PWD}/_build
+ISO_ROOT=${BUILD_ROOT}/root
+IMAGES_ROOT=${ISO_ROOT}/images
+
+PRODUCT_NAME=Zenko-Base
+PRODUCT_LOWERNAME=zenko-base
+BUILD_TIMESTAMP=$(date -u +"%Y-%m-%dT%H:%M:%SZ")
+BUILD_HOST=$(hostname)
+
+VERSION_SHORT=$(git describe --abbrev=0)
+VERSION_FULL=${VERSION_SHORT}-dev
+GIT_REVISION=$(git describe --long --always --tags --dirty)
+ISO=${BUILD_ROOT}/${PRODUCT_LOWERNAME}-${VERSION_FULL}.iso
+
+DOCKER=docker
+DOCKER_OPTS=
+DOCKER_SOCKET=${DOCKER_SOCKET:-unix:///var/run/docker.sock}
+HARDLINK=hardlink
+SKOPEO=skopeo
+SKOPEO_OPTS="--override-os linux --insecure-policy"
+OPERATOR_TAG=$(grep /operator: deps.txt | awk -F ':' '{print $2}')
+SOLUTION_REGISTRY=metalk8s-registry-from-config.invalid/${PRODUCT_LOWERNAME}-${VERSION_FULL}
+
+KUBEDB_SCRIPT_BRANCH_TAG=89fab34cf2f5d9e0bcc3c2d5b0f0599f94ff0dca
+
+export KUBEDB_OPERATOR_TAG=${OPERATOR_TAG}
+export KUBEDB_NAMESPACE=SOLUTION_ENV
+export KUBEDB_SERVICE_ACCOUNT=kubedb-operator
+export KUBEDB_OPERATOR_NAME=operator
+export KUBEDB_DOCKER_REGISTRY=${SOLUTION_REGISTRY}
+export KUBEDB_PRIORITY_CLASS=system-cluster-critical
+
+# grab our dependencies from our deps.txt file as an array
+readarray -t DEP_IMAGES < deps.txt
+
+function clean()
+{
+    echo cleaning
+    rm -rf ${BUILD_ROOT}
+    rm -rf ca.crt ca.key server.crt server.key
+}
+
+function mkdirs()
+{
+    echo making dirs
+    mkdir -p ${ISO_ROOT}
+    mkdir -p ${IMAGES_ROOT}
+}
+
+function kubedb_yamls()
+{
+    echo merging kubedb yamls
+    OPERATOR_PATH=${ISO_ROOT}/operator.yaml
+
+    yamls=(
+        certs
+        operator
+        service-account
+        rbac-list
+        user-roles
+        appcatalog-user-roles
+        validating-webhook
+        mutating-webhook
+        psp-operator
+        psp-mongodb
+        psp-redis
+        kubedb-catalog-mongodb
+        kubedb-catalog-redis
+    )
+
+    for y in "${yamls[@]}"; do
+        cat kubedb/${y}.yaml | envsubst >> ${OPERATOR_PATH}
+        echo --- >> ${OPERATOR_PATH}
+    done
+}
+
+function gen_manifest_yaml()
+{
+    cat > ${ISO_ROOT}/manifest.yaml <<EOF
+apiVersion: solutions.metalk8s.scality.com/v1alpha1
+kind: Solution
+metadata:
+  annotations:
+    solutions.metalk8s.scality.com/display-name: ${PRODUCT_NAME}
+    solutions.metalk8s.scality.com/git: ${GIT_REVISION}
+    solutions.metalk8s.scality.com/development-release: true
+    solutions.metalk8s.scality.com/build-timestamp: ${BUILD_TIMESTAMP}
+    solutions.metalk8s.scality.com/build-host: ${BUILD_HOST}
+  name: ${PRODUCT_LOWERNAME}
+spec:
+  version: ${VERSION_FULL}
+EOF
+}
+
+# function copy_yamls()
+# {
+    # no yamls currently but other dependencies may require them
+    # cp -R -f operator/ ${ISO_ROOT}/operator
+# }
+
+function copy_image()
+{
+    IMAGE_NAME=${1##*/}
+    FULL_PATH=${IMAGES_ROOT}/${IMAGE_NAME/:/\/}
+    mkdir -p ${FULL_PATH}
+    ${SKOPEO} ${SKOPEO_OPTS} copy \
+        --format v2s2 --dest-compress \
+        --src-daemon-host ${DOCKER_SOCKET} \
+        docker-daemon:${1} \
+        dir:${FULL_PATH}
+}
+
+function dedupe()
+{
+    ${HARDLINK} -c ${IMAGES_ROOT}
+}
+
+function build_registry_config()
+{
+    docker run \
+        --name static-oci-registry \
+        --mount type=bind,source=${ISO_ROOT}/images,destination=/var/lib/images \
+        --mount type=bind,source=${ISO_ROOT},destination=/var/run \
+        --rm \
+        docker.io/nicolast/static-container-registry:latest \
+            python3 static-container-registry.py \
+            --name-prefix '{{ repository }}' \
+            --server-root '{{ registry_root }}' \
+            --omit-constants \
+            /var/lib/images > ${ISO_ROOT}/registry-config.inc.j2
+    rm ${ISO_ROOT}/static-container-registry.conf -f
+}
+
+function build_iso()
+{
+    mkisofs -output ${ISO} \
+        -quiet \
+        -rock \
+        -joliet \
+        -joliet-long \
+        -full-iso9660-filenames \
+        -volid "${PRODUCT_NAME} ${VERSION_FULL}" \
+        --iso-level 3 \
+        -gid 0 \
+        -uid 0 \
+        -input-charset iso8859-1 \
+        -output-charset iso8859-1 \
+        ${ISO_ROOT}
+    sha256sum ${ISO} > ${ISO_ROOT}/SHA256SUM
+    echo ISO File at ${ISO}
+    echo SHA256 for ISO:
+    cat ${ISO_ROOT}/SHA256SUM
+}
+
+# run everything in order
+clean
+mkdirs
+kubedb_yamls
+gen_manifest_yaml
+# copy_yamls
+for img in "${DEP_IMAGES[@]}"; do
+    # only pull if the image isnt already local
+    echo downloading ${img}
+    ${DOCKER} image inspect ${img} > /dev/null 2>&1 || ${DOCKER} ${DOCKER_OPTS} pull ${img}
+    copy_image ${img}
+done
+dedupe
+build_registry_config
+build_iso
+echo DONE
--- a/solution-base/deps.txt
+++ b/solution-base/deps.txt
@@ -1,0 +1,7 @@
+kubedb/operator:v0.13.0-rc.0
+kubedb/redis:5.0.3-v1
+kubedb/redis_exporter:v0.21.1
+kubedb/mongo:4.0-v1
+kubedb/mongo-tools:4.0-v1
+kubedb/busybox:latest
+kubedb/mongodb_exporter:v1.0.0
--- a/solution-base/kubedb/certs.yaml
+++ b/solution-base/kubedb/certs.yaml
@@ -1,0 +1,33 @@
+apiVersion: cert-manager.io/v1alpha2
+kind: Issuer
+metadata:
+  name: selfsigned-issuer
+  namespace: ${KUBEDB_NAMESPACE}
+spec:
+  selfSigned: {}
+---
+apiVersion: cert-manager.io/v1alpha2
+kind: Certificate
+metadata:
+  name: kubedb-operator-apiserver-cert
+  namespace: ${KUBEDB_NAMESPACE}
+spec:
+  # Secret names are always required.
+  secretName: kubedb-operator-apiserver-cert
+  # TODO: How does kubedb handle cert regens?
+  duration: 2160h # 90d
+  renewBefore: 360h # 15d
+  isCA: false
+  keyAlgorithm: rsa
+  keyEncoding: pkcs1
+  keySize: 2048
+  # At least one of a DNS Name, URI, Email Address, or IP address is required.
+  dnsNames:
+  - kubedb-${KUBEDB_OPERATOR_NAME}
+  - kubedb-${KUBEDB_OPERATOR_NAME}.${KUBEDB_NAMESPACE}.svc
+  # Issuer references are always required.
+  issuerRef:
+    name: selfsigned-issuer
+    # We can reference ClusterIssuers by changing the kind here.
+    # The default value is Issuer (i.e. a locally namespaced Issuer)
+    kind: Issuer
--- a/solution-base/kubedb/kubedb-catalog-mongodb.yaml
+++ b/solution-base/kubedb/kubedb-catalog-mongodb.yaml
@@ -1,0 +1,19 @@
+apiVersion: catalog.kubedb.com/v1alpha1
+kind: MongoDBVersion
+metadata:
+  name: "4.0-v1"
+  labels:
+    app: kubedb
+spec:
+  version: "4.0.5"
+  db:
+    image: "${KUBEDB_DOCKER_REGISTRY}/mongo:4.0-v1"
+  exporter:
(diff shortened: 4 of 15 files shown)


solution-base/build.sh
```diff
@@ -0,0 +1,205 @@
...
+HARDLINK=hardlink
+SKOPEO=skopeo
+SKOPEO_OPTS="--override-os linux --insecure-policy"
+OPERATOR_TAG=$(grep /operator: deps.txt | awk -F ':' '{print $2}')
+SOLUTION_REGISTRY=metalk8s-registry-from-config.invalid/${PRODUCT_LOWERNAME}-${VERSION_FULL}
+
+ONESSL_VERSION=v0.13.1
+
+export KUBEDB_SCRIPT_LOCATION="curl -fsSL https://raw.githubusercontent.com/kubedb/installer/v0.13.0-rc.0/"
+export KUBEDB_NAMESPACE=kube-system
```
miniscruff: Kubedb defaults to using the kube-system namespace, I don't think we can specify the solution environment here as that is an unknown value. But it may not be ideal to use kube-system in this way.

solution-base/build.sh
```diff
@@ -0,0 +1,205 @@
...
+    onessl create server-cert server --domains=kubedb-$KUBEDB_OPERATOR_NAME.$KUBEDB_NAMESPACE.svc
+    export SERVICE_SERVING_CERT_CA=$(cat ca.crt | onessl base64)
+    export TLS_SERVING_CERT=$(cat server.crt | onessl base64)
+    export TLS_SERVING_KEY=$(cat server.key | onessl base64)
+}
+
+function kubedb_yamls()
+{
+    ${KUBEDB_SCRIPT_LOCATION}deploy/operator.yaml | onessl envsubst > ${DEPLOY_ROOT}/operator.yaml
+    # ${KUBEDB_SCRIPT_LOCATION}deploy/kubedb-catalog/mongodb.yaml | onessl envsubst > deploy/crds/mongodb.yaml
```
miniscruff: the kubedb catalog for mongo includes a lot of different docker images, I am not sure if I should include the entire catalog or not as we would need to also grab the images for all versions we pulled and it seemed like a lot. 
( here is the full file for reference https://raw.githubusercontent.com/kubedb/installer/v0.13.0-rc.0/deploy/kubedb-catalog/mongodb.yaml )

solution-base/build.sh
alexanderchan-scality: we probably shouldn't deploy in  the `kube-system` namespace anyways.
Since we're going to manually run `kubectl apply`, this can probably be left as a placeholder until then

solution-base/build.sh
alexanderchan-scality: it should be safe to apply the full catalog. As long as those unused versions aren't being pulled, it is fine to leave them in. (🤔 though it does create useless resources)

solution-base/build.sh
```diff
@@ -0,0 +1,205 @@
...
+    # ${KUBEDB_SCRIPT_LOCATION}deploy/kubedb-catalog/redis.yaml | onessl envsubst > deploy/crds/redis.yaml
+    ${KUBEDB_SCRIPT_LOCATION}deploy/service-account.yaml | $ONESSL envsubst > ${DEPLOY_ROOT}/service-account.yaml
+    ${KUBEDB_SCRIPT_LOCATION}deploy/rbac-list.yaml | $ONESSL envsubst > ${DEPLOY_ROOT}/rbac-list.yaml
+    ${KUBEDB_SCRIPT_LOCATION}deploy/user-roles.yaml | $ONESSL envsubst > ${DEPLOY_ROOT}/user-roles.yaml
+    ${KUBEDB_SCRIPT_LOCATION}deploy/appcatalog-user-roles.yaml | $ONESSL envsubst > ${DEPLOY_ROOT}/appcatalog-user-roles.yaml
```
alexanderchan-scality: missing the `pod security policy` resources.

solution-base/build.sh
miniscruff: Ok so use the full catalog but only include one set of images?

solution-base/build.sh
miniscruff: hmm that sounds reasonable, does kubectl have any sort of placeholder or env var completion built in? Or just edit the file? I haven't seen anything about it before.

solution-base/build.sh
miniscruff: ooh right, adding that now

solution-base/build.sh
alexanderchan-scality: I don't believe the `kubectl` has any placeholders (without bringing in a different tool like `kustomize`).
I was thinking of using `envsubst` or remove the `namespace` line to allow for `kubectl apply` with the `--namespace` option

solution-base/build.sh
alexanderchan-scality: ^ yeah. I believe that will work

solution-base/build.sh
```diff
@@ -69,42 +67,46 @@ onessl_found() {
...
 
 function gen_certs()
 {
     echo generating certs
     onessl create ca-cert
-    onessl create server-cert server --domains=kubedb-$KUBEDB_OPERATOR_NAME.$KUBEDB_NAMESPACE.svc
+    onessl create server-cert server --domains=kubedb-$KUBEDB_OPERATOR_NAME.svc
     export SERVICE_SERVING_CERT_CA=$(cat ca.crt | onessl base64)
     export TLS_SERVING_CERT=$(cat server.crt | onessl base64)
     export TLS_SERVING_KEY=$(cat server.key | onessl base64)
```
alexanderchan-scality: I might be wrong, but it might be unsafe to build the certs during packaging stage. The ISO will have a cert associated with it, and the cert can possible be used in multiple deployments

@miniscruff @tcarmet @RachedBenMustapha what are your thoughts on this? any downside with leaving this step to when we manually apply yaml files

solution-base/build.sh
miniscruff: I did some searching after I asked and didn't find anything kubectl related, I replaced it with "SOLUTION_ENV" for now which should be an easy replace. We can decide how later on as we might need to replace more than one thing in the end.

solution-base/build.sh
miniscruff: ok done, did some clean up as well

solution-base/build.sh
miniscruff: added

solution-base/build.sh
```diff
@@ -0,0 +1,207 @@
...
+DOCKER_SOCKET=${DOCKER_SOCKET:-unix:///var/run/docker.sock}
+HARDLINK=hardlink
+SKOPEO=skopeo
+SKOPEO_OPTS="--override-os linux --insecure-policy"
+OPERATOR_TAG=$(grep /operator: deps.txt | awk -F ':' '{print $2}')
+SOLUTION_REGISTRY=metalk8s-registry-from-config.invalid/${PRODUCT_LOWERNAME}-${VERSION_FULL}
+
+ONESSL_VERSION=v0.13.1
+
+export KUBEDB_SCRIPT_LOCATION="curl -fsSL https://raw.githubusercontent.com/kubedb/installer/v0.13.0-rc.0/"
```
tcarmet: Should we use `KUBEDB_OPERATOR_TAG` in the url?

tcarmet (approved):
Only one comment to be looked at, as per the rest LGTM

solution-base/build.sh
alexanderchan-scality: also there is a bugfix in `89fab34cf2f5d9e0bcc3c2d5b0f0599f94ff0dca` that we're using, so might be good to have the `branch/tag` as a variable

solution-base/build.sh
miniscruff: I did think it was a little strange to package a cert this way but at the same time wasn't sure if it was insecure or an alternative if it was.

Edit: I guess it could also be another envsubst value like kubedb was using orginally. But it will add more steps to deploying.

rachedbenmustapha: Yes, certs should not be packaged. Alternatives would be to generate them into a `Secret` by a `Job` at install time, or ship certificate resources to be handled by `cert-manager` (which we have to make sure is there anyway), and use that secret.

tcarmet: @alexanderchan-scality Depends on the utility of the certificate, I don't know for what it's used but indeed, my guess would be to generate those certificates at installation time.

Also the following comment lacks a lot of product knowledge but, I believe we will use cert-manager, is it possible to create a cert-manager certificate object that creates it the way we want at install time? that way we just have need to expose a secret name instead of the certificate itself.

solution-base/operator.yaml
```diff
@@ -0,0 +1,162 @@
...
+  # TODO: How does kubedb handle cert regens?
+  duration: 2160h # 90d
+  renewBefore: 360h # 15d
+  isCA: false
+  keyAlgorithm: rsa
+  keyEncoding: pkcs1
+  keySize: 2048
+  # At least one of a DNS Name, URI, Email Address, or IP address is required.
+  dnsNames:
+  - kubedb-operator
```
alexanderchan-scality: this should be the operator service FQDN?

solution-base/build.sh
```diff
@@ -0,0 +1,180 @@
...
+OPERATOR_TAG=$(grep /operator: deps.txt | awk -F ':' '{print $2}')
+SOLUTION_REGISTRY=metalk8s-registry-from-config.invalid/${PRODUCT_LOWERNAME}-${VERSION_FULL}
+
+KUBEDB_SCRIPT_BRANCH_TAG=89fab34cf2f5d9e0bcc3c2d5b0f0599f94ff0dca
+
+export KUBEDB_OPERATOR_TAG=${OPERATOR_TAG}
+export KUBEDB_SCRIPT_LOCATION="curl -fsSL https://raw.githubusercontent.com/kubedb/installer/${KUBEDB_SCRIPT_BRANCH_TAG}/"
+export KUBEDB_NAMESPACE=SOLUTION_ENV
+export KUBEDB_SERVICE_ACCOUNT=kubedb-operator
+export KUBEDB_OPERATOR_NAME=operator
```
alexanderchan-scality, line 33:
```suggestion
export KUBEDB_OPERATOR_NAME=kubedb-operator
```

this would remove the need to prefix `kubedb-` in many parts in the yaml. 
(unless I had missed something)

solution-base/operator.yaml
miniscruff: oh hey, I just pushed an additional name but Thomas and I were unable to resolve an error that may or may not be related to this.
```yaml
▶ k describe apiservices v1alpha1.mutators.kubedb.com
Name:         v1alpha1.mutators.kubedb.com
Namespace:    
Labels:       app=kubedb
Annotations:  cert-manager.io/inject-ca-from: zenko-env/kubedb-operator-apiserver-cert
API Version:  apiregistration.k8s.io/v1
Kind:         APIService
Metadata:
  Creation Timestamp:  2020-11-04T23:41:07Z
  Resource Version:    8630
  Self Link:           /apis/apiregistration.k8s.io/v1/apiservices/v1alpha1.mutators.kubedb.com
  UID:                 a25dbe7e-d7a9-4b8e-b761-f079be1798b0
Spec:
  Ca Bundle:  INJECTED_VALUE
  Group:                   mutators.kubedb.com
  Group Priority Minimum:  1000
  Service:
    Name:            kubedb-operator
    Namespace:       zenko-env
    Port:            443
  Version:           v1alpha1
  Version Priority:  15
Status:
  Conditions:
    Last Transition Time:  2020-11-04T23:41:07Z
    Message:               failing or missing response from https://10.96.157.182:443/apis/mutators.kubedb.com/v1alpha1: bad status from https://10.96.157.182:443/apis/mutators.kubedb.com/v1alpha1: 404
    Reason:                FailedDiscoveryCheck
    Status:                False
    Type:                  Available
Events:                    <none>
```

solution-base/build.sh
miniscruff: Maybe, I simply copy pasted this from the kubedb operator yaml. I tried to make as few changes as possible.

solution-base/build.sh
miniscruff: @RachedBenMustapha updated to use cert-manager

solution-base/build.sh
```diff
@@ -0,0 +1,180 @@
...
+    OPERATOR_PATH=${BUILD_ROOT}/operator.yaml
+
+    yamls=(
+        kubedb-catalog/mongodb
+        service-account
+        rbac-list
+        user-roles
+        appcatalog-user-roles
+        psp/operator
+        psp/mongodb
```
alexanderchan-scality: might also need `psp/redis` since it is still an option at the moment

miniscruff: oh, do i need any more of the redis data like the catalog?

alexanderchan-scality: yes. those will be needed as well

solution-base/deps.txt
```diff
@@ -0,0 +1,9 @@
+kubedb/operator:v0.13.0-rc.0
+kubedb/redis:5.0.3-v1
+kubedb/mongo:4.0-v1
+kubedb/mongo-tools:4.0-v1
+kubedb/busybox:latest
+kubedb/mongodb_exporter:v1.0.0
+kubedb/redis:5.0.3-v1
```
alexanderchan-scality: duplicate line

solution-base/build.sh
```diff
@@ -0,0 +1,182 @@
...
+    yamls=(
+        service-account
+        rbac-list
+        user-roles
+        appcatalog-user-roles
+        psp/operator
+        psp/mongodb
+        psp/redis
+        kubedb-catalog/mongodb
+        kubedb-catalog/redis
```
rachedbenmustapha: I think the catalog yaml files should be either provided by us directly or patched because they're what define the images - and the offline nature of the iso makes it necessary to point the local registry for these images.

miniscruff: the yaml files included in kubedbs install process include variable substitution that I modify using the KUBEDB_* values above, the image name is using the metalk8s prefixed image paths. I also specify a SOLUTION_ENV that will need to be replaced during the install process ( will probably want a readme for this later on ).

rachedbenmustapha: Ah sorry, I missed that.

solution-base/build.sh
```diff
@@ -0,0 +1,182 @@
...
+ISO=${BUILD_ROOT}/${PRODUCT_LOWERNAME}-${VERSION_FULL}.iso
+
+DOCKER=docker
+DOCKER_OPTS=
+DOCKER_SOCKET=${DOCKER_SOCKET:-unix:///var/run/docker.sock}
+HARDLINK=hardlink
+SKOPEO=skopeo
+SKOPEO_OPTS="--override-os linux --insecure-policy"
+OPERATOR_TAG=$(grep /operator: deps.txt | awk -F ':' '{print $2}')
+SOLUTION_REGISTRY=metalk8s-registry-from-config.invalid/${PRODUCT_LOWERNAME}-${VERSION_FULL}
```
rachedbenmustapha, line 26: @gdemonet being able to rely on a stable and predictable registry name like here is very convenient; is this scheme/name reliable? can we make it a "contract"?

rachedbenmustapha (commented):
LGTM, just wondering if the `operator.yaml` file will be appended to for other dependencies, if not it should be named `kubedb-operator.yaml`.

miniscruff:
> LGTM, just wondering if the `operator.yaml` file will be appended to for other dependencies, if not it should be named `kubedb-operator.yaml`.

my intent was to append all dependencies to one file yes, not sure if I am sold on operator.yaml specifically though if a better name does come up ( this is just the kubedb file I started from really )

solution-base/deps.txt
```diff
@@ -0,0 +1,9 @@
...
+redis:5.0.3
```
alexanderchan-scality: not sure where this is used.
if by the zenko deployed redis, wouldn't it be better to have this in the zenko solution?

miniscruff: I pulled it from the zenko operators zenko-version YAML, but it does make sense to put it in the zenko operator iso. Does this mean I would want the MongoDB stuff as well in the zenko ISO? Not really sure where to draw the line I guess is my question.

alexanderchan-scality: the MongoDB images can stay (kubedb prefixed + redis exporter).  it's only `redis:5.0.3`/`oliver006/redis_exporter:v1.8.0` that will need to be copied to the `Zenko` solution.

miniscruff: ok will do ( setting up a metal cluster atm )

solution-base/operator.yaml
```diff
@@ -0,0 +1,390 @@
...
```
tcarmet: Adding this here before I forget: should be modified to 
```suggestion
  - kubedb-${KUBEDB_OPERATOR_NAME}
```
Because it corresponds to the service name, and that's how we templated it's value on line 117

solution-base/kubedb/kubedb-catalog-redis.yaml
```diff
@@ -0,0 +1,219 @@
+---
```
alexanderchan-scality: since we're managing this file, we can probably just create one version for the version we use. This would reduce the line changes.

solution-base/kubedb/kubedb-catalog-mongodb.yaml
```diff
@@ -0,0 +1,602 @@
+---
```
alexanderchan-scality: same here

solution-base/deps.txt
```diff
@@ -0,0 +1,6 @@
+kubedb/operator:v0.13.0-rc.0
+kubedb/redis:5.0.3-v1
```
alexanderchan-scality, line 2: probably my mistake from being unclear in a previous comment, but missing the `redis_exporter` image  also needed by `kubedb`

solution-base/kubedb/kubedb-catalog-redis.yaml
rachedbenmustapha: I agree since we have never released yet, but once we have more than 1 release, previous catalogs should be available to avoid hypothetical issues.

solution-base/kubedb/certs.yaml
```diff
@@ -0,0 +1,33 @@
...
+  duration: 2160h # 90d
+  renewBefore: 360h # 15d
+  isCA: false
+  keyAlgorithm: rsa
+  keyEncoding: pkcs1
+  keySize: 2048
+  # At least one of a DNS Name, URI, Email Address, or IP address is required.
+  dnsNames:
+  - kubedb-${KUBEDB_OPERATOR_NAME}
+  - kubedb-${KUBEDB_OPERATOR_NAME}.${KUBEDB_NAMESPACE}.svc
```
alexanderchan-scality, line 27:
```suggestion
  - kubedb-${KUBEDB_OPERATOR_NAME}.${KUBEDB_NAMESPACE}.svc.cluster.local
```

alexanderchan-scality (commented):
.

alexanderchan-scality (approved):

solution-base/build.sh
```diff
@@ -0,0 +1,174 @@
...
+
+KUBEDB_SCRIPT_BRANCH_TAG=89fab34cf2f5d9e0bcc3c2d5b0f0599f94ff0dca
+
+export KUBEDB_OPERATOR_TAG=${OPERATOR_TAG}
+export KUBEDB_NAMESPACE=SOLUTION_ENV
```
alexanderchan-scality, line 31:
```suggestion
export KUBEDB_NAMESPACE='${SOLUTION_ENV}'
```

`envsubst` would otherwise not be able to replace the placeholder

miniscruff: oh sure, I was just gonna add a `sed` example to a readme file but if we want to use envsubst instead then we should do this.

alexanderchan-scality: ah...`sed` can work too. was under the impression that `envsubst` was the tool we would be using.
if that is the case, the suggestion can be ignored.

PR #1071, opened 2020-11-16
Feature/zenko 2887/package kafka solution

alexanderchan-scality:
adds kafka-operator build step to zenko-operator dependency solution

fixes ZENKO-2887

### DIFF
--- a/eve/workers/build-iso/Dockerfile
+++ b/eve/workers/build-iso/Dockerfile
@@ -2,6 +2,7 @@ FROM centos:7
 WORKDIR /workdir
 
 ENV PATH=$PATH:/usr/local/go/bin
+ENV HELM_VERSION v3.2.3
 
 RUN yum install -y yum-utils && \
     yum-config-manager \
@@ -9,6 +10,8 @@ RUN yum install -y yum-utils && \
         https://download.docker.com/linux/centos/docker-ce.repo
 RUN yum install -y python3 make skopeo wget mkisofs git docker-ce docker-ce-cli
 RUN wget https://dl.google.com/go/go1.14.3.linux-amd64.tar.gz
+RUN curl -sSL https://get.helm.sh/helm-${HELM_VERSION}-linux-amd64.tar.gz | \
+    tar -xvz && install linux-amd64/helm /usr/local/bin
 RUN tar -C /usr/local -xzf go1.14.3.linux-amd64.tar.gz
 
 # install python + buildbot worker
--- a/solution-base/build.sh
+++ b/solution-base/build.sh
@@ -22,14 +22,15 @@ DOCKER_SOCKET=${DOCKER_SOCKET:-unix:///var/run/docker.sock}
 HARDLINK=hardlink
 SKOPEO=skopeo
 SKOPEO_OPTS="--override-os linux --insecure-policy"
-OPERATOR_TAG=$(grep /operator: deps.txt | awk -F ':' '{print $2}')
 SOLUTION_REGISTRY=metalk8s-registry-from-config.invalid/${PRODUCT_LOWERNAME}-${VERSION_FULL}
 
 KUBEDB_SCRIPT_BRANCH_TAG=89fab34cf2f5d9e0bcc3c2d5b0f0599f94ff0dca
 OPERATOR_PATH=${ISO_ROOT}/operator.yaml
 
-export KUBEDB_OPERATOR_TAG=${OPERATOR_TAG}
-export KUBEDB_NAMESPACE=SOLUTION_ENV
+SOLUTION_ENV='SOLUTION_ENV'
+
+export KUBEDB_OPERATOR_TAG=$(grep kubedb/operator: deps.txt | awk -F ':' '{print $2}')
+export KUBEDB_NAMESPACE=${SOLUTION_ENV}
 export KUBEDB_SERVICE_ACCOUNT=kubedb-operator
 export KUBEDB_OPERATOR_NAME=operator
 export KUBEDB_DOCKER_REGISTRY=${SOLUTION_REGISTRY}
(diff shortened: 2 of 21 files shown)


solution-base/build.sh
```diff
@@ -22,18 +22,30 @@ DOCKER_SOCKET=${DOCKER_SOCKET:-unix:///var/run/docker.sock}
 HARDLINK=hardlink
 SKOPEO=skopeo
 SKOPEO_OPTS="--override-os linux --insecure-policy"
-OPERATOR_TAG=$(grep /operator: deps.txt | awk -F ':' '{print $2}')
+OPERATOR_TAG=$(grep kubedb/operator: deps.txt | awk -F ':' '{print $2}')
 SOLUTION_REGISTRY=metalk8s-registry-from-config.invalid/${PRODUCT_LOWERNAME}-${VERSION_FULL}
 
 KUBEDB_SCRIPT_BRANCH_TAG=89fab34cf2f5d9e0bcc3c2d5b0f0599f94ff0dca
 
+SOLUTION_ENV="zenko"
```
miniscruff: I was using a replacable value here as I think the intention was to sed the value for whatever you choose as your metalk8s solution name. As we will not be able to define that here.

alexanderchan-scality: you're correct; Forgotten that we had a conversation about this.

solution-base/kafka/build.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
```
miniscruff: I guess we don't need this file if everything is in the root

solution-base/build.sh
```diff
@@ -0,0 +1,202 @@
...
+KAFKA_METRICS_IMAGE_REPO_TAG=$(grep ${KAFKA_METRICS_IMAGE_NAME}: deps.txt | awk -F ':' '{print $2}')
+
+# grab our dependencies from our deps.txt file as an array
+readarray -t DEP_IMAGES < deps.txt
+
+function clean()
+{
+    echo cleaning
+    rm -rf ${BUILD_ROOT}
+    rm -rf ca.crt ca.key server.crt server.key
```
miniscruff: I forgot to remove this line when switching to cert manager if you could

solution-base/build.sh
```diff
@@ -0,0 +1,202 @@
...
+    for y in "${yamls[@]}"; do
+        cat kubedb/${y}.yaml | envsubst >> ${OPERATOR_PATH}
+        echo --- >> ${OPERATOR_PATH}
+    done
+}
+
+function kafka_yamls()
+{
+    echo merging kafka operator yamls
+    CHART_PATH="$(dirname $0)/kafka/charts/kafka-operator"
```
miniscruff, line 114: is dirname $0 the same as PWD?

solution-base/kafka/charts/kafka-operator/templates/operator-validating-webhook.yaml
```diff
@@ -0,0 +1,63 @@
...
+---
+apiVersion: cert-manager.io/v1alpha2
+kind: Issuer
+metadata:
+  name: {{ include "kafka-operator.name" . }}-issuer
+  namespace: {{ .Release.Namespace }}
+spec:
+  selfSigned: {}
+---
+apiVersion: cert-manager.io/v1alpha2
```
miniscruff: should we set this to `cert-manager.io/v1` now?

eve/workers/build-iso/Dockerfile
```diff
@@ -9,6 +9,8 @@ RUN yum install -y yum-utils && \
         https://download.docker.com/linux/centos/docker-ce.repo
 RUN yum install -y python3 make skopeo wget mkisofs git docker-ce docker-ce-cli
 RUN wget https://dl.google.com/go/go1.14.3.linux-amd64.tar.gz
+RUN curl -sSL https://get.helm.sh/helm-${HELM_VERSION}-linux-amd64.tar.gz | \
```
miniscruff, line 13: I think it is worth noting the solution-base ISO is not yet part of eve

solution-base/build.sh
```diff
@@ -0,0 +1,202 @@
...
+export KUBEDB_SERVICE_ACCOUNT=kubedb-operator
+export KUBEDB_OPERATOR_NAME=operator
+export KUBEDB_DOCKER_REGISTRY=${SOLUTION_REGISTRY}
+export KUBEDB_PRIORITY_CLASS=system-cluster-critical
+
+KAFKA_NAME='kafka-operator'
+KAFKA_NAMESPACE=${SOLUTION_ENV}
+KAFKA_CERT_NAME='kafka-operator-serving-cert'
+KAFKA_IMAGE_NAME="/kafka-operator"
+KAFKA_IMAGE_REPO_NAME="${SOLUTION_REGISTRY}${KAFKA_IMAGE_NAME}"
```
tcarmet: nitpicking, I don't know if I like the `/` in the image name, I rather do it like that
```suggestion
KAFKA_IMAGE_NAME="kafka-operator"
KAFKA_IMAGE_REPO_NAME="${SOLUTION_REGISTRY}/${KAFKA_IMAGE_NAME}"
```

But, we don't need to take this comment into account, you can skip it :smile: 

I'm also starting to think there's a reason behind this :thinking:

solution-base/kafka/charts/kafka-operator/Chart.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
+name: kafka-operator
+version: 0.2.18
+description: kafka-operator manages Kafka deployments on Kubernetes
+sources:
+  - https://github.com/banzaicloud/kafka-operator
+appVersion: v0.11.2
```
tcarmet: I can see that you made the choice of committing the whole chart in here, this will avoid us some trouble in the future because we're not downloading it from the internet every time, which is nice.

Now I would consider to make the task of "upgrading" this chart easier, standardize. meaning, for this chart and the next ones that we are going to pick here, if any, could be nice to have a helper script, makefile, (whatever fits) that does the right `helm fetch kafka-operator --version $VERSION && tar -xf` command. In order to:
* Track where did this chart came from
* How did we retrieved it
* How do we make the update if needed

I also don't think it's required to do in this PR. If you have the time, great, if not, let me know what you think and we can create a ticket for it.

solution-base/kafka/charts/kafka-operator/templates/operator-validating-webhook.yaml
tcarmet: So here are some thoughts of mine, if we start patching the chart, I would rather track that for the day we need to upgrade the upstream. could be with some `.patch` files.

Or actually, make a pr on the real kafka operator chart to `v1`

tcarmet (approved):
LGTM, but leaving some thoughts and open questions.

solution-base/kafka/charts/kafka-operator/templates/operator-validating-webhook.yaml
alexanderchan-scality: `.patch` files would definitely help.
the webhook cert-manager resource template does not exist in `kafka-operator` .

solution-base/kafka/charts/kafka-operator/Chart.yaml
alexanderchan-scality: will create a script to pull and apply custom patches

solution-base/build.sh
```diff
@@ -0,0 +1,202 @@
...
+KAFKA_IMAGE_REPO_TAG=$(grep ${KAFKA_IMAGE_NAME}: deps.txt | awk -F ':' '{print $2}')
+KAFKA_METRICS_IMAGE_NAME="/kubebuilder/kube-rbac-proxy"
```
miniscruff: this needs to just be the last part as metal solutions only support one level

```suggestion
KAFKA_METRICS_IMAGE_NAME="/kube-rbac-proxy"
```

solution-base/kafka/charts/kafka-operator/templates/operator-validating-webhook.yaml
tcarmet: oh if it doesn't exist, in that case we can do whatever we want with this file

miniscruff (approved):

miniscruff (approved):

solution-base/kafka/Makefile
```diff
@@ -0,0 +1,23 @@
+ROOT_DIR:=$(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
```
tcarmet, line 1: Thanks for doing this one, it's simple and I hope it will be very helpful for future maintenance :)

tcarmet (approved):
LGTM :) :rocket:

PR #1073, opened 2020-11-18
feature: ZENKO-2885 Package zookeeper in solution-base solution

miniscruff:

Adds zookeeper to base solution

fixes ZENKO-2885

Will need to resolve conflicts with Kafka PR when that is merged.
Also fixed a typo

### DIFF
--- a/solution-base/kubedb/certs.yaml
+++ b/solution-base/kubedb/certs.yaml
@@ -1,4 +1,4 @@
-apiVersion: cert-manager.io/v1alpha2
+apiVersion: cert-manager.io/v1
 kind: Issuer
 metadata:
   name: selfsigned-issuer
@@ -6,7 +6,7 @@ metadata:
 spec:
   selfSigned: {}
 ---
-apiVersion: cert-manager.io/v1alpha2
+apiVersion: cert-manager.io/v1
 kind: Certificate
 metadata:
   name: kubedb-operator-apiserver-cert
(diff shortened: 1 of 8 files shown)


solution-base/kubedb/certs.yaml
```diff
@@ -1,12 +1,12 @@
-apiVersion: cert-manager.io/v1alpha2
+apiVersion: cert-manager.io/v1
```
miniscruff, line 1: kubedb cert-manager is behind a paywall now so I do not believe they will update to 1.0 for the version we are using.

solution-base/zookeeper/operator.yaml
```diff
@@ -0,0 +1,37 @@
...
+  template:
+    metadata:
+      labels:
+        name: zookeeper-operator
+    spec:
+      serviceAccountName: zookeeper-operator
+      containers:
+        - name: zookeeper-operator
+          # Replace this with the built image name
+          image: ${ZOOKEEPER_IMAGE_PREFIX}/zookeeper-operator:${ZOOKEEPER_OPERATOR_TAG}
```
tcarmet: I saw that we used `KUBEDB_IMAGE_NAME` or `KAFKA_OPERATOR_IMAGE_NAME` (don't know if that was the exact name but the idea was there) should we make one for `zookeeper-operator` as well?

miniscruff: Sure, will do.

miniscruff: done*

tcarmet (approved):

alexanderchan-scality (approved):

PR #1074, opened 2020-11-20
documentation/ZENKO-2294_rename_product_XDM

wabernat:
changed Zenko to XDM in Operation and Installation, and Reference.

Updated Release No. to 2.0.0

Still need to rebrand Orbit.


wabernat:
wrong name.

PR #1079, opened 2020-11-30
feature: ZENKO-3005 Add base ISO to eve

miniscruff:
Add base ISO building to eve

fixes ZENKO-3005

### DIFF
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -69,6 +69,7 @@ stages:
         stage_names:
         - build-doc
         - build-iso
+        - build-base-iso
     - TriggerStages:
         name: Launch end2end tests
         haltOnFailure: true
@@ -103,7 +104,7 @@ stages:
           - "latex/*.pdf"
 
   build-iso:
-    worker:
+    worker: &build-iso-worker
       type: kube_pod
       path: eve/workers/build-iso/pod.yaml
       images:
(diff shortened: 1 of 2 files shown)


eve/main.yml
```diff
@@ -125,6 +126,29 @@ stages:
         urls:
           - "*.iso"
 
+  build-base-iso:
+    worker:
+      type: kube_pod
+      path: eve/workers/build-iso/pod.yaml
+      images:
+        build-iso: eve/workers/build-iso
```
tcarmet: I don't know if you want to use that weird yaml pointer thing (forgot the technical name of it), but in case you do. I suggest that you set on line 107

```yaml
    worker: &build-iso-worker
```

so that you can do the following:
```suggestion
    worker: *build-iso-worker
```

tcarmet (approved):
LGTM

I left two suggestions regarding some yaml tricks, but as they can be a pain to read for human sometimes, I consider them 100% optional / up to you.

eve/main.yml
```diff
@@ -125,6 +126,29 @@ stages:
...
+        haltOnFailure: true
+        workdir: build/solution-base
+        command: ./build.sh
+        env:
+          DOCKER_SOCKET: 'http://localhost:2375'
+    - Upload:
+        name: Upload ISO
+        source: 'solution/_build'
+        urls:
+          - "*.iso"
```
tcarmet: same yaml trick here:

line 123
```yaml
    - Upload: &upload-iso
```
So that you can:
```suggestion
    - Upload: *upload-iso
```

eve/main.yml
miniscruff: oh let me try these

eve/main.yml
miniscruff: the source is different in each ISO so not sure if this works

alexanderchan-scality (approved):

eve/main.yml
tcarmet: woops, true

PR #1089, opened 2020-12-07
bf: ZENKO-2889 Reverting vendored zookeeper chart

tcarmet:


NicolasT:
This is not a fix, because it'd still mean `zookeeper` will be deployed, no? Next to `zenko-quorum`. That's the root of the issue. This `zookeeper` vendored chart should likely not exist in the first place.

tcarmet:
> This is not a fix, because it'd still mean `zookeeper` will be deployed, no? Next to `zenko-quorum`. That's the root of the issue. This `zookeeper` vendored chart should likely not exist in the first place.

hmm interesting, that's not what I understood, thanks for this feedback, I'll dig into a little bit more and make the appropriate correction.

NicolasT:
We were not depending on some online copy of the `zookeeper` chart in the first place: it was already vendored (but called `zenko-quorum` for whatever reason).

tcarmet:
> We were not depending on some online copy of the `zookeeper` chart in the first place: it was already vendored (but called `zenko-quorum` for whatever reason).

Ah ha! that's very confusing, because on the `zenko-queue-manager` chart, there was the dependency of the upstream zookeeper chart defined on the `requirements.yaml` 

So now, I'm thinking about reverting the commit 40dce3b1454086165423b3a63577a925a7f9fb58
removing the reference on the `requirements.yaml`
and call it done, therefore the ZENKO-2889 ticket was already fixed in the first place.

NicolasT:
Reverting 40dce3b is likely the right approach indeed. Not sure the requirements file needs to be changed, since  things worked just fine before...

tcarmet:
> Reverting [40dce3b](https://github.com/scality/Zenko/commit/40dce3b1454086165423b3a63577a925a7f9fb58) is likely the right approach indeed. Not sure the requirements file needs to be changed, since things worked just fine before...

I still think they need to be removed because looking at the content of it:
```yaml
dependencies:                                                                                     
- name: zookeeper                                                                                 
  version: 1.2.0                                                                                  
  repository: https://kubernetes-charts-incubator.storage.googleapis.com/                         
  condition: zookeeper.enabled                                                                    
```

I don't see how it could have "worked" before, or it did because the chart was never called out.

gdemonet:
> it did because the chart was never called out.

At least, it was not used with the default configuration (from what I could see)

NicolasT:
> > Reverting [40dce3b](https://github.com/scality/Zenko/commit/40dce3b1454086165423b3a63577a925a7f9fb58) is likely the right approach indeed. Not sure the requirements file needs to be changed, since things worked just fine before...
> 
> I still think they need to be removed because looking at the content of it:
> 
> ```yaml
> dependencies:                                                                                     
> - name: zookeeper                                                                                 
>   version: 1.2.0                                                                                  
>   repository: https://kubernetes-charts-incubator.storage.googleapis.com/                         
>   condition: zookeeper.enabled                                                                    
> ```
> 
> I don't see how it could have "worked" before, or it did because the chart was never called out.

As long as `zenko-queue-manager.zookeeper.enabled` is not set to `true` in the top-level `values.yaml` (or passed on `helm` invocation or with a custom values file and such), it's not enabled anyway. So things worked just fine, no need to change, I believe.

tcarmet:
@NicolasT @gdemonet  Alright then, so I'm not going to remove it. At least for now.

NicolasT:
Did you try deploying on a K8s 1.16 (or later) cluster? Could be Kind or whatever...

tcarmet:
> Did you try deploying on a K8s 1.16 (or later) cluster? Could be Kind or whatever...

I did:
* the version with my commit https://github.com/scality/Zenko/commit/40dce3b1454086165423b3a63577a925a7f9fb58 worked if I patched the zookeeper chart statefulset version to `app/v1`
* the version without my commit worked.
* I also tested removing the `requirements.yaml` worked fine as well on kind, with default values.

NicolasT (approved):

miniscruff (approved):

PR #1097, opened 2020-12-08
feature: ZENKO-3007 Add zenko version for solution

miniscruff:

Add zenko version to generated solution iso

fixes ZENKO-3007

### DIFF
--- a/solution/zenkoversion.yaml
+++ b/solution/zenkoversion.yaml
@@ -1,0 +1,67 @@
+---
+apiVersion: zenko.io/v1alpha1
+kind: ZenkoVersion
+metadata:
+  name: ${VERSION_FULL}
+  namespace: SOLUTION_ENV
+spec:
+  imagePullPolicy: Always
+  versions:
+    management:
+      ui:
+        image: ${SOLUTION_REGISTRY}/zenko-ui
+        tag: 2f10bac
+      api:
+        image: ${SOLUTION_REGISTRY}/pensieve-api
+        tag: 611b9ca
+      pushAPI:
+        image: ${SOLUTION_REGISTRY}/pensieve-api
+        tag: 611b9ca
+    cloudserver:
+      image: ${SOLUTION_REGISTRY}/cloudserver
+      tag: latest-8.1.2
+    backbeat:
+      image: ${SOLUTION_REGISTRY}/backbeat
+      tag: 161b9c1
+    blobserver:
+      image: ${SOLUTION_REGISTRY}/blobserver
+      tag: 1.0.8
+    jabba:
+      image: ${SOLUTION_REGISTRY}/jabba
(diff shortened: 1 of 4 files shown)


solution/zenkoversion.yaml
```diff
@@ -0,0 +1,67 @@
+---
+apiVersion: zenko.io/v1alpha1
+kind: ZenkoVersion
+metadata:
+  name: 1.1-dev
```
tcarmet: should it be `2.0.0-dev`?

miniscruff: Uh maybe? I am not sure where the names or versions are decided. I also am not sure if this needs a namespace, going to add it a small commit though.

tcarmet: I don't think the namespace is mandatory, by default kubectl will set it in whatever is set in the kubeconfig context. I like to treat it as a deployment specific config. and it can be overwritten via `kubectl apply --namespace zenkoversion.yaml` command by the user deploying. So that each person that will use this file can deploy it in the namespace that fits. 

As per the version, it should match IMO the version of Zenko that we are targetting, today we are aiming for `2.0.0-alpha1`. Something along those lines, @terrydervaux can confirm. Now, that's just my opinion, I'm open to any thoughts.

miniscruff: oh, yea always forget about --namespace, I was doing a sed replace approach as it sort of went inline with what kubedb was doing with envsubst. But I am just doing it for the namespace so I could use --namespace instead. Maybe when we write some docs we can make it all uniform.

tcarmet (approved):

solution/zenkoversion.yaml
alexanderchan-scality: this would need to be `${PRODUCT_LOWERNAME}-${VERSION_FULL}` ? the solution config seems to be using that.

miniscruff: I think it just needs to match the "version" value from the Solution manifest. Odd how one is called version while the other is name though.

miniscruff:
Will be updating some images after the other PRs are merged

alexanderchan-scality (approved):

solution/zenkoversion.yaml
```diff
@@ -0,0 +1,67 @@
+---
+apiVersion: zenko.io/v1alpha1
+kind: ZenkoVersion
+metadata:
+  name: ${VERSION_FULL}
+  namespace: SOLUTION_ENV
```
alexanderchan-scality, line 6: minor: this can probably be removed and use the '--namespace' option

miniscruff: created follow up story to also add docs https://scality.atlassian.net/browse/ZENKO-3089

PR #1102, opened 2020-12-15
ZENKO-3033 add version file

tcarmet:
In order to promote our artifacts, we need to provide to our CI a `get_product_version.sh` script that outputs the version of our product.

This script will be used to identify which version is this build working on. And allow us to promote them.


VERSION
```diff
@@ -0,0 +1,6 @@
+VERSION_MAJOR=2
+VERSION_MINOR=0
+VERSION_PATCH=0
```
miniscruff: I would probably merge these into one value, maybe just VERSION or VERSION_SHORT as it would be much easier to sed or replace later on in a script. Plus I think most people read it as one set of values, not 3 individual values.

miniscruff (commented):
Edit: Nvm it is...

eve/get_product_version.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+SCRIPT_FULL_PATH=$(readlink -f "$0")
+REPOSITORY_DIR=$(dirname "$SCRIPT_FULL_PATH")/..
+source "$REPOSITORY_DIR/VERSION" && \
+    echo "${VERSION_MAJOR}.${VERSION_MINOR}.${VERSION_PATCH}"
```
miniscruff: Does this mean artifacts can not promote RC's or alphas?

tcarmet: it can, it's just that artifacts only care about digits, then you can add whatever you want after them.

tcarmet:
```suggestion
    echo "${VERSION}"
```

miniscruff: My review is this change above ^

VERSION
tcarmet: done

miniscruff (approved):

VERSION
```diff
@@ -0,0 +1,4 @@
+VERSION="2.0.0"
+VERSION_SUFFIX=-alpha1
```
terrydervaux: I think you should use
```suggestion
VERSION_SUFFIX=-alpha.1
```
If we want to follow  https://semver.org/#spec-item-9

tcarmet: Thanks!

this review is very appreciated :')

naren-rajendran (approved):

tmacro (approved):

terrydervaux (approved):

PR #1105, opened 2020-12-19
Documentation/zenko 3058 xdm UI create a bucket

wabernat:
This PR documents the create a bucket feature in the new UI.

Ignore the garbage in the transient source file. It's a place-holder.


solution-base/build.sh
```diff
@@ -32,9 +32,7 @@ SOLUTION_ENV='SOLUTION_ENV'
 export KUBEDB_OPERATOR_TAG=$(grep kubedb/operator: deps.txt | awk -F ':' '{print $2}')
```
miniscruff, line 32: a few of these changes seem to be from my PR

miniscruff (changes requested):

solution-base/build.sh
wabernat: This looks to be some Git branch mismanagement, and odds are I'm the goat.

wabernat: Okay... I merged documentation/ZENKO-3058 with development/2.0 on my local machine, and the changes to stuff outside of Zenko/docs/docsource/ no longer appear when I do a git show. My theory was these changes happened to dev/2.0 while I had Zenko-3058 up on the lift, and that's why they were showing up here. Also, I believe I was trying to build my own changes cumulatively (i.e., docs/ZENKO-3058 on top of ZENKO-3057, and so forth) which appears to be a hurtin' strategy. 

Anyway... I have what looks to be a clean-ish branch (my changes only, and none of yours show up on a git show) but now when I try to push/force-push, I get an "already up to date" no-news-here response from git. 

So... I don't know if I've done the right thing here or screwed everything up. Please re-review what you can see and let me know if I need to revert.

PR #1120, opened 2021-01-21
feature: ZENKO-3134-update-vault-image-v8.2.0-alpha.1

anurag4DSB:
Update Vault image with v8.2.0-alpha.1

Image pull command: `docker pull registry.scality.com/vault/vault:8.2.0-alpha.1`

For instructions on docker login: https://github.com/scality/Vault/blob/development/8.2/docs/RELEASE.md#how-to-pull-docker-images

Respective Operator PR merged: https://github.com/scality/zenko-operator/pull/128/files

### DIFF
--- a/solution/zenkoversion.yaml
+++ b/solution/zenkoversion.yaml
@@ -50,8 +50,8 @@ spec:
       image: kafka
       tag: 2.12-2.3.0
     vault:
-      image: vault
-      tag: zenko-2.0.0
+      image: ${SOLUTION_REGISTRY}/vault
+      tag: 8.2.0-alpha.1
     shell:
       image: busybox
       tag: latest
--- a/solution/deps.txt
+++ b/solution/deps.txt
@@ -7,7 +7,7 @@ registry.scality.com/sf-eng/blobserver:1.0.8
 registry.scality.com/sf-eng/jabba:940d
 registry.scality.com/sf-eng/pensieve-api:611b9ca
 registry.scality.com/sf-eng/utapi:zenko-1.0.0
-registry.scality.com/sf-eng/vault:zenko-2.0.0
+registry.scality.com/vault/vault:8.2.0-alpha.1
 registry.scality.com/sf-eng/zenko-operator:0a7b2f0
 registry.scality.com/zenko-dev/backbeat:161b9c1
 registry.scality.com/zenko-dev/cloudserver:latest-8.1.2


anurag4DSB:
ping

tcarmet:
I believe we also need to change the registry url in the file `solution/deps.txt`

miniscruff:
ZENKO-3155 is going to update all the images, plus I think we still need to update zenko-operator to handle the new vault before putting it in zenko, which is ZENKO-3146 that I will be working on starting tomorrow.

anurag4DSB:
> [ZENKO-3155](https://scality.atlassian.net/browse/ZENKO-3155)

@terrydervaux as this will be done in Zenko-3155
Should I close my ticket?

anurag4DSB:
> I believe we also need to change the registry url in the file `solution/deps.txt`

@tcarmet changed the registry URL here in the squash commit :)

anurag4DSB:
> [ZENKO-3155](https://scality.atlassian.net/browse/ZENKO-3155) is going to update all the images, plus I think we still need to update zenko-operator to handle the new vault before putting it in zenko, which is [ZENKO-3146](https://scality.atlassian.net/browse/ZENKO-3146) that I will be working on starting tomorrow.

@miniscruff This should not block me from updating vault dependency. If the zenko CI failed because of the operator or some other reason, yes the PR should not be merge until that thing is fixed. So I do not see a reason why this PR should be blocked or put on hold till we update the operator.

If CI is green we should be able to merge our dep update. So that we can move on to the next thing.

That said I know the CI is not in place properly yet. For now Zenko2.x will not work(already doesn't) and the CI is not reliable. But we will get there :) 

Per the process discussed in the various meetings, I should be able to update my dependency if it is ready. What do you think? :)

miniscruff:
> If CI is green we should be able to merge our dep update. So that we can move on to the next thing.

The build will not fail because the e2e test does not use either of the files you updated. That config is placed inside a shell script so your change is not being tested at all. ( Updating this file is also related to the issue I posted and will be merged into the single source later on ).

> Per the process discussed in the various meetings, I should be able to update my dependency if it is ready. What do you think? :)

Typically yes, but, and correct me if I am wrong here, your latest change requires a new configuration. Meaning, if we do not add this new config vault will not work properly. That was my concern.

anurag4DSB:
forced pushed review commits and new vault version with fixes.

@miniscruff I agree with you, in a real world scenario(near future) when zenko-CI is working, I won't update vault Image till dependencies like Operator is also updated.

PS: Second thoughts...

anurag4DSB:
The pending PR has been merged: https://github.com/scality/zenko-operator/pull/128/files

solution/zenkoversion.yaml
```diff
@@ -50,8 +50,8 @@ spec:
       image: kafka
       tag: 2.12-2.3.0
     vault:
-      image: vault
-      tag: zenko-2.0.0
+      image: ${SOLUTION_REGISTRY}/vault
```
miniscruff, line 53:
```suggestion
      image: vault
```

anurag4DSB: I was following what zenko-ui has on line # 11 of the same file.
and the same is mentioned in the summary of this ticket: https://scality.atlassian.net/browse/ZENKO-3134

If I specify just vault, won't it look for it in dockerhub and not solution registry?

CC 
@tcarmet for confirmation as he reported the ticket.
@nicolas2bert for awareness

anurag4DSB:
Update: closing this PR as ZENKO-3163 will handle image update of vault as well (per discussion with @miniscruff)

PR #1124, opened 2021-01-30
improvement: ZENKO-3127 set image pull policy

rahulreddy:
Set image pull policy to `IfNotPresent`

### DIFF
--- a/eve/workers/end2end/scripts/deploy-zenko.sh
+++ b/eve/workers/end2end/scripts/deploy-zenko.sh
@@ -61,7 +61,7 @@ metadata:
   name: ${ZENKO_VERSION_NAME}
   namespace: ${NAMESPACE}
 spec:
-  imagePullPolicy: Always
+  imagePullPolicy: IfNotPresent
   versions:
     management:
       ui:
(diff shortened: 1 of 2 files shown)


rahulreddy:
There are other references like the zookeeper operator
```
          command:
          - zookeeper-operator
          imagePullPolicy: Always

```
Should I change this as well?

miniscruff:
> 
> 
> There are other references like the zookeeper operator
> 
> ```
>           command:
>           - zookeeper-operator
>           imagePullPolicy: Always
> ```
> 
> Should I change this as well?

I think so

NicolasT:
I'm a bit confused as to why we have to (ever) specify an `imagePullPolicy`. The default policy is `IfNotPresent`, *except* if the tag is `latest`, in which case it is `Always`. So, setting a (custom) `imagePullPolicy` (likely setting it to `Always` for non-`latest` tags) is a sign of a flawed delivery pipeline (note, I'm aware that's *not* what this PR is doing). Except if there's a security policy which dictates to use `Always` (which would then always verify the digest of the local image with the remote one, preventing image cache injection attacks).
So, instead of setting this to `IfNotPresent`, wouldn't it be better to rely on the (sensible) defaults and not specify anything at all?

rachedbenmustapha:
Already relies on the default value provided by zenko-operator which is "if not present"

PR #1125, opened 2021-02-01
Improvement/zenko 3164/update ci keycloak

alexanderchan-scality:
add keycloak to CI

ZENKO-3164

fixes #<ISSUE>


miniscruff (approved):

eve/main.yml
```diff
@@ -7,10 +7,16 @@ branches:
...
   - env: &kind-env
+      WORKER_COUNT: '2'
       OPERATOR_REPO: git@github.com:scality/zenko-operator.git
       OPERATOR_IMAGE: '%(prop:OPERATOR_IMAGE_NAME)s:%(prop:OPERATOR_IMAGE_TAG)s'
       KIND_NODE_IMAGE: "kindest/node:v1.17.5@sha256:ab3f9e6ec5ad8840eeb1f76c89bb7948c77bbf76bcebe1a8b59790b8ae9a283a"
       VOLUME_ROOT: /artifacts
+      OIDC_REALM: "zenko"
+      OIDC_CLIENT_ID: "zenko-ui"
+      OIDC_USERNAME: 'zenko-end2end'
+      OIDC_ENDPOINT: 'ui.zenko.local'
```
nicolas2bert: Can we call this endpoint: `UI_ENDPOINT` and set `OIDC_ENDPOINT:"keycloak.zenko.local"`?

eve/main.yml
```diff
@@ -176,42 +182,67 @@ stages:
...
     - ShellCommand:
         name: Deploy Zenko Instance
         env:
           <<: *kind-env
-        command: scripts/deploy-zenko.sh "end2end" ${VOLUME_ROOT} "default"
+        command: bash scripts/deploy-zenko.sh end2end default
+        workdir: build/eve/workers/end2end
+        haltOnFailure: true
+    - ShellCommand:
+        name: Add user
```
nicolas2bert: NIT: `Add keycloak user` ?

eve/main.yml
```diff
@@ -63,13 +72,13 @@ stages:
...
         property: launch_end2end
         command: eve/scripts/launch-end2end.sh
-    - TriggerStages:
-        name: trigger iso and documentation build
-        haltOnFailure: true
-        stage_names:
-        - build-doc
-        - build-iso
-        - build-base-iso
+    # - TriggerStages:
```
miniscruff: assume this is a temp thing but just want to comment

eve/main.yml
```diff
@@ -47,7 +56,7 @@ models:
   - SetProperties: &set_operator_properties
       properties:
         OPERATOR_IMAGE_NAME: registry.scality.com/sf-eng/zenko-operator
-        OPERATOR_IMAGE_TAG: 004087b
+        OPERATOR_IMAGE_TAG: '3879889'
```
miniscruff, line 59: We may want to pull this from deps.txt or the one source of truth later. Having to update this tag when the operator is updated will be hard.

eve/main.yml
miniscruff: apparently I took too long...

eve/main.yml
alexanderchan-scality: I had thought of using a floating tag like `latest`. Will probably discuss this in the `Zenko: how do we test?` meeting next Tuesday

nicolas2bert (approved):

anurag4DSB (approved):

eve/main.yml
rachedbenmustapha: The operator version that runs should be the exact one that would be embedded in the ISO if we built it directly from this exact tree. That version is available in the `solution/deps.txt` file. Otherwise the tests don't really reflect what would be built.

PR #1127, opened 2021-02-02
feature: ZENKO-3163 Update Zenko Version CRs

miniscruff:
* Update dependency versions
* Update operator CRDs

### DIFF
--- a/solution/zenkoversion.yaml
+++ b/solution/zenkoversion.yaml
@@ -8,7 +8,7 @@ spec:
   versions:
     management:
       ui:
-        image: ${SOLUTION_REGISTRY}/zenko-ui
+        image: zenko-ui
         tag: 0.1.0
       api:
         image: pensieve-api
@@ -18,10 +18,10 @@ spec:
         tag: 611b9ca
     cloudserver:
       image: cloudserver
-      tag: latest-8.1.2
+      tag: 8.2.8
     backbeat:
       image: backbeat
-      tag: 161b9c1
+      tag: 8.2.5
     blobserver:
       image: blobserver
       tag: 1.0.8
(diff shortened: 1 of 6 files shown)


eve/workers/end2end/scripts/deploy-zenko.sh
```diff
@@ -66,10 +66,10 @@ spec:
     management:
       ui:
         image: registry.scality.com/sf-eng/zenko-ui
-        tag: 'c091051'
+        tag: '0.1.0'
```
tcarmet: Here I think it's `image: registry.scality.com/zenko-ui/zenko-ui`

eve/workers/end2end/scripts/deploy-zenko.sh
```diff
@@ -65,11 +65,11 @@ spec:
   versions:
     management:
```
anurag4DSB: Do we need to add vault in this file as well?
I do not know how this works yet. But still checking

miniscruff: Vault is optional and has to be enabled which is not the case in the zenko CR below. Now if you think we should be using it here that is a different question that I am not really the one to answer :|

anurag4DSB: I might not have the complete context, clarifying just to learn more.
Are you saying Vault is optional in Zenko 2.0.0-alpha.X?
or 
Vault is optional in end2end tests?

miniscruff: > Are you saying Vault is optional in Zenko 2.0.0-alpha.X?

Currently yes, it must be enabled through the zenko CR.

> Vault is optional in end2end tests?

Also yes as currently, it seems to be disabled.

alexandre-merle:
should we wait for https://github.com/scality/cloudserver/pull/3312 as well ?

miniscruff:
> should we wait for [scality/cloudserver#3312](https://github.com/scality/cloudserver/pull/3312) as well ?

I am not sure I can decide if we should, it sounds like a good idea to me and only takes a few minutes to update here once the tag is made.

alexandre-merle:
It's merged, so if you want to update the tag, you can :)

miniscruff:
> It's merged, so if you want to update the tag, you can :)

Is there a new release or just a git hash? We are trying to use tags when they exist. ( a few projects don't have any tags yet )

solution/zenkoversion.yaml
```diff
@@ -33,25 +33,32 @@ spec:
...
-      tag: '1.2'
+      tag: '1.8.2'
     zookeeper:
       image: zookeeper
       tag: 0.2.4
     kafka:
-      image: kafka
-      tag: 2.12-2.3.0
+      cluster:
+        image: wurstmeister/kafka
```
alexanderchan-scality:
```suggestion
        image: kafka
```

alexanderchan-scality (approved):

naren-rajendran (approved):

PR #1131, opened 2021-02-06
ZENKO-3163: update pensieve api

alexanderchan-scality:
update pensieve-api image in zenko solutions
ZENKO-3163
fixes #<ISSUE>


alexanderchan-scality:
in favor of #1132

PR #1168, opened 2021-03-22
ZENKO-3304 fixup regex to match semver input on nightly release script

tcarmet:


alexanderchan-scality (approved):

eve/scripts/release-nightly.sh
```diff
@@ -33,7 +33,7 @@ HELP
 }
 
 parse_semver() {
-  local RE='[^0-9]*\([0-9]*\)\.\([0-9]*\)\.\([0-9]*\)[-+]\([0-9A-Za-z\-\.]*\)'
+  local RE='[^0-9]*\([0-9]*\)\.\([0-9]*\)\.\([0-9]*\)[-+]\?\([0-9A-Za-z\-\.]*\)'
```
jonathan-gramain, line 36: likely minor: it might be better to include the `[-+]\?` part inside an optional group matching the entire special suffix, and nesting another group to capture the special version:

```
\([-+]\([0-9A-Za-z\-\.]*\)\)\?
```

And then use `\5` (the nested parenthesis) instead of `\4` to retrieve the special value.

The minor benefit is to avoid capturing any letter/dash/dot in the patch version since the match would fail.

tcarmet: alright, I'll apply this change

saitharunthodupunuru (approved):

PR #1179, opened 2021-03-30
run end2end tests in series

alexanderchan-scality:
Modify the CI config to run the end-to-end tests in series; improves CI stability by only running a single zenko instance at a time. 3 parallel instances seems to be taxing the k8s environment, timing out on stabilization checks. 

fixes ZENKO-3290

### DIFF
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -211,12 +211,22 @@ stages:
         - build-doc
         - test-cosmos
     - TriggerStages:
-        name: trigger end2end tests
+        name: trigger python end2end tests
         haltOnFailure: true
         doStepIf: '%(prop:launch_end2end)s'
         stage_names:
         - python-tests
+    - TriggerStages:
+        name: trigger node 1 end2end tests
+        haltOnFailure: true
+        doStepIf: '%(prop:launch_end2end)s'
+        stage_names:
         - node-tests-01
+    - TriggerStages:
+        name: trigger node 2 end2end tests
+        haltOnFailure: true
+        doStepIf: '%(prop:launch_end2end)s'
+        stage_names:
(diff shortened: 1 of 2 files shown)


anurag4DSB (approved):

nicolas2bert (approved):

saitharunthodupunuru (approved):

eve/main.yml
```diff
@@ -211,12 +211,22 @@ stages:
...
+        name: trigger python end2end tests
         haltOnFailure: true
         doStepIf: '%(prop:launch_end2end)s'
         stage_names:
         - python-tests
+    - TriggerStages:
+        name: trigger node 1 end2end tests
+        haltOnFailure: true
+        doStepIf: '%(prop:launch_end2end)s'
+        stage_names:
```
tcarmet, line 223: By doing this we're installing and deleting a instance of Zenko three time. I believe we can install it once and run all the test on it.

That would mean we would only have one stage that we call something like `end2end` and let it run all the tests.

(I feel like this would save us around 20 minutes of time in the CI build). Should we try it?

tcarmet: Here's what I meant https://github.com/scality/Zenko/compare/development/1.2...feature/sequencial-testing

Let's see if it actually work before going towards my idea. Because you got a green build here so I don't want to introduce a "breaking change :D

tcarmet: for what is worth, it actually worked :tada: 
https://eve.devsca.com/github/scality/zenko/#/builders/4/builds/16399

the build lasted in total one hour

tcarmet (approved):
Left an open comment, but as discussed, the content of the PR as it is LGTM

PR #1190, opened 2021-04-04
bugfix: dedupe image info

rachedbenmustapha:
- add formal model to list image versions
- only store placeholders in zenkoversion.yaml
- fix missing mongodb image download


alexanderchan-scality (approved):
lgtm. can we update the e2e tests use the `zenkoversion` generation as well? or is that out of the scope of this PR?

rachedbenmustapha:
@alexanderchan-scality this is more of a hotfix (with root cause fix) than a refactor. For sure the e2e tests should use the artifacts generated by the iso build but my understanding is that it's a more involved planned effort.

jonathan-gramain (approved):

PR #1191, opened 2021-04-06
ZENKO-3329: update release script

alexanderchan-scality:
Update release script and Zenko Chart values

ZENKO-3329

### DIFF
--- a/eve/scripts/release
+++ b/eve/scripts/release
@@ -105,5 +105,14 @@ for chart in CHARTS:
         args.version, [zenko_version['minor'], patch], verbose=args.verbose)
     save_file(chart, chart_path)
     save_file(values, values_path)
+
 # update Zenko
-update_chart(open_file('%s/Chart.yaml' % args.path, verbose=args.verbose), None, args.version, None, verbose=args.verbose, dep=False)
+chart_path = '%s/Chart.yaml' % (args.path)
+values_path= '%s/values.yaml' % (args.path)
+chart, values = update_chart(open_file(chart_path, verbose=args.verbose),
+                             open_file(values_path, verbose=args.verbose),
+                             args.version,
+                             None,
+                             verbose=args.verbose,
+                             dep=False)
+save_file(chart, chart_path)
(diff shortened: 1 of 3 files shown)


eve/scripts/release
```diff
@@ -105,5 +105,16 @@ for chart in CHARTS:
...
+values_path= '%s/values.yaml' % (args.path)
+chart, values = update_chart(open_file(chart_path, verbose=args.verbose),
+                             open_file(values_path, verbose=args.verbose),
+                             args.version,
+                             None,
+                             verbose=args.verbose,
+                             dep=False)
+# only update Charts.yaml
+save_file(chart, chart_path)
+# save_file(values, values_path)
```
tcarmet: Why are we not saving the values file?

alexanderchan-scality: I had skipped that because `zenko/values.yaml` does not contain the root-level `image.*` fields. (It seems to be specifically written for updating the tags of dependencies in `zenko/charts/*`

tcarmet: alrighty

tcarmet: thanks for that info

tcarmet (approved):

eve/scripts/release
tcarmet: Now that I think about it a little bit. Should we remove the commented line then?

kubernetes/zenko/Chart.yaml
```diff
@@ -1,14 +1,14 @@
...
-  - https://github.com/Scality/Zenko
+- https://github.com/Scality/Zenko
 maintainers:
-  - name: rahulreddy
-  - name: tmacro
+- name: rahulreddy
+- name: tmacro
 engine: gotpl
-tillerVersion: ">=2.9.1"
+tillerVersion: '>=2.9.1'
```
saitharunthodupunuru, line 14: Just a curious question. Does this quote change required, does it matters?

saitharunthodupunuru (approved):
LGTM only needed clarification on quote changes

kubernetes/zenko/Chart.yaml
alexanderchan-scality: the changes are generated. also `"` vs `'` doesn't matter.

PR #1192, opened 2021-04-08
ZENKO-3289 implement kustomize for yaml generation

tcarmet:


solution/build.sh
```diff
@@ -82,7 +82,7 @@ function dependencies_versions_env()
 function copy_yamls()
 {
     cp -R -f operator/ ${ISO_ROOT}/operator
-    env $(dependencies_versions_env) envsubst < zenkoversion.yaml > ${ISO_ROOT}/zenkoversion.yaml
+    kustomize build -o ${ISO_ROOT}/zenkoversion.yaml
```
tcarmet, line 85: The output here is overall the same, but now we use kustomize to do it.

miniscruff (approved):

PR #1234, opened 2021-05-26
Changing helm release to stable 3.5.4 and new official repo location

mrepop:
Moving the helm binary location to the current stable release location at helm.sh as it's no longer being published to the GCP storage.googleapis.com bucket.  

Updated helm install version to stable 3.5.4 build.


jonathan-gramain (approved):

tmacro (approved):

mrepop:
Updated to allow for intermediary step between  helm2 and helm3 tillerless install to allow for end2end testing with and without tiller.

PR #1262, opened 2021-09-02
Feature/zenko 3586/add bitnami mongo db charts

alexanderchan-scality:

fixes ZENKO-3586

commit#1: makefile commit to fetch charts
commit#2: adding bitnami charts (no review needed?)
commit#3: modifies charts to be usable as pre-rendered yamls


solution-base/build.sh
```diff
@@ -138,6 +150,59 @@ function kafka_yamls()
...
+    echo merging mongodb yamls
+    CHART_PATH="$(dirname $0)/mongodb/charts/mongodb"
+    
+    echo "${CHART_PATH}/values-production.yaml" 
+    if [[ "$NODE_COUNT" == "3" ]]; then
+        helm template ${MONGODB_NAME} ${CHART_PATH} -n ${MONGODB_NAMESPACE} \
+            -f "${CHART_PATH}/values-production.yaml" \
+            --set "separateSecrets=true" \
+            --set 'existingUserSecret=$USER_SECRET_PLACEHOLDER' \
+            --set 'existingRootSecret=$ROOT_SECRET_PLACEHOLDER' \
```
alexanderchan-scality: could this placeholder possibly be substituted by the artesca-installer

solution-base/deps.txt
```diff
@@ -1,10 +1,6 @@
...
-kubedb/redis_exporter:v0.21.1
-kubedb/mongo:4.0-v1
-kubedb/mongo-tools:4.0-v1
-kubedb/busybox:latest
-kubedb/mongodb_exporter:v1.0.0
 pravega/zookeeper-operator:0.2.7
 banzaicloud/kafka-operator:v0.11.2
 gcr.io/kubebuilder/kube-rbac-proxy:v0.4.0
+bitnami/mongodb:4.0.14-debian-9-r24
+bitnami/minideb:stretch
```
rachedbenmustapha, line 5: Is there a versioned alternative? Floating tags don't allow for reproducible builds.

solution-base/build.sh
```diff
@@ -138,6 +150,59 @@ function kafka_yamls()
...
+            --set "volumePermission.image.repository=${MONGODB_INIT_IMAGE_NAME}" \
+            --set "volumePermission.image.tag=${MONGODB_INIT_IMAGE_TAG}" \
+            --set "metrics.image.registry=${MONGODB_REGISTRY}" \
+            --set "metrics.image.repository=${MONGODB_EXPORTER_IMAGE_NAME}" \
+            --set "metrics.image.tag=${MONGODB_EXPORTER_IMAGE_TAG}" \
+            --set "replicaSet.enabled=true" \
+            --set "replicaSet.pdb.minAvailable.secondary=1" \
+            --set "replicaSet.pdb.minAvailable.arbiter=0" \
+            --set "replicaSet.replicas.secondary=2" \
+            --set "replicaSet.replicas.arbiter=0" >> ${OUTPUT_PATH}
```
rachedbenmustapha: Are these the same values as were set by kubedb?

solution-base/build.sh
rachedbenmustapha: It seems in the upgrade case, the secret would already exist (although might need to be copied in case it's garbage collected), and in the fresh install case, the secret would need to be create by the installer.

solution-base/build.sh
```diff
@@ -34,6 +34,8 @@ KAFKA_OPERATOR_PATH=${ISO_ROOT}/deploy/kafka.yaml
 KUBEDB_OPERATOR_PATH=${ISO_ROOT}/deploy/kubedb.yaml
 KUBEDB_CATALOGS_PATH=${ISO_ROOT}/deploy/kubedb-catalogs.yaml
 ZOOKEEPER_OPERATOR_PATH=${ISO_ROOT}/deploy/zookeeper.yaml
+MONGODB_SINGLE_NODE_PATH=${ISO_ROOT}/deploy/mongodb-1-node.yaml
+MONGODB_THREE_NODE_PATH=${ISO_ROOT}/deploy/mongodb-3-node.yaml
```
rachedbenmustapha, line 38: How about just including the chart in the archive and deploying from the installer? Helm3 is tiller-less so the end result would be identical. The issue with number of nodes presets is that we'd need to keep updating each time there's a new configuration to support.

solution-base/deps.txt
alexanderchan-scality: there isn't a versioned alternative provided. what can be done is to create a fixed vesion to be referenced in  'registry.scality.com`

solution-base/build.sh
alexanderchan-scality: the default kubedb configuration for 3 nodes is: 1 primary + 2 secondaries.

solution-base/mongodb/charts/mongodb/templates/configmap.yaml
```diff
@@ -0,0 +1,14 @@
...
+  labels:
+    app: {{ template "mongodb.name" . }}
+    chart: {{ template "mongodb.chart" . }}
+    heritage: {{ .Release.Service }}
+    release: {{ .Release.Name }}
+  name: {{ template "mongodb.fullname" . }}
+data:
+  mongodb.conf: |-
+{{ toYaml .Values.configmap | indent 4 }}
+{{- end }}
```
naren-rajendran, line 14: Missing empty line

solution-base/mongodb/Makefile
```diff
@@ -0,0 +1,23 @@
+ROOT_DIR:=$(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
+CHART_DIR:="${ROOT_DIR}/charts"
+
+CHART_REPO:="https://charts.bitnami.com/bitnami"
+CHART_VERSION:="7.8.0"
+
+PATCH_DIR:="${ROOT_DIR}/patches"
+PATCH_FILES:="$(shell ls -d ${PATCH_DIR}/*)"
+
+HELM=helm
```
miniscruff, line 10: Curious what the reason for this is? Just in anticipation of changing later on?

miniscruff (approved):

solution-base/deps.txt
rachedbenmustapha: Please add a followup ticket for this

solution-base/mongodb/Makefile
rachedbenmustapha: Typically it would be `?=` to allow temporarily using another `helm` binary (with `make ... HELM=/my/other/helm`) than the default one in the path, to quickly test something or work around a bug without changing the makefile.

rachedbenmustapha (approved):
I'm anticipating more changes once we test actual migrations and fresh installs with this, but looks good as the baseline.

solution-base/mongodb/Makefile
miniscruff: gotcha, ok. That is more or less what I was thinking.

solution-base/deps.txt
alexanderchan-scality: created ticket: ZENKO-3614

solution-base/deps.txt
```diff
@@ -1,10 +1,6 @@
-kubedb/operator:v0.13.0-rc.0
-kubedb/redis:5.0.3-v1
-kubedb/redis_exporter:v0.21.1
-kubedb/mongo:4.0-v1
-kubedb/mongo-tools:4.0-v1
-kubedb/busybox:latest
-kubedb/mongodb_exporter:v1.0.0
```
gdemonet, line 7: Since these images are removed, shouldn't the KubeDB manifests be removed as well? Not that it would hurt keeping them around :slightly_smiling_face:

solution-base/deps.txt
gdemonet: Another approach you could use is to rely on the SHA256, instead of a version tag (which is not a reproducible entity either, FWIW).

It can be easily retrieved with:

```bash
MINIDEB_DIGEST=$(skopeo inspect docker://bitnami/minideb:stretch | jq -r '.Digest')
docker pull "bitnami/minideb@$MINIDEB_DIGEST"
```

solution-base/deps.txt
alexanderchan-scality: @gdemonet I had left the manifests in order to delete them with `kubectl delete -f`

gdemonet: Makes sense!

PR #1278, opened 2021-09-28
Update Zenko-operator and backbeat

nicolas2bert:
Update Zenko operator and backbeat.

### DIFF
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -68,7 +68,7 @@ models:
   - SetProperties: &set_operator_properties
       properties:
         OPERATOR_IMAGE_NAME: registry.scality.com/sf-eng/zenko-operator
-        OPERATOR_IMAGE_TAG: '1.0.2'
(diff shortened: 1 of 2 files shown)


eve/main.yml
```diff
@@ -68,7 +68,7 @@ models:
   - SetProperties: &set_operator_properties
       properties:
         OPERATOR_IMAGE_NAME: registry.scality.com/sf-eng/zenko-operator
```
miniscruff, line 70: we should switch to using a different namespace for zenko operator, maybe for the next release @tcarmet

miniscruff (approved):

nicolas2bert:
@bert-e reset

lavigne958 (approved):

eve/main.yml
miniscruff: after talking to Alex C there might be an issue given zenko operator is private while zenko is public, so where to place the image is an issue

nicolas2bert:
@bert-e approve

PR #1280, opened 2021-09-28
ZENKO-3620 docker upgrade now requires double quote on volumes

tcarmet:


nicolas2bert (approved):

eve/workers/end2end.yaml
```diff
@@ -4,6 +4,18 @@ kind: Pod
 metadata:
   name: "zenko-operator-worker-smoke"
 spec:
+  initContainers:
```
miniscruff, line 8: Do you think you could add some context to this, it looks pretty magical. Or maybe a follow up ticket to remove it when we can.

alexanderchan-scality (approved):

eve/workers/end2end.yaml
tcarmet: Just added the link to the incident

miniscruff (approved):

PR #1289, opened 2021-10-06
ZENKO-3629: use helm3 for CI tests

alexanderchan-scality:
use `helm3`

fixes ZENKO-3629

### DIFF
--- a/eve/workers/end2end/requirements.sh
+++ b/eve/workers/end2end/requirements.sh
@@ -1,8 +1,7 @@
 #!/bin/bash
 
-export HELM_VER=2.16.7
+export HELM_VER=3.5.4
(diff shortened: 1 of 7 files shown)


eve/workers/end2end/requirements.sh
```diff
@@ -1,9 +1,9 @@
 #!/bin/bash
 
 export HELM_VER=2.16.7
+export HELM3_VER=3.5.4
```
tcarmet: Any particular reason to not fully remove helm2?

alexanderchan-scality: will remove after confirming that the change works.

miniscruff (approved):

naren-rajendran (approved):

PR #1300, opened 2021-10-07
base/mongodb: Ensure datadir is not a mountpoint

gdemonet:
Mounting a filesystem to /bitnami/mongodb/data/db directly will lead to
initialization checks bailing out, since the datadir being non empty (a
"lost+found" dir always exists for recovery purposes, at the root of any
mounted Linux FS) means no initialization of the database and users
should be done.
This commit reset the value to the default in the original chart's
`values.yaml` file.

Fixes: ZENKO-3630

---

To do:

- [ ] Update migration docs from KubeDB (db path in the PVs changes)


alexanderchan-scality (commented):
change lgtm. waiting on migration guide step update

alexanderchan-scality (approved):

rahulreddy (approved):

PR #1301, opened 2021-10-07
ZENKO-3631: add bitnami mongodb init scripts

alexanderchan-scality:

the bitnami default init scripts creates the user in the non-admin db; this will require clients to authenticate against the non-admin db as opposed to the default db, admin, breaking compatibility with the zenko components.

fixes #<ISSUE>

### DIFF
--- a/solution-base/mongodb/charts/mongodb/templates/statefulset-primary-rs.yaml
+++ b/solution-base/mongodb/charts/mongodb/templates/statefulset-primary-rs.yaml
@@ -121,21 +121,23 @@ spec:
           - name: MONGODB_REPLICA_SET_NAME
             value: {{ .Values.replicaSet.name | quote }}
             {{- if .Values.replicaSet.useHostnames }}
+          - name: MONGODB_HOSTS
+            value: {{ template "mongodb.hosts". }}
           - name: MONGODB_ADVERTISED_HOSTNAME
             value: "$(MONGODB_POD_NAME).{{ template "mongodb.fullname" . }}-headless.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain }}"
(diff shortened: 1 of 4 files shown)


solution-base/mongodb/charts/mongodb/templates/statefulset-primary-rs.yaml
```diff
@@ -121,21 +121,23 @@ spec:
...
             value: {{ .Values.replicaSet.name | quote }}
             {{- if .Values.replicaSet.useHostnames }}
+          - name: MONGODB_HOSTS
+            value: {{ template "mongodb.hosts". }}
           - name: MONGODB_ADVERTISED_HOSTNAME
             value: "$(MONGODB_POD_NAME).{{ template "mongodb.fullname" . }}-headless.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain }}"
             {{- end }}
           {{- if .Values.usePassword }}
-          - name: MONGODB_USERNAME
+          - name: MONGODB_APP_USERNAME
```
alexanderchan-scality, line 130: renamed to avoid being picked up by the default init scripts but still be usable by the `create-user` script.

gdemonet:
This PR should target `development/2.0` I believe

alexanderchan-scality:
@gdemonet 
> This PR should target development/2.0 I believe

the Bitnami works should not have been in development/2.0, and the reverts has been done. 

blockers: 
~fixing bad propagation: https://github.com/scality/Zenko/pull/1312~

nicolas2bert (approved):

naren-rajendran (approved):

saitharunthodupunuru (approved):

PR #1362, opened 2021-10-28
Enable mongodb metrics capture by prometheus

francoisferrand:
Since we are using metalk8s' prometheus stack, we must set the
`metalk8s.scality.com/monitor: ''` label on the ServiceMonitor.

Issue: ZENKO-3691

### DIFF
--- a/solution-base/mongodb/charts/mongodb/custom-values.yaml
+++ b/solution-base/mongodb/charts/mongodb/custom-values.yaml
@@ -493,7 +493,8 @@ metrics:
 
     ## Used to pass Labels that are used by the Prometheus installed in your cluster to select Service Monitors to work with
     ## ref: https://github.com/coreos/prometheus-operator/blob/master/Documentation/api.md#prometheusspec
-    additionalLabels: {}
+    additionalLabels:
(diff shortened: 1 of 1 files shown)


anurag4DSB (approved):

solution-base/mongodb/charts/mongodb/custom-values.yaml
```diff
@@ -493,7 +493,7 @@ metrics:
 
     ## Used to pass Labels that are used by the Prometheus installed in your cluster to select Service Monitors to work with
     ## ref: https://github.com/coreos/prometheus-operator/blob/master/Documentation/api.md#prometheusspec
-    additionalLabels: {}
+    additionalLabels: { app.kubernetes.io/name: '' }
```
miniscruff: Could use the more yaml syntax, might need to quote the key not quite sure
```suggestion
    additionalLabels: 
        app.kubernetes.io/name: ''
```

francoisferrand: 👍

miniscruff (approved):

PR #1366, opened 2021-10-29
ZENKO-3700: fix creat-user script to fail with non-zero

alexanderchan-scality:

update create-user script to exit with non-zero on error.
currently, the script will only soft fail, printing an error message if the user cannot be provisioned; This would allow the deployment to complete without having provisioned the required user needed by the data services.
The non-zero exit is needed to stop the deployment process when the user creation init step fails.

fixes ZENKO-3700

### DIFF
--- a/eve/workers/end2end/scripts/configs/mongodb_init_scripts/create-app-user.sh
+++ b/eve/workers/end2end/scripts/configs/mongodb_init_scripts/create-app-user.sh
@@ -30,7 +30,10 @@ retry() {
         sleep 5
     done
 
-    [ $count -lt 10 ] || echo "Failed to create app user."
+    if [ $count -ge 10 ]; then
+        echo "Failed to create app user."
+        exit 1
+    fi
 }
 
 retry create_user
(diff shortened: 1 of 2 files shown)


miniscruff (approved):

naren-rajendran (approved):

eve/workers/end2end/scripts/configs/mongodb_init_scripts/create-app-user.sh
```diff
@@ -30,7 +30,12 @@ retry() {
         sleep 5
     done
 
-    [ $count -lt 10 ] || echo "Failed to create app user."
+    if [ $count -ge 10 ]; then
+        echo "Failed to create app user."
+        exit 1
+    fi
+
+    exit 0
```
francoisferrand: the 'exit 0' at the end of the function is not needed, and would prevent calling it multiple times in a row, like this:

```
retry setup
retry create_user
```

Exit'ing in case of error is nice, but I think we should let the script continue otherwise.

alexanderchan-scality: I can make this change as it make sense to allow multiple calls. Didn't catch this in the single-command use case.

francoisferrand (approved):

PR #1373, opened 2021-11-03
Documentation/zenko 3656/remove s3 endpoint

John-VanHook:

fixes #<ISSUE>


nloewensen (approved):
Commands for CLI actions would need to be reviewed by a developer (not me ;) )

cathydossantospinto (approved):

John-VanHook:
Niels said: 

`I think we should add an explanation here of what will happen if the location is still in use by a bucket. Honestly I have not tried but will the deletion fail or succeed?`

Tech Pubs: We have no way to test this. Can you look into it and come back with an answer?

PR #1375, opened 2021-11-05
Improvement/zenko 3661 setup dependency mocks for replication tests

naren-rajendran:
Enable backbeat replication end 2 end tests.


tests/zenko_tests/e2e-config.yaml
```diff
@@ -1,8 +1,25 @@
 accounts:
   - "zenko"
-endpoints: []
-locations: []
+endpoints:
+  - hostname: "http://aws-mock"
+    locationName: "awsbackendmismatch"
+  - hostname: "http://azure-mock/devstoreaccount1"
+    locationName: "azurebackendmismatch"
```
alexanderchan-scality: the endpoints seems incorrect.
`endpoints` is the list of hostnames the `s3api` can be reached by

the mock service endpoints should be created when they are deployed

nicolas2bert (approved):
Let's move away from the python test set up and use a JS swagger client instead. 
Let's add a comment referencing a JIRA ticket. 
This can be done after the "clean-up non-running tests" step.

naren-rajendran:
> Let's move away from the python test set up and use a JS swagger client instead. Let's add a comment referencing a JIRA ticket. This can be done after the "clean-up non-running tests" step.

Created this ticket to address this, https://scality.atlassian.net/browse/ZENKO-3732

miniscruff (approved):

naren-rajendran:
@bert-e reset

PR #1393, opened 2021-11-12
ZENKO-3727 ensure we read zenko-operator version from deps.yaml

tcarmet:


eve/main.yml
```diff
@@ -72,10 +72,14 @@ models:
...
         exit 1'
       haltOnFailure: true
-  - SetProperties: &set_operator_properties
-      properties:
-        OPERATOR_IMAGE_NAME: registry.scality.com/sf-eng/zenko-operator
-        OPERATOR_IMAGE_TAG: '1.1.6'
+  - SetPropertyFromCommand: &set_operator_image_name
+      name: Set operator image name
+      property: OPERATOR_IMAGE_NAME
+      command: yq '."zenko-operator".sourceRegistry'
```
rachedbenmustapha: I think it's  missing the input file name, `solution/deps.yaml`

tcarmet: Ah yes indeed, good catch, thank you!

rachedbenmustapha: from build.sh, the syntax is `yq eval '.zenko-operator.tag' deps.yaml` (yq had to be different from jq)

tcarmet: I don't know if there's two kind of yq but from the one I have installed here's the output I get:
```
 ➜  yq '."zenko-operator".image' solution/deps.yaml         
"zenko-operator"
 ➜  yq eval '.zenko-operator.image' solution/deps.yaml
usage: yq [-h] [--yaml-output] [--yaml-roundtrip] [--width WIDTH] [--indentless-lists] [--in-place] [--version] [jq_filter] [files ...]
yq: error: argument files: can't open '.zenko-operator.image': [Errno 2] No such file or directory: '.zenko-operator.image'
(gh-actions-exporter-Evj7LpRq-py3.9) ➜  zenko git:(bugfix/ZENKO-3727-read-zenko-operator-version-from-deps) ✗ 
```

eve/main.yml
```diff
@@ -72,10 +72,14 @@ models:
...
         exit 1'
       haltOnFailure: true
-  - SetProperties: &set_operator_properties
-      properties:
-        OPERATOR_IMAGE_NAME: registry.scality.com/sf-eng/zenko-operator
-        OPERATOR_IMAGE_TAG: '1.1.6'
+  - SetPropertyFromCommand: &set_operator_image_name
+      name: Set operator image name
+      property: OPERATOR_IMAGE_NAME
+      command: yq '."zenko-operator".sourceRegistry' solution/deps.yaml
```
alexanderchan-scality: missing the image name: 
```suggestion
      command: yq eval '."zenko-operator" | .sourceRegistry + "/" + .image' solution/deps.yaml
```

eve/main.yml
rachedbenmustapha: Ah now that you mention it, I think there's 2 different `yq`s. But it's strange that you're getting an error, as the `eval` command works in `solution/build.sh`

eve/main.yml
tcarmet: thanks for the catch, just added it.

eve/main.yml
tcarmet: The `yq` version I had on my computer (which is what I was using to test) was different from the one used on the `solution/build.sh`

Just updated the commands accordingly to suite with the version of yq we use in the CI. All good!

alexanderchan-scality (approved):

rachedbenmustapha (approved):

PR #1400, opened 2021-11-18
ZENKO-3731 - add open source redis dashboard

lavigne958:
User dashboard from redis exporter repository
[here](https://github.com/oliver006/redis_exporter/blob/master/contrib/grafana_prometheus_redis_dashboard.json)

used version: 1.29.0, matches the version of redis exporter used in the product.

added an extra variable to let the user select the namespace filter.

### DIFF
--- a/monitoring/dashboards/redis.json
+++ b/monitoring/dashboards/redis.json
@@ -588,7 +588,7 @@
                 {
                     "expr": "rate(redis_net_input_bytes_total{namespace=\"${namespace}\", job=\"${job}\"}[5m])",
                     "intervalFactor": 2,
-                    "legendFormat": "{{ input }}",
+                    "legendFormat": "input",
                     "refId": "A",
                     "step": 240
                 },
@@ -596,7 +596,7 @@
                     "expr": "rate(redis_net_output_bytes_total{namespace=\"${namespace}\", job=\"${job}\"}[5m])",
                     "interval": "",
                     "intervalFactor": 2,
-                    "legendFormat": "{{ output }}",
+                    "legendFormat": "output",
                     "refId": "B",
                     "step": 240
                 }
@@ -1085,7 +1085,8 @@
                     "expr": "redis_connected_clients{namespace=\"${namespace}\", job=\"${job}\"}",
                     "format": "time_series",
                     "intervalFactor": 1,
-                    "refId": "A"
+                    "refId": "A",
+                    "legendFormat": "clients"
                 }
             ],
             "thresholds": [],
@@ -1167,7 +1168,7 @@
     },
     "refresh": "30s",
     "schemaVersion": 12,
-    "version": 52,
+    "version": 1,
     "links": [],
     "gnetId": 37
 }
--- a/monitoring/dashboards/redis.json
+++ b/monitoring/dashboards/redis.json
@@ -1,0 +1,1173 @@
+{
+    "__inputs": [
+        {
+            "name": "DS_PROMETHEUS",
+            "label": "Prometheus",
+            "description": "",
+            "type": "datasource",
+            "pluginId": "prometheus",
+            "pluginName": "Prometheus"
+        },
+        {
+            "name": "namespace",
+            "type": "constant",
+            "label": "namespace",
+            "description": "Namespace associated with the Zenko instance",
+            "value": "zenko"
+        },
+        {
+            "name": "job",
+            "type": "constant",
+            "label": "job",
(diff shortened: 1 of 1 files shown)


monitoring/dashboards/redis.json
```diff
@@ -0,0 +1,1470 @@
...
+            "id": "singlestat",
+            "name": "Singlestat",
+            "version": ""
+        }
+    ],
+    "annotations": {
+        "list": [
+            {
+                "builtIn": 1,
+                "datasource": "-- Grafana --",
```
francoisferrand: should use DS_DATASOURCE as alert source

monitoring/dashboards/redis.json
```diff
@@ -0,0 +1,1470 @@
...
+            "6h",
+            "12h",
+            "24h",
+            "2d",
+            "7d",
+            "30d"
+        ]
+    },
+    "timezone": "browser",
+    "title": "Redis Dashboard for Prometheus Redis Exporter 1.x",
```
francoisferrand: we probably don't about the "for Prometheus Redis Exporter 1.x" : just `Redis` is likely enough

monitoring/dashboards/redis.json
```diff
@@ -0,0 +1,1470 @@
...
+            {
+                "allValue": null,
+                "current": {},
+                "datasource": "${DS_PROMETHEUS}",
+                "definition": "label_values(redis_up, instance)",
+                "hide": 0,
+                "includeAll": true,
+                "label": null,
+                "multi": true,
+                "name": "instance",
```
francoisferrand: Should be an __input

monitoring/dashboards/redis.json
```diff
@@ -0,0 +1,1470 @@
...
+                "current": {},
+                "datasource": "${DS_PROMETHEUS}",
+                "definition": "label_values(redis_up, namespace)",
+                "description": "identify running instance by namespace",
+                "error": null,
+                "hide": 0,
+                "includeAll": true,
+                "label": null,
+                "multi": false,
+                "name": "namespace",
```
francoisferrand: Should be an __input

monitoring/dashboards/redis.json
lavigne958: will do, I though we just put the dashboard, then we'll look at these details later when integrating them, but making them clean now is better you're right.

monitoring/dashboards/redis.json
lavigne958: I need to take a second look at this section to understand what it does.

monitoring/dashboards/redis.json
```diff
@@ -0,0 +1,1470 @@
...
+                "options": [],
+                "query": "label_values(redis_up, namespace)",
+                "refresh": 1,
+                "regex": "",
+                "skipUrlSync": false,
+                "sort": 1,
+                "type": "query"
+            }
+        ]
+    },
```
lavigne958: Will remove that section, this becomes a selector in the UI where the user can choose between the running redis server or we don't want this, we'll move it to the `__input` section and make it a constant value: the redis service name and the zenko namespace.

lavigne958:
forced push to create a new first commit with a better JSON file with the appropriate inputs, and not variables open to the user to select.

Set versioning to 1 to start versioning this file, set a couple of legends on some metrics instead of the raw labels.

monitoring/dashboards/redis.json
```diff
@@ -0,0 +1,1174 @@
...
+            ],
+            "yaxis": {
+                "align": false,
+                "alignLevel": null
+            }
+        },
+        {
+            "aliasColors": {
+                "evicts": "#890F02",
+                "memcached_items_evicted_total{instance=\"172.17.0.1:9150\",job=\"prometheus\"}": "#890F02",
```
francoisferrand: the "job" filter seems fishy, with the hard-coded `prometheus` value.
And the instance IP should definitly not be hard-coded

francoisferrand (changes requested):

monitoring/dashboards/redis.json
lavigne958: this is alias colors, I saw it somewhere else, I will remove the entire section `aliasColors` I think we don't need it and we don't care so much if red (#890F02) becomes some shade of grass green (#890F02) :laughing:

lavigne958:
Final result looks like this:

image
image

francoisferrand (approved):

XinLi0207 (approved):

PR #1401, opened 2021-11-19
Add dashboard for zookeeper (base-quorum)

francoisferrand:
Issue: ZENKO-3736

### DIFF
--- a/monitoring/dashboards/zookeeper.json
+++ b/monitoring/dashboards/zookeeper.json
@@ -170,7 +170,7 @@
           "instant": false,
           "interval": "",
           "intervalFactor": 2,
-          "legendFormat": "{{instance}} znode_count",
+          "legendFormat": "{{pod}} znode_count",
           "refId": "A"
         },
         {
@@ -179,7 +179,7 @@
           "instant": false,
           "interval": "",
           "intervalFactor": 2,
-          "legendFormat": "{{instance}} ephemerals",
+          "legendFormat": "{{pod}} ephemerals",
           "refId": "B"
         }
       ],
@@ -271,7 +271,7 @@
           "instant": false,
           "interval": "",
           "intervalFactor": 2,
-          "legendFormat": "{{instance}} znode_count_rate",
+          "legendFormat": "{{pod}} znode_count_rate",
           "refId": "A"
         },
         {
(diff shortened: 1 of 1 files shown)


monitoring/dashboards/zookeeper.json
```diff
@@ -0,0 +1,9030 @@
...
+        "current": {},
+        "datasource": "${DS_PROMETHEUS}",
+        "definition": "label_values({job=~\"${job}\",namespace=~\"${namespace}\"}, pod)",
+        "hide": 0,
+        "includeAll": true,
+        "label": "Instance",
+        "multi": true,
+        "name": "pod",
+        "options": [],
+        "query": "label_values({job=~\"${job}\",namespace=~\"${namespace}\"}, pod)",
```
lavigne958: se should not need to use regexp to filter namespace or job. we already know the exact name and don't use regexp for matching, we could simply use `job=\"${job}\"`

monitoring/dashboards/zookeeper.json
```diff
@@ -0,0 +1,9030 @@
...
+      "12h",
+      "24h",
+      "2d",
+      "7d",
+      "30d"
+    ]
+  },
+  "timezone": "",
+  "title": "ZooKeeper",
+  "uid": "SDE76m7Zzz",
```
lavigne958: I think you tooke the dashboard from grafana, I mean: you import a dashboard then you check it in grafana then you copy/paste the section "JSON model", but this contains some _runtime_ attributes, like the UID, which we don't want to set, I think it is generated by grafana when you import your dashboard.

monitoring/dashboards/zookeeper.json
francoisferrand: not for instances (e..g `pod`) though, otherwise viewing "All" instances does not work.

monitoring/dashboards/zookeeper.json
francoisferrand: indeed, uid should not be set, good catch! Should we just remove it? Or set it to `null` ?

I am wondering also what we should do about `version`, `revision` and `iteration`...

lavigne958: `version`: 1, we increment as we update it
`revision`: can be removed I suppose, we have not interest in this one
`iteration`: same no interest I think, can be removed if grafana does not require it.

lavigne958 (approved):

XinLi0207 (approved):

PR #1405, opened 2021-11-23
ZENKO-3737 - MongoDB dashboard

jbertran:
* Add a MongoDB dashboard
* Add a README for future contributors to dashboards
* Enforce JSON dashboard version increments in CI

NOTE: this PR introduces the expectations that the version of a dashboard is the number of commits that change it. If you get a failure in CI, this might be the culprit.

### DIFF
--- a/monitoring/dashboards/README.md
+++ b/monitoring/dashboards/README.md
@@ -1,0 +1,31 @@
+# Building new dashboards
+
+## From OSS
+
+1. Download the OSS JSON definition
+2. Import it in a deployed Grafana
+3. Tweak to your liking
+
+## From scratch
+
+1. Create a new dashboard in a deployed Grafana
+2. Create and tweak panels
+
+
+## Final steps
+
+1. Copy the JSON model
+2. Apply the following modifications:
+   * add the `__inputs` of the original download to the model
+   * add the `__requires` of the original download to the model
+   * move `id` from int to `null` in the model
+   * move `uid` from int to `null` in the model
+   * set the `version` of the model to `1`
+   * remove `iteration` entirely from the model
+3. Comnit in `monitoring/dashboards`
+
+
+# Modifying existing dashboards
+
+If a commit modifies a dashboard, it **must** increment the top-level
+`version` field in the dashboard. This is enforced by the CI.
--- a/monitoring/dashboards/README.md
+++ b/monitoring/dashboards/README.md
@@ -16,12 +16,15 @@
 
 1. Copy the JSON model
 2. Apply the following modifications:
-   * add the `__inputs` of the original download to the model
+   * add the `__inputs` of the original download to the model.
+     Inputs are named values populated at dashboard upload time, used to match
+     the dashboard with the deployment.
    * add the `__requires` of the original download to the model
    * move `id` from int to `null` in the model
    * move `uid` from int to `null` in the model
    * set the `version` of the model to `1`
    * remove `iteration` entirely from the model
+   * set `annotations` to `{"list": []}`
 3. Comnit in `monitoring/dashboards`
 
 
--- a/monitoring/dashboards/mongodb.json
+++ b/monitoring/dashboards/mongodb.json
@@ -49,7 +49,7 @@
       "version": "5.0.0"
     }
   ],
-  "description": "MongoDB Prometheus Exporter Dashboard. A fork inspired from https://grafana.com/grafana/dashboards/2583 which is made to work well with https://github.com/percona/mongodb_exporter.",
+  "description": "MongoDB Dashboard.",
   "editable": true,
   "gnetId": 12079,
   "graphTooltip": 1,
@@ -1832,5 +1832,5 @@
   },
   "timezone": "browser",
   "title": "MongoDB",
-  "version": 1
+  "version": 10
 }
--- a/monitoring/dashboards/mongodb.json
+++ b/monitoring/dashboards/mongodb.json
@@ -14,6 +14,13 @@
       "label": "namespace",
       "description": "Namespace associated with the Zenko instance",
       "value": "zenko"
+    },
+    {
+      "name": "job",
+      "type": "constant",
+      "label": "job",
+      "description": "Name of the MongoDB job, used to filter only the MongoDB instances.",
+      "value": "mongodb"
     }
   ],
   "__requires": [
@@ -42,10 +49,7 @@
       "version": "5.0.0"
     }
   ],
-  "annotations": {
-    "list": []
-  },
-  "description": "MongoDB Prometheus Exporter Dashboard.",
+  "description": "MongoDB Prometheus Exporter Dashboard. A fork inspired from https://grafana.com/grafana/dashboards/2583 which is made to work well with https://github.com/percona/mongodb_exporter.",
   "editable": true,
   "gnetId": 12079,
   "graphTooltip": 1,
@@ -68,7 +72,7 @@
       "id": 20,
       "panels": [],
       "repeat": "pod",
-      "title": "Query Metrics for $pod",
+      "title": "Query Metrics for ${pod}",
       "type": "row"
     },
     {
@@ -115,7 +119,7 @@
       "targets": [
         {
           "exemplar": true,
-            "expr": "rate(mongodb_op_counters_total{pod=~\"$pod\", namespace=\"${namespace}\"}[$interval])",
+          "expr": "rate(mongodb_op_counters_total{pod=\"${pod}\", namespace=\"${namespace}\", job=\"${job}\"}[10m])",
           "format": "time_series",
           "interval": "",
           "intervalFactor": 2,
@@ -220,7 +224,7 @@
       "targets": [
         {
           "exemplar": true,
-            "expr": "rate(mongodb_mongod_metrics_document_total{pod=~\"$pod\", namespace=\"${namespace}\"}[$interval])",
+          "expr": "rate(mongodb_mongod_metrics_document_total{pod=\"${pod}\", namespace=\"${namespace}\", job=\"${job}\"}[10m])",
           "format": "time_series",
           "interval": "",
           "intervalFactor": 2,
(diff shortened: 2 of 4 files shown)


monitoring/dashboards/README.md
```diff
@@ -0,0 +1,30 @@
...
+2. Create and tweak panels
+
+
+## Final steps
+
+1. Copy the JSON model
+2. Apply the following modifications:
+   * add the `__inputs` of the original download to the model
+   * add the `__requires` of the original download to the model
+   * move `id` from int to `null` in the model
```
lavigne958, line 23: I would add: `* remove "uid" entry`

monitoring/dashboards/mongodb.json
```diff
@@ -0,0 +1,1909 @@
...
+        },
+        "datasource": "Prometheus",
+        "definition": "label_values(mongodb_connections, pod)",
+        "description": null,
+        "error": null,
+        "hide": 0,
+        "includeAll": true,
+        "label": "env",
+        "multi": true,
+        "name": "env",
```
lavigne958: is this the right name to define the pod we want metrics from ?

monitoring/dashboards/mongodb.json
```diff
@@ -0,0 +1,1909 @@
...
+            "selected": false,
+            "text": "30d",
+            "value": "30d"
+          }
+        ],
+        "query": "1m,10m,30m,1h,6h,12h,1d,7d,14d,30d",
+        "refresh": 2,
+        "skipUrlSync": false,
+        "type": "interval"
+      }
```
lavigne958: Do we need this one  ? we already configured diferent time intervals and options bellow.

monitoring/dashboards/mongodb.json
```diff
@@ -0,0 +1,1909 @@
...
+      "12h",
+      "24h",
+      "2d",
+      "7d",
+      "30d"
+    ]
+  },
+  "timezone": "browser",
+  "title": "MongoDB",
+  "uid": "AyWQt9jWk",
```
lavigne958: this should be removed.

monitoring/dashboards/mongodb.json
```diff
@@ -0,0 +1,1909 @@
+{
+  "__inputs": [
```
lavigne958, line 2: This section is missing the `namespace` contant input. we want to enforce the filter on only mongodb in zenko namespace.

monitoring/dashboards/mongodb.json
```diff
@@ -0,0 +1,1909 @@
...
+        "builtIn": 1,
+        "datasource": "-- Grafana --",
+        "enable": true,
+        "hide": true,
+        "iconColor": "rgba(0, 211, 255, 1)",
+        "name": "Annotations & Alerts",
+        "type": "dashboard"
+      }
+    ]
+  },
```
lavigne958: I think this can be removed, it will be added by grafana automatically with the appropriate values.

monitoring/dashboards/mongodb.json
```diff
@@ -0,0 +1,1909 @@
...
+      "points": false,
+      "renderer": "flot",
+      "seriesOverrides": [],
+      "spaceLength": 10,
+      "stack": false,
+      "steppedLine": false,
+      "targets": [
+        {
+          "exemplar": true,
+          "expr": "rate(mongodb_op_counters_total{pod=~\"$env\", namespace=\"zenko\"}[$interval])",
```
lavigne958: I see, you should not set the namespace in the metrics but keep in the inputs, this is easier to change if we deploy zenko in a different namespace, we only change 1 occurrence.

monitoring/dashboards/README.md
jbertran: Is this OK? I remember we weren't sure about removing or leaving as is yesterday on Slack.

monitoring/dashboards/mongodb.json
jbertran: No, indeed, this should be `pod` I guess.

monitoring/dashboards/mongodb.json
jbertran: Fair. I'll update the readme to reflect it.

monitoring/dashboards/README.md
francoisferrand: I set it to `null`, like `id`...

monitoring/dashboards/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+1. Create a new dashboard in a deployed Grafana
+2. Create and tweak panels
+
+
+## Final steps
+
+1. Copy the JSON model
+2. Apply the following modifications:
+   * add the `__inputs` of the original download to the model
```
francoisferrand: would it be relevant to document what should (typically) be as `inputs` and what should be as `variables` ?
Something like :

* inputs should be used match the deployment: e.g. datasource, namespace & service name to filter on...
* variables should be used to select between nodes or instances of a services

monitoring/dashboards/mongodb.json
```diff
@@ -0,0 +1,1909 @@
...
+      "id": "singlestat",
+      "name": "Singlestat",
+      "version": "5.0.0"
+    }
+  ],
+  "annotations": {
+    "list": [
+      {
+        "builtIn": 1,
+        "datasource": "-- Grafana --",
```
francoisferrand: should be prometheus, since we use prometheus' alert manager alerts

monitoring/dashboards/mongodb.json
```diff
@@ -0,0 +1,1909 @@
...
+        "enable": true,
+        "hide": true,
+        "iconColor": "rgba(0, 211, 255, 1)",
+        "name": "Annotations & Alerts",
+        "type": "dashboard"
+      }
+    ]
+  },
+  "description": "MongoDB Prometheus Exporter Dashboard. A fork inspired from https://grafana.com/grafana/dashboards/2583 which is made to work well with https://github.com/percona/mongodb_exporter.",
```
francoisferrand: update description to something more end-user friendly ?

monitoring/dashboards/mongodb.json
```diff
@@ -0,0 +1,1909 @@
...
+  "editable": true,
+  "gnetId": 12079,
+  "graphTooltip": 1,
+  "id": null,
+  "links": [],
+  "panels": [
+    {
+      "collapsed": false,
+      "datasource": null,
```
francoisferrand: ${DS_PROMETHEUS}

monitoring/dashboards/mongodb.json
francoisferrand: we may also want to further filter by service (or similar field), so we only get mongodb pods here

monitoring/dashboards/mongodb.json
```diff
@@ -0,0 +1,1906 @@
...
+          "text": "All",
+          "value": "$__all"
+        },
+        "datasource": "Prometheus",
+        "definition": "label_values(mongodb_connections, pod)",
+        "description": null,
+        "error": null,
+        "hide": 0,
+        "includeAll": true,
+        "label": "pod",
```
francoisferrand: for the label (and for the label only!), I used "instance", which is less specific to the way it is implemented and will most likely speak to the user...

monitoring/dashboards/mongodb.json
```diff
@@ -0,0 +1,1906 @@
...
+          "refId": "StandardVariableQuery"
+        },
+        "refresh": 1,
+        "regex": "",
+        "skipUrlSync": false,
+        "sort": 1,
+        "type": "query",
+        "useTags": false
+      },
+      {
```
francoisferrand: what is this second variable?
i don't think i have seen it in other dashboards, and "interval" is managed by grafana anwyay.... maybe it is not required?

francoisferrand (changes requested):

monitoring/dashboards/mongodb.json
jbertran: We're removing it entirely and letting Grafana do it.

monitoring/dashboards/mongodb.json
```diff
@@ -0,0 +1,1906 @@
...
+      "repeat": "pod",
+      "title": "Query Metrics for $pod",
+      "type": "row"
+    },
+    {
+      "aliasColors": {},
+      "bars": false,
+      "dashLength": 10,
+      "dashes": false,
+      "datasource": "Prometheus",
```
francoisferrand: ${DS_PROMETHEUS}

(basically, any time the `datasource` is specified...)

monitoring/dashboards/README.md
lavigne958: I don't see any mentions of *UID* but only *ID* did you mixed them up perhaps ?

monitoring/dashboards/mongodb.json
lavigne958: it was a variable with a lot of different time range, they are then replaced in range vector like `rate(mongodb_metric[$interval])` this way you can adjust the precision of the metrics in live, but we don't need this.

lavigne958:
Here is the resulting dashboard:

image
image

lavigne958 (approved):

monitoring/dashboards/README.md
francoisferrand: nop, there are 2 fields. But maybe it depends on where the dashboard comes from, or the version...
(See lines 109 and 9028 in https://github.com/scality/Zenko/pull/1401)

francoisferrand:
@lavigne958 : One comment about the description was not adressed (maybe still on jordi's computer)... And you were bitten by the CI script that this commit introduces (thanks for testing it!), you do not have the right dashboard version 😀

lavigne958:
> 

I don't understand, I don't have the right version ? 
I took the dashboard left by Jordi and updated it regarding our comments.

About the description yes I will update it to something simpler like: "MongdoDB dashboard"

francoisferrand (approved):

PR #1406, opened 2021-11-23
Feature/zenko 3735 kafka dashboard

lavigne958:
This PR introduces 2 new dashboards for grafana

None, it is an improvement.


francoisferrand:
can we/should we merge both into a single Kafka dashboard ?

lavigne958:
> can we/should we merge both into a single Kafka dashboard ?

yes, but that's gonna be a BIG dashboard.

francoisferrand:
It may be a bit bigger, but the second dashboard is quite small, and there are duplicated metrics between both (jvm_memory_bytes_used for exemple, though this is not an exaustive list : just the first one i looked...). On top of that, it may be easier for user to know where to look (only one dashboard) and use (with the ability to view multiple metrics in correlation). To make the size manageable, there may be "sections" which can be collapsed as needed.

My opinion on this is that (generally) we should have separate dashboard only if this is required for 2 independent purposes, and avoid having too many dashboards which would IMHO bring more confusion than help... I am not sure if this is the case here, but you should definitly check with @ImaneSAniss and/or @anurag4DSB (or possibly some in-house kafka experts?

monitoring/dashboards/kafka_system.json
```diff
@@ -0,0 +1,588 @@
...
+    "list": []
+  },
+  "annotations": {
+    "list": []
+  },
+  "schemaVersion": 12,
+  "version": 1,
+  "links": [],
+  "gnetId": 721,
+  "description": "Kafka resource usage and throughput"
```
francoisferrand: throughput is not a "system" metrics (as in the title of diagram), and the associated metrics (`kafka_server_brokertopicmetrics_bytes*_total`) seems to be in the other dashboard as well.

lavigne958: done, I removed the first dashboard, in fact all its metrics are already in the second dashboard.

francoisferrand (approved):

miniscruff:
Can you include screenshots for dashboard PRs? Reviewing raw JSON dashboards can be quite hard.

lavigne958:
> Can you include screenshots for dashboard PRs? Reviewing raw JSON dashboards can be quite hard.

Sure here it is.

For the context we discussed the following points for the dashboards, you can see the README in[ jordi's PR](https://github.com/scality/Zenko/pull/1405).

The screenshots:
image
image
image
image
image
image
image
image
image

Cuervino (approved):

PR #1410, opened 2021-11-30
ZENKO-3748 - include third party dashboards in ISO

lavigne958:
Include the third party dashboards that are located in
`monitoring/dashboards` in the ISO file.

This is the first step of introducing the deployment of dashboards.

The current dashboards won't be deployed as the tool
necessary to prepare the file `registry-config.inc.j2`
must be updated.


solution/build.sh
```diff
@@ -97,6 +97,72 @@ function copy_image()
...
+    "config": {
+        "mediaType": "application/vnd.oci.image.config.v1+json",
+        "digest": "sha256:${config_sum}",
+        "size": 2
+    },
+    "layers": [
+        {
+            "mediaType": "application/grafana-dashboard+json",
+            "digest": "sha256:${digest}",
+            "size": ${size},
```
francoisferrand: i would suggest introducing a function which generates this output from an input stream, computes the sha256sum, renames the file and generates this block:

```
# $1 : mime-type of the file
# $2 : optionally, name of the file
# input : the content of the file
# output : the manifest fragment defining the file
function processFile {
   local tmp=$(mktemp xxxxx)
   cat > $tmp
    digest=$(sha256sum  ${tmp} | cut -d " " -f 1)     # get sha256, sha256sum prints the checksum and the filename, keep the checksum only
    size=$(stat --printf "%s" ${tmp})     # get only the size (in bytes) of the file. format "%s" only prints the size
    mv $tmp $digest
    
    echo "mediaType": "$1",
    echo "digest": "sha256:${digest}",
    echo "size": ${size},
    [ "$2" ] && echo "\"annotations\": { \"org.opencontainers.image.title\": \"$2\" }"
}
```

It can then be used both the the config and the dashboard:

```
$(processFile "application/vnd.oci.image.config.v1+json" <<< '{}')
$(processFile "application/grafana-dashboard+json" "$filename" < "$file")
```

(note that this will become more useful if/when we have multiple dashboards in the same image)

solution/build.sh
```diff
@@ -97,6 +97,72 @@ function copy_image()
...
+    dashboard_base_dir=${IMAGES_ROOT}/${filename%.json}-dashboard/${VERSION}/
+    mkdir -p ${dashboard_base_dir}
+
+    # get sha256, sha256sum prints the checksum and the filename, keep the checksum only
+    digest=$(sha256sum  ${file} | cut -d " " -f 1)
+    # get only the size (in bytes) of the file. format "%s" only prints the size
+    size=$(stat --printf "%s" ${file})
+
+    # generate empty config file
+    config_file=${dashboard_base_dir}/config.json
```
francoisferrand, line 120: should not "blindly" a (temp) file in the source code, it may conflict with an existing file (we never know...) : use mktemp instead

francoisferrand:
also, i'm wondering why you need to change the version on each "renamed" dashboard, except on kafka... should be the same behavior for all 🧐

francoisferrand (approved):

eve/main.yml
```diff
@@ -2,7 +2,7 @@
 version: "0.2"
 
 branches:
-  documentation/*, feature/*, improvement/*, bugfix/*, w/*, q/*, hotfix/*:
+  documentation/*, feature/*, improvement/*, bugfix/*, w/*, q/*, hotfix/*, user/*:
```
francoisferrand, line 5: should this really be commited ?

lavigne958: of course it should, we forgot to have the `user/*` branches covered in this repo. I needed to add it each time I needed to run the cI. :laughing:

lavigne958:
> 

I have no clue, but the script that checks the version of the dashboards says it should be version 6. so I just put the expected number :lab_coat:

solution/build.sh
lavigne958: I thought I'll do this later, for now we don't have any dashboards for others than third parties so once we are clear on the number of dashboards, if we have multiple dashboards, if we need to etc we can re-work the function that produces the manifest file. for now we don't know exactly what we are gonna need.

miniscruff (approved):

solution/build.sh
lavigne958: done

lavigne958:
Updated the build.sh to use the proposal from Francois.
Made a function to produce a single layer of an image.

PR #1425, opened 2021-12-21
bugfix: ZENKO-3070 Set pull policy to if not present for busybox

miniscruff:
Create a new variable for the pull policy to set for busy box. Note that the default pull policy is always so one needs to be added.

### DIFF
--- a/kubernetes/zenko/charts/s3-data/templates/deployment.yaml
+++ b/kubernetes/zenko/charts/s3-data/templates/deployment.yaml
@@ -22,6 +22,7 @@ spec:
       initContainers:
         - name: {{ .Chart.Name }}-init
           image: busybox
+          imagePullPolicy: {{ .Values.image.initPullPolicy }}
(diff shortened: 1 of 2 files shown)


rachedbenmustapha (approved):

NicolasT:
It's not entirely true that the default is `Always`. The default is `Always` if no image tag is provided, or the tag is `latest`. If a non-`latest` image tag *is* provided, the default pull policy is `IfNotPresent` ([source](https://kubernetes.io/docs/concepts/containers/images/#imagepullpolicy-defaulting)).

One could argue the actual bug here is we're using an image without a specific tag.

miniscruff:
> It's not entirely true that the default is `Always`. The default is `Always` if no image tag is provided, or the tag is `latest`. If a non-`latest` image tag _is_ provided, the default pull policy is `IfNotPresent` ([source](https://kubernetes.io/docs/concepts/containers/images/#imagepullpolicy-defaulting)).
> 
> One could argue the actual bug here is we're using an image without a specific tag.

True, I guess I was more leaning towards "the default in this particular instance" rather than the only possible instance.

nicolas2bert (approved):

PR #1430, opened 2021-12-22
zenko-3747 bump cloudserver version, embed dashboards

lavigne958:
- Bump cloudserver version
- Using new version, deploy dashboard
- In order to deploy dashboard use skopeo to pull dashboard from registry and deploy it as OCI image
- update skopeo version to handle OCI image
- upgrade the `build-iso` dockerfile to use `curl` only (remove use of `wget`).
- sort the list of dependencies so it is easier to find an entry by looking at it.


ghost (approved):
LGTM :smile:

lavigne958:
rebased to solve conflicts.

jbertran (approved):

PR #1439, opened 2021-12-30
Cleanup & improve 3rd party component's dashboards

francoisferrand:
Issue: ZENKO-3799

Screenshot 2021-12-30 at 23 03 31
Screenshot 2022-01-03 at 14 49 46
Screenshot 2021-12-30 at 23 04 40

### DIFF
--- a/eve/workers/end2end/scripts/configs/zenkoversion.yaml
+++ b/eve/workers/end2end/scripts/configs/zenkoversion.yaml
@@ -11,19 +11,19 @@ spec:
     #   tag: '${CLOUDSERVER_TAG}'
     # commented out to unblock k8s version testing
     # kafka:
-    #   image: kafka_grafana_dashboard
+    #   image: kafka-dashboard
     #   tag: '${ZENKO_VERSION_NAME}'
     # mongodb:
-    #   image: mongodb_grafana_dashboard
+    #   image: mongodb-dashboard
     #   tag: '${ZENKO_VERSION_NAME}'
     # redis:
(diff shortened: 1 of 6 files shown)


miniscruff:
Small thing here but i would split hits and misses. Because they will likely have very different values it can be hard to see any significant changes in the smaller one.

monitoring/dashboards/redis.json
```diff
@@ -61,6 +61,66 @@
...
+                    "values": false
+                },
+                "text": {},
+                "textMode": "auto"
+            },
+            "pluginVersion": "8.0.6",
+            "targets": [
+                {
+                    "exemplar": true,
+                    "expr": "sum(up{namespace=\"zenko\", job=\"artesca-data-base-cache-metrics\"})",
```
lavigne958: In this file the label `job` is an input value, we should replace the constant string with the variable `${job}`.

monitoring/dashboards/redis.json
```diff
@@ -148,6 +208,66 @@
...
+                "values": false
+                },
+                "text": {},
+                "textMode": "auto"
+            },
+            "pluginVersion": "8.0.6",
+            "targets": [
+                {
+                "exemplar": true,
+                "expr": "sum(redis_db_keys{namespace=\"zenko\", job=\"artesca-data-base-cache-metrics\"})",
```
lavigne958: same here we should replace the constant for job with the variable `${job}`.

lavigne958 (approved):

jbertran (approved):

PR #1472, opened 2022-01-14
Feature/zenko 3786/add sharded bitnami charts

alexanderchan-scality:

fixes #<ISSUE>


solution-base/build.sh
```diff
@@ -50,15 +50,23 @@ MONGODB_NAME="mongodb"
...
+MONGODB_SHARDED_IMAGE_TAG=$(yq eval ".mongodb-sharded.tag" deps.yaml)
+MONGODB_SHARDED_EXPORTER_IMAGE_NAME="mongodb-exporter"
+MONGODB_SHARDED_EXPORTER_IMAGE_TAG=$(yq eval ".mongodb-sharded-exporter.tag" deps.yaml)
+MONGODB_SHARDED_SHELL_IMAGE_NAME="bitnami-shell"
+MONGODB_SHARDED_SHELL_IMAGE_TAG=$(yq eval ".mongodb-sharded-shell.tag" deps.yaml)
 MONGODB_STORAGE_CLASS="MONGODB_STORAGE_CLASS"
 
-# grab our dependencies from our deps.txt file as an array
-readarray -t DEP_IMAGES < deps.txt
+function flatten_source_images()
```
rachedbenmustapha, line 68: This is a bit too non-trivial to be copy-pasted.

eve/workers/end2end/scripts/install-kind-dependencies.sh
```diff
@@ -74,30 +85,73 @@ stringData:
...
+        --set "replicaSet.pdb.minAvailable.arbiter=0" \
+        --set "replicaSet.replicas.secondary=0" \
+        --set "replicaSet.replicas.arbiter=0"
+
+    kubectl rollout status statefulset dev-db-mongodb-primary
+    kubectl rollout status statefulset dev-db-mongodb-secondary
+}
+
+mongodb_sharded() {
+    local CHART_PATH="$(dirname $0)/../../../../solution-base/mongodb/charts/mongodb-sharded"
```
rachedbenmustapha: Any way to use the the manifest generated for the solution instead, to ensure it works?

alexanderchan-scality: that may be a bit difficult as the manifest generated for the solution uses the metalk8s registry and placeholder.

solution-base/build.sh
alexanderchan-scality: I am not sure that I understand the issue here.
the solution-base `deps` file has been modified to be similar to that of the solution minus the `ENVSUBST` names

eve/workers/end2end/scripts/install-kind-dependencies.sh
rachedbenmustapha: If the only difference is the image (or any other param that kubectl can patch), it can be easily patched prior to applying with something like this:
```shell
echo "resources: [ $GENERATED_YAML_FILE ]" > kustomization.yaml
kustomize edit set image mongodb=$(get full image source with tag from deps.yaml) mongodb-exporter=$(...) mongodb-minideb=$(...) ...
kubectl apply -k .
```

rachedbenmustapha (approved):

naren-rajendran (approved):

alexanderchan-scality:
ping

PR #1487, opened 2022-02-07
ZENKO-3911 - fix prometheus alert rule syntax

jbertran:
**What does this PR do, and why do we need it?** Syntax errors in alert rules rendered. The checking tool is not merged yet because I want to make it a GH action, but these corrections need to be merged ASAP

fixes ZENKO-3911

### DIFF
--- a/monitoring/kafka/alerts.yaml
+++ b/monitoring/kafka/alerts.yaml
@@ -105,7 +105,7 @@ groups:
       summary: 'Zookeeper Sync Disconected'
       description: 'Kafka Zookeeper Sync Disconected'
 
-  - alert: ConsummerLagWarning
+  - alert: ConsumerLagWarning
     expr: |
       kafka_consumergroup_group_max_lag{namespace="${namespace}",cluster_name="${cluster}",group!=""}
           > ${maxConsumerLagMessagesWarningThreshold}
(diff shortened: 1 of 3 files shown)


lavigne958 (approved):

monitoring/kafka/alerts.yaml
```diff
@@ -121,7 +121,7 @@ groups:
         Kafka consumer lag has been more more than ${maxConsumerLagSecondsWarningThreshold} seconds
         or more than ${maxConsumerLagMessagesWarningThreshold} messages for 5 minutes.
 
-        Current time lag is {{ with printf 'kafka_consumergroup_group_max_lag_seconds{namespace="${namespace}",cluster_name="${cluster}",group="%s"}'
+        Current time lag is {{ with printf "kafka_consumergroup_group_max_lag_seconds{namespace='${namespace}',cluster_name='${cluster}',group='%s'}"
```
francoisferrand: outer double-quote is needed because we are in a Go template (so it is a string).
according to proemtheus docs, single and double quotes are the same : so we can indeed use the inner single quotes ; but it kind of looks different than all the other prometheus expressions in this file (and mostly everywhere else... So not sure if we should strive for uniformity or simplicity :-/

francoisferrand (approved):

monitoring/kafka/alerts.yaml
jbertran: Escaped the inner quotes for simplicity's sake

PR #1498, opened 2022-02-11
ZENKO-3923 Adding e2e tests for iam policies on iam user

KillianG:
e2e tests for iam policies over metadatasearch
We are only testing if call is authorized or not with the policy.
The real ft_tests are already tested in cloudserver

We won't add our tests in end2end_https is because we can see in Zenko main.yaml, the only difference between HTTP tests and HTTPS test is if keycloak is HTTP or not, we already have keycloak https tests in Vault. And the main objective here is to test cloudserver and vault interact correctly with each other, so we don't need the whole test to run again in https mode.

### DIFF
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -374,6 +374,13 @@ stages:
         command: bash scripts/run-e2e-test.sh "end2end" ${E2E_IMAGE_NAME}:${E2E_IMAGE_TAG} "end2end" "default"
         workdir: build/eve/workers/end2end
         haltOnFailure: true
+    - ShellCommand: &run_iam_policies_test
+        name: Run iam policies tests
+        env:
+          <<: [ *kind-env, *e2e-env, *http-env, *secrets-env ]
+        command: bash scripts/run-e2e-test.sh "end2end" ${E2E_IMAGE_NAME}:${E2E_IMAGE_TAG} "iam-policies" "default"
+        workdir: build/eve/workers/end2end
+        haltOnFailure: true
     - ShellCommand: &run_smoke_test
         name: Run smoke tests
         env:
@@ -456,6 +463,10 @@ stages:
         <<: *configure_end2end_tests
         env:
           <<: [*kind-env, *e2e-env, *https-env, *secrets-env]
+    - ShellCommand:
+        <<: *run_iam_policies_test
+        env:
+          <<: [ *kind-env, *e2e-env, *http-env, *secrets-env ]
     - ShellCommand:
         <<: *run_end2end_tests
         env:
@@ -568,6 +579,10 @@ stages:
         <<: *configure_end2end_tests
         env:
           <<: [*kind-env, *e2e-env, *http-env, *secrets-env]
+    - ShellCommand:
+        <<: *run_iam_policies_test
+        env:
+          <<: [ *kind-env, *e2e-env, *http-env, *secrets-env ]
     - ShellCommand:
         <<: *run_end2end_tests
         env:
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -464,10 +464,6 @@ stages:
         env:
           <<: [*kind-env, *e2e-env, *https-env, *secrets-env]
     - ShellCommand:
-        <<: *run_iam_policies_test
-        env:
-          <<: [ *kind-env, *e2e-env, *http-env, *secrets-env ]
-    - ShellCommand:
         <<: *run_end2end_tests
         env:
           <<: [*kind-env, *e2e-env, *https-env, *secrets-env]
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -576,10 +576,6 @@ stages:
         env:
           <<: [*kind-env, *e2e-env, *http-env, *secrets-env]
     - ShellCommand:
-        <<: *run_iam_policies_test
-        env:
-          <<: [ *kind-env, *e2e-env, *http-env, *secrets-env ]
-    - ShellCommand:
         <<: *run_end2end_tests
         env:
           <<: [*kind-env, *e2e-env, *http-env, *secrets-env]
--- a/tests/zenko_tests/node_tests/iam_policies/cloudserver/IAMUser.js
+++ b/tests/zenko_tests/node_tests/iam_policies/cloudserver/IAMUser.js
@@ -6,7 +6,7 @@ const {makeGETRequest} = require("../../utils/request");
 const iam = require('../../s3SDK').scalityIAMClient;
 const s3 = require('../../s3SDK').scalityS3Client;
 
-const userName = "username-policy-test-metadatasearch";
+const userName = "username-policy-test-metadatasearch2";
 
 const bucketName = "bucket-test";
 const listBucketPolicyName = "listBucketPolicy";
--- a/tests/zenko_tests/node_tests/iam_policies/cloudserver/IAMUser.js
+++ b/tests/zenko_tests/node_tests/iam_policies/cloudserver/IAMUser.js
@@ -1,64 +1,62 @@
 const assert = require('assert');
 const async = require('async');
-const uuidV4 = require('uuid/v4');
-const {makeGETRequest, getResponseBody} = require("../../utils/request");
-const {stat} = require("fs");
 const { errors } = require('arsenal');
+const { makeGETRequest, getResponseBody } = require('../../utils/request');
 
 const iam = require('../../s3SDK').scalityIAMClient;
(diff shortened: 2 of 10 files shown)


tests/zenko_tests/node_tests/utils/request.js
```diff
@@ -36,12 +36,12 @@ function getResponseBody(res, cb) {
...
  * @return {undefined}
  */
-function makeGETRequest(path, cb) {
+function makeGETRequest(path, cb, userCredentials) {
     let options = Object.assign({}, defaultOptions, {
         method: 'GET',
         path,
     });
-    options = aws4.sign(options, credentials);
+    options = aws4.sign(options, userCredentials ? userCredentials : credentials);
```
ghost: `userCredentials || credentials` ?

tests/zenko_tests/node_tests/iam_policies/cloudserver/IAMUser.js
```diff
@@ -0,0 +1,257 @@
...
+            next => iam.deleteUser({UserName: userName}, next),
+            next => s3.deleteBucket({Bucket: bucketName}, next),
+            next => s3.deleteBucket({Bucket: bucketName2}, next),
+        ], err => {
+            assert.ifError(err);
+            done();
+        });
+    })
+
+    it('should not be able to metadata search without the policy s3:MetadataSearch', done => {
```
ghost: Nitpicking : permission instead of policy

tests/zenko_tests/node_tests/iam_policies/cloudserver/IAMUser.js
```diff
@@ -0,0 +1,257 @@
...
+        return metadataSearchResponseCode(userCredentials, bucketName, (err, res) => {
+            if (err) {
+                assert.ifError(err)
+                return done(err)
+            }
+            assert.strictEqual(res, 403);
```
ghost: Do you think we can also check if the result is AccessDenied? So we don't only check the http error code (because the bucket ACLs can also lead to 403 responses, maybe?)

ghost: Also asking because this work is great and will help adding more tests to verify that cloudServer and vault are working well together (e.g. we miss tests with S3+seeded roles and with IAM users)

tests/zenko_tests/node_tests/iam_policies/cloudserver/IAMUser.js
```diff
@@ -0,0 +1,257 @@
...
+                if (err)
+                    return next(err);
+                return iam.attachUserPolicy({UserName: userName, PolicyArn: res.Policy.Arn}, next);
+            }),
+            next => metadataSearchResponseCode(userCredentials, bucketName, (err, res) => {
+                    if (err) {
+                        assert.ifError(err)
+                        return next(err)
+                    }
+                    assert.strictEqual(res, 200);
```
ghost: Can we also check here if we get the right information from the bucket? To make sure having an allowed permission to perform MD search is not affecting the behaviour of the API.

tests/zenko_tests/node_tests/iam_policies/cloudserver/IAMUser.js
```diff
@@ -0,0 +1,257 @@
...
+                    return next();
+                }),
+            next => iam.listPolicies({}, (err, res) => {
+                    if (err) {
+                        return next(err);
+                    }
+                    const p = res.Policies.find(p => p.PolicyName === allowMdSearchPolicyName);
+                    if (p)
+                        return iam.detachUserPolicy({UserName: userName, PolicyArn: p.Arn}, (err, res) => {
+                            if (err)
```
ghost: Use brackets please 🙂

tests/zenko_tests/node_tests/iam_policies/cloudserver/IAMUser.js
```diff
@@ -0,0 +1,257 @@
+const assert = require('assert');
```
ghost, line 1: Global comment: can we have a linter for these JS files? We can add an extra step in eve as well. This will help us write consistent code.

ghost (changes requested):
Mostly nitpicking, feel free to let me know if you disagree with some comments!

tests/zenko_tests/node_tests/utils/request.js
KillianG: this would return true or false
But I assume you wanted to say `userCredentials | credentials`, I will change this :)

tests/zenko_tests/node_tests/iam_policies/cloudserver/IAMUser.js
KillianG: Yeah sure I will take a look if it is possible to be more precise (for sure it is)

tests/zenko_tests/node_tests/iam_policies/cloudserver/IAMUser.js
KillianG: Actually, the ft_tests are already done in cloudserver, I don't think we need to double check this, here we just want to check if the policy give the permissions to bucket search

tests/zenko_tests/node_tests/iam_policies/cloudserver/IAMUser.js
KillianG: I will take a look into this yeah

tests/zenko_tests/node_tests/utils/request.js
ghost: If the first variable is null or undefined, it will return the second variable:

```
undefined || 'test'
> 'test'
```

tests/zenko_tests/node_tests/iam_policies/cloudserver/IAMUser.js
```diff
@@ -0,0 +1,257 @@
...
+            })
+    });
+
+    it('should be able to metadata search with the policy s3:MetadataSearch allowed' +
+        'on specific bucket but not on others', done => {
+        async.series([
+                next => iam.createPolicy({PolicyName: allowMdSearchPolicyName, PolicyDocument: createPolicy("s3:MetadataSearch", true, `arn:aws:s3:::${bucketName}`)}, (err, res) => {
+                    if (err)
+                        return next(err);
+                    return iam.attachUserPolicy({UserName: userName, PolicyArn: res.Policy.Arn}, next);
```
XinLi0207: I would suggest creating a null variable `policyArn` outside of async.series() and assign `res.Policy.Arn` to it here.
Then in line 142, we don't need to `listPolicies`, we can directly do the `detachUserPolicy` and `deletePolicy` using the `policyArn` saved before

eve/main.yml
```diff
@@ -374,6 +374,20 @@ stages:
...
         workdir: build/eve/workers/end2end
         haltOnFailure: true
+    - ShellCommand: &run_iam_policies_test
+        name: Run iam policies tests
+        env:
+          <<: [ *kind-env, *e2e-env, *http-env, *secrets-env ]
+        command: bash scripts/run-e2e-test.sh "end2end" ${E2E_IMAGE_NAME}:${E2E_IMAGE_TAG} "iam-policies" "default"
+        workdir: build/eve/workers/end2end
+        haltOnFailure: true
+    - ShellCommand: &run_lint_test
```
XinLi0207: Can we put the `run_lint_test`right after `git_pull`, so that we don't need to execute the following if there is any lint issue? like what we did in Vault's main.yaml https://github.com/scality/Vault/blob/development/8.4/eve/main.yml#L89

XinLi0207 (approved):
LGTM

eve/main.yml
```diff
@@ -568,6 +575,10 @@ stages:
         <<: *configure_end2end_tests
         env:
           <<: [*kind-env, *e2e-env, *http-env, *secrets-env]
+    - ShellCommand:
```
XinLi0207: oh here, sorry forget to mention, as discussed with William, there is also no need to test in sharded mongo. If our tests work in mongo, then they should work in sharded mongo. Sharded mongo is working or not is already tested by other tests.

XinLi0207 (changes requested):

eve/main.yml
KillianG: Sure I will remove

XinLi0207 (approved):

anurag4DSB (approved):

ghost (approved):

PR #1491, opened 2022-02-10
Improvement/zenko 3924 add e2e iam policies test assume role

XinLi0207:
context:
the current PR is based on [this PR](https://github.com/scality/Zenko/pull/1498)
we created 3 test files under node_tests/iam-policies/cloudserver: one for iam user, one for assumeRole session users, one for assumeRoleWWI session users. Later for all e2e tests between cloudserver and vault, we can put them here.

commit1: 
add some necessary env variables in `run_e2e_test.sh`

commit2:
implement e2e tests for metadataSearch API:
for testing, we will have two account, we want to test the user under account2 should be able to perform mdsearch on buckets of account1 after assuming a role with attached mdsearch policy under account1, etc.
to achieve this, first of all we need to :
- create account1, generateAccountAccessKey for it and get account1's iam client
- create account2, generateAccountAccessKey for it, get account2's iam client
- create a user under account2, create access key for the user and get the user's sts client
- account2's iam client creates allowAssumeRole policy and attach it to the user created above
- get account1's s3 client, create 2 buckets under account1

then, for each test, create a Role and a policy respectively, attach them then assumeRole, then using temporary credentials to perform mdsearch operations

PR #1499, opened 2022-02-14
Improvement/zenko 3925 add e2e iam policies tests assumr role wii

XinLi0207:
context:
the current PR is based on [this PR](https://github.com/scality/Zenko/pull/1491)
we created 3 test files under node_tests/iam-policies/cloudserver: one for iam user, one for assumeRole session users, one for assumeRoleWWI session users. Later for all e2e tests between cloudserver and vault, we can put them here.

commit1:
add some necessary env variables in run_e2e_test.sh

commit2:
implement e2e tests for assumrRoleWithWebIdentity session users for metadataSearch API:
for testing, we will create an account named `AccountTest`, we want to test the keycloak users `storage_manager`, `storage_account_onwer` and `data_consumer` should be able to perform mdsearch on buckets of this account after assuming `storage-manager-role`, `storage-account-owner-role` and `data-consumer-role` respectively.

to achieve this, first of all we need to :

create account `AccountTest`, generateAccountAccessKey for it, get iam client, sts client and s3 client of this account
use s3 client to create 2 buckets under this account
then, for each test, get web identity token using keycloak user's credentials, then assume the corresponding role, then using temporary credentials to perform mdsearch operations

### DIFF
--- a/tests/zenko_tests/node_tests/iam_policies/cloudserver/AssumeRoleWithWebIdentity.js
+++ b/tests/zenko_tests/node_tests/iam_policies/cloudserver/AssumeRoleWithWebIdentity.js
@@ -1,0 +1,137 @@
+const assert = require('assert');
+const async = require('async');
+const VaultClient = require('../../VaultClient');
+const { getS3Client } = require('../../s3SDK');
+const { getSTSClient } = require('../../stsSDK');
+const { getTokenForIdentity } = require('../../utils/getWebIdentityToken');
+const { metadataSearchResponseCode } = require('./utils');
+
+let iamClient = null;
+let stsClient = null;
+let s3Client = null;
+
+const clientAdmin = VaultClient.getAdminClient();
+const accountName = 'AccountTest';
+const accountInfo = {
+    email: `${accountName}@test.com`,
+    password: 'test',
+};
+const externalAccessKey = 'DZMMJUPWIUK8IWXRP0HQ';
+const externalSecretKey = 'iTuJdlidzrLipymvAGrLP66Yxghl4NQxLZR3cLlu';
+
+const duration = '1000';
+
+const bucket1 = 'bucket1';
+
+const storageManagerName = 'storage_manager';
+const storageAccountOwnerName = 'storage_account_owner';
+const dataConsumerName = 'data_consumer';
+const storageManagerRole = 'storage-manager-role';
+const storageAccountOwnerRole = 'storage-account-owner-role';
+const dataConsumerRole = 'data-consumer-role';
+
+describe('iam policies - cloudserver - AssumeRoleWithWebIdentity - Metadata', () => {
+
+    before(done => {
+        async.series([
+            // create an account, generateAccountAccessKey for it
+            // get iam client, sts client and s3 client of this account
+            next => clientAdmin.createAccount(accountName, accountInfo, next),
+            next => clientAdmin.generateAccountAccessKey(accountName, next, { externalAccessKey, externalSecretKey }),
+            next => {
+                iamClient = VaultClient.getIamClient(externalAccessKey, externalSecretKey);
+                stsClient = getSTSClient(externalAccessKey, externalSecretKey);
+                s3Client = getS3Client(externalAccessKey, externalSecretKey);
+                next();
+            },
+            // use s3 client to create a bucket and put 2 objects
+            next => {
+                async.series([
+                    next => s3Client.createBucket({ Bucket: bucket1 }, next),
+                    next => s3Client.putObject({ Bucket: bucket1, Key: 'file1' }, next),
+                    next => s3Client.putObject({ Bucket: bucket1, Key: 'file2' }, next),
+                ], next);
+            },
+        ], done);
+    });
+
+    after(done => {
+        async.series([
+            next => s3Client.deleteObjects({
+                Bucket: bucket1,
+                Delete: {
+                    Objects: [{ Key: 'file1' }, { Key: 'file2' }],
+                    Quiet: false,
+                },
+            }, next),
+            next => s3Client.deleteBucket({ Bucket: bucket1 }, next),
+            next => clientAdmin.deleteAccount(accountName, next),
+        ], done);
+    });
+
+
+    const tests = [
+        {
+            name: 'should be able to perform metadata search on all buckets for storage manager role',
+            oidcIdentity: storageManagerName,
+            roleName: storageManagerRole,
+            assertion: result => assert.strictEqual(result.statusCode, 200),
+        },
+        {
+            name: 'should be able to perform metadata search on all buckets for storage account owner role',
+            oidcIdentity: storageAccountOwnerName,
+            roleName: storageAccountOwnerRole,
+            assertion: result => assert.strictEqual(result.statusCode, 200),
+        },
+        {
+            name: 'should be able to perform metadata search on all buckets for data consumer role',
+            oidcIdentity: dataConsumerName,
+            roleName: dataConsumerRole,
+            assertion: result => assert.strictEqual(result.statusCode, 200),
+        },
+    ];
+
+    tests.forEach((test, i) => {
+        it(test.name, done => {
+            let jwtToken = null;
+            let roleArn = null;
+            async.waterfall([
+                next => getTokenForIdentity(test.oidcIdentity, (err, token) => {
+                    assert.ifError(err);
+                    jwtToken = token;
+                    next();
+                }),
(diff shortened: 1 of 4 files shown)


tests/zenko_tests/node_tests/iam_policies/cloudserver/AssumeRoleWithWebIdentity.js
```diff
@@ -0,0 +1,144 @@
...
+const storageAccountOwnerRole = 'storage-account-owner-role';
+const dataConsumerRole = 'data-consumer-role';
+
+describe('iam policies - cloudserver - AssumeRoleWithWebIdentity - Metadata', () => {
+
+    before(done => {
+        async.series([
+            // create an account, generateAccountAccessKey for it, get iam client, sts client and s3 client of this account
+            next => clientAdmin.createAccount(accountName, accountInfo, next),
+            next => clientAdmin.generateAccountAccessKey(accountName, next, { externalAccessKey, externalSecretKey }),
```
ghost, line 40: Can't we directly use the Zenko credentials as root account access keys? According to what I saw in Killian's PR, we can create IAM users using these credentials, so we may not need these two steps?

ghost: `process.env.ZENKO_ACCESS_KEY`

tests/zenko_tests/node_tests/utils/getWebIdentityToken.js
```diff
@@ -0,0 +1,85 @@
...
+ * @param {string} password - password of user requesting token
+ * @param {string} host - host URL of keycloak service
+ * @param {number} port - port of keycloak service
+ * @param {string} path - path of keycloak service authentication API
+ * @param {string} client_id - id of the client of the user
+ * @param {string} grant_type - grant of the user
+ * @param {function} callback - callback function called with error or result
+ * @returns {undefined} undefined
+ */
+function getWebIdentityToken(username, password, host, port, path,
```
ghost, line 36: This function is already in VaultClient that is in the dependencies, how about using the VaultClient function instead? If we want to change the function, it is easier to change it in one place only

ghost: If the current VaultClient version does not embedded the function, we can bump it in package.json

ghost (changes requested):
Some comments, otherwise LGTM. Same comment I sent to Killian, I am not a big fan of only checking http response codes, because this does not formally confirm that the API behaves as we expect (and http codes are also affected by the bucket ACL)

XinLi0207:
I change keycloak request from HTTPS to HTTP is to keep in line with [`end2end_http` step ](https://github.com/scality/Zenko/blob/development/2.3/eve/main.yml#L273), our new added tests are all under `end2end_http`

we won't add our tests in `end2end_https` is because we can see in Zenko main.yaml, the only difference of HTTP tests and HTTPS test is if keycloak is HTTP or not, we already have tests for that in Vault. and the main objective here is to test cloudserver and vault interact correctly with each other, so we don't need the whole test to run again in https mode.

tests/zenko_tests/node_tests/iam_policies/cloudserver/AssumeRoleWithWebIdentity.js
XinLi0207: Have discussed with William privately, `ZENKO_ACCESS_KEY` is the storage_manager_role user of an account named `zenko`, I can't use this user to assumeRoleWithWII of the 3 default roles of account `zenko`. Because we can see from the[ keycloak configure file](https://github.com/scality/Zenko/blob/development/2.3/eve/workers/end2end/scripts/configs/keycloak_config.json#L6), there are only 3 default users defined in keycloak under the account `AccountTest`, so I have to create a new account named `AccountTest` to be able to use the 3 existing users in keycloak.

tests/zenko_tests/node_tests/utils/getWebIdentityToken.js
XinLi0207: Have discussed with William privately as well, `getWebIdentityToken` is not in VaultClient

ghost (approved):
LGTM!

tests/zenko_tests/node_tests/iam_policies/cloudserver/AssumeRoleWithWebIdentity.js
```diff
@@ -0,0 +1,144 @@
...
+                },
+            }, next),
+            next => s3Client.deleteBucket({Bucket: bucket1}, next),
+            next => s3Client.deleteBucket({Bucket: bucket2}, next),
+            next => clientAdmin.deleteAccount(accountName, next),
+        ], done)
+    });
+
+
+    tests = [
```
KillianG: I feel like you are only testing seeded roles, maybe you can test custom roles as well ? 
And when you explicitly deny the operation, for example, you are allowed to mdsearch on all bucket but you have a deny on `bucket1`
And if you only have the right, let's say to listbucket, will you get a 403 when you mdsearch ?

tests/zenko_tests/node_tests/utils/getWebIdentityToken.js
```diff
@@ -0,0 +1,85 @@
...
```
KillianG, line 36: I think there is a lot of code duplicate here, you can probably use
`makeGetRequest` and `getResponseBody` from [request.js](https://github.com/scality/Zenko/blob/development/2.3/tests/zenko_tests/node_tests/utils/request.js)

tests/zenko_tests/node_tests/iam_policies/cloudserver/AssumeRoleWithWebIdentity.js
```diff
@@ -0,0 +1,144 @@
...
+let stsClient = null;
+let s3Client = null;
+
+const clientAdmin = VaultClient.getAdminClient();
+const accountName = 'AccountTest';
+const accountInfo = {
+    email: `${accountName}@test.com`,
+    password: 'test',
+};
+const externalAccessKey = 'DZMMJUPWIUK8IWXRP0HQ';
```
KillianG, line 19: Is this really hardcoded ?

KillianG (changes requested):

tests/zenko_tests/node_tests/iam_policies/cloudserver/AssumeRoleWithWebIdentity.js
ghost: No, the generateAccountAccessKeys is a custom API (superadmin) where we can provide our own access keys. It's not generated.

tests/zenko_tests/node_tests/iam_policies/cloudserver/AssumeRoleWithWebIdentity.js
XinLi0207: for testing custom roles, we need to change the default seting in [keycloak configuration](https://github.com/scality/Zenko/blob/development/2.3/eve/workers/end2end/scripts/configs/keycloak_config.json) . by default keycloak only has `AccountTest::DataConsumer` group, `AccountTest::StorageAccountOwner`group and `StorageManager` role and 3 corresponding users: `data_consumer`, `storage_account_owner` and `storage_manager`. So I can only test using these 3 keycloak users

tests/zenko_tests/node_tests/iam_policies/cloudserver/AssumeRoleWithWebIdentity.js
KillianG: I don't understand why not just using the ones generated by `generateAccountAccessKeys`

tests/zenko_tests/node_tests/utils/getWebIdentityToken.js
XinLi0207: this `getWebIdentityToken` function is I copied from Vault project, I think it's fine.
`makeGetRequest` from request.js is mainly for cloudserver calls, here this call is for keycloak endpoint, if we really want to reuse this func, we may break a lot of current codes

KillianG (approved):

tests/zenko_tests/node_tests/iam_policies/cloudserver/AssumeRoleWithWebIdentity.js
KillianG: I approve the PR because this is not blocking

XinLi0207:
ping

PR #1531, opened 2022-03-09
ZENKO-3998 - fix kafka error in alert

lavigne958:
When no kafka instance is running there is no metric
in prometheus. The function `count` does not return 0
when there is no metrics, instead use the function
`absent` will check that and return 1 if no metric exists.


lavigne958:
hey bert-e

miniscruff (approved):

francoisferrand (approved):

PR #1545, opened 2022-03-21
improvement: ZENKO-4043 bump cloudserver and vault

XinLi0207:
the issue is described in this ticket https://scality.atlassian.net/browse/ARTESCA-2237
post fix arsenal,  cloudserver and vault


solution/deps.yaml
```diff
@@ -20,7 +20,7 @@ cloudserver:
   sourceRegistry: registry.scality.com/cloudserver
   dashboard: cloudserver-dashboards
   image: cloudserver
-  tag: 8.4.3
+  tag: 8.4.4
```
francoisferrand: if you bump cloudserver to 8.4.4 (and get slow listing), you must bump backbeat as well so it supports the format (v1) of newly created buckets .

backbeat should be at least 8.3.9.

francoisferrand (changes requested):

solution/deps.yaml
XinLi0207: thanks, will do!

XinLi0207:
closed cause the clouserver version will be updated in other PR

PR #1569, opened 2022-04-14
ZENKO-4101: add mongodb configserver service

alexanderchan-scality:

* add mongodb configserver service resource
* modify mongos sts to point mongodb configserver service instead of a specific node

fixes #<ISSUE>

### DIFF
--- a/solution-base/mongodb/charts/mongodb-sharded/templates/_helpers.tpl
+++ b/solution-base/mongodb/charts/mongodb-sharded/templates/_helpers.tpl
@@ -49,6 +49,10 @@ metadata:
   {{- end -}}
 {{- end -}}
 
+{{- define "mongodb-sharded.configServer.serviceName" -}}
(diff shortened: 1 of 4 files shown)


francoisferrand (changes requested):
This should IMHO be added to the `patches`, not directly in the charts...
otherwise it will be difficult to ensure we don't break anything when upgrading to newer version of the charts.

francoisferrand (approved):

francoisferrand:
do we need to also keep the changes "applied" ? are the patches not applied automatically during Zenko image build?

alexanderchan-scality:
@francoisferrand the patches are not applied on build time. the changes are always applied to the charts. 
patching in build time can be something I can do after the RC release.

miniscruff (approved):

alexanderchan-scality:
ping

PR #1570, opened 2022-04-15
ZENKO-4103: bump Zenko 2.3.12

JBWatenbergScality:

fixes #<ISSUE>


solution/deps.yaml
```diff
@@ -90,7 +90,7 @@ zenko-operator:
 zenko-ui:
   sourceRegistry: registry.scality.com/zenko-ui
   image: zenko-ui
-  tag: 1.4.3
+  tag: 1.4.4
```
francoisferrand, line 93: should be in a separate commit or at least mentionned in commit message : so that it is easier to see in change log...

francoisferrand (approved):

ghost (approved):

PR #1607, opened 2022-05-17
improvement: ZENKO-4143-restoreObject-e2e-iam-policies-test

XinLi0207:
commit1: Fix test image build flakiness with node10
commit2:  refactor permission tests for CLDSRV and add RestoreObject tests

refactored to use the same test suites for iam policy tests for the two APIs(mdsearch and restore), just two APIs use different checkResponseCode function, so later any new APIs that need to check iam policies we can just add the new API in this list

different from other APIs, restoreObject we cannot have end2end test cases other than 403AccessDenied, no success cases nor no other failed cases. so for resotreObject's success cases, we just assert it's not equal to 403 AccessDenied

commit3: bump vault and arsenal
commit4: update general request call function


ghost (approved):

tests/zenko_tests/node_tests/iam_policies/cloudserver/AssumeRoleWithWebIdentity.js
```diff
@@ -21,117 +22,149 @@ const externalSecretKey = 'iTuJdlidzrLipymvAGrLP66Yxghl4NQxLZR3cLlu';
...
+const testAPIs = [
+    {
+        API: 'MetadataSearch',
+        checkResponse: metadataSearchResponseCode,
+    },
+    {
+        API: 'RestoreObject',
+        checkResponse: restoreObjectResponseCode,
+    },
+];
```
anurag4DSB, line 41: A comment would be helpful in explaining, that we can add new APIs here. (Intent)

anurag4DSB (approved):
Ideally it would be great to have tests independent from each other by using beforeEach and afterEach.
But otherwise LGTM.

KillianG (approved):
LGTM :)

PR #1614, opened 2022-05-20
ZENKO-4202: bump zenko

alexanderchan-scality:
**What does this PR do, and why do we need it?**

**Which issue does this PR fix?**

fixes #<ISSUE>

**Special notes for your reviewers**:

PR #1615, opened 2022-05-24
ZENKO-4206: Bump zenko ui 1.4.6

ChengYanJin:
Bump zenko ui 1.4.6
- Includes the bugfix of regression of uploading file with Chrome and FF.
- Data service role authentication in ZENKO UI


ChengYanJin:
bump zenko-ui 1.4.6 within PR: https://github.com/scality/Zenko/pull/1614

PR #1660, opened 2022-06-28
CI: Add a retry pattern arround end to end test so that we can bypass a few flakiness when the platform is underload and avoid to rebuild the whole project

JBWatenbergScality:
This change assume that the end to end tests are idempotent (so they don't have side effects).


francoisferrand:
> This change assume that the end to end tests are idempotent (so they don't have side effects).

I am not sure we can make that assumption: esp. in case of error, there may be some left-over, e.g. users/buckets/objects...

PR #1696, opened 2022-07-27
Feature/zenko 4275 add oob replication tests

welansari:
Issue: [ZENKO-4275](https://scality.atlassian.net/browse/ZENKO-4275)

- adding e2e tests for replication in an oob bucket

[ZENKO-4275]: https://scality.atlassian.net/browse/ZENKO-4275?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jbertran:
The way we launch the orchestrator script, we _always_ run the cleanup script.

Since we rely on the exit code of the last command to determine if the step is successful, in CI we always report that the tests are successful, although they are not. We should fix this before going further.

PR #1717, opened 2022-09-06
ZENKO-4273: Migration to Github actions

ghost:
Migration to GitHub Actions!


.github/workflows/end2end.yaml
```diff
@@ -0,0 +1,406 @@
+---
+name: build-iso-and-end2end-test
+
+concurrency:
+  group: "zenko-tests"
```
tcarmet: You only want one build at a time? Or one build per branch?
```suggestion
  group: "zenko-tests-${{ github.ref_name }}"
```

.github/workflows/end2end.yaml
```diff
@@ -0,0 +1,406 @@
...
+          url: https://artifacts.scality.net
+          user: ${{ secrets.ARTIFACTS_USER }}
+          password: ${{ secrets.ARTIFACTS_PASSWORD }}
+          source: ${{ env.ROOT_DOCUMENTATION_BUILDDIR }}
+
+  build-iso:
+    runs-on:
+      - self-hosted
+      - centos7
+      - large
```
tcarmet, line 142: Do you need centos7 here?

I would recommend building the iso with a docker image. That will allow you to use it locally to build the iso as well when required.

You can use then the `container` parameter in a job to use that image inside the pipeline.

.github/workflows/end2end.yaml
```diff
@@ -0,0 +1,406 @@
...
+          rm -rf /tmp/.buildx-cache
+          mv /tmp/.buildx-cache-new /tmp/.buildx-cache
+  
+  end2end-http:
+    needs: [build-test-image, check-dashboard-versions]
+    runs-on:
+      - self-hosted
+      - ubuntu
+      - focal
+      - xlarge
```
tcarmet: Keep an eye on a blog post I'll post in a few. If you remove the `self-hosted` tag you're in for a treat.

.github/workflows/end2end.yaml
```diff
@@ -0,0 +1,406 @@
...
+          user: ${{ secrets.ARTIFACTS_USER }}
+          password: ${{ secrets.ARTIFACTS_PASSWORD }}
+          source: "solution-base/_build"
+
+  build-test-image:
+    runs-on:
+      - self-hosted
+      - centos7
+      - large
+    steps:
```
tcarmet: Why centos7 here? I think `runs-on: ubuntu-20.04` would do the job.

.github/workflows/end2end.yaml
```diff
@@ -0,0 +1,406 @@
...
+          source: /tmp/artifacts
+        if: always()
+
+  end2end-https:
+    needs: [build-test-image, check-dashboard-versions]
+    runs-on:
+      - self-hosted
+      - ubuntu
+      - focal
+      - xlarge
```
tcarmet: same as above

.github/workflows/release.yaml
```diff
@@ -0,0 +1,67 @@
...
+          fetch-depth: 0
+      
+      - name: Fetch tags
+        run: git fetch --tags
+      
+      - name: 'Check if tag was already created'
+        shell: bash
+        run: >
+          git show-ref --tags v${{ github.event.inputs.tag }} --quiet \
+          && exit 1 || exit 0
```
tcarmet, line 31: Why don't you create it here? ;)

```yaml
      - name: Create Release
        uses: softprops/action-gh-release@v1
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
        with:
          name: Release ${{ github.event.inputs.tag }}
          tag_name: ${{ github.event.inputs.tag }}
          generate_release_notes: true
          target_commitish: ${{ github.sha }}
```          

Haven't checked but if there's a release document, update it. If there's none. Create one.

tcarmet (approved):
couple of suggestions that I'll let you do whatever you'd like. LGTM

.github/workflows/end2end.yaml
ghost: I am not sure if this can help, considering the changes required to be able to send the generated iso file from a docker container to the artifacts... It will require a bunch of secrets to be set locally if we want to use it as well as ability to write the file in the host system, with the support of both the ISO and Base-ISO files. Also, this will reduce our ability to reuse actions, if needed. Maybe a manual workflow to create ISOs can be helpful? Otherwise, we can just push to the repo to generate an ISO file, it's faster!

So I won't do this in this PR as this requires proper end2end testing and I do not have the network bandwidth to test that locally currently.

For the Centos7 part, I guess we can get rid of it and use ubuntu, but I do not see any benefit of doing so...

.github/workflows/release.yaml
ghost: I am not sure I understand your comment: this first step is here to ensure we don't recreate an existing tag, before creating the release and promoting the artifacts, so it needs to be done before both. AFAIK, this is not supported by `softprops/action-gh-release@v1`, is it?

I will update the documentation!

.github/workflows/end2end.yaml
ghost: This looks amazing! So I just removed the `self-hosted` line (visual studio is saying we have a syntax error but I guess it's because I'm not up to date or it is still a beta).
Is it enough to be able to use these new runners? I still see gcloud & openstack in the builds.

For the backbeat tests, having a more powerful environment will greatly help (for the other, the current CPU is enough).

.github/workflows/end2end.yaml
ghost: Yes you are right, I updated that, thanks!

.github/workflows/release.yaml
ghost: Here we go: https://github.com/scality/Zenko/pull/1717/commits/c060b83e3948d84166f9e0a68900dfde1b99d3ea#diff-87479ac9ea16cec5ceafce00b3b049d7d5920cb663afae42901f20e7c4e6705dR7

This is not yet tested, but I think that's the first thing I will do once we merge this PR, with test tags. Do you think it's wise, or should I test this workflow before? (and if yes, how? I cannot start it manually for now, although I could force it in a commit...)

ghost: Also created the Releng PR: https://github.com/scality/releng-it/pull/504

.github/workflows/end2end.yaml
tcarmet: Feel free to schedule a session with me some time next week. I think you're missing some GitHub Actions knowledge here. There won't be any change to upload the iso to artifacts, and managing secrets will remain the same. Checkout https://docs.github.com/en/actions/using-jobs/running-jobs-in-a-container for more information

.github/workflows/end2end.yaml
tcarmet: Should be enough, I'll have a look and let you know

.github/workflows/release.yaml
tcarmet: I think you can force the workflow via the `gh` cli if it's not available via the ui.

But overall, it's up to you, I'm confident you will make sure it's working so, feel free to merge and check afterwards.

tcarmet: and my bad about the comment I think I missed a part of your file when I reviewed it 😂

.github/workflows/end2end.yaml
```diff
@@ -0,0 +1,400 @@
...
+  push:
+    branches:
+      - "documentation/**"
+      - "feature/**"
+      - "improvement/**"
+      - "bugfix/**"
+      - "w/**"
+      - "hotfix/**"
+      - "ci-dev/**"
+      - "user/**"
```
tcarmet, line 18: Just noticed a queue branch missing, make sure it's available on other workflows you expect to be triggered before merging (if any)
```suggestion
on:
  push:
    branches:
      - "documentation/**"
      - "feature/**"
      - "improvement/**"
      - "bugfix/**"
      - "w/**"
      - "hotfix/**"
      - "ci-dev/**"
      - "user/**"
      - "q/*/**"
```

.github/workflows/end2end.yaml
ghost: Just to confirm: do you mean having a dockerfile + an entrypoint where everything is done (so that we can just run it locally and call it in the CI with 1 command), or just running the steps not with `ubuntu-latest` but inside a container where all the depds are already installed?

I understood the first, but maybe you wanted the latter...

ghost: I'll try to schedule something next week then, thanks!

.github/workflows/end2end.yaml
ghost: Oh thanks! I will add that.

ghost: Added in https://github.com/scality/Zenko/pull/1717/commits/fa83c2be495163f23e8d9c0d5437f606af9f61b4

.github/workflows/end2end.yaml
ghost: I got a build on the new runner! https://github.com/scality/Zenko/actions/runs/3066374814/jobs/4951502412#step:1:2

anurag4DSB (approved):

.github/workflows/end2end.yaml
tcarmet: Uhul!! 🎉 

My theory was, as the self-hosted are pre-created and pre attached, if we're not under a lot of load. GitHub actions will pick those as they are there and available. but if we reach a point where they are all busy, or being created, it will trigger a GitHub Hosted one.

I have nothing to back this up, but it feels right 😅 🤷

PR #1726, opened 2022-09-26
Fix Zenko Critical CVEs

ghost:
Fix doc for 2.4 branch, bump vault and cloudserver to integrate the new S3 permissions fixes, and disable utapi from the build to get rid of CVEs as utapi is not used in Zenko 2.x today.

Also, add more verbosity to vault & cloudserver in the CI as this will be helpful in case of bugs, to ease the debugging.


ghost:
ping

anurag4DSB:
Question: How come we need a fix in ARTESCA v1.5.X as well?
As this maintenance should go in the next minor release only right ie. Artesca v1.6.X

ghost:
@anurag4DSB as we use the same projects branch versions and this PR includes the S3 permissions fix from Ring, it was discussed that the 2.4 branch would be targeted by this PR. I will update 2.5 only here as written in Jira, but at some point, the 2.4 branch will be updated with the changes.
If we want to only fix in 2.5, then we should create all the Artesca 1.6 branches in our projects (Vault, Cloudserver, ...).

ghost (approved):

anurag4DSB (approved):

PR #1727, opened 2022-09-26
feature: ZENKO-4063 Add bucket notification e2e tests

welansari:
Issue: [ZENKO-4063](https://scality.atlassian.net/browse/ZENKO-4063)

- Integrated CTST into a new CI stage
- Added bucket notification end to end CTST tests

[ZENKO-4063]: https://scality.atlassian.net/browse/ZENKO-4063?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/ctst/world/Zenko.ts
```diff
@@ -0,0 +1,158 @@
...
+    STORAGE_ACCOUNT_OWNER = 'STORAGE_ACCOUNT_OWNER',
+    DATA_CONSUMER = 'DATA_CONSUMER',
+}
+
+/**
+ * Cucumber custom World implementation to support Zenko.
+ * This World is reponsible for AWS CLI calls.
+ * Shared between all tests (S3, IAM, STS).
+ */
+export default class Zenko extends World {
```
ghost, line 25: Just to say that thanks to your (awesome :smile: ) work, we will soon be able to have all our http tests in CTST as well!

.github/scripts/end2end/run-e2e-ctst.sh
```diff
@@ -0,0 +1,78 @@
...
+      type: Directory
+  - name: world
+    hostPath:
+      path: /Zenko/tests/ctst/world
+      type: Directory
+EOF
+# Waiting for pod stabilization
+kubectl wait --for=condition=ready pod cli-testing --timeout 5m
+# Running end2end ctst tests
+kubectl exec -it cli-testing -- ./run $COMMAND $WORLD_PARAMETERS
```
ghost: Just curious here, why do we need to pre-create the pod and then run the command? Can't we just run the pod with the right command directly?

.github/workflows/end2end.yaml
```diff
@@ -430,6 +434,65 @@ jobs:
           source: /tmp/artifacts
         if: always()
 
+  ctst-end2end-sharded:
+    needs: [build-test-image, check-dashboard-versions]
+    runs-on:
+      # Enable this for Ring-based tests
+      # - scality-cloud
```
ghost: I believe this is not needed here, it was mainly kept in case we wanted to enable-back ring backend in our tests, but CTST is only about Zenko tests, no storage, unless you plan to have Ring as a backend for bucket notification tests?

.github/scripts/end2end/configure-e2e-ctst.sh
```diff
@@ -0,0 +1,23 @@
...
+
+# Add bucket notification target
+envsubst < ./configs/notification_destinations.yaml | kubectl apply -f -
+# Wait for service stabilization
+kubectl wait --for condition=DeploymentInProgress=true --timeout 10m zenko/${ZENKO_NAME}
+kubectl wait --for condition=DeploymentFailure=false --timeout 10m zenko/${ZENKO_NAME}
+kubectl wait --for condition=DeploymentInProgress=false --timeout 10m zenko/${ZENKO_NAME}
+
+# Creating bucket notification topic in kafka
+kubectl run -i --rm kafka-topics --image=ghcr.io/banzaicloud/kafka:2.13-2.7.0-bzc.1 --restart=Never -- \
```
ghost: Could we maybe add this image in the github config file? I guess it is something that may be updated at some point, so let's have the tag somewhere easy to edit!

tests/ctst/common/common.ts
```diff
@@ -0,0 +1,30 @@
+import { Then } from '@cucumber/cucumber';
+
+const assert = require('assert').strict;
+
+Then('we must get "no such entity"', async function () {
```
ghost: I was wondering if this file could be embedded by default in CTST, as this is very common... What do you think?

tests/ctst/package.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "artesca-ctst",
+  "version": "1.0.0",
```
ghost, line 3: Please note that with the way we dynamically get the CTST version from this file, you will have to use 0.2.0, I guess, depending on the version you want to use after merging the Kafka module.

ghost: See https://github.com/scality/Zenko/pull/1727/files#diff-a8a4d241ff23517544a6ae0ddf6d27875543114739b9ba34b36fd0c7f270e78aR9

tests/ctst/package.json
```diff
@@ -0,0 +1,26 @@
...
+  "repository": "git+https://github.com/scality/artesca.git",
+  "author": "Scality",
+  "license": "ISC",
+  "private": true,
+  "scripts": {
+    "setup": "PREPARE=1 ./run.sh",
+    "dry-run": "./run.sh all --dry-run",
+    "premerge": "./run.sh premerge",
+    "only": "./run.sh only",
+    "postmerge": "./run.sh all"
```
ghost:
```suggestion
    "setup": "PREPARE=1 ./run.sh",
    "dry-run": "./run.sh all --dry-run",
    "premerge": "./run.sh premerge",
    "only": "./run.sh only",
```

We don't have any post merge in Zenko.

Also, do you think we could document how to run these CTST tests locally, considering we have a Zenko running locally or in a VM?

ghost (commented):
A quick review while your PR is still a draft. Thanks for this effort, it's great to see CTST in Zenko as well. It will help us having something more homogeneous between our projects.

.github/scripts/end2end/run-e2e-ctst.sh
ghost: You can override the default `--parallel 4` and use a greater number, if you believe this won't add flakiness, it will help you greatly reduce the tests duration.

tests/ctst/package.json
welansari: We don't actually use any of those commands, so i'll remove them all

tests/ctst/common/common.ts
welansari: For convenience i think we can leave it here for now, but i do agree that when we start getting too many similar steps the file should probably move to CTST, we could even have two common files one global shared between all the projects and one local shared between all tests for project specific steps

.github/scripts/end2end/configure-e2e-ctst.sh
welansari: Even better we can extract it from the Zenko version file, we already have kafka there

.github/scripts/end2end/run-e2e-ctst.sh
welansari: I had issues with `kubectl run` so to not waste time i just used this while the PR was still in progress.
Now we build a CTST image with the required folders copied into it, making it simple to use `kubectl run`

welansari: I actually had to reduce it to 2, as there is some flakiness in the tests.
The flakiness is due to a bug, that will be fixed in this PR https://github.com/scality/backbeat/pull/2321

tests/ctst/common/common.ts
ghost: Yes, I agree.

tests/ctst/steps/notifications.ts
```diff
@@ -0,0 +1,321 @@
...
+    // Getting and adapting previous notification configuration
+    this.addCommandParameter({ bucket: this.saved.bucketName });
+    const result = await S3.getBucketNotificationConfiguration(this.getCommandParameters());
+    try {
+        const notificationConfig = JSON.parse(result.stdout);
+        // Update old config and update
+        notificationConfig.QueueConfigurations.push(destinationConfig.QueueConfigurations[0])
+        this.addCommandParameter({ notificationConfiguration: `'${JSON.stringify(notificationConfig)}'` });
+    } catch (error) {
+        // Put new config it old doesn't exist
```
ghost, line 212:
```suggestion
        // Put new config if old doesn't exist
```

tests/ctst/features/notifications.feature
```diff
@@ -0,0 +1,173 @@
+Feature: Bucket notifications
+    In order to receive notifications
+    As an Artesca User
+    I want to activate notifications
+    And to subscribe to events I want to be notified on
+    And to receive notifications on buckets/objects activities I have subscribed to
+
+  @2.5.0
+  @PreMerge
```
ghost, line 9: You can use a `@BucketNotification` tag if you want, it will ease the filtering if we want to add more tests in the future in multiple jobs where not all tests are expected to run.

ghost (approved):
I went through all the PR several times and it looks good. I did not reviewed fully the content of the tests (the logic tested), but the CTST integration and github actions part. For the tests, I'm trusting the outputs of your last commit's build. As I am lacking time before being OOO, I am approving the CTST/GHA part only here :slightly_smiling_face: 

Thanks for the great work!

Just a quick note: I am note sure having this number of tests at Zenko level for one feature is something we will keep in the future, but that's still something to discuss. Usually we would focus on scenarios for the usual usage of the feature, with some corner cases if needed. But maybe we can't test Bucket Notifications elsewhere, and in this case, it's fine.

.github/scripts/end2end/run-e2e-ctst.sh
ghost: It would be great to see if having something like 8 or 16 parallel tests is stable in the CI, if yes, we will greatly reduce the build time for the CTST part :slightly_smiling_face:

welansari: Yes i completely agree

francoisferrand:
i am wondering if we should not put this PR in 2.4 : since all the code is already there, to "ensure" it works there, since it may be activated...

tests/ctst/features/notifications.feature
francoisferrand: should we not be able (as a "core" feature of CTST) to filter tests by Feature?

(aka it seems redundant to specify a tag matching the name of the feature for *all* scenarios of a feature)

francoisferrand (approved):

PR #1736, opened 2022-10-10
Fix errors due to Kubectl

francoisferrand:
- Use setup-kubectl action to install kubectl
- Replace depreacted serviceaccount flag

Issue: ZENKO-4351


francoisferrand:
will be merged in https://github.com/scality/Zenko/pull/1737

PR #1741, opened 2022-10-14
improvement: ZENKO-4381 upgrade mongodb to fix CVE

naren-rajendran:
Fixes, 
CVE-2017-12424 
CVE-2017-14062
CVE-2021-3520

### DIFF
--- a/.github/scripts/end2end/configs/mongodb_options.yaml
+++ b/.github/scripts/end2end/configs/mongodb_options.yaml
@@ -18,7 +18,7 @@ image:
   ## Bitnami MongoDB image tag
   ## ref: https://hub.docker.com/r/bitnami/mongodb/tags/
   ##
-  tag: 4.0.14-debian-9-r24
+  tag: 4.0.27-debian-9-r118
   ## Specify a imagePullPolicy
   ## ref: http://kubernetes.io/docs/user-guide/images/#pre-pulling-images
   ##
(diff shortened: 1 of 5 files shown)


tmacro (approved):

.github/scripts/end2end/configs/mongodb_options.yaml
```diff
@@ -18,7 +18,7 @@ image:
   ## Bitnami MongoDB image tag
   ## ref: https://hub.docker.com/r/bitnami/mongodb/tags/
   ##
-  tag: 4.0.14-debian-9-r24
+  tag: 4.0.27-debian-9-r118
```
francoisferrand, line 21: I believe changing only deps.yaml is enough, right?

Also, please remember this fixes the CVE only in non-sharded mongo, which we mostly do not use (we deploy sharded mongo)

welansari (approved):

.github/scripts/end2end/configs/mongodb_options.yaml
naren-rajendran: @francoisferrand There is another pr handling the sharded mongo bump, https://github.com/scality/Zenko/pull/1745

naren-rajendran: And I am not sure if changing only deps.yaml is sufficient. So I made changes in all places non-sharded mongodb version is referred.

nicolas2bert (approved):

naren-rajendran:
@bert-e ping

PR #1744, opened 2022-10-19
ZENKO-4374 update bitnami/mongodb-exporter

nicolas2bert:
We use "--compatible-mode" flag to export old metric: "mongodb_mongod_replset_oplog_head_timestamp" for mongoDB and backbeat dashboards.

**Note**: The backbeat dashboard seems to only use the `mongodb_mongod_replset_oplog_head_timestamp` metric accessible thanks to the compatible mode.

### DIFF
--- a/.github/scripts/end2end/configs/mongodb_options.yaml
+++ b/.github/scripts/end2end/configs/mongodb_options.yaml
@@ -437,7 +437,7 @@ metrics:
   image:
     registry: docker.io
     repository: bitnami/mongodb-exporter
-    tag: 0.10.0-debian-9-r104
+    tag: 0.34.0-debian-11-r24
     pullPolicy: IfNotPresent
     ## Optionally specify an array of imagePullSecrets.
     ## Secrets must be manually created in the namespace.
@@ -448,7 +448,10 @@ metrics:
 
   ## String with extra arguments to the metrics exporter
   ## ref: https://github.com/percona/mongodb_exporter/blob/master/mongodb_exporter.go
-  extraArgs: ""
+  ##
+  ## We use "--compatible-mode" flag to export old metric: "mongodb_mongod_replset_oplog_head_timestamp"
+  ## for mongoDB and backbeat dashboards.
+  extraArgs: "--collector.diagnosticdata --collector.replicasetstatus --collector.dbstats --collector.topmetrics --compatible-mode"
 
   ## Metrics exporter resource requests and limits
   ## ref: http://kubernetes.io/docs/user-guide/compute-resources/
--- a/monitoring/mongodb/dashboard.json
+++ b/monitoring/mongodb/dashboard.json
@@ -139,7 +139,7 @@
       "targets": [
         {
           "exemplar": true,
-          "expr": "mongodb_instance_uptime_estimate_seconds{pod=\"${pod}\", namespace=\"${namespace}\"}",
+          "expr": "mongodb_ss_uptime{pod=\"${pod}\", namespace=\"${namespace}\"}",
           "format": "time_series",
           "instant": true,
           "interval": "",
@@ -218,13 +218,13 @@
       "targets": [
         {
           "exemplar": true,
-          "expr": "sum by(pod) (mongodb_connections{pod=\"${pod}\", namespace=\"${namespace}\", state=\"available\"})",
+          "expr": "sum by(pod) (mongodb_ss_connections{pod=\"${pod}\", namespace=\"${namespace}\", conn_type=\"available\"})",
           "format": "time_series",
           "interval": "",
           "instant": false,
           "intervalFactor": 2,
           "legendFormat": "",
-          "metric": "mongodb_connections",
+          "metric": "mongodb_ss_connections",
           "refId": "A",
           "step": 1800
         }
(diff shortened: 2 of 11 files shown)


nicolas2bert:
ping

monitoring/mongodb/dashboard.json
```diff
@@ -666,11 +666,11 @@
       "targets": [
         {
           "exemplar": true,
-          "expr": "sum by(type) (rate(mongodb_op_counters_repl_total{pod=\"${mongod}\", namespace=\"${namespace}\"}[$__rate_interval]))",
+          "expr": "sum by(legacy_op_type) (rate(mongodb_ss_opcountersRepl{pod=\"${mongod}\", namespace=\"${namespace}\"}[$__rate_interval]))",
           "format": "time_series",
           "interval": "",
           "intervalFactor": 2,
-          "legendFormat": "{{type}}",
+          "legendFormat": "{{legacy_op_type}}",
```
francoisferrand, line 673: should we upgrade this to use the "new" op_type (whatever that is)?

.github/scripts/end2end/configs/mongodb_options.yaml
```diff
@@ -448,7 +448,10 @@ metrics:
 
   ## String with extra arguments to the metrics exporter
   ## ref: https://github.com/percona/mongodb_exporter/blob/master/mongodb_exporter.go
-  extraArgs: ""
+  ##
+  ## We use "--compatible-mode" flag to export old metric: "mongodb_mongod_replset_oplog_head_timestamp"
+  ## for mongoDB and backbeat dashboards.
+  extraArgs: "--collect-all --compatible-mode"
```
francoisferrand: looking at the code for mongodb-exporter, the collect-all flag will activate the scanning of every collection and each index metric : i.e. it will list the collections, then iterate over them to get collection-specific metrics and expose this to prometheus.

this may be an issue:
* This may increase load (scan all collections), in case we have many collections/buckets
* This may generate high-cardinality metrics, if the collection is used as a label

Can you check & confirm this?

francoisferrand: note: there are multiple more granular --collect-xxxx flags we may use instead

monitoring/mongodb/dashboard.json
```diff
@@ -774,18 +774,7 @@
...
-          "intervalFactor": 2,
-          "legendFormat": "# members",
-          "refId": "A",
-          "step": 600,
-          "instant": true
-        },
-        {
-          "exemplar": true,
-          "expr": "sum(mongodb_mongod_replset_member_health{pod=\"${mongod}\", namespace=\"${namespace}\"})",
+          "expr": "sum(mongodb_rs_members_health{pod=\"${mongod}\", namespace=\"${namespace}\"})",
```
francoisferrand, line 788: `mongodb_mongod_replset_number_of_members` got removed?

monitoring/mongodb/dashboard.json
```diff
@@ -1084,11 +1073,11 @@
       "targets": [
         {
           "exemplar": true,
-          "expr": "mongodb_mongod_replset_oplog_size_bytes{pod=~\"${mongod}\", namespace=\"${namespace}\"}",
+          "expr": "mongodb_oplog_stats_size{pod=~\"${mongod}\", namespace=\"${namespace}\"}",
           "format": "time_series",
           "interval": "",
           "intervalFactor": 2,
-          "legendFormat": "{{type}}",
+          "legendFormat": "{{pod}}",
```
francoisferrand: the "pod" may vary between members of replset, this may look very strange...
is there a way to get the replset name instead?

monitoring/mongodb/dashboard.json
nicolas2bert: no, it is "back" with the compatible flag.

monitoring/mongodb/dashboard.json
nicolas2bert: Good catch, thanks.

monitoring/mongodb/dashboard.json
nicolas2bert: I don't see any new label from: https://github.com/scality/Zenko/pull/1744/files#diff-b6c135d2f216d51b1b595ceb7676ee028c5c55994ce5b2c966e4c654391db3beR2613

`mongodb_ss_opcountersRepl{cl_id="6350129adfb0b7e3db4773bb",cl_role="",legacy_op_type="command",rs_nm="rs0",rs_state="1"} 0
`

.github/scripts/end2end/configs/mongodb_options.yaml
nicolas2bert: After discussion, I went for the "more granular" options.

monitoring/mongodb/mongo_metrics_with_compatible_mode.example
```diff
@@ -0,0 +1,4859 @@
+# HELP collector_scrape_time_ms Time taken for scrape by collector
```
francoisferrand: should we store this doc here?

if we keep it, maybe it should have the version of mongodb-exporter in its name, to avoid issues when there is a newer upgrade which changes the metrics again.

francoisferrand (approved):

monitoring/mongodb/mongo_metrics_with_compatible_mode.example
nicolas2bert: It will help us figure out dashboard/alert issues if any occur.

alexanderchan-scality (approved):

welansari (approved):

nicolas2bert:
@bert-e approve

nicolas2bert:
ping

PR #1745, opened 2022-10-20
ZENKO-4384: bump mongodb-sharded

alexanderchan-scality:


solution-base/mongodb/containers/mongodb-sharded/Dockerfile
```diff
@@ -0,0 +1,32 @@
+FROM docker.io/bitnami/minideb:bullseye
+LABEL maintainer "Scality"
+
+ENV HOME="/" \
+    OS_ARCH="amd64" \
+    OS_FLAVOUR="debian-9" \
```
francoisferrand: Should be 11, since this is the baseline you are using.

solution-base/mongodb/containers/mongodb-sharded/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+RUN . /opt/bitnami/scripts/libcomponent.sh && component_unpack "gosu" "1.14.0-0" --checksum d177bc3da307f83d83ac8284be8062fc8b428e8f38a3b697876b5efc73ebf74f
+RUN chmod g+rwX /opt/bitnami
+RUN ln -s /opt/bitnami/scripts/mongodb-sharded/entrypoint.sh /entrypoint.sh
+RUN ln -s /opt/bitnami/scripts/liblog.sh /liblog.sh
+RUN ln -s /opt/bitnami/scripts/mongodb-sharded/run.sh /run.sh
+
+COPY rootfs /
+RUN /opt/bitnami/scripts/mongodb-sharded/postunpack.sh
+ENV BITNAMI_APP_NAME="mongodb-sharded" \
+    BITNAMI_IMAGE_VERSION="4.0.27-debian-9-r112" \
```
francoisferrand: Here as well

francoisferrand (commented):
Unfortunately, I doubt we can do this upgrade safely.

Mongodb is built (and was supported) only on Debian 9, and depends on packages which have been updated in an incompatible fashion on Debian 9+ : libcurl3, openssl1.0

While bumping Debian would address the CVEs, I am not sure this is an option here : unfortunately we may need to wait until we can upgrade to mongo 4.2

solution-base/deps.yaml
```diff
@@ -8,8 +8,9 @@ mongodb-exporter:
   image: bitnami/mongodb-exporter
   tag: 0.11.2-debian-10-r410
 mongodb-sharded:
-  image: bitnami/mongodb-sharded
-  tag: 4.0.27-debian-9-r111
```
francoisferrand, line 12: What we may do is update (or build) against the latest "debian-9-r118" (or more?) bitnami minideb.
Not sure it will solve much CVEs though

naren-rajendran (approved):

nicolas2bert (commented):
Is this change fixing any CVEs? If so, which one(s)? If not, we can always close this ticket stating that the upgrade can be done for XYZ reasons.

PR #1799, opened 2023-02-06
ZENKO-4485: ARWWI iam policies ctst tests

KazToozs:
Adds ARWWI tests to Zenko CTST suite using cucumber steps that are [common to iam policy CLI tests](https://github.com/scality/Zenko/pull/1809)

### DIFF
--- a/tests/ctst/features/iam-policies/AssumeRoleWithWebIdentity.feature
+++ b/tests/ctst/features/iam-policies/AssumeRoleWithWebIdentity.feature
@@ -1,0 +1,190 @@
+Feature: Assume Role with Web Identity
+    In order to interact with restricted APIs
+    As an Artesca User
+    I want to use a web identity
+    And to succeed in accessing the API
+
+  @2.6.0
+  @PreMerge
+  @IAM-Policies-ARWWI
+  Scenario Outline: Assume Role with Web Identity
+    Given an existing bucket "" "<withVersioning>" versioning, "without" ObjectLock "without" retention mode
+    And an object that "<objectExists>"
+    And a <type> type
+    When the user tries to perform "<action>" on the bucket
+    Then the user should be able to perform successfully the "<action>" action
+
+    Examples:
+      | action                          | type                  | withVersioning | objectExists   |
+      | MetadataSearch                  | STORAGE_MANAGER       | without        | does not exist |
+      | PutObject                       | STORAGE_MANAGER       | without        | does not exist |
+      | PutObjectAcl                    | STORAGE_MANAGER       | without        | does not exist |
+      | GetObject                       | STORAGE_MANAGER       | without        | exists         |
+      | GetObject                       | STORAGE_MANAGER       | with           | exists         |
+      | GetObjectAcl                    | STORAGE_MANAGER       | without        | exists         |
+      | DeleteObject                    | STORAGE_MANAGER       | without        | exists         |
+      | DeleteObject                    | STORAGE_MANAGER       | with           | exists         |
+      | GetBucketVersioning             | STORAGE_MANAGER       | with           | does not exist |
+      | GetBucketAcl                    | STORAGE_MANAGER       | without        | does not exist |
+      | ListObjectsV2                   | STORAGE_MANAGER       | without        | exists         |
+      | ListObjectVersions              | STORAGE_MANAGER       | with           | exists         |
+      | DeleteObjects                   | STORAGE_MANAGER       | without        | exists         |
+      | HeadObject                      | STORAGE_MANAGER       | without        | exists         |
+      | CopyObject                      | STORAGE_MANAGER       | without        | exists         |
+      | GetObjectTagging                | STORAGE_MANAGER       | without        | exists         |
+      | GetObjectTagging                | STORAGE_MANAGER       | with           | exists         |
+      | PutObjectTagging                | STORAGE_MANAGER       | without        | exists         |
+      | PutBucketLifecycleConfiguration | STORAGE_MANAGER       | without        | does not exist |
+      | GetObjectTagging                | STORAGE_MANAGER       | with           | exists         |
+      | DeleteObjectTagging             | STORAGE_MANAGER       | with           | exists         |
+      | DeleteObjectTagging             | STORAGE_MANAGER       | without        | exists         |
+      | PutObjectTagging                | STORAGE_MANAGER       | without        | exists         |
+      | PutObjectTagging                | STORAGE_MANAGER       | with           | exists         |
+      | GetObjectAcl                    | STORAGE_MANAGER       | with           | exists         |
+      | GetObjectAcl                    | STORAGE_MANAGER       | without        | exists         |
+      | PutObjectAcl                    | STORAGE_MANAGER       | with           | exists         |
+      | PutObjectAcl                    | STORAGE_MANAGER       | without        | exists         |
+      | PutBucketTagging                | STORAGE_MANAGER       | without        | does not exist |
+      | DeleteBucketTagging             | STORAGE_MANAGER       | without        | does not exist |
+      | PutBucketReplication            | STORAGE_MANAGER       | with           | does not exist |
+      | MetadataSearch                  | STORAGE_ACCOUNT_OWNER | without        | does not exist |
+      | PutObject                       | STORAGE_ACCOUNT_OWNER | without        | does not exist |
+      | PutObjectAcl                    | STORAGE_ACCOUNT_OWNER | without        | does not exist |
+      | GetObject                       | STORAGE_ACCOUNT_OWNER | without        | exists         |
+      | GetObject                       | STORAGE_ACCOUNT_OWNER | with           | exists         |
+      | GetObjectAcl                    | STORAGE_ACCOUNT_OWNER | without        | exists         |
+      | DeleteObject                    | STORAGE_ACCOUNT_OWNER | without        | exists         |
+      | DeleteObject                    | STORAGE_ACCOUNT_OWNER | with           | exists         |
+      | GetBucketVersioning             | STORAGE_ACCOUNT_OWNER | with           | does not exist |
+      | GetBucketAcl                    | STORAGE_ACCOUNT_OWNER | without        | does not exist |
+      | ListObjectsV2                   | STORAGE_ACCOUNT_OWNER | without        | exists         |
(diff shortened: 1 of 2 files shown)


tests/ctst/features/iam-policies/AssumeRoleWithWebIdentity.feature
```diff
@@ -0,0 +1,144 @@
...
+    I want to use a web identity
+    And to subscribe to events I want to be notified on
+    And to receive notifications on buckets/objects activities I have subscribed to
+
+  @2.6.0
+  @PreMerge
+  @IAM-Policies-ARWWI
+  Scenario Outline: Assume Role with Web Identity
+    Given an existing bucket "<bucketName>" "<withVersioning>" versioning
+    And we "<haveObject>" an existing object "<objectName>" "<withVersionId>" version-Id "<objectVersionId>"
```
ghost: As discussed in Xin's PR, I would suggest not specifying the name of entities we create, unless we really want to test them: the names **must** be unique for each test and managed by the code only. This will let us run CTST with multiple workers to speed up the testing time, if needed.
The only aspect we can alter is the entities existence (e.g., if the bucket is versioned or not)

```suggestion
    And an object that "<exists>" "<with>" version-Id"
```

Where `exists` can be "does not exist" or "exist" , and `with` can be "with" or "without", then the steps definition can be generic and re-usable easily, and we can also be consistent across scenarios

ghost: if you really need to name the entities yourself, you can leverage the use of data tables. It was not done untill now so not tested as well, but this is supported by the framework that powers CTST: CucumberJS. A nice article to have examples : https://vinodmuralidharan.medium.com/cucumber-js-data-table-guide-a2447b74c86e

tests/ctst/features/iam-policies/AssumeRoleWithWebIdentity.feature
```diff
@@ -0,0 +1,144 @@
...
+    And a <type> type
+    When the user tries to perform "<ifS3Standard>" "<action>" on the bucket
+    Then the user should be able to perform successfully
```
ghost: Not asking for change here, but just an idea to reduce the number of lines here:
Related to the above comment, if we want to test the APIs with all identities, we can use a data table for all types we want, and then, in the code, we can call sequentially the right prepareForType() and then the API call in a loop, and store everything in `this.results`, the benefit is that we can, this way, automatically test any API as all identity types

tests/ctst/steps/iam-policies/common.ts
```diff
@@ -0,0 +1,212 @@
...
+});
+
+Then('the user should be able to perform successfully', function () {
+    this.endForType();
+    if (this.saved.ifS3Standard) {
+        assert.strictEqual(this.result?.err?.includes("AccessDenied"), false);
+    } else {
+        assert.strictEqual(this.result?.statusCode, 200);
+    }
+});
```
ghost: Can we check instead that we got a valid response? In case the API crashes with a given identity, we won't get an AccessDenied, but not a valid response either...

tests/ctst/world/Zenko.ts
```diff
@@ -147,18 +472,242 @@ export default class Zenko extends World {
...
+        } catch (err: any) {
+            console.log(err.response);
+            return { statusCode: err.response.status, data: err.response.data };
+        }
+    }
+
+    async restoreObjectResponseCode() {
+        this.addCommandParameter({ restoreRequest: "Days=1" });
+        return await S3.restoreObject(this.getCommandParameters());
+    }
```
ghost: This would go into another utils file IMHO, this is not really related to the Zenko namespace, but more a convenient way to easily call all APIs with the right parameters using the same step definition.

ghost (commented):
See related comments in https://github.com/scality/Zenko/pull/1802#pullrequestreview-1299676396, this is a first review during the draft phase to give some feedback

tests/ctst/steps/iam-policies/common.ts
XinLi0207: we did consider this part, some API calls just can't return a valid response, like restoreObject, like putObjectRetention, the condition required to get a successfully response for these calls  is not easy to provide, as we are not testing the functionalities of these APIs, so we are thinking of specifying the expected error that is returned for each API call. like for restoreObject it's `invalidObjectState`, for putObjectRetention, it's `InvalidRequest, bucket is missing object lock configuration`. we only accept the expected error returned, if other error returns, we just fail the tests

solution/deps.yaml
```diff
@@ -88,7 +88,7 @@ vault:
   sourceRegistry: registry.scality.com/vault
   dashboard: vault-dashboards
   image: vault
-  tag: 8.7.2
+  tag: 8.7.4
```
ghost: I did it already : https://github.com/scality/Zenko/blob/development/2.6/solution/deps.yaml#L92

It should be gone after rebasing

tests/ctst/features/iam-policies/AssumeRoleWithWebIdentity.feature
```diff
@@ -0,0 +1,192 @@
...
+      | PutObjectAcl                    | STORAGE_MANAGER       | with           | exists         |
+      | PutObjectAcl                    | STORAGE_MANAGER       | without        | exists         |
+      | PutBucketTagging                | STORAGE_MANAGER       | without        | does not exist |
+      | DeleteBucketTagging             | STORAGE_MANAGER       | without        | does not exist |
+      | PutBucketReplication            | STORAGE_MANAGER       | with           | does not exist |
+      | MetadataSearch                  | STORAGE_ACCOUNT_OWNER | without        | does not exist |
+      | PutObject                       | STORAGE_ACCOUNT_OWNER | without        | does not exist |
+      | PutObjectAcl                    | STORAGE_ACCOUNT_OWNER | without        | does not exist |
+      | GetObject                       | STORAGE_ACCOUNT_OWNER | without        | exists         |
+      | GetObject                       | STORAGE_MANAGER       | with           | exists         |
```
ghost: This line is duplicated; I can already see it at line 24 and 39 (and here, 56)

Might be worth checking if other lines are not affected?

tests/ctst/world/Zenko.ts
```diff
@@ -53,19 +61,19 @@ export default class Zenko extends World {
...
+                AccessKeyId: this.parameters.AccountAccessKey,
                 SecretAccessKey: this.parameters.AccountSecretKey,
                 SessionToken: this.parameters.AccountSessionToken,
-            }
-            this.cliMode.parameters.AssumedSession = CacheHelper.ARWWI[CacheHelper.AccountName];
-            this.cliMode.assumed = true;
+            };
+            CacheHelper.parameters.AccessKey = this.parameters.AccountAccessKey;
+            CacheHelper.parameters.SecretKey = this.parameters.AccountSecretKey;
+            CacheHelper.parameters.SessionToken = this.parameters.AccountSessionToken;
```
ghost: Is it still used? I believe we removed the need to pass an ARWWI session to CTST but an account instead, as "AccountSessionToken" does not make sense, as accounts do not have any session tokens

tests/ctst/world/Zenko.ts
```diff
@@ -99,6 +107,322 @@ export default class Zenko extends World {
...
+        const savedParameters = JSON.parse(JSON.stringify(this.cliOptions));
+        this.resetGlobalType();
+        switch (type) {
+            case EntityType.ACCOUNT:
+                await this.prepareRootUser();
+                this.saved.type = EntityType.IAM_USER;
+                break;
+            case EntityType.IAM_USER:
+                await this.prepareIamUser();
+                this.saved.type = EntityType.ACCOUNT;
```
ghost:
```suggestion
                await this.prepareRootUser();
                this.saved.type = EntityType.ACCOUNT;
                break;
            case EntityType.IAM_USER:
                await this.prepareIamUser();
                this.saved.type = EntityType.IAM_USER;
```

tests/ctst/world/Zenko.ts
```diff
@@ -99,6 +107,322 @@ export default class Zenko extends World {
...
+                break;
+            case EntityType.STORAGE_MANAGER:
+                await this.prepareARWWI(
+                    this.parameters.StorageManagerUsername || 'storage_manager',
+                    this.parameters.KeycloakTestPassword || '123',
```
ghost: I think I already reviewed this file but in another PR, so putting the comments here again:

We can store the default password in a const

tests/ctst/features/iam-policies/AssumeRoleWithWebIdentity.feature
KazToozs: This was due to a set of lines we were copy pasting to add versioned/non-versioned cases for the same APIs, I thought I was being careful but I missed this, good catch, thanks

tests/ctst/world/Zenko.ts
KazToozs: Yes this was because the branch wasn't rebased

tests/ctst/world/Zenko.ts
KazToozs: branch wasn't rebased

tests/ctst/world/Zenko.ts
KazToozs: yes branch wasn't rebased

tests/ctst/features/iam-policies/AssumeRoleWithWebIdentity.feature
KazToozs: 892d1348d0be9ca19bd3682978ec006da01430a4

tests/ctst/features/iam-policies/AssumeRoleWithWebIdentity.feature
```diff
@@ -0,0 +1,191 @@
+Feature: Assume Role with Web Identity
+    In order to interact with restricted APIs
+    As an Artesca User
+    I want to use a web identity
+    And to subscribe to events I want to be notified on
+    And to receive notifications on buckets/objects activities I have subscribed to
```
ghost: should be adapted for this file

ghost (changes requested):

ghost (approved):

XinLi0207 (approved):

PR #1802, opened 2023-02-07
Improvement/zenko 4483 ctst iam policies for iam users

XinLi0207:
based on [this PR](https://github.com/scality/Zenko/pull/1799)
translated the [s3 iam-policies end2end tests for IAM users](https://github.com/scality/Zenko/blob/development/2.6/tests/zenko_tests/node_tests/iam_policies/cloudserver/IAMUser.js) to ctst, will eventually remove the original tests.

### DIFF
--- a/tests/ctst/common/common.ts
+++ b/tests/ctst/common/common.ts
@@ -45,11 +45,11 @@ Given('an existing bucket {string} {string} versioning, {string} ObjectLock {str
     }
 });
 
-Given('an object that {string}',
-    async function (objectExists) {
+Given('an object {string} that {string}',
+    async function (objectName, objectExists) {
         this.resetCommand();
         if (objectExists === 'exists') {
-            this.saved.objectName = 'x'.repeat(10);
+            this.saved.objectName = objectName || Utils.randomString();
             this.addCommandParameter({ key: this.saved.objectName });
             this.addCommandParameter({ bucket: this.saved.bucketName });
             this.saved.versionId = extractPropertyFromResults(
--- a/tests/ctst/features/iam-policies/IAMUser.feature
+++ b/tests/ctst/features/iam-policies/IAMUser.feature
@@ -1,0 +1,78 @@
+Feature: IAM Policies for IAM Users
+  This feature allows you to create and attach IAM policies for IAM users.
+  IAM users should have the permissions to perform the actions that they are granted in their IAM policies.
+
+  @2.6.0
+  @PreMerge
+  @IamPoliciesIamUsers
+  Scenario Outline: User is not authorized to perform the actions without IAM policy
+    Given an existing bucket "" "without" versioning, "without" ObjectLock "" retention mode
+    And an object "" that "<objectExists>"
+    And a IAM_USER type
+    When the user tries to perform "<action>" on the bucket
+    Then the user should receive "AccessDenied" error
+
+    Examples:
+      | action         | objectExists   |
+      | MetadataSearch | does not exist |
+      | GetObject      | exists         |
+
+  @2.6.0
+  @PreMerge
+  @IamPoliciesIamUsers
+  Scenario Outline: User is authorized to perform the actions that are granted in the IAM policy
+    Given an existing bucket "<bucketName>" "without" versioning, "without" ObjectLock "" retention mode
+    And an object "<objectName>" that "<objectExists>"
+    And a IAM_USER type
+    And an IAM policy attached to the user with "Allow" effect to perform "<action>" on "<resource>"
+    When the user tries to perform "<action>" on the bucket
+    Then the user should be able to perform successfully the "<action>" action
+
+    Examples:
+      | action         | resource                | bucketName    | objectExists   | objectName |
+      | MetadataSearch | *                       |               | does not exist |            |
+      | GetObject      | *                       |               | exists         |            |
+      | MetadataSearch | iu-md-bucket1           | iu-md-bucket1 | does not exist |            |
+      | GetObject      | iu-go-bucket1/*         | iu-go-bucket1 | exists         |            |
+      | GetObject      | iu-go-bucket2/go-object | iu-go-bucket2 | exists         | go-object  |
+
+  @2.6.0
+  @PreMerge
+  @IamPoliciesIamUsers
+  Scenario Outline: User is not authorized to perform the actions on the resource that they don't have permissions for or explicitly denied
+    Given an existing bucket "<bucketName>" "without" versioning, "without" ObjectLock "" retention mode
+    And an object "<objectName>" that "<objectExists>"
+    And a IAM_USER type
+    And an IAM policy attached to the user with "<effect>" effect to perform "<action>" on "<resource>"
+    When the user tries to perform "<action>" on the bucket
+    Then the user should receive "AccessDenied" error
+
+    Examples:
+      | action         | effect | resource                 | bucketName    | objectExists   | objectName |
+      | MetadataSearch | Allow  | iu-md-bucket3-1          | iu-md-bucket3 | does not exist |            |
+      | MetadataSearch | Deny   | *                        |               | does not exist |            |
+      | MetadataSearch | Deny   | iu-md-bucket4            | iu-md-bucket4 | does not exist |            |
+      | GetObject      | Allow  | iu-go-bucket3-1/*        | iu-go-bucket3 | exists         |            |
+      | GetObject      | Allow  | iu-go-bucket4/go-object1 | iu-go-bucket4 | exists         | go-object  |
+      | GetObject      | Deny   | *                        | iu-go-bucket5 | exists         |            |
+      | GetObject      | Deny   | iu-go-bucket5/*          | iu-go-bucket5 | exists         |            |
+      | GetObject      | Deny   | iu-go-bucket6/go-object  | iu-go-bucket6 | exists         | go-object  |
+
+  @2.6.0
+  @PreMerge
+  @IamPoliciesIamUsers
+  Scenario Outline: User is not authorized to perform the actions on the resource when Allow and Denied are both specified
+    Given an existing bucket "<bucketName>" "without" versioning, "without" ObjectLock "" retention mode
+    And an object "<objectName>" that "<objectExists>"
+    And a IAM_USER type
+    And an IAM policy attached to the user with "Allow" effect to perform "<action>" on "<resource>"
+    And an IAM policy attached to the user with "Deny" effect to perform "<action>" on "<resource>"
+    When the user tries to perform "<action>" on the bucket
+    Then the user should receive "AccessDenied" error
+    Examples:
+      | action         | resource                | bucketName    | objectExists   | objectName |
+      | MetadataSearch | *                       | iu-md-bucket5 | does not exist |            |
+      | MetadataSearch | iu-md-bucket6           | iu-md-bucket6 | does not exist |            |
+      | GetObject      | *                       | iu-go-bucket7 | exists         |            |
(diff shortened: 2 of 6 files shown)


tests/ctst/features/iam-policies/IAMUser.feature
```diff
@@ -0,0 +1,82 @@
+Feature: IAM Policies for IAM Users
+  This feature allows you to create and attach IAM policies for IAM users.
+  IAM users should have the permissions to perform the actions that they are granted in their IAM policies.
+
+  @2.6.0
+  @PreMerge
+  @IAM-Policies-IAMUsers
+  Scenario Outline: User is not authorized to perform the actions without IAM policy
+    Given an existing bucket "<bucketName>" "without" versioning
```
ghost: Don't we want to test both with and without versioning enabled?

```suggestion
    Given an existing bucket "<bucketName>" "<enableversioning>" versioning
```

tests/ctst/features/iam-policies/IAMUser.feature
```diff
@@ -0,0 +1,82 @@
...
+    And we "<haveObject>" an existing object "<objectName>" "without" version-Id ""
+    And a IAM_USER type
+    When the user tries to perform "<ifS3Standard>" "<action>" on the bucket "<haveObject>" the object
```
ghost: I am not sure to understand this sentence, plus the "haveObject" should not be needed, it can be retrieved from previous steps with a default behavior

```suggestion
    When the user tries to perform a "<ifS3Standard>" "<action>" on the bucket

File: tests/ctst/features/iam-policies/IAMUser.feature [hunk 3]
```diff
@@ -0,0 +1,82 @@
+Feature: IAM Policies for IAM Users
+  This feature allows you to create and attach IAM policies for IAM users.
+  IAM users should have the permissions to perform the actions that they are granted in their IAM policies.
+
+  @2.6.0
+  @PreMerge
+  @IAM-Policies-IAMUsers
+  Scenario Outline: User is not authorized to perform the actions without IAM policy
+    Given an existing bucket "<bucketName>" "without" versioning
+    And we "<haveObject>" an existing object "<objectName>" "without" version-Id ""
```
ghost: Not sure to understand a sentence like "And we havenot an existing object ObjectName without version-Id"

how about this:
```suggestion
    And a object that "<objectExists>" "<without>" version-Id
```

* The `"<objectExists>"` can be either `exists` or `does not exists`
* The object name is not needed unless we want to test a specific object name, it can be generated by the test
* Same for the version id

tests/ctst/features/iam-policies/IAMUser.feature
```diff
@@ -0,0 +1,82 @@
...
+    And a IAM_USER type
+    When the user tries to perform "<ifS3Standard>" "<action>" on the bucket "<haveObject>" the object
+    Then the user should receive an access denied error
+
+    Examples:
+      | action         | ifS3Standard  | bucketName | haveObject | objectName |
+      | MetadataSearch | notS3Standard |            | havenot       |            |
+      | RestoreObject  | S3Standard    |            | have    |            |
```
ghost: You can use spaces in the values, to have real sentences:

```suggestion
      | action         | ifS3Standard  | bucketName | haveObject | objectName |
      | MetadataSearch | notS3Standard |            |   have not |            |
      | RestoreObject  | S3Standard    |            |       have |            |
```

ghost: + my comments above also apply to the other test definitions

tests/ctst/features/iam-policies/IAMUser.feature
```diff
@@ -0,0 +1,82 @@
...
+
+  @2.6.0
+  @PreMerge
+  @IAM-Policies-IAMUsers
+  Scenario Outline: User is authorized to perform the actions that are granted in the IAM policy
+    Given an existing bucket "<bucketName>" "without" versioning
+    And we "<haveObject>" an existing object "<objectName>" "without" version-Id ""
+    And a IAM_USER type
+    And an existing IAM policy with "<effect>" effect to perform "<action>" on "<resource>"
```
ghost:
```suggestion
    And an existing IAM policy granting the user to perform "<action>" on "<resource>"
```

tests/ctst/features/iam-policies/IAMUser.feature
```diff
@@ -0,0 +1,82 @@
...
+    And attach the IAM policy to the user
```
ghost: This should not describe an action as a `Given`, instead, we should specify in the line above if the policy should be attached or not.

Regarding reusability of the steps definitions, I would expect different checks if we just want to setup the environment VS if we want to test the actual API; so I think having a single `Given` step definition for an IAM Policy creation + attach would be better

tests/ctst/features/iam-policies/IAMUser.feature
```diff
@@ -0,0 +1,82 @@
...
+    When the user tries to perform "<ifS3Standard>" "<action>" on the bucket "<haveObject>" the object
+    Then the user should be able to perform successfully
```
ghost: We should be more explicit about what is expected, in the sentence:

```suggestion
    Then the user should be able to perform successfully the "<action>" action
```

tests/ctst/steps/iam-policies/IAMUser.ts
```diff
@@ -0,0 +1,50 @@
...
+        }
+        return null;
+    } catch (err) {
+        return null;
+    }
+}
+
+Given('an existing IAM policy with {string} effect to perform {string} on {string}', async function (effect: string, action: string, resource: string) {
+    this.endForType();
+    this.resetCommand();
```
ghost: Maybe you can create a helper for these 2 actions, if we want to run them in multiple steps, it will allow to centralize place where we reset things, if needed

tests/ctst/steps/iam-policies/common.ts
```diff
@@ -209,4 +210,14 @@ Then('the user should be able to perform successfully', function () {
...
+
+Then('the user should receive an access denied error', function () {
+    this.endForType();
+    if (this.saved.ifS3Standard) {
+        assert.strictEqual(this.result?.err?.includes("AccessDenied"), true);
+    } else {
+        assert.strictEqual(this.result?.statusCode, 403);
+        assert.strictEqual(this.result?.data?.includes("AccessDenied"), true);
+    }
+});
```
ghost: Suggestion: make this function generic:

```suggestion
Then('the user should receive a {string} error', function (errorName) {
    this.endForType();
    if (!this.result) {
        throw new Error('Expecting a result after calling the API, but got undefined.');
    }
    if (this.saved.ifS3Standard) {
        assert.strictEqual(this.result?.err?.includes(errorName), true);
    } else {
        assert.strictEqual(this.result?.statusCode, 403);
        assert.strictEqual(this.result?.data?.includes(errorName), true);
    }
});
```

Note: I added a precheck for the `this.result` existence, in case something goes wrong, so that the test fails also in this case

ghost (commented):
cc @KazToozs as some comments are on common things with your work

tests/ctst/features/iam-policies/IAMUser.feature
KazToozs: This might have been a typo by Xin, as indeed I think it makes sense to have this as a variable, but we'll have to see if she has a logic here.

tests/ctst/features/iam-policies/IAMUser.feature
KazToozs: I can confirm this has been [fixed](https://github.com/scality/Zenko/pull/1799/files#diff-284f11d6681baa8186a18ff40a18801ef2f353d3d23a6d8be5d52e3367ed92c9R5) in my PR, so this will need updating

tests/ctst/features/iam-policies/IAMUser.feature
KazToozs: Will update in common PR, starting here ed5d2bd2e26929a6a1decb7086b79ade19e1e988

tests/ctst/features/iam-policies/IAMUser.feature
KazToozs: updating in common PR, starting here ed5d2bd2e26929a6a1decb7086b79ade19e1e988

tests/ctst/features/iam-policies/IAMUser.feature
XinLi0207: I am thinking that we are testing s3 iam policy auth here not functional tests of these s3 apis, versioning doesn't really have an impact on policy check, right?

ghost: Yes, but given that we can reuse the step definitions, we can already implement a version-aware step that we could reuse in the future, even if we do not see any use case for this option today.

tests/ctst/common/common.ts
```diff
@@ -33,11 +33,11 @@ Given('an existing bucket {string} {string} versioning', async function (bucketN
...
 });
 
-Given('an object that {string} {string} version-Id',
-    async function (objectExists, withVersioning) {
+Given('an object {string} that {string} {string} version-Id',
+    async function (objectName, objectExists, withVersioning) {
     this.resetCommand();
     if (objectExists === 'exists') {
-        this.saved.objectName = 'x'.repeat(10);
+        this.saved.objectName = objectName || 'x'.repeat(10);
```
ghost: IMHO the default value here should be something random, the repeat could be used as an object body though. It will allow us not to have any conflict in the scenario where the object name is not specified, what do you think?

tests/ctst/steps/iam-policies/IAMUser.ts
```diff
@@ -0,0 +1,48 @@
...
+setDefaultTimeout(Constants.DEFAULT_TIMEOUT);
+
+/**
+ * Policy are using ARN, not names. This helper will dynamically extract a policy
+ * ARN from a CLI result
+ * @param {object} results - results from the command line
+ * @return {string} - the policy arn, or null if an error occured when
+ * parsing results.
+ */
+function extractPolicyArnFromResults(results: any) {
```
ghost, line 13: Note for reviewers: removed in https://github.com/scality/Zenko/pull/1807, in favor of a more modular function

ghost, line 13: Note for reviewers: removed in https://github.com/scality/Zenko/pull/1807, in favor of a more modular function

ghost (commented):
LGTM otherwise, but I will wait for all the commits to be all green and rebased before approving

tests/ctst/common/common.ts
ghost: Reopening as I do not see any change for this comment (or reason why not to change the logic)

tests/ctst/common/common.ts
```diff
@@ -45,11 +45,11 @@ Given('an existing bucket {string} {string} versioning, {string} ObjectLock {str
...
 });
 
-Given('an object that {string}',
-    async function (objectExists) {
+Given('an object {string} that {string}',
+    async function (objectName, objectExists) {
         this.resetCommand();
         if (objectExists === 'exists') {
-            this.saved.objectName = 'x'.repeat(10);
+            this.saved.objectName = objectName || 'x'.repeat(10);
```
ghost:
```suggestion
            this.saved.objectName = objectName || Utils.randomString();
```

ghost (approved):
LGTM! Don't forget to squash the fixup & tmate commits :smile:

KazToozs (approved):

PR #1806, opened 2023-02-13
zenko-base: Upgrade PodDisruptionBudget to stable API

francoisferrand:
Migrate from `policy/v1beta1` to `policy/v1`, to support running on k8s
1.25. This changes the minimum supported k8s version to 1.21.

Issue: ZENKO-4496


welansari (approved):

solution-base/mongodb/charts/mongodb/templates/poddisruptionbudget-arbiter-rs.yaml
```diff
@@ -1,5 +1,5 @@
 {{- if and (and .Values.replicaSet.enabled .Values.replicaSet.pdb.enabled) (gt .Values.replicaSet.replicas.arbiter 0) }}
-apiVersion: policy/v1beta1
+apiVersion: policy/v1
```
KillianG, line 2: Why not using an include here as well ?

francoisferrand: the included-file is in the `mongodb-sharded` chart, but not in the `mongodb` chart

alexanderchan-scality (approved):

PR #1807, opened 2023-02-14
Improvement/zenko 4484 ctst iam policies for assume role user

XinLi0207:
based on [this PR](https://github.com/scality/Zenko/pull/1802)
translated the [s3 iam-policies end2end tests for AssumeRole User](https://github.com/scality/Zenko/blob/development/2.6/tests/zenko_tests/node_tests/iam_policies/cloudserver/AssumeRole.js) to ctst, will eventually remove the original tests.

and add a world function `prepareAssumeRole` to do all the preparation work for assuming a role within an account or cross account

### DIFF
--- a/tests/ctst/features/iam-policies/AssumeRole.feature
+++ b/tests/ctst/features/iam-policies/AssumeRole.feature
@@ -1,0 +1,99 @@
+Feature: IAM Policies for Assume Role Session Users
+  This feature allows you to create and attach IAM policies for IAM users.
+  IAM users should have the permissions to perform the actions that they are granted in their IAM policies.
+
+
+  @2.6.0
+  @PreMerge
+  @IamPoliciesAssumeRole
+  Scenario Outline: Assume Role User is not authorized to perform the actions with no IAM policy attached to the role
+    Given an existing bucket "" "without" versioning, "without" ObjectLock "" retention mode
+    And an object "" that "<objectExists>"
+    And a "<ifCrossAccount>" AssumeRole user
+    When the user tries to perform "<action>" on the bucket
+    Then the user should receive "AccessDenied" error
+
+    Examples:
+      | action         | objectExists   | ifCrossAccount |
+      | MetadataSearch | does not exist |                |
+      | MetadataSearch | does not exist | cross account  |
+      | GetObject      | exists         |                |
+      | GetObject      | exists         | cross account  |
+
+  @2.6.0
+  @PreMerge
+  @IamPoliciesAssumeRole
+  Scenario Outline: Assume Role User is authorized to perform the actions if the IAM policies that attached to the role have the right permission
+    Given an existing bucket "<bucketName>" "without" versioning, "without" ObjectLock "" retention mode
+    And an object "<objectName>" that "<objectExists>"
+    And a "<ifCrossAccount>" AssumeRole user
+    And an IAM policy attached to the entity "role" with "Allow" effect to perform "<action>" on "<resource>"
+    When the user tries to perform "<action>" on the bucket
+    Then the user should be able to perform successfully the "<action>" action
+
+    Examples:
+      | action         | resource                | bucketName    | objectExists   | objectName | ifCrossAccount |
+      | MetadataSearch | *                       |               | does not exist |            |                |
+      | MetadataSearch | *                       |               | does not exist |            | cross account  |
+      | GetObject      | *                       |               | exists         |            |                |
+      | GetObject      | *                       |               | exists         |            | cross account  |
+      | MetadataSearch | ar-md-bucket1           | ar-md-bucket1 | does not exist |            |                |
+      | MetadataSearch | ar-md-bucket2           | ar-md-bucket2 | does not exist |            | cross account  |
+      | GetObject      | ar-go-bucket1/*         | ar-go-bucket1 | exists         |            |                |
+      | GetObject      | ar-go-bucket2/*         | ar-go-bucket2 | exists         |            | cross account  |
+      | GetObject      | ar-go-bucket3/go-object | ar-go-bucket3 | exists         | go-object  |                |
+      | GetObject      | ar-go-bucket4/go-object | ar-go-bucket4 | exists         | go-object  | cross account  |
+
+  @2.6.0
+  @PreMerge
+  @IamPoliciesAssumeRole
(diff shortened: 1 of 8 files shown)


tests/ctst/common/utils.ts
```diff
@@ -0,0 +1,31 @@
...
+        return null;
+    } catch (err) {
+        return null;
+    }
+}
+
+
+export function extractPropertyFromResults(results: { err: null; stdout: string } | any, ...propertyChain: string[]) : any | null {
+    try {
+        console.log(results);
```
ghost: just so that we don't forget to remove it

tests/ctst/world/Zenko.ts
```diff
@@ -280,6 +288,78 @@ export default class Zenko extends World {
...
+     */
+    async prepareAssumeRole(crossAccount: boolean = false) {
+        // Getting account ID
+        const account = await SuperAdmin.getAccount({
+            name: this.parameters.AccountName || Constants.ACCOUNT_NAME,
+        });
+        Zenko.additionalAccountsCredentials[account.name] = {
+            AccessKey: CacheHelper.parameters.AccessKey,
+            SecretKey: CacheHelper.parameters.SecretKey,
+        };
```
ghost, line 306: It might be a good idea to save the current account in the local list in Zenko init functions, so that we can use any account ing a generic way across all tests, what do you think?

ghost (commented):
Beyond common remarks as in https://github.com/scality/Zenko/pull/1802 on the test definitions, it looks good to me! What is tested is great as well

I just wonder if testing all S3 APIs in this context would make sense, as a safety approach so that we are confident about all our APIs in an AssumeRole context? If you agree, you will be able to use Will's work to call all APIs with the baseline parameters easily!

tests/ctst/features/iam-policies/AssumeRole.feature
```diff
@@ -0,0 +1,99 @@
...
+  This feature allows you to create and attach IAM policies for IAM users.
+  IAM users should have the permissions to perform the actions that they are granted in their IAM policies.
+
+
+  @2.6.0
+  @PreMerge
+  @IAM-Policies-AssumeRole
+  Scenario Outline: Assume Role User is not authorized to perform the actions with no IAM policy attached to the role
+    Given an existing bucket "" "without" versioning
+    And an object "" that "<objectExists>" "without" version-Id
```
ghost: do we really need the "objectExists" option? It is fine creating the file even if we don't use it, but having a "Given" that possibility do nothing feels a bit off IMHO; if this is really not needed, we could just split the scenario.

.github/scripts/end2end/run-e2e-ctst.sh
```diff
@@ -54,4 +54,4 @@ kubectl run $POD_NAME \
         --image-pull-policy=Always \
         --env=TARGET_VERSION=$VERSION  \
         --env=VERBOSE=1 \
-        -- ./run "$COMMAND" $WORLD_PARAMETERS "--parallel $PARALLEL_RUNS --retry $RETRIES --retry-tag-filter @Flaky --tags @IAM-Policies-IAMUsers"
+        -- ./run "$COMMAND" $WORLD_PARAMETERS "--parallel $PARALLEL_RUNS --retry $RETRIES --retry-tag-filter @Flaky --tags  @IAM-Policies-AssumeRole"
```
ghost:
```suggestion
        -- ./run "$COMMAND" $WORLD_PARAMETERS "--parallel $PARALLEL_RUNS --retry $RETRIES --retry-tag-filter @Flaky --tags @IAM-Policies-AssumeRole"
```

tests/ctst/common/utils.ts
```diff
@@ -0,0 +1,24 @@
...
+ */
+export function extractPropertyFromResults(results: { err: null; stdout: string } | any, ...propertyChain: string[]) : any | null {
+    try {
+        if (results.stdout) {
+            const jsonResults = JSON.parse(results.stdout);
+            let res = jsonResults;
+            if (jsonResults) {
+                while (propertyChain.length > 0) {
+                    // @ts-ignore
+                    res = jsonResults[propertyChain.shift()];
```
ghost: We should be able to do better with types here, even if this is with `any`, to at least avoid this ts-ignore comment

tests/ctst/common/utils.ts
```diff
@@ -0,0 +1,24 @@
+/**
+ * This helper will dynamically extract a property from a CLI result
+ * @param {object} results - results from the command line
+ * @param {string[]} propertyChain - the property chain to extract, like Policy, Arn
+ * @return {string} - the expected property, or null if an error occurred when parsing results.
+ */
+export function extractPropertyFromResults(results: { err: null; stdout: string } | any, ...propertyChain: string[]) : any | null {
```
ghost: According to the jsdoc, we return a string (or null), "any" seem wrong here

```suggestion
export function extractPropertyFromResults(results: { err: null; stdout: string } | any, ...propertyChain: string[]) : string | null {
```

tests/ctst/features/iam-policies/AssumeRole.feature
```diff
@@ -0,0 +1,99 @@
+Feature: IAM Policies for Assume Role Session Users
+  This feature allows you to create and attach IAM policies for IAM users.
+  IAM users should have the permissions to perform the actions that they are granted in their IAM policies.
+
+
+  @2.6.0
+  @PreMerge
+  @IAM-Policies-AssumeRole
```
ghost: DO we want to use camelCase here, as tag naming convention?

.github/workflows/end2end.yaml
```diff
@@ -553,6 +553,8 @@ jobs:
       - name: Configure E2E test environment
         run: bash configure-e2e-ctst.sh
         working-directory: ./.github/scripts/end2end
+      - name: Setup tmate session
+        uses: mxschmitt/action-tmate@v3
```
ghost: A comment not to forget to remove this before merging

tests/ctst/package.json
```diff
@@ -14,6 +14,6 @@
     "prometheus-query": "^3.2.0"
   },
   "devDependencies": {
-    "cli-testing": "github:scality/cli-testing.git#0.2.2"
+    "cli-testing": "github:scality/cli-testing.git#449ae371265505b55e3f2894a8195e24d38708a2"
```
ghost: A comment not to forget to use a proper tag here

ghost (approved):

KazToozs (approved):

XinLi0207:
ping

PR #1809, opened 2023-02-21
Improvement/zenko 4499 iam policy common

KazToozs:
Adding functionality for iam policy testing with CTST.

Part of https://scality.atlassian.net/jira/software/c/projects/OS/boards/214?modal=detail&selectedIssue=ZENKO-4480&assignee=61360619c425a20068e9dfcb&assignee=602aa4981bf1ce006b9de203&assignee=6113892de6e6f8007168633b

This PR relies on CTST v0.2.2 which is not yet released and will fail at first.

### DIFF
--- a/.github/actions/deploy/action.yaml
+++ b/.github/actions/deploy/action.yaml
@@ -52,7 +52,7 @@ runs:
       shell: bash
       run: bash deploy-zenko.sh end2end default
       working-directory: ./.github/scripts/end2end
-    - name: Add Keycloak user
+    - name: Add Keycloak user and assign StorageManager role
       shell: bash
       run: bash keycloak-helper.sh add-user default
       working-directory: ./.github/scripts/end2end
--- a/.github/scripts/end2end/configs/keycloak_config.json
+++ b/.github/scripts/end2end/configs/keycloak_config.json
@@ -19,7 +19,7 @@
       "subGroups": []
     }
   ],
-  "defaultRoles" : [ "uma_authorization", "StorageManager", "offline_access" ],
+  "defaultRoles" : [ "uma_authorization", "offline_access" ],
   "roles": {
     "realm": [
       {
--- a/.github/scripts/end2end/keycloak-helper.sh
+++ b/.github/scripts/end2end/keycloak-helper.sh
@@ -37,6 +37,12 @@ case $COMMAND in
             -r ${OIDC_REALM} \
             --username ${OIDC_USERNAME} \
             --new-password ${OIDC_PASSWORD}
+
+        # attach StorageManager role to user
+        ${KEYCLOAK_EXEC} /opt/jboss/keycloak/bin/kcadm.sh add-roles \
+          -r ${OIDC_REALM} \
+          --uusername ${OIDC_USERNAME} \
+          --rolename "StorageManager"
         ;;
 
     *)
--- a/.github/scripts/end2end/run-e2e-ctst.sh
+++ b/.github/scripts/end2end/run-e2e-ctst.sh
@@ -9,12 +9,20 @@ RETRIES=${4:-3}
 # Zenko Version
 VERSION=$(cat ../../../VERSION | grep -Po 'VERSION="\K[^"]*')
 
-# Zenko Access Keys
-ZENKO_ACCOUNT_NAME="zenko"
-ZENKO_ACCESS_KEY=$(kubectl get secret end2end-account-zenko -o jsonpath='{.data.AccessKeyId}' | base64 -d)
-ZENKO_SECRET_KEY=$(kubectl get secret end2end-account-zenko -o jsonpath='{.data.SecretAccessKey}' | base64 -d)
-ZENKO_SESSION_TOKEN=$(kubectl get secret end2end-account-zenko -o jsonpath='{.data.SessionToken}' | base64 -d)
+# Zenko Environment
+ZENKO_ACCOUNT_NAME="zenko-ctst"
+ADMIN_ACCESS_KEY_ID=$(kubectl get secret end2end-management-vault-admin-creds.v1 -o jsonpath='{.data.accessKey}' | base64 -d)
+ADMIN_SECRET_ACCESS_KEY=$(kubectl get secret end2end-management-vault-admin-creds.v1  -o jsonpath='{.data.secretKey}' | base64 -d)
+STORAGE_MANAGER_USER_NAME="ctst_storage_manager"
+STORAGE_ACCOUNT_OWNER_USER_NAME="ctst_storage_account_owner"
+DATA_CONSUMER_USER_NAME="ctst_data_consumer"
 ZENKO_PORT="80"
+KEYCLOAK_TEST_PASSWORD=${OIDC_PASSWORD}
+KEYCLOAK_TEST_HOST=${OIDC_HOST}
+KEYCLOAK_TEST_PORT="80"
+KEYCLOAK_TEST_REALM_NAME=${OIDC_REALM}
+KEYCLOAK_TEST_CLIENT_ID=${OIDC_CLIENT_ID}
+KEYCLOAK_TEST_GRANT_TYPE="password"
 
 # Extracting kafka host from bacbeat's config
 KAFKA_HOST_PORT=$(kubectl get secret -l app.kubernetes.io/name=backbeat-config,app.kubernetes.io/instance=end2end \
@@ -22,7 +30,7 @@ KAFKA_HOST_PORT=$(kubectl get secret -l app.kubernetes.io/name=backbeat-config,a
 KAFKA_HOST_PORT=${KAFKA_HOST_PORT:1:-1}
 
 # Setting CTST world params
-WORLD_PARAMETERS='{"subdomain":"'${SUBDOMAIN}'","ssl":false,"port":"'${ZENKO_PORT}'","AccountName":"'${ZENKO_ACCOUNT_NAME}'","AccountAccessKey":"'${ZENKO_ACCESS_KEY}'","AccountSecretKey":"'${ZENKO_SECRET_KEY}'","AccountSessionToken":"'${ZENKO_SESSION_TOKEN}'","NotificationDestination":"'${NOTIF_DEST_NAME}'","NotificationDestinationTopic":"'${NOTIF_DEST_TOPIC}'","NotificationDestinationAlt":"'${NOTIF_ALT_DEST_NAME}'","NotificationDestinationTopicAlt":"'${NOTIF_ALT_DEST_TOPIC}'","KafkaHosts":"'${KAFKA_HOST_PORT}'"}'
+WORLD_PARAMETERS='{"subdomain":"'${SUBDOMAIN}'","ssl":false,"port":"'${ZENKO_PORT}'","AccountName":"'${ZENKO_ACCOUNT_NAME}'","AdminAccessKey":"'${ADMIN_ACCESS_KEY_ID}'","AdminSecretKey":"'${ADMIN_SECRET_ACCESS_KEY}'","NotificationDestination":"'${NOTIF_DEST_NAME}'","NotificationDestinationTopic":"'${NOTIF_DEST_TOPIC}'","NotificationDestinationAlt":"'${NOTIF_ALT_DEST_NAME}'","NotificationDestinationTopicAlt":"'${NOTIF_ALT_DEST_TOPIC}'","KafkaHosts":"'${KAFKA_HOST_PORT}'","KeycloakPassword":"'${KEYCLOAK_TEST_PASSWORD}'","KeycloakHost":"'${KEYCLOAK_TEST_HOST}'","KeycloakPort":"'${KEYCLOAK_TEST_PORT}'","keycloakRealm":"'${KEYCLOAK_TEST_REALM_NAME}'","keycloakClientId":"'${KEYCLOAK_TEST_CLIENT_ID}'","keycloakGrantType":"'${KEYCLOAK_TEST_GRANT_TYPE}'","StorageManagerUsername":"'${STORAGE_MANAGER_USER_NAME}'","StorageAccountOwnerUsername":"'${STORAGE_ACCOUNT_OWNER_USER_NAME}'","DataConsumerUsername":"'${DATA_CONSUMER_USER_NAME}'"}'
 
 E2E_IMAGE=$E2E_CTST_IMAGE_NAME:$E2E_IMAGE_TAG
 POD_NAME="${ZENKO_NAME}-ctst-tests"
@@ -32,7 +40,7 @@ CTST_VERSION=$(sed 's/.*"cli-testing": ".*#\(.*\)".*/\1/;t;d' ../../../tests/cts
 docker run \
     --rm \
     --network=host \
-    registry.scality.com/cli-testing/cli-testing:"${CTST_VERSION}" /bin/bash \
+    "${E2E_IMAGE}" /bin/bash \
     -c "SUBDOMAIN=${SUBDOMAIN} CONTROL_PLANE_INGRESS_ENDPOINT=${OIDC_ENDPOINT} ACCOUNT=${ZENKO_ACCOUNT_NAME} KEYCLOAK_REALM=${KEYCLOAK_TEST_REALM_NAME} STORAGE_MANAGER=${STORAGE_MANAGER_USER_NAME} STORAGE_ACCOUNT_OWNER=${STORAGE_ACCOUNT_OWNER_USER_NAME} DATA_CONSUMER=${DATA_CONSUMER_USER_NAME} /ctst/bin/seedKeycloak.sh"; [[ $? -eq 1 ]] && exit 1 || echo 'Keycloak Configured!'
 
 
--- a/.github/scripts/end2end/run-e2e-ctst.sh
+++ b/.github/scripts/end2end/run-e2e-ctst.sh
@@ -27,6 +27,15 @@ WORLD_PARAMETERS='{"subdomain":"'${SUBDOMAIN}'","ssl":false,"port":"'${ZENKO_POR
 E2E_IMAGE=$E2E_CTST_IMAGE_NAME:$E2E_IMAGE_TAG
 POD_NAME="${ZENKO_NAME}-ctst-tests"
 
+CTST_VERSION=$(sed 's/.*"cli-testing": ".*#\(.*\)".*/\1/;t;d' ../../../tests/ctst/package.json)
+# Configure keycloak
+docker run \
+    --rm \
+    --network=host \
+    registry.scality.com/cli-testing/cli-testing:"${CTST_VERSION}" /bin/bash \
+    -c "SUBDOMAIN=${SUBDOMAIN} CONTROL_PLANE_INGRESS_ENDPOINT=${OIDC_ENDPOINT} ACCOUNT=${ZENKO_ACCOUNT_NAME} KEYCLOAK_REALM=${KEYCLOAK_TEST_REALM_NAME} STORAGE_MANAGER=${STORAGE_MANAGER_USER_NAME} STORAGE_ACCOUNT_OWNER=${STORAGE_ACCOUNT_OWNER_USER_NAME} DATA_CONSUMER=${DATA_CONSUMER_USER_NAME} /ctst/bin/seedKeycloak.sh"; [[ $? -eq 1 ]] && exit 1 || echo 'Keycloak Configured!'
+
+
 # Running end2end ctst tests
 kubectl run $POD_NAME \
         --pod-running-timeout=5m \
--- a/.github/workflows/end2end.yaml
+++ b/.github/workflows/end2end.yaml
@@ -46,6 +46,7 @@ env:
   # http-env
   UI_ENDPOINT: 'http://ui.zenko.local'
   OIDC_ENDPOINT: 'http://keycloak.zenko.local'
+  OIDC_HOST: 'keycloak.zenko.local'
   NAVBAR_ENDPOINT: 'http://shell-ui.zenko.local'
   ENABLE_KEYCLOAK_HTTPS: 'false'
   # mock-env
--- a/tests/ctst/Dockerfile
+++ b/tests/ctst/Dockerfile
@@ -7,6 +7,11 @@ COPY ./common /ctst/common
 COPY ./steps /ctst/steps
 COPY ./world /ctst/world
 
+USER root
+
+RUN chmod 0777 -R /tmp/
+RUN chmod 0777 -R /ctst/
+
 WORKDIR /ctst
 
 CMD ["./run"]
--- a/tests/ctst/common/common.ts
+++ b/tests/ctst/common/common.ts
@@ -1,5 +1,8 @@
-import { Given} from '@cucumber/cucumber';
+import { Given, setDefaultTimeout } from '@cucumber/cucumber';
 import { Constants, S3, Utils } from 'cli-testing';
+import { extractPropertyFromResults } from './utils';
+
+setDefaultTimeout(Constants.DEFAULT_TIMEOUT);
 
 Given('a {string} bucket', async function (versioning) {
     this.resetCommand();
@@ -11,7 +14,46 @@ Given('a {string} bucket', async function (versioning) {
     this.saved.bucketVersioning = versioning;
     if (versioning !== 'Non versioned') {
         const versioningConfiguration = versioning === 'Versioned' ? 'Enabled' : 'Suspended';
-        this.addCommandParameter({ versioningConfiguration: `Status=${versioningConfiguration}`});
+        this.addCommandParameter({ versioningConfiguration: `Status=${versioningConfiguration}` });
         await S3.putBucketVersioning(this.getCommandParameters());
     }
 });
+
+Given('an existing bucket {string} {string} versioning, {string} ObjectLock {string} retention mode', async function
+    (bucketName, withVersioning, withObjectLock, retentionMode) {
+    this.resetCommand();
+    const preName = this.parameters.AccountName || Constants.ACCOUNT_NAME;
+    bucketName = bucketName || `${preName}${Constants.BUCKET_NAME_TEST}${Utils.randomString()}`.toLocaleLowerCase();
+    this.saved.bucketName = bucketName;
+    this.addCommandParameter({ bucket: bucketName });
+    await S3.createBucket(this.getCommandParameters());
+    if (withVersioning === 'with') {
+        this.addCommandParameter({ versioningConfiguration: `Status=Enabled` });
+        await S3.putBucketVersioning(this.getCommandParameters());
+    }
+    if (withObjectLock === 'with') {
+        this.resetCommand();
+        this.addCommandParameter({ bucket: bucketName });
+        this.addCommandParameter({ 
+            objectLockConfiguration: '{ ' +
+            '"ObjectLockEnabled": "Enabled",' +
+            '"Rule": {' +
+            '"DefaultRetention":' +
+            `{ "Mode": "${retentionMode}", "Days": 50 }}}`
+        });
+        await S3.putObjectLockConfiguration(this.getCommandParameters());
+    }
+});
+
+Given('an object that {string}',
+    async function (objectExists) {
+        this.resetCommand();
+        if (objectExists === 'exists') {
+            this.saved.objectName = 'x'.repeat(10);
+            this.addCommandParameter({ key: this.saved.objectName });
+            this.addCommandParameter({ bucket: this.saved.bucketName });
+            this.saved.versionId = extractPropertyFromResults(
+                await S3.putObject(this.getCommandParameters()), 'VersionId'
+            );
+        }
+    });
--- a/tests/ctst/common/utils.ts
+++ b/tests/ctst/common/utils.ts
@@ -1,0 +1,74 @@
+/**
+ * This helper will dynamically extract a property from a CLI result
+ * @param {object} results - results from the command line
+ * @param {string[]} propertyChain - the property chain to extract, like Policy, Arn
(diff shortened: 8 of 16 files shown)


.github/scripts/end2end/run-e2e-ctst.sh
```diff
@@ -36,4 +53,5 @@ kubectl run $POD_NAME \
         --attach=True \
         --image-pull-policy=Always \
         --env=TARGET_VERSION=$VERSION  \
-        -- ./run "$COMMAND" $WORLD_PARAMETERS "--parallel $PARALLEL_RUNS --retry $RETRIES --retry-tag-filter @Flaky"
+        --env=VERBOSE=1 \
+        -- ./run "$COMMAND" $WORLD_PARAMETERS "--parallel $PARALLEL_RUNS --retry $RETRIES --retry-tag-filter @Flaky --tags @IAM-Policies-ARWWI"
```
XinLi0207: let's remove this tag  `--tags @IAM-Policies-ARWWI"`

.github/workflows/end2end.yaml
```diff
@@ -634,4 +631,4 @@ jobs:
           url: https://artifacts.scality.net
           user: ${{ secrets.ARTIFACTS_USER }}
           password: ${{ secrets.ARTIFACTS_PASSWORD }}
-          source: artifacts
+          source: artifacts
```
XinLi0207: let's bring back the blank line

tests/ctst/Dockerfile
```diff
@@ -1,12 +1,19 @@
 ARG CTST_TAG
-FROM registry.scality.com/cli-testing/cli-testing:$CTST_TAG
+FROM registry.scality.com/cli-testing-dev/cli-testing:$CTST_TAG
```
XinLi0207: leave a note here, so that we don't forget to change here once ctst is released

tests/ctst/common/common.ts
```diff
@@ -15,3 +17,33 @@ Given('a {string} bucket', async function (versioning) {
         await S3.putBucketVersioning(this.getCommandParameters());
     }
 });
+
+Given('an existing bucket {string} {string} versioning', async function (bucketName, hasVersioning) {
+    this.resetCommand();
+    const preName = this.parameters.AccountName || Constants.ACCOUNT_NAME;
+    if (bucketName === '') {
+        bucketName = `${preName}${Constants.BUCKET_NAME_TEST}${Utils.randomString()}`.toLocaleLowerCase();
+    }
```
XinLi0207:
```suggestion
    bucketName = bucketName || `${preName}${Constants.BUCKET_NAME_TEST}${Utils.randomString()}`.toLocaleLowerCase();
```

tests/ctst/common/common.ts
```diff
@@ -15,3 +17,33 @@ Given('a {string} bucket', async function (versioning) {
...
+        this.saved.objectName = 'x'.repeat(10);
+        this.addCommandParameter({key: this.saved.objectName});
+        this.addCommandParameter({bucket: this.saved.bucketName});
+        if (withVersioning === 'with') {
+            this.saved.versionId = '1';
+            this.addCommandParameter({versionId: this.saved.versionId});
+        }
+        await S3.putObject(this.getCommandParameters());
+    }
+});
```
XinLi0207: let's add the last empty line

tests/ctst/steps/iam-policies/common.ts
```diff
@@ -0,0 +1,217 @@
...
+        },
+    ];
+    
+    let userCredentials;
+    if ([EntityType.IAM_USER, EntityType.ACCOUNT].includes(this.saved.type)) {
+        userCredentials = this.parameters.IAMSession;
+    } else {
+        userCredentials = this.parameters.AssumedSession;
+    }
+    const api = testAPIs.find(api => api.API === action);
```
XinLi0207: let's raise an error if api is not found in the list

tests/ctst/world/Zenko.ts
```diff
@@ -53,19 +60,21 @@ export default class Zenko extends World {
...
         if (this.parameters.AccountSessionToken) {
             CacheHelper.ARWWI[CacheHelper.AccountName] = {
-                AccessKeyId:  this.parameters.AccountAccessKey,
+                AccessKeyId: this.parameters.AccountAccessKey,
                 SecretAccessKey: this.parameters.AccountSecretKey,
                 SessionToken: this.parameters.AccountSessionToken,
             }
-            this.cliMode.parameters.AssumedSession = CacheHelper.ARWWI[CacheHelper.AccountName];
-            this.cliMode.assumed = true;
+            // this.cliMode.parameters.AssumedSession = CacheHelper.ARWWI[CacheHelper.AccountName];
```
XinLi0207: we can remove the comments

.github/scripts/end2end/run-e2e-ctst.sh
```diff
@@ -9,24 +9,41 @@ RETRIES=${4:-3}
...
 
 E2E_IMAGE=$E2E_CTST_IMAGE_NAME:$E2E_IMAGE_TAG
 POD_NAME="${ZENKO_NAME}-ctst-tests"
 
+CTST_VERSION=$(sed 's/.*"cli-testing": ".*#\(.*\)".*/\1/;t;d' ../../../tests/ctst/package.json)
+# Configure keycloak
+docker run \
+    --rm \
+    --network=host \
+    registry.scality.com/cli-testing-dev/cli-testing:"${CTST_VERSION}" /bin/bash \
```
ghost: Leaving a comment here as well, to rely on the real ctst image, not the dev one 
```suggestion
    "${E2E_IMAGE}":"${CTST_VERSION}" /bin/bash \
```

.github/workflows/end2end.yaml
```diff
@@ -237,12 +238,10 @@ jobs:
         run: |-
           sudo curl --fail -L https://github.com/mikefarah/yq/releases/download/${{ env.YQ_VERSION}}/${{ env.YQ_BINARY }} -o /usr/bin/yq
           sudo chmod +x /usr/bin/yq
-
           get_image_from_deps() {
               local dep_name=$1
               yq eval ".$dep_name | (.sourceRegistry // \"docker.io\") + \"/\" + .image" deps.yaml
           }
-
           cd solution/
```
ghost: Do we really need to remove lines? It also helps keeping the code readable

.github/scripts/end2end/run-e2e-ctst.sh
```diff
@@ -36,4 +53,5 @@ kubectl run $POD_NAME \
         --attach=True \
         --image-pull-policy=Always \
         --env=TARGET_VERSION=$VERSION  \
-        -- ./run "$COMMAND" $WORLD_PARAMETERS "--parallel $PARALLEL_RUNS --retry $RETRIES --retry-tag-filter @Flaky"
+        --env=VERBOSE=1 \
+        -- ./run "$COMMAND" $WORLD_PARAMETERS "--parallel $PARALLEL_RUNS --retry $RETRIES --retry-tag-filter @Flaky"
```
ghost, line 56: Suggestion: CTST has a report command we could use to generate an html page with the test results `yarn result`. The file is already generated in `reports/report.html` with the `yarn base` command, so I am wondering if having it in the build artifacts would be helpful.

tests/ctst/steps/iam-policies/common.ts
```diff
@@ -0,0 +1,218 @@
...
+    assert(api);
+    if (ifS3Standard === 'notS3Standard'){
```
ghost: We could just filter the `MetadataSearch` action if we only use the standard field for this API, as this is implicit and we already have `if`s to filter this API specifc name, we can just have some kind of switch for these non standard APIs and follow the default logic for standard APIs.

tests/ctst/steps/iam-policies/common.ts
```diff
@@ -0,0 +1,218 @@
...
+        }
+        // Handle other cases
+    } else if (api) {
+        this.resetCommand();
+        this.saved.ifS3Standard = true;
+        this.addCommandParameter({bucket: this.saved.bucketName});
+        if (this.saved.versionId && this.saved.versionId != '') {
+            this.addCommandParameter({versionId: this.saved.versionId});
+        }
+        if (this.saved.objectName && this.saved.objectName != '') {
```
ghost:
```suggestion
        if (this.saved.objectName) {
```

As `!!''` return false already (an empty string is falsy).

tests/ctst/world/Zenko.ts
```diff
@@ -99,6 +106,326 @@ export default class Zenko extends World {
...
+                this.saved.type = EntityType.IAM_USER;
+                break;
+            case EntityType.IAM_USER:
+                await this.prepareIamUser();
+                this.saved.type = EntityType.ACCOUNT;
+                break;
+            case EntityType.STORAGE_MANAGER:
+                await this.prepareARWWI(
+                    this.parameters.StorageManagerUsername || 'storage_manager',
+                    this.parameters.KeycloakTestPassword || '123',
```
ghost: We can store the default password as a const to avoid duplication

tests/ctst/world/Zenko.ts
```diff
@@ -99,6 +106,326 @@ export default class Zenko extends World {
...
+            this.addCommandParameter({ userName: Zenko.IAMUserName });
+            this.addCommandParameter({ policyArn });
+            // Save the attached policy for cleanup
+            Zenko.IAMUserAttachedPolicy = policyArn;
+            await IAM.attachUserPolicy(this.getCommandParameters());
+            this.resetCommand();
+            // Create credentials for the user
+            this.addCommandParameter({ userName: Zenko.IAMUserName });
+            const accessKey = await IAM.createAccessKey(this.getCommandParameters());
+            this.parameters.IAMSession = JSON.parse(accessKey.stdout).AccessKey;
```
ghost: Do we want to properly handle errors here? If the previous API call fails for some reason, the error message might be misleading here (and in IAM)

tests/ctst/steps/iam-policies/common.ts
KazToozs: https://github.com/scality/Zenko/pull/1809/files#diff-284f11d6681baa8186a18ff40a18801ef2f353d3d23a6d8be5d52e3367ed92c9R180

this ok?

tests/ctst/common/common.ts
```diff
@@ -15,3 +17,31 @@ Given('a {string} bucket', async function (versioning) {
...
+    async function (objectExists, withVersioning) {
+    this.resetCommand();
+    if (objectExists === 'exists') {
+        this.saved.objectName = 'x'.repeat(10);
+        this.addCommandParameter({key: this.saved.objectName});
+        this.addCommandParameter({bucket: this.saved.bucketName});
+        if (withVersioning === 'with') {
+            this.saved.versionId = '3sL4kqtJlcpXroDTDmJ+rmSpXd3dIbrHY+MTRCxf3vjVBH40Nr8X8gdRQBpUMLUo';
+            this.addCommandParameter({versionId: this.saved.versionId});
+        }
```
ghost: Should be removed, we cannot specify a version ID when putting an object, this is done by the backend. If the bucket is version-enabled, it will be returned by the first command.

.github/scripts/end2end/run-e2e-ctst.sh
```diff
@@ -9,24 +9,41 @@ RETRIES=${4:-3}
...
+WORLD_PARAMETERS='{"subdomain":"'${SUBDOMAIN}'","ssl":false,"port":"'${ZENKO_PORT}'","AccountName":"'${ZENKO_ACCOUNT_NAME}'","AdminAccessKey":"'${ADMIN_ACCESS_KEY_ID}'","AdminSecretKey":"'${ADMIN_SECRET_ACCESS_KEY}'","NotificationDestination":"'${NOTIF_DEST_NAME}'","NotificationDestinationTopic":"'${NOTIF_DEST_TOPIC}'","NotificationDestinationAlt":"'${NOTIF_ALT_DEST_NAME}'","NotificationDestinationTopicAlt":"'${NOTIF_ALT_DEST_TOPIC}'","KafkaHosts":"'${KAFKA_HOST_PORT}'","KeycloakPassword":"'${KEYCLOAK_TEST_PASSWORD}'","KeycloakHost":"'${KEYCLOAK_TEST_HOST}'","KeycloakPort":"'${KEYCLOAK_TEST_PORT}'","keycloakRealm":"'${KEYCLOAK_TEST_REALM_NAME}'","keycloakClientId":"'${KEYCLOAK_TEST_CLIENT_ID}'","keycloakGrantType":"'${KEYCLOAK_TEST_GRANT_TYPE}'","StorageManagerUsername":"'${STORAGE_MANAGER_USER_NAME}'","StorageAccountOwnerUsername":"'${STORAGE_ACCOUNT_OWNER_USER_NAME}'","DataConsumerUsername":"'${DATA_CONSUMER_USER_NAME}'"}'
```
ghost, line 33: Just a formatting comment, given the number of parameters we could better display this, to make the review/edit process easy, what do you think? (not a blocker here)

tests/ctst/common/common.ts
```diff
@@ -15,3 +17,31 @@ Given('a {string} bucket', async function (versioning) {
         await S3.putBucketVersioning(this.getCommandParameters());
     }
 });
+
+Given('an existing bucket {string} {string} versioning', async function (bucketName, hasVersioning) {
```
ghost: Do we want a bit more options in this base scenario? Like the following parameters:

* Object lock enabled
* If yes, if we enable or not the retention
* If retention is enabled, whether we want it in governance or compliance mode.

This way, we can better reuse this common step definition. Another alternative would be to create other steps for these scenarios, but it duplicates the code.

.github/scripts/end2end/run-e2e-ctst.sh
```diff
@@ -36,4 +53,5 @@ kubectl run $POD_NAME \
         --attach=True \
         --image-pull-policy=Always \
         --env=TARGET_VERSION=$VERSION  \
-        -- ./run "$COMMAND" $WORLD_PARAMETERS "--parallel $PARALLEL_RUNS --retry $RETRIES --retry-tag-filter @Flaky"
+        --env=VERBOSE=1 \
```
ghost: Note for other reviewers: I think we should keep this flag set in the CI, t ease debugging in case of error, as this flag will display all the aws cli commands issued to Zenko, with all details about the responses.

tests/ctst/common/common.ts
ghost: Closing as resoved in latest commit.

.github/actions/deploy/action.yaml
```diff
@@ -52,7 +52,7 @@ runs:
       shell: bash
       run: bash deploy-zenko.sh end2end default
       working-directory: ./.github/scripts/end2end
-    - name: Add Keycloak user
+    - name: Add Keycloak user and assign StorageManager role
```
ghost, line 55: Picking random line:

It would be great to create a markdown page that explains how to contribute to the CTST tests in Zenko CI, so that we can also explain how they are supposed to work, i.e.:

- The tests are idempotent: they must not conflict with each other, here by using random bucket/object names.
- It should be possible to rerun the tests after the end of the CTST test suite (so, tests are not dependent on the environment).
And then maybe some other rules to follow, what do you think?

.github/scripts/end2end/run-e2e-ctst.sh
KazToozs: E2E_IMAGE comes out as `registry.scality.com/zenko/zenko-e2e-ctst` from the end2end.yaml file, so for the moment we don't have an appropriate env variable. Any other suggestion here?

.github/scripts/end2end/configs/keycloak_config.json
```diff
@@ -19,7 +19,7 @@
       "subGroups": []
     }
   ],
-  "defaultRoles" : [ "uma_authorization", "StorageManager", "offline_access" ],
+  "defaultRoles" : [ "uma_authorization", "offline_access" ],
```
XinLi0207, line 22: leave a comment for reviewers
`StorageManager` should not be set as defaultRole for keycloak users, but it didn't affect existing keycloak users, because existing keycloak users overwrite their roles. but it does affect new added keycloak users, so need to fix here

tests/ctst/world/Zenko.ts
```diff
@@ -35,11 +36,17 @@ export default class Zenko extends World {
 
     private options: any = {};
 
-    private saved: object = {};
+    private saved: any = {};
+
+    private static IAMUserName = '';
```
anurag4DSB, line 42: (adding in one place but multiple places have same occurrence)
why aren't we declaring types here?
Example this can also be written as 

```js
private static IAMUserName: string = '';
```

tests/ctst/world/Zenko.ts
```diff
@@ -99,6 +106,318 @@ export default class Zenko extends World {
...
+        const savedParameters = JSON.parse(JSON.stringify(this.cliOptions));
+        this.resetGlobalType();
+        switch (type) {
+            case EntityType.ACCOUNT:
+                await this.prepareRootUser();
+                this.saved.type = EntityType.IAM_USER;
+                break;
+            case EntityType.IAM_USER:
+                await this.prepareIamUser();
+                this.saved.type = EntityType.ACCOUNT;
```
anurag4DSB: Why are these flipped?
for `case EntityType.ACCOUNT` we are using `this.saved.type = EntityType.IAM_USER;`
and vice versa

tests/ctst/world/Zenko.ts
```diff
@@ -99,6 +106,318 @@ export default class Zenko extends World {
...
 
+    /**
+     * This function will dynamically prepare credentials based on the type of
+     * entity provided to let the test run the AWS CLI command using this particular
+     * type of entity.
+     * @param {ScenarioCallerType} type - type of entity, can be 'account', 'storage manager',
+     * 'storage account owner', 'data consumer' or 'iam user'
+     * @returns {undefined}
+     */
+    async prepareForType(type: string): Promise<void> {
```
anurag4DSB: nit: worth rewording to make it more intutive
`prepareForType` -> `setupEntity`
`type` parameter -> `entityType`

tests/ctst/world/Zenko.ts
```diff
@@ -99,6 +106,318 @@ export default class Zenko extends World {
...
+            this.options.webIdentityToken = token;
+            if (!this.options.webIdentityToken) {
+                throw new Error('Error when trying to get a WebIdentity token.');
+            }
+            // Getting account ID
+            const account = await SuperAdmin.getAccount({
+                name: this.parameters.AccountName || Constants.ACCOUNT_NAME,
+            });
+            // Getting roles with GetRolesForWebIdentity
+            const roles = await SuperAdmin.getRolesForWebIdentity(this.options.webIdentityToken);
```
anurag4DSB: nit: worth destructuring for readability and reduce repeatability later on, for roles.data.data
```
const { data { data: roles } } = await SuperAdmin.getRolesForWebIdentity(this.options.webIdentityToken);
```

tests/ctst/world/Zenko.ts
```diff
@@ -99,6 +106,318 @@ export default class Zenko extends World {
...
+        const config: AxiosRequestConfig = {
+            method: 'post',
+            url: `${baseUrl}${host}:${port}${path}`,
+            headers: {
+                'Content-Type': 'application/x-www-form-urlencoded',
+            },
+            data,
+        };
+        const result: AxiosResponse = await axios(config);
+        return result.data.access_token;
```
anurag4DSB: Do we not care about any error handling here, considering we are making an API request for returning an access_token and we possibly have many functions relying on it?

PS: Not reviewed how it is used yet.

tests/ctst/steps/iam-policies/common.ts
KazToozs: 3c6d7c63e9e9d6f3ef640927915e92067e49a2a9

This seem ok?

tests/ctst/world/Zenko.ts
```diff
@@ -99,6 +106,318 @@ export default class Zenko extends World {
...
+            this.parameters.IAMSession = this.cliMode.parameters.IAMSession;
+            this.cliMode.env = true;
+        }
+    }
+
+    /**
+     * Erases all the environment configuration from prepareForType
+     * @returns {undefined}
+     */
+    endForType(): void {
```
anurag4DSB: nit: worth rewording to use the keyword `cleanup` as a part of the name of the function to make it more intuitive 
adding to the comment above where I suggested rewording the `prepareForType` function

tests/ctst/world/Zenko.ts
KazToozs: d88ee6e72f4271d89dd51cfa6a48de5d20df8ce5

tests/ctst/world/Zenko.ts
```diff
@@ -99,6 +106,318 @@ export default class Zenko extends World {
...
+        this.cliMode.assumed = false;
+        this.cliMode.env = false;
+    }
+
     /**
      * Helper function to change the default aws cli command type (iam|sts|s3).
```
anurag4DSB, line 432: might be out of scope, but here we are using a service, and `type` for service, hence a bit confusing.

.github/scripts/end2end/run-e2e-ctst.sh
KazToozs: f87c3c881062fc60e0b746ebce2e0cc542345dc5

tests/ctst/common/utils.ts
```diff
@@ -0,0 +1,74 @@
...
+            return res;
+        }
+        return null;
+    } catch (err) {
+        return null;
+    }
+}
+
+export const s3FunctionExtraParams : { [key: string]: Object } = {
+    'restoreObject': { restoreRequest: 'Days=1' },
```
anurag4DSB, line 26: we did merge recently Days can be zero in this PR: https://github.com/scality/pensieve-api/pull/242
Not sure if its usable right away as it might depend on releases.

.github/scripts/end2end/keycloak-helper.sh
```diff
@@ -37,6 +37,12 @@ case $COMMAND in
             -r ${OIDC_REALM} \
             --username ${OIDC_USERNAME} \
             --new-password ${OIDC_PASSWORD}
+
+        # attach StorageManager role to user
+        ${KEYCLOAK_EXEC} /opt/jboss/keycloak/bin/kcadm.sh add-roles \
+          -r ${OIDC_REALM} \
+          --uusername ${OIDC_USERNAME} \
```
anurag4DSB, line 44: Is this a typo, there are 2 "u" in username

anurag4DSB (commented):
Iteration 1 of review

.github/scripts/end2end/keycloak-helper.sh
XinLi0207: there are 2 u
this is how keycloak defined this command
https://github.com/keycloak/keycloak-documentation/blob/main/server_admin/topics/admin-cli.adoc#adding-client-roles-to-a-user

tests/ctst/common/utils.ts
XinLi0207: that's good to know. Actually in our cases we don't care about valid or invalid values for specific parameters, we are testing permission check, as long as we provide any valid parameter and the request don't get `wrong parameter error`  we should be good.
so for some specific parameters, we hardcoded them

tests/ctst/world/Zenko.ts
XinLi0207: 😅  will switch them back, thanks for catching

tests/ctst/common/common.ts
KazToozs: 22d2f30b793b72bbd174ea27a9693b1e6900d79e

this look ok?

tests/ctst/world/Zenko.ts
KazToozs: TypeScript infers a type at initialisation
https://www.tutorialsteacher.com/typescript/type-inference#:~:text=TypeScript%20infers%20types%20of%20variables,values%20are%20set%20for%20parameters

But I'm open to making changes to variables you feel are worth typing explicitly if you insist on this

tests/ctst/world/Zenko.ts
KazToozs: 669a8fddbdcf3a3d1f57ee3645ade64ad2a03996

tests/ctst/world/Zenko.ts
KazToozs: ef214455bb346bf5cde53f71ac79e06eca513be3

tests/ctst/world/Zenko.ts
KazToozs: 2491f6211da707d8a7ad043996d61ee10e8fbbf8

tests/ctst/world/Zenko.ts
ghost: Usually the approach for tests, from my experience, is "let it throw", as it will return an error anyway...

It may be better to throw a known and human-readable error though, to ease the debugging, as long as enough information is logged along with the message

tests/ctst/world/Zenko.ts
KazToozs: fa26d853481674bb9c2b12689a0f06340526a4cb

.github/scripts/end2end/run-e2e-ctst.sh
KazToozs: ec9c50025db03abec6891bab3547478d580353e8

initial solution was for JS, not shell, oops

tests/ctst/world/Zenko.ts
anurag4DSB: yeah! aiming for easy debugging as well.
If you think error is clear enough, we should be good.
Feel free to resolve the comment if that's the case.

tests/ctst/world/Zenko.ts
anurag4DSB: added one more comment: https://github.com/scality/Zenko/commit/fa26d853481674bb9c2b12689a0f06340526a4cb#r102400941

.github/scripts/end2end/keycloak-helper.sh
anurag4DSB: Thanks for educating me :)

anurag4DSB (approved):
LGTM, we already have a document in progress for the way to add tests, which is good.
Added one small comment on `this.cleanupEntity()` method.

Reference: https://github.com/scality/Zenko/pull/1809#discussion_r1114298765

tests/ctst/world/Zenko.ts
KazToozs: 70220c200e70dad2d88d35cd04696e8c4e3c8cc1

No found occurrences in the code of this function and it's variable being used, so changes were minimal

XinLi0207 (approved):

tests/ctst/world/Zenko.ts
KazToozs: 455f9416a39b4307accf57abb26b7b87a86b3662

thoughts?

tests/ctst/world/Zenko.ts
KazToozs: Addressed https://github.com/scality/Zenko/commit/fa26d853481674bb9c2b12689a0f06340526a4cb#r102419022

.github/scripts/end2end/run-e2e-ctst.sh
KazToozs: Could not find reasonable solution for this, resolving as not blocking

.github/scripts/end2end/run-e2e-ctst.sh
KazToozs: a1872b532f5fd6ea0924a55df3623d859e43baba

thoughts?

.github/actions/deploy/action.yaml
KazToozs: a1872b532f5fd6ea0924a55df3623d859e43baba

thoughts?

tests/ctst/steps/iam-policies/common.ts
KazToozs: https://github.com/scality/Zenko/pull/1809/commits/bcb2e258ea0b45c91357224397bf7796eab99f1c#diff-284f11d6681baa8186a18ff40a18801ef2f353d3d23a6d8be5d52e3367ed92c9R15

ok?

ghost: All god, the scope inside `{ ... }` and the last `break` are not formally needed, but resolving the comment, it's a minor suggestion

tests/ctst/common/common.ts
ghost: yes!

ghost: yes!

.github/actions/deploy/action.yaml
ghost: > Beyond usual Cucumber.js practice of using worlds, steps and features, this test suite has some specifics to consider:

This is something all tests should have in common, this is not specific to CTST, but an important reminder yes!

You could also add that CTST comes along with the AWS-CLI tool and abstraction layers to run it from TS code, but can be extended if needed (iwht iether cli tools or pure http tools, like axios).

tests/ctst/world/Zenko.ts
ghost: perfect!

ghost (approved):
LGTM, beyond the last minor comments

.github/workflows/end2end.yaml
```diff
@@ -555,6 +556,8 @@ jobs:
       - name: Run CTST end to end tests
         run: bash run-e2e-ctst.sh
         working-directory: ./.github/scripts/end2end
+      - name: Setup tmate session
```
XinLi0207: let's remove the tmate session

KazToozs:
ping

PR #1818, opened 2023-03-14
Improvement/zenko 4522 lint world file

KazToozs:
The Zenko world file is required for setting up the environment of CTST tests. This PR adds linting to it.


KazToozs:
On the TODO that needs to be done [here](https://github.com/scality/Zenko/pull/1818/files#diff-371f6719b5883f728d9d3218b4ed2eaf1b7ada8faf798acf95b686a7c347c7ddR487):

VaultClient is recovered as an 'any' [here](https://github.com/scality/cli-testing/blob/main/utils/utils.ts#L95) and is [not explicitly typed](https://github.com/scality/vaultclient/blob/development/8.3/lib/IAMClient.js#L19).

I'm not sure what I can do to handle this one, any suggestions are appreciated

XinLi0207:
seems like in ctst tests TS failed to compile, try to run `tsc` locally to check anywhere failed to compile

tests/ctst/world/Zenko.ts
```diff
@@ -424,58 +526,68 @@ export default class Zenko extends World {
...
+     * @param {Object.<string,*>} parameters - the client-provided parameters
      * @returns {undefined}
      */
-    static async init(parameters: any) {
+    static async init(parameters: { [key: string]: unknown }) {
+
         if (!CacheHelper.accountAccessKeys) {
+            // TODO will need to see how VaultClient is typed
+            // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
             CacheHelper.adminClient = await Utils.getAdminCredentials(parameters);
```
ghost: > VaultClient is recovered as an 'any' [here](https://github.com/scality/cli-testing/blob/main/utils/utils.ts#L95) and is [not explicitly typed](https://github.com/scality/vaultclient/blob/development/8.3/lib/IAMClient.js#L19).
> I'm not sure what I can do to handle this one, any suggestions are appreciated

A usual way of handling JS projects in TS is to have `.d.ts` files that will add types to the library. See [this](https://www.typescriptlang.org/docs/handbook/declaration-files/dts-from-js.html). It's a quick way to support TS wihtout having to rewrite everything.

ghost: (and yes it would require a PR in VaultClient!)

tests/ctst/world/Zenko.ts
```diff
@@ -625,21 +745,47 @@ ${JSON.stringify(policy)}\n${err.message}\n`);
         };
     }
 
+    /**
+     * Get all saved parameters
+     * @returns {Object.<string,*>} - an object with any saved parameters
+     */
+    public getSaved(): { [key: string]: unknown } {
+        return this.saved;
+    }
```
ghost: Souldn't this return a `Saved` type? From there, it solves other lines like this one: https://github.com/scality/Zenko/pull/1818/files#diff-371f6719b5883f728d9d3218b4ed2eaf1b7ada8faf798acf95b686a7c347c7ddR221

ALternatively, if "this.saved" already has the `Saved` type, TSC will dynamically infer the type of thie function and all the variables that you can create using it.

ghost: Can be changed in https://github.com/scality/Zenko/blob/d5982766b6b8fe4db16470290b9a4143fd5e622a/tests/ctst/world/Zenko.ts#L120

tests/ctst/world/Zenko.ts
```diff
@@ -1,15 +1,94 @@
...
+    objectNameSufix: string,
+    [key: string]: unknown
+}
+
+interface Parameters {
+    [key: string]: unknown;
+}
+
+export interface Result {
+    err: null;
```
ghost: if err is `null` maybe we don't need it at all in the document, or maybe it can have another value?

tests/ctst/world/Zenko.ts
```diff
@@ -58,38 +137,40 @@ export default class Zenko extends World {
...
 
         // store service users credentials from world parameters
-        if (this.parameters.ServiceUsersCredentials) {
-             Object.entries(JSON.parse(this.parameters.ServiceUsersCredentials))
-                .forEach((entry : [string, any]) => {
-                    Zenko.serviceUsersCredentials[entry[0] as string] = {
-                        AccessKeyId: entry[1].accessKey as string,
-                        SecretAccessKey: entry[1].secretKey as string,
-
+        if ((this.parameters as Parameters).ServiceUsersCredentials) {
```
ghost: Why not setting the `Parameters` type directly in `parameters`?

tests/ctst/world/Zenko.ts
```diff
@@ -58,38 +137,40 @@ export default class Zenko extends World {
...
-        this.cliMode.parameters = this.parameters;
-
-        if (this.parameters.AccountSessionToken) {
-            CacheHelper.ARWWI[CacheHelper.AccountName] = {
-                AccessKeyId: this.parameters.AccountAccessKey,
-                SecretAccessKey: this.parameters.AccountSecretKey,
-                SessionToken: this.parameters.AccountSessionToken,
+        CacheHelper.parameters = worldParameters;
+        const cacheParameters = CacheHelper.parameters as Parameters;
+        this.cliMode.parameters = worldParameters as ClientOptions;
```
ghost: This looks a bit too verbose to be properly typed, all these variables should already be of the right type, otherwise, casting might lead to missed transpiler errors, as it "changes" the type (with some checks). What is the reason the `worldParameters` is created as `Parameters` and then casted to `ClientOptions`?

Here, we create three variables out of one, with different types, which is hard to read.

tests/ctst/world/Zenko.ts
```diff
@@ -58,38 +137,40 @@ export default class Zenko extends World {
...
-            CacheHelper.parameters.SecretKey = this.parameters.AccountSecretKey;
-            CacheHelper.parameters.SessionToken = this.parameters.AccountSessionToken;
+            cacheParameters.AccessKey = worldParameters.AccountAccessKey;
+            cacheParameters.SecretKey = worldParameters.AccountSecretKey;
+            cacheParameters.SessionToken = worldParameters.AccountSessionToken;
         } else {
-            CacheHelper.AccountName = CacheHelper.parameters.AccountName;
-            CacheHelper.parameters.AccessKey = this.parameters.AccountAccessKey;
-            CacheHelper.parameters.SecretKey = this.parameters.AccountSecretKey;
+            CacheHelper.AccountName = cacheParameters.AccountName as string;
```
ghost: From lines above
```
CacheHelper.parameters = worldParameters
```

So why not using `worldParameters` direclty? Also, AccountName should already be a string in the type.

tests/ctst/world/Zenko.ts
```diff
@@ -101,19 +182,22 @@ export default class Zenko extends World {
      * @param {object} result - contains both stderr and stdout from the CLI command.
      * @returns {boolean} - if the result is a success or a failure
      */
-    checkResult(result: any[]): boolean {
+    checkResult(result: Result | Utils.Command | Result[]): boolean {
+        let usedResult: Result | Utils.Command | Result[];
```
ghost:
```suggestion
    checkResult(result: Result |Result[]): boolean {
        let usedResult: Result | Result[];
```

I do not see any use of `Utils.Command` in this function, as `result` is always casted to `result`, it might be needed to update the caller functions here, it seems strange to mix types...

ghost:
```suggestion
    checkResult(result: Result | Result[]): boolean {
        let usedResult: Result | Result[];
```

I do not see any use of `Utils.Command` in this function, as `result` is always casted to `result`, it might be needed to update the caller functions here, it seems strange to mix types...

tests/ctst/world/Zenko.ts
```diff
@@ -101,19 +182,22 @@ export default class Zenko extends World {
...
         if (!Array.isArray(result)) {
-            result = [result];
+            usedResult = [result] as Result[];
+        } else {
+            usedResult = result;
         }
         let decision = true;
-        result.forEach(res => {
+        usedResult.forEach(res => {
```
ghost: It feels strange that a variable potentially not an array has a `forEach`, why not updating as follows:

```suggestion
        let usedResult: Result[] = Array.isArray(result) ? result : [result];
        let decision = true;
        usedResult.forEach(res => {
```

tests/ctst/world/Zenko.ts
```diff
@@ -134,47 +218,49 @@ export default class Zenko extends World {
...
-                this.saved.type = EntityType.ACCOUNT;
-                break;
-            case EntityType.IAM_USER:
-                await this.prepareIamUser();
-                this.saved.type = EntityType.IAM_USER;
-                break;
-            case EntityType.STORAGE_MANAGER:
-                await this.prepareARWWI(
-                    this.parameters.StorageManagerUsername || 'storage_manager',
+        case EntityType.ACCOUNT:
```
ghost, line 201: is the formatting change needed?

tests/ctst/world/Zenko.ts
```diff
@@ -134,47 +218,49 @@ export default class Zenko extends World {
...
+            await this.prepareRootUser();
+            saved.type = EntityType.ACCOUNT;
+            break;
+        case EntityType.IAM_USER:
+            await this.prepareIamUser();
+            saved.type = EntityType.IAM_USER;
+            break;
+        case EntityType.STORAGE_MANAGER:
+            await this.prepareARWWI(
+                    usedParameters.StorageManagerUsername as string || 'storage_manager',
```
ghost: This should be already typed in the `Parameter` interface as a known type.

tests/ctst/package.json
ghost, line 1: Do we already run the linter in the CI? I would expect that each linter PR would have like a special linter rule to only chack specific files, till all are handled

tests/ctst/world/Zenko.ts
```diff
@@ -197,37 +283,41 @@ export default class Zenko extends World {
...
+                `/auth/realms/${usedParameters.keycloakRealm as string || 'zenko'}/protocol/openid-connect/token`,
+                usedParameters.keycloakClientId as string || Constants.K_CLIENT,
+                usedParameters.keycloakGrantType as string || 'password',
             );
             this.options.webIdentityToken = token;
             if (!this.options.webIdentityToken) {
                 throw new Error('Error when trying to get a WebIdentity token.');
             }
             // Getting account ID
             const account = await SuperAdmin.getAccount({
```
ghost, line 276: Should be already of type `Account`

tests/ctst/world/Zenko.ts
```diff
@@ -1,15 +1,94 @@
...
+    Name: string;
+    Arn: string;
+}
+
+interface Account {
+    id: string;
+    Arn: string;
+    name: string;
+    Roles: Role[];
+}
```
ghost: This (and maybe other types) might make more sense to define in CTST directly. I am not sure it makes sense to define them here because if tomorrow we want to enforce this kind of linter wherever CTST is running, we might end up duplicating this code. Plus if this requires to cast some of the CTST functions everytime we call them, it makes it hard to read.

If this is related to a generated file in CTST, it is fine to edit it,a s long as we handle the change in the generator script accordingly (so ideally we do it only in the generator script and we just need to regenerate the file(s)).

tests/ctst/world/Zenko.ts
```diff
@@ -197,37 +283,41 @@ export default class Zenko extends World {
...
                     (roleArn: string) => roleArn.includes(ARWWITargetRole) && roleArn.includes(account.id),
                 );
             } else {
-                roles.Accounts.forEach((_account: any) => {
+                data.roles.Accounts.forEach((_account: Account) => {
                     roleToAssume = _account.Roles.find(
                         (role: { Name: string; Arn: string }) =>
                             role.Arn.includes(ARWWITargetRole) && role.Arn.includes(account.id),
-                    )?.Arn || roleToAssume;
+                    )?.Arn as string || roleToAssume;
```
ghost: I don't see why this is needed if the interface `Role` is already defined.

I will also stop here the comments for this kind of issue, as I see 152 ` as ` appearance; which is not really a good practise as you are telling TS "hey, I know what type this is, so trust me", but that could hide other issues and misses the advantages of using TS.

As a general strategy I would suggest to type the objects/functions you have when they are created, instead of casting all the variables almost in every file, as it translates either that we are not using the right types, or that we should merge some of them.

ghost (changes requested):

tests/ctst/world/Zenko.ts
ghost: THat might be helpful for some of the variables: https://github.com/cucumber/cucumber-js/blob/main/docs/support_files/world.md#world-parameters-1

To be confirmed considering our verison of cucumberjs

tests/ctst/world/Zenko.ts
KazToozs: The `Saved` type is an interface I created before fully understanding typing solutions I think, adding fields to the interface as they're added to the object; the `saved` object is meant to be able to take anything, so I think it actually makes more sense to remove the `Saved` interface and working with it in a generalised way, but thank you for picking up on this.

KazToozs: thoughts ? : b974e868c1cef25abaddc7e91df392ba48a27bc9

tests/ctst/world/Zenko.ts
KazToozs: Unfortuneatly because parameters is an `IWorld` attribute from cucumber that is `any`; is there a better way given this? 

Edit: thinking of it, could an override work? 🤔 Will try it.
Done: dc288de684d1e8b4d648061a4d1b9589b66cca93

tests/ctst/world/Zenko.ts
KazToozs: Responding to here and above: admittedly I have no idea why we make these assignments as the variables are not even used later in the code. I'll drop them and let's see what happens.
b777f3979d70fe34f44b7939b05fd8751d4d61f9

tests/ctst/world/Zenko.ts
KazToozs: Responding for here and above
Let's see what this does, I'll rebase the other PRs to see if anything breaks after all these changes

9103dbb8a3921d72239868a91c336abca0294aa2

tests/ctst/world/Zenko.ts
KazToozs: apparently for the linter, yes

tests/ctst/world/Zenko.ts
KazToozs: Responded below

tests/ctst/world/Zenko.ts
KazToozs: responded below

KazToozs: responded below

tests/ctst/world/Zenko.ts
KazToozs: Since changing to Records above, this has changed. Let me know if this is still an issue

tests/ctst/package.json
KazToozs: No it doesn't already run in the linter in this PR, as indeed it would fail due to the other files

tests/ctst/world/Zenko.ts
KazToozs: `SuperAdmin.getAccount()` returns Promise<any>, not sure what you're saying here?

tests/ctst/world/Zenko.ts
KazToozs: Agreed, this is what I was expecting as feedback to this PR. 

I'm not sure I understand your comment on the second paragraph however; are you suggesting how we should handle adding these types to CTST?

tests/ctst/world/Zenko.ts
KazToozs: The typing for world parameters described lower would be useful, but is for v8.1 and we're on 7.3. We could bump ?

tests/ctst/world/Zenko.ts
KazToozs: On this topic, I have identified (atleast for this file) the types that need defining in the cli-testing project:
- CacheHelper.parameters
- CacheHelper.ARWWI (list containing ClientOptions['AssumedSession'] type objects)
- Account (SuperAdmin.ts->getAccount return value)
- add Roles array to Account type in utils.ts
- I currently use the 'Result' interface to evaluate to the { err: null; stdout: string } type returned by the cli IAM APIs: it might be worth adding that to cli-testing (make buildAwsCliCommand return Result)

ghost: > I'm not sure I understand your comment on the second paragraph however; are you suggesting how we should handle adding these types to CTST?

It's just that some of CTST files are generated and maybe not typed as you need, in this case, it's important to handle it! The good way is to update the generator scripts to include the right types you want, if this is needed.

tests/ctst/package.json
```diff
@@ -14,13 +14,13 @@
...
-    "cli-testing": "github:scality/cli-testing.git#0.2.4",
+    "cli-testing": "github:scality/cli-testing.git#v0.2.6",
     "@typescript-eslint/eslint-plugin": "^5.45.0",
     "@typescript-eslint/parser": "^5.45.0",
     "babel-jest": "^29.3.1",
     "eslint": "^8.28.0"
   },
   "scripts": {
-    "lint": "eslint ."
+    "lint": "eslint ./world/Zenko.ts"
```
ghost, line 26: To enforce it, we need to call it in the CI at some point, IMO in the workflow where ctst is built: https://github.com/scality/Zenko/blob/daa12076a4802461c60c760decc9e66cbe373597/.github/workflows/end2end.yaml#L320

tests/ctst/world/Zenko.ts
```diff
@@ -52,44 +83,41 @@ export default class Zenko extends World {
...
-                    Zenko.serviceUsersCredentials[entry[0] as string] = {
-                        AccessKeyId: entry[1].accessKey as string,
-                        SecretAccessKey: entry[1].secretKey as string,
-
+            Object.entries(JSON.parse(
+                this.parameters.ServiceUsersCredentials as string) as UserCredentials)
+                .forEach(entry => {
+                    Zenko.serviceUsersCredentials[entry[0]] = {
+                        AccessKeyId: (entry[1] as { accessKey: string }).accessKey,
+                        SecretAccessKey: (entry[1] as { secretKey: string }).secretKey,
```
ghost: This looks strange, the object is force-casted into a `UserCredentials` object, which is an object, then we iterate over it, and access its values as an array...

```suggestion
                this.parameters.ServiceUsersCredentials as string) as Record<string, UserCredentials>)
                .forEach(([key, value]) => {
                    Zenko.serviceUsersCredentials[key] = {
                        AccessKeyId: value.accessKey,
                        SecretAccessKey: value.secretKey,
```

(not tested but to illustrate the idea)

tests/ctst/world/Zenko.ts
```diff
@@ -52,44 +83,41 @@ export default class Zenko extends World {
...
         }
 
         // Workaround to be able to access global parameters in BeforeAll/AfterAll hooks
         CacheHelper.parameters = this.parameters;
-        this.cliMode.parameters = this.parameters;
+        this.cliMode.parameters = this.parameters as ClientOptions;
 
         if (this.parameters.AccountSessionToken) {
-            CacheHelper.ARWWI[CacheHelper.AccountName] = {
+            ((CacheHelper.ARWWI as Record<string, unknown>)[CacheHelper.AccountName]) = {
```
ghost: The type of this object should be already known and should not be of type "unknown", but `Record<string, ClientOptions['AssumedSession']>` (according to your recent changes)

tests/ctst/world/Zenko.ts
```diff
@@ -52,44 +83,41 @@ export default class Zenko extends World {
...
                 SessionToken: this.parameters.AccountSessionToken,
             };
-            CacheHelper.parameters.AccessKey = this.parameters.AccountAccessKey;
-            CacheHelper.parameters.SecretKey = this.parameters.AccountSecretKey;
-            CacheHelper.parameters.SessionToken = this.parameters.AccountSessionToken;
         } else {
-            CacheHelper.AccountName = CacheHelper.parameters.AccountName;
-            CacheHelper.parameters.AccessKey = this.parameters.AccountAccessKey;
-            CacheHelper.parameters.SecretKey = this.parameters.AccountSecretKey;
+            CacheHelper.AccountName = this.parameters.AccountName as string;
```
ghost: I would argue that all the parameters should be already known, as we handle them with our own logic, so the parameters type should be properly typed, to avoid `unknwon` that leads to many `as` in the code logic, making it hard to maintain.

tests/ctst/world/Zenko.ts
ghost: Yes, I still see force-casts in the logic for this object, which should be enforced, i.e. we want to know in advance all the parameters we support, with their type. This applies to all the changes related to `this.parameters`

tests/ctst/world/Zenko.ts
ghost: Any promise will hold a type that will be infered dynamically when you `await` the function, to avoid casting the function result to something that might not work, in case of an error, I still see that this is needed here

tests/ctst/world/Zenko.ts
```diff
@@ -197,36 +224,39 @@ export default class Zenko extends World {
...
-            if (roles.ListOfRoleArns) {
-                roleToAssume = roles.find(
-                    (roleArn: string) => roleArn.includes(ARWWITargetRole) && roleArn.includes(account.id),
+            // Get the first role with the storage-manager-role name
+            const data: GetRolesForWIResponse =
+                (await SuperAdmin.getRolesForWebIdentity(this.options.webIdentityToken)).data as GetRolesForWIResponse;
+            let roleToAssume: string | undefined = '';
+            if (data.roles.ListOfRoleArns) {
+                roleToAssume = data.roles.ListOfRoleArns.find(
+                    (roleArn: string) => roleArn.includes(ARWWITargetRole) && roleArn.includes(account.id as string),
```
ghost: If the `account` object is already of the right type, the `as string` is not needed)

ghost (changes requested):

tests/ctst/world/Zenko.ts
KazToozs: `SuperAdmin.getAccount()` now returns `Promise<Command>`, as that is the most compatible return value for how the code works on cli-testing's side.

tests/ctst/world/Zenko.ts
KazToozs: TL;DR below, most of this is rubber ducking:
What I'm foreseeing here is another issue.
Let's take this example: `parameters.StorageManagerUsername`.
The issue here is we are casting to a string.
So to fix this, we put :
```
interface ZenkoWorldParameters {
   StargeManagerUsername?: string;
}
```
The cast is still necessary to avoid TypeScript complaining that the variable may be undefined, as the linter doesn't like use of non-null assertion (`!` operator).
For strings this can be avoided by defaulting the parameter to empty string I suppose, but number values such as for ports, or booleans such as `ssl` I'm not so sure.
I feel like the purpose of these parameters are to allow for flexible inputting to the World object, so I somewhat disagree with the notion that we should be hard typing this, but I could be wrong.


TL;DR
In any case, after rubber ducking here and exploring the parameters we use, everything is pretty much a string apart from ssl, maybe just something like this will work:
```
interface ZenkoWorldParameters {
   [key: string]: string;
   ssl: boolean;
}
```
And any other variable that has a different type can be specified, what do you think?

Edit: this does not work, as TS doesn't allow `ssl` as a key have value type `boolean` next to `[key: string]: string;`, so the solution seems to have a variable name for each expected variable, which I'm not wholly on board with.

tests/ctst/world/Zenko.ts
KazToozs: The problem is the account object has a type of 
```
AccountObject {
    id?: string;
    name?: string;
    arn?: string;
}
```
so `id` can be undefined, so must be cast here. Do you suggest in CTST making these variables non optional?

tests/ctst/package.json
KazToozs: this was an oopsie on my part, didn't mean to commit this line. 

but yes, I will add the CI call in the last PR ZENKO-4525

tests/ctst/world/Zenko.ts
KazToozs: Upgraded cucumber version

tests/ctst/world/Zenko.ts
KazToozs: addressed by adding world [parameters](https://github.com/cucumber/cucumber-js/blob/main/docs/support_files/world.md#world-parameters-1) with cucumber

tests/ctst/world/Zenko.ts
KazToozs: This one took some searching to figure out lol
https://github.com/scality/Zenko/pull/1818/commits/aabb5c887a39f344bdb5052278b42ee3a9e97124

tests/ctst/world/Zenko.ts
KazToozs: Done when adding World parameters from cucumber bump

tests/ctst/world/Zenko.ts
```diff
@@ -665,16 +810,21 @@ ${JSON.stringify(policy)}\n${err.message}\n`);
...
-            return { statusCode: response.status, data: response.data }
-        } catch (err: any) {
-            return { statusCode: err.response.status, err: err.response.data };
+            return { statusCode: response.status, data: response.data as unknown };
+        } catch (err: unknown) {
+            return {
+                statusCode:
+                    (err as { response: { status: string } }).response.status,
+                err: (err as { response: { data: unknown } }).response.data,
+            };
```
ghost: could this work?

```suggestion
        } catch (err: { response: { status: string } }) {
            return {
                statusCode: err.response.status,
                err: err.response.data,
            };
```

tests/ctst/world/Zenko.ts
```diff
@@ -546,15 +651,18 @@ ${JSON.stringify(policy)}\n${err.message}\n`);
...
         const accessKey = await IAM.createAccessKey(this.getCommandParameters());
-        this.parameters.IAMSession = JSON.parse(accessKey.stdout).AccessKey;
-        this.cliMode.parameters.IAMSession = this.parameters.IAMSession;
-        CacheHelper.parameters.IAMSession = this.parameters.IAMSession;
+        this.parameters.IAMSession =
+            (JSON.parse(accessKey.stdout) as { AccessKey: string })?.AccessKey;
+        this.cliMode.parameters.IAMSession =
+            this.parameters.IAMSession as ClientOptions['IAMSession'];
+        (CacheHelper.parameters as { IAMSession: ClientOptions['IAMSession'] }).IAMSession =
+            this.parameters.IAMSession as ClientOptions['IAMSession'];
```
ghost: We can define IAMSession in `ZenkoWorldParameters` as a string to avoid `as` here, as `ClientOptions['IAMSession']` is a string. (same for all other appearances of this)

tests/ctst/world/Zenko.ts
```diff
@@ -546,15 +651,18 @@ ${JSON.stringify(policy)}\n${err.message}\n`);
     async prepareIamUser() {
         this.saved.userName = `iamusertest${Utils.randomString()}`;
         // Create IAM user
-        this.addCommandParameter({ userName: this.saved.userName });
+        this.addCommandParameter({ userName: this.saved.userName as string });
```
ghost: Is this really needed for the `this.saved` object that is defined as `[key: string]: any`? It should accept anything, and we should not enforce the `any` rule outside of Zenko (i.e., in CTST).

tests/ctst/world/Zenko.ts
```diff
@@ -497,20 +598,21 @@ export default class Zenko extends World {
...
                 this.addCommandParameter({ policyDocument: IAMUserPolicy });
             }
             const policy = await IAM.createPolicy(this.getCommandParameters());
             const account = await SuperAdmin.getAccount({
-                name: this.parameters.AccountName || Constants.ACCOUNT_NAME,
-            });
-            let policyArn = `arn:aws:iam::${account.id}:policy/IAMUserPolicy-${Zenko.IAMUserName}}`;
+                accountName: this.parameters.AccountName || Constants.ACCOUNT_NAME,
+            }) as Utils.Account['account'];
+            let policyArn = `arn:aws:iam::${account.id as string}:policy/IAMUserPolicy-${Zenko.IAMUserName}}`;
```
ghost: if `account` is already casted, the `as` in line 609 should not be needed, as it should be part of the `account` type already. If it complains because id can be null, you can do

```ts
            let policyArn = `arn:aws:iam::${account.id!}:policy/IAMUserPolicy-${Zenko.IAMUserName}}`;
```

(same for all related)

tests/ctst/world/Zenko.ts
```diff
@@ -424,35 +508,46 @@ export default class Zenko extends World {
...
-     * @param {object} parameters - the client-provided parameters
+     * @param {Object.<string,*>} parameters - the client-provided parameters
      * @returns {undefined}
      */
-    static async init(parameters: any) {
+    static async init(parameters: Record<string, unknown>) {
+        CacheHelper.parameters ??= {};
         if (!CacheHelper.accountAccessKeys) {
+            // TODO will need to see how VaultClient is typed
+            // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
```
ghost: Still needed?

tests/ctst/world/Zenko.ts
```diff
@@ -304,92 +377,100 @@ export default class Zenko extends World {
...
+            CacheHelper.parameters.SecretKey = account2Credentials.value as string;
 
             accountToBeAssumedFrom = account2.account;
         }
 
         // Creating a user in the account to be assumed from
         this.resetCommand();
-        const userName = `${accountToBeAssumedFrom.name}${Constants.USER_NAME_TEST}${Utils.randomString()}`;
-        this.addCommandParameter({userName});
+        const userName = `${accountToBeAssumedFrom.name as string}${Constants.USER_NAME_TEST}${Utils.randomString()}`;
```
ghost:
```suggestion
        const userName = `${accountToBeAssumedFrom.name!}${Constants.USER_NAME_TEST}${Utils.randomString()}`;
```

tests/ctst/world/Zenko.ts
ghost: See my comment from the latest review about the use of `!`

tests/ctst/world/Zenko.ts
```diff
@@ -197,13 +257,14 @@ export default class Zenko extends World {
      * @returns {undefined}
      */
     async prepareARWWI(ARWWIName: string, ARWWIPassword: string, ARWWITargetRole: string) {
+
         if (!(ARWWIName in CacheHelper.ARWWI)) {
-            const token = await this.getWebIdentityToken(
+            const token: string = await this.getWebIdentityToken(
```
ghost: We can specify this type in the function instead
```suggestion
            const token = await this.getWebIdentityToken(
```

tests/ctst/world/Zenko.ts
KazToozs: A specific TS error states:
`Catch clause variable type annotation must be 'any' or 'unknown' if specified.ts(1196)`

tests/ctst/world/Zenko.ts
KazToozs: `ClientOptions['IAMSession']` here to me is an object as follows:
```
IAMSession?: {
        AccessKeyId: string;
        SecretAccessKey: string;
        SessionToken: string;
    };
```
Am I missing something?

tests/ctst/world/Zenko.ts
KazToozs: Indeed it is not necessary for the linter; removing.

tests/ctst/world/Zenko.ts
KazToozs: Using the exclamation mark operator creates warnings from the linter, not that I agree with the rule as it would've made things easier to remove it in many places. 
```
  609:45  warning  Forbidden non-null assertion  @typescript-eslint/no-non-null-assertion
```
What do you think?

tests/ctst/world/Zenko.ts
KazToozs: same as above

tests/ctst/world/Zenko.ts
KazToozs: No, thanks for the catch; removing

tests/ctst/world/Zenko.ts
```diff
@@ -134,47 +193,48 @@ export default class Zenko extends World {
      * @returns {undefined}
      */
     async setupEntity(entityType: string): Promise<void> {
-        const keycloakPassword = this.parameters.KeycloakPassword || '123';
-        const savedParameters = JSON.parse(JSON.stringify(this.cliOptions));
+        const usedParameters = this.parameters;
```
XinLi0207: the name `usedParameters` is not very appropriate, maybe let's call it `defaultParameters`

XinLi0207 (approved):

tests/ctst/.eslintrc.js
```diff
@@ -50,7 +50,7 @@ module.exports = {
             },
         }],
         'linebreak-style': [1, 'unix'],
-        'new-cap': [1, { newIsCap: true }],
+        'new-cap': 'off',
```
ghost, line 53: I would like to see a linter step in the `build-ctst-test-image` step (I am not able to see/don't remember where we currently check that, but it should be done before any Zenko deployment), so we can quickly get notified about any error when pushing new code

tests/ctst/world/Zenko.ts
ghost: You can define this type in `ZenkoWorldParameters` to avoid multiple `as`, in general it's better to define the variables with their type than casting them everytime we use them

tests/ctst/world/Zenko.ts
ghost: I think we can just disable this rule for this particular line, it feels strange to cast to a specific object like that

tests/ctst/world/Zenko.ts
ghost: You can disable this linter rule, we cannot control all our dependencies so at some point if we reject `any`s and all `!` we'll be blocked, or we'll have to rely on `as` which is not a good practise. It is fine disabling this rule, but keeping the `any` one.

tests/ctst/world/Zenko.ts
```diff
@@ -134,47 +193,48 @@ export default class Zenko extends World {
      * @returns {undefined}
      */
     async setupEntity(entityType: string): Promise<void> {
-        const keycloakPassword = this.parameters.KeycloakPassword || '123';
-        const savedParameters = JSON.parse(JSON.stringify(this.cliOptions));
+        const defaultParameters = this.parameters;
+        const keycloakPassword = defaultParameters.KeycloakTestPassword as string || '123';
```
ghost: We can define `KeycloakTestPassword` as string in `ZenkoWorldParameters` and then

```suggestion
        const keycloakPassword = defaultParameters.KeycloakTestPassword || '123';
```

tests/ctst/world/Zenko.ts
```diff
@@ -213,20 +274,25 @@ export default class Zenko extends World {
...
-            let roleToAssume = '';
-            if (roles.ListOfRoleArns) {
-                roleToAssume = roles.find(
-                    (roleArn: string) => roleArn.includes(ARWWITargetRole) && roleArn.includes(account.id),
+            // Get the first role with the storage-manager-role name
+            const data: GetRolesForWIResponse =
+                (await SuperAdmin.getRolesForWebIdentity(this.options.webIdentityToken)).data as GetRolesForWIResponse;
+            let roleToAssume: string | undefined = '';
+
+            // TODO see necessity of if check: from my understanding, and from what I've seen in both Vault projects
```
ghost: to be removed maybe? I think this test is needed

tests/ctst/world/Zenko.ts
```diff
@@ -240,17 +306,23 @@ export default class Zenko extends World {
...
+                if (parsedOutput && parsedOutput.Credentials) {
+                    this.parameters.AssumedSession = parsedOutput.Credentials;
+                }
             } else {
                 throw new Error('Error when trying to Assume Role With Web Identity.');
             }
             // Save the session for future scenarios (increases performance)
-            CacheHelper.ARWWI[ARWWIName] = this.parameters.AssumedSession;
-            this.cliMode.parameters.AssumedSession = CacheHelper.ARWWI[ARWWIName];
+            CacheHelper.ARWWI[ARWWIName] = this.parameters.AssumedSession as ClientOptions['AssumedSession'];
```
ghost: Same here, it can be set in the paramaters type above, to avoid casting here

tests/ctst/world/Zenko.ts
```diff
@@ -240,17 +306,23 @@ export default class Zenko extends World {
...
+            this.cliMode.parameters.AssumedSession =
+                CacheHelper.ARWWI[ARWWIName];
             this.cliMode.assumed = true;
         } else {
-            this.parameters.AssumedSession = CacheHelper.ARWWI[ARWWIName];
-            this.cliMode.parameters.AssumedSession = CacheHelper.ARWWI[ARWWIName];
+            this.parameters.AssumedSession =
+                CacheHelper.ARWWI[ARWWIName] as object;
```
ghost: This should already be of type `ClientOptions['AssumedSession']`, so maybe the caste is not needed

tests/ctst/world/Zenko.ts
```diff
@@ -304,92 +377,100 @@ export default class Zenko extends World {
...
-            name: this.parameters.AccountName || Constants.ACCOUNT_NAME,
-        });
-        Zenko.additionalAccountsCredentials[account.name] = {
-            AccessKey: CacheHelper.parameters.AccessKey,
-            SecretKey: CacheHelper.parameters.SecretKey,
+            accountName: this.parameters.AccountName || Constants.ACCOUNT_NAME,
+        }) as Utils.Account['account'];
+        Zenko.additionalAccountsCredentials[account.name as string] = {
+            AccessKey: CacheHelper.parameters?.AccessKey as string,
+            SecretKey: CacheHelper.parameters?.SecretKey as string,
```
ghost: Same here about the type to be defined above and then:

```suggestion
            AccessKey: CacheHelper.parameters?.AccessKey,
            SecretKey: CacheHelper.parameters?.SecretKey,
```

tests/ctst/world/Zenko.ts
```diff
@@ -304,92 +377,100 @@ export default class Zenko extends World {
...
+            const account2Credentials = await SuperAdmin.generateAccountAccessKey({
+                accountName: account2.account.name,
+            }) as Utils.AccessKeys;
 
             // Set the credentials of the second account as the default credentials
-            CacheHelper.parameters.AccessKey = account2Credentials.id;
-            CacheHelper.parameters.SecretKey = account2Credentials.value;
+            CacheHelper.parameters ??= {};
+            CacheHelper.parameters.AccessKey = account2Credentials.id as string;
+            CacheHelper.parameters.SecretKey = account2Credentials.value as string;
```
ghost:
```suggestion
            CacheHelper.parameters.AccessKey = account2Credentials.id!;
            CacheHelper.parameters.SecretKey = account2Credentials.value!;
```

tests/ctst/world/Zenko.ts
```diff
@@ -304,92 +377,100 @@ export default class Zenko extends World {
...
-        this.addCommandParameter({userName});
+        const userName = `${accountToBeAssumedFrom.name!}${Constants.USER_NAME_TEST}${Utils.randomString()}`;
+        this.addCommandParameter({ userName });
         await IAM.createUser(this.getCommandParameters());
 
         // Creating a policy to allow it to AssumeRole
         this.resetCommand();
-        this.addCommandParameter({ policyName: `${accountToBeAssumedFrom.name}${Constants.POLICY_NAME_TEST}${Utils.randomString()}` });
+        this.addCommandParameter({
+            policyName: `${(accountToBeAssumedFrom).name as string }` +
```
ghost:
```suggestion
            policyName: `${(accountToBeAssumedFrom).name! }` +
```

tests/ctst/world/Zenko.ts
```diff
@@ -304,92 +377,100 @@ export default class Zenko extends World {
...
         this.cliMode.env = false;
 
+        CacheHelper.parameters ??= {};
         // reset the credentials of default account as the defualt credentials
-        CacheHelper.parameters.AccessKey = Zenko.additionalAccountsCredentials[account.name].AccessKey;
-        CacheHelper.parameters.SecretKey = Zenko.additionalAccountsCredentials[account.name].SecretKey;
+        CacheHelper.parameters.AccessKey =
+            Zenko.additionalAccountsCredentials[account.name as string].AccessKey;
+        CacheHelper.parameters.SecretKey =
+            Zenko.additionalAccountsCredentials[account.name as string].SecretKey;
```
ghost:
```suggestion
        CacheHelper.parameters.AccessKey =
            Zenko.additionalAccountsCredentials[account.name!].AccessKey;
        CacheHelper.parameters.SecretKey =
            Zenko.additionalAccountsCredentials[account.name!].SecretKey;
```

tests/ctst/world/Zenko.ts
```diff
@@ -424,35 +508,44 @@ export default class Zenko extends World {
...
+    static async init(parameters: Record<string, unknown>) {
+        CacheHelper.parameters ??= {};
         if (!CacheHelper.accountAccessKeys) {
             CacheHelper.adminClient = await Utils.getAdminCredentials(parameters);
 
             let account = null;
             // Create the account if already exist will not throw any error
             await SuperAdmin.createAccount({
-                name: parameters.AccountName || Constants.ACCOUNT_NAME,
+                accountName: parameters.AccountName as string || Constants.ACCOUNT_NAME,
```
ghost: to be defined above and then
```suggestion
                accountName: parameters.AccountName || Constants.ACCOUNT_NAME,
```

tests/ctst/world/Zenko.ts
```diff
@@ -424,35 +508,44 @@ export default class Zenko extends World {
...
     }
 
     /**
      * Hook Zenko is a utility function to prepare a Zenko
-     * @param {object} parameters - the client-provided parameters
+     * @param {Object.<string,*>} parameters - the client-provided parameters
      * @returns {undefined}
      */
-    static async init(parameters: any) {
+    static async init(parameters: Record<string, unknown>) {
```
ghost:
```suggestion
    static async init(parameters: ZenkoWorldParameters) {
```

tests/ctst/world/Zenko.ts
```diff
@@ -527,12 +627,15 @@ ${JSON.stringify(policy)}\n${err.message}\n`);
                 throw new Error(`Error creating the IAM User's access key.\n
                  ${accessKey.err}`);
             }
-            this.parameters.IAMSession = extractPropertyFromResults(accessKey, 'AccessKey');
-            this.cliMode.parameters.IAMSession = this.parameters.IAMSession;
+            this.parameters.IAMSession =
+                extractPropertyFromResults(accessKey, 'AccessKey');
+            this.cliMode.parameters.IAMSession =
+                this.parameters.IAMSession as ClientOptions['IAMSession'];
```
ghost: Can be defined above

tests/ctst/world/Zenko.ts
```diff
@@ -591,7 +697,7 @@ ${JSON.stringify(policy)}\n${err.message}\n`);
      * @param {object} param - an object with a key and a value
      * @returns {undefined}
      */
-    addCommandParameter(param: AwsCliObjectParameters): void {
+    addCommandParameter(param: Record<string, unknown>): void {
```
ghost, line 706: I don't get why this change is needed as the previous type `AwsCliParameters` was already embedding everything? ANything not in this type means something not supported.

tests/ctst/.eslintrc.js
KazToozs: Yes, but again, I think this should be done on the last linting PR
https://github.com/scality/Zenko/pull/1821

ghost: I would prefer having it starting this PR, and only enable the checks for the files currently linted, and then include more files in each PR, otherwise I cannot approve any PR with confidence that the linting is complete

(you can also rebase your PR to benefit from all the recent improvements in the CI, it will make your builds a bit more reliable and CTST will be way faster)

tests/ctst/world/Zenko.ts
KazToozs: Here, the parameters type is defined in ctst as `parameters : Record<string, unknown> | undefined = undefined;`.
While `CacheHelper.parameters` is indeed set to `this.parameters` in the `init`, the linter doesn't recognize this, so here the casting seems necessary.

tests/ctst/world/Zenko.ts
KazToozs: Not sure I understand what you mean? If we look at the initial typing:
```
interface AwsCliObjectParameters {
    [key: string]: number | string | undefined | object;
```
Why is this better than a `Record<string, unknown>` ?
A Record also has better compatibility with potential assignments such as 
```
        this.addCommandParameter({ userName: this.saved.userName });
```
Where the input `this.saved.userName` is an `unknown`

tests/ctst/world/Zenko.ts
KazToozs: Seems to be same problem as [here](https://github.com/scality/Zenko/pull/1818#discussion_r1179232321)

tests/ctst/world/Zenko.ts
KazToozs: b7460b1c75e76aefa5df57099a8cbecb7c46e999

tests/ctst/world/Zenko.ts
KazToozs: b7460b1c75e76aefa5df57099a8cbecb7c46e999

tests/ctst/world/Zenko.ts
KazToozs: b7460b1c75e76aefa5df57099a8cbecb7c46e999

tests/ctst/world/Zenko.ts
KazToozs: b7460b1c75e76aefa5df57099a8cbecb7c46e999

tests/ctst/world/Zenko.ts
KazToozs: 02d2a7109fdea473e90ed07623ee8b058e4ab026

tests/ctst/world/Zenko.ts
KazToozs: b7460b1c75e76aefa5df57099a8cbecb7c46e999

tests/ctst/world/Zenko.ts
KazToozs: b7460b1c75e76aefa5df57099a8cbecb7c46e999

tests/ctst/world/Zenko.ts
KazToozs: b7460b1c75e76aefa5df57099a8cbecb7c46e999

tests/ctst/world/Zenko.ts
KazToozs: b7460b1c75e76aefa5df57099a8cbecb7c46e999

tests/ctst/world/Zenko.ts
KazToozs: b7460b1c75e76aefa5df57099a8cbecb7c46e999

tests/ctst/world/Zenko.ts
KazToozs: bbd52cc625f601a7a4b292ad5358b61912d3b551

tests/ctst/world/Zenko.ts
KazToozs: b7460b1c75e76aefa5df57099a8cbecb7c46e999

.github/workflows/end2end.yaml
```diff
@@ -318,10 +319,37 @@ jobs:
...
+      # #   run: git clone git@github.com:scality/Zenko.git
+      # - name: Test git token on private repo
+      #   run: git clone git@github.com:scality/vault2.git
+      
+      - name: Install subfolder dependencies
+        working-directory: tests/ctst
+        run: yarn install
+      - name: Lint subfolder
+        working-directory: tests/ctst
+        run: yarn lint
```
anurag4DSB: The workflow step says `build-ctst-test-image`. We are doing much more than that in this step, hence the name can be mis-leading and someone facing the issue will have to check the this workflow to figure out whats going wrong.
Either we should rename the step of use another one for this static test.

Similarly, we should change the name from `name: Lint subfolder` to something more logical as reading this I do not know which subfolder and will have to check the workflow.
May be `name: Lint ctst tests`

.github/workflows/end2end.yaml
```diff
@@ -318,10 +319,37 @@ jobs:
...
+          git config --global url.https://x-access-token:${{ secrets.GIT_ACCESS_TOKEN }}@github.com/.insteadOf ssh://git@github.com/
+        env:
+          GIT_ACCESS_TOKEN: ${{ secrets.GIT_ACCESS_TOKEN }}
+        
+      # - name: Test git token on cli-testing repo
+      #   run: git clone git@github.com:scality/cli-testing.git
+      # # - name: Test git token on public repo
+      # #   run: git clone git@github.com:scality/Zenko.git
+      # - name: Test git token on private repo
+      #   run: git clone git@github.com:scality/vault2.git
```
anurag4DSB: These comments need to be removed as well.

anurag4DSB: I knot cleanup is already in progress, just commented on what I saw currently.

.github/workflows/end2end.yaml
```diff
@@ -317,11 +317,34 @@ jobs:
...
+        working-directory: tests/ctst
+        run: yarn lint
+
   build-ctst-test-image:
     runs-on: ubuntu-20.04
     steps:
       - name: Checkout
         uses: actions/checkout@v3
+        with:
+          token: ${{ secrets.GIT_ACCESS_TOKEN }}
```
ghost: I think we should set the static tests as a condition to run the ctst build, if we have an error, we don't run CTST. If there is a linting error, a new commit will be required anyway, so there is no gain to logically separating it from the build (i.e. a build error or a linter error = no ctst tests started).
I would also prefer to see this step in this build-ctst-test-image rather than a separate step in this case. You can then rename it `lint-and-build-ctst`.

tests/ctst/world/Zenko.ts
ghost: In this case, it should be better to define this as a record in CTST directly, as this will benefit all other projects using it in the future.

tests/ctst/world/Zenko.ts
```diff
@@ -213,20 +282,23 @@ export default class Zenko extends World {
             }
             // Getting account ID
             const account = await SuperAdmin.getAccount({
-                name: this.parameters.AccountName || Constants.ACCOUNT_NAME,
-            });
+                accountName: this.parameters.AccountName || Constants.ACCOUNT_NAME,
+            }) as Utils.AccountObject;
```
ghost: Reopening a new one here, this cast must be done in CTST, not Zenko, for the function to return the right "Account" type. No need for a 'as' here. This is to be specified on CTST side, in https://github.com/scality/cli-testing/blob/main/utils/api/SuperAdmin.ts#L52

These functions are relying on VaultClient, not aws-cli, hence they do not return anything like stderr, stdout, but the actual object. In this case (and this is different from the other IAM and S3 APIs), we should return the right type, but not `Command`. This should be updated to return either an Account type, or an error/something empty in case of error. Then here, we can check for the success or failure of this function call. For now, there is no such check, and the `as` invalidates the typescript checks on the `account` properties, i.e. `Property 'id' does not exist on type 'Command'.ts(2339)`.

ghost: In Zenko, the `as` should only be necessary when parsing outputs (like JSON.parse).

tests/ctst/world/Zenko.ts
```diff
@@ -275,7 +353,8 @@ export default class Zenko extends World {
         path: string,
         clientId: string,
         grantType: string,
-    ) {
+    ): Promise<string> {
+        this.parameters;
```
ghost, line 349: not sure this line is expected

tests/ctst/world/Zenko.ts
```diff
@@ -616,7 +728,7 @@ ${JSON.stringify(policy)}\n${err.message}\n`);
 
     /**
      * Get all mapped parameters
-     * @returns {AwsCliObjectParameters} - an object with the cli command options
+     * @returns {Record<string, unknown>} - an object with the cli command options
```
ghost, line 731: This should return an object of type `AwsCliOptions` (not just as a jsdoc). This is the one I wanted to comment earlier but I commented the wrong one.
This type has all the elements we support, and is being expected by all the IAM and S3 functions of CTST as inputs. Hence,  we should use this type to avoid unecessary casts like in 
https://github.com/scality/Zenko/pull/1818/files#diff-371f6719b5883f728d9d3218b4ed2eaf1b7ada8faf798acf95b686a7c347c7ddR408
https://github.com/scality/Zenko/pull/1818/files#diff-371f6719b5883f728d9d3218b4ed2eaf1b7ada8faf798acf95b686a7c347c7ddR447
https://github.com/scality/Zenko/pull/1818/files#diff-371f6719b5883f728d9d3218b4ed2eaf1b7ada8faf798acf95b686a7c347c7ddR460
https://github.com/scality/Zenko/pull/1818/files#diff-371f6719b5883f728d9d3218b4ed2eaf1b7ada8faf798acf95b686a7c347c7ddR467

And also all the test files, for the upcoming PRs.

tests/ctst/world/Zenko.ts
```diff
@@ -462,20 +561,25 @@ export default class Zenko extends World {
                 CacheHelper.isPreloadedAccount = true;
             } else {
                 const accessKeys = await SuperAdmin.generateAccountAccessKey({
-                    name: parameters.AccountName || Constants.ACCOUNT_NAME,
+                    accountName: parameters.AccountName  || Constants.ACCOUNT_NAME,
```
ghost:
```suggestion
                    accountName: parameters.AccountName || Constants.ACCOUNT_NAME,
```

ghost (commented):
LGTM otherwise

tests/ctst/world/Zenko.ts
KazToozs: The initial type didn't come from CTST, it was declared at the top of the Zenko world file just to facilitate typing. I think you might be confusing `AwsCliObjectParameters` which was used for this typing and `AwsCliParameters` which is used within CTST specifically, but also `AwsCliOptions` which IS the type that embeds all possible parameters (generated from API). 

If I add a `[key: string]: string` to the generator for the `AwsCliOptions` type, I can use it as parameter type here, what do you think? 
While at it, maybe I could replace the `AwsCliParameters` occurences/functionality in CTST with `AwsCliOptions`, but it might be a bit of work (and out of scope); atleast worth a ticket?

ghost: Please see https://github.com/scality/Zenko/pull/1818#discussion_r1184762254 for the initial comment, I was commenting the wrong line, it indeed does not make sense for this command, but it does for `getCommandParameters`.

The new point here is that we can rework this type you added in CTST to be more flexible, to avoid casting here

ghost (approved):

KazToozs:
ping

KazToozs:
ping

KazToozs:
ping

ghost:
ping

PR #1819, opened 2023-03-14
Improvement/zenko 4523 lint common files

KazToozs:
this 0859738976080339bb576b839db3ce4b1d398487 commit related to installing a package to make the CI functional might be better in a different PR as it's not originally in the scope of this one, thoughts?

### DIFF
--- a/tests/ctst/common/common.ts
+++ b/tests/ctst/common/common.ts
@@ -1,17 +1,18 @@
 import { Given, setDefaultTimeout } from '@cucumber/cucumber';
 import { Constants, S3, Utils } from 'cli-testing';
+import Zenko from 'world/Zenko';
 import { extractPropertyFromResults } from './utils';
 
 setDefaultTimeout(Constants.DEFAULT_TIMEOUT);
 
-Given('a {string} bucket', async function (versioning) {
+Given('a {string} bucket', async function (this: Zenko, versioning: string) {
     this.resetCommand();
     const preName = this.parameters.AccountName || Constants.ACCOUNT_NAME;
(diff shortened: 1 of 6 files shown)


tests/ctst/common/common.ts
```diff
@@ -1,59 +1,64 @@
...
+                '"ObjectLockEnabled": "Enabled",' +
+                '"Rule": {' +
+                '"DefaultRetention":' +
+                `{ "Mode": "${retentionMode}", "Days": 50 }}}`,
         });
         await S3.putObjectLockConfiguration(this.getCommandParameters());
     }
 });
 
+// eslint-disable-next-line new-cap
```
ghost: Same as in other PR, we can disable this globally for CTST

tests/ctst/common/hooks.ts
```diff
@@ -4,15 +4,22 @@ import Zenko, { EntityType } from '../world/Zenko';
 // HTTPS should not cause any error for CTST
 process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0';
 
-const worlds : any = {
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const worlds : any = { // TODO: World doesn't work here becuase Zenko is missing some properties, what to do ?
     ZENKO: Zenko,
 };
```
ghost: It's expected with TS that extending a class won't work when trying to infer types compatibility. I'm not sure we can easily handle these eslint rules in this hook file, which is a bit special regarding which World to use. I would argue that it's maybe even not needed to route the right `World` as I don't see the need for any other world, as we are in Zenko...

ghost (approved):

welansari (approved):

KazToozs:
ping

PR #1820, opened 2023-03-14
Improvement/zenko 4524 lint iam policies

KazToozs:
Linting of Iam policy files

PR #1821, opened 2023-03-14
Improvement/zenko 4525 lint other tests

KazToozs:
~~Discussion with @williamlardier asks whether we should remove the `@typescript-eslint/no-unsafe-assignment` rule from the linter~~

### DIFF
--- a/tests/ctst/package.json
+++ b/tests/ctst/package.json
@@ -23,6 +23,6 @@
     "eslint": "^8.28.0"
   },
   "scripts": {
-    "lint": "eslint ./world/Zenko.ts ./common/* ./steps/iam-policies/*"
+    "lint": "eslint ."
   }
 }
--- a/tests/ctst/steps/notifications.ts
+++ b/tests/ctst/steps/notifications.ts
@@ -1,6 +1,11 @@
-import { Then, Given, When, setDefaultTimeout, IWorld } from '@cucumber/cucumber';
+import { Then, Given, When, setDefaultTimeout } from '@cucumber/cucumber';
 import { strict as assert } from 'assert';
+import Zenko, {
+    ApiResult,
+    AWSVersionObject,
+} from '../world/Zenko';
 import { Constants, S3, Utils, KafkaHelper } from 'cli-testing';
+import { Message } from 'node-rdkafka';
 
 setDefaultTimeout(Constants.DEFAULT_TIMEOUT);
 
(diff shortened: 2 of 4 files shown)


tests/ctst/steps/notifications.ts
```diff
@@ -49,41 +76,51 @@ async function emptyVersionedBucket(world: IWorld) {
     }));
 }
 
-async function putObject(world: IWorld) {
+async function putObject(world: Zenko) {
```
ghost, line 84: Mixed with [the logic we currently have](https://github.com/scality/Zenko/pull/1820/files#diff-504fd6b6f773cd1c920c1a4789386463c448655cdeb9ca7395ae4108cd787d6aR8), the World may not always be Zenko...

tests/ctst/steps/notifications.ts
```diff
@@ -49,41 +76,51 @@ async function emptyVersionedBucket(world: IWorld) {
...
     world.addCommandParameter({ key: objName });
-    world.addCommandParameter({ copySource: `${world.saved.bucketName}/${world.saved.objectName}` });
-    world.saved.objectName = objName;
+    world.addCommandParameter({
+        copySource:
+            `${world.getSaved().bucketName as string}/${world.getSaved().objectName as string}`,
+    });
+    // TODO this rule honestly just seems dumb, especially in this case
+    // eslint-disable-next-line no-param-reassign
+    world.getSaved().objectName = objName;
```
ghost: We should have a `setSaved()` method in this case, it does not feel standard to edit the returned variable of a `get` function in the same line

The approach before looked more standard

tests/ctst/package.json
ghost, line 1: About the linter rule: if it works well today, it's ok to keep it. With the latest changes you enforces a code style that averyone must follow, so this should be all good. Let's wait and see if the rule is too restrictive

tests/ctst/world/Zenko.ts
```diff
@@ -169,8 +170,8 @@ export default class Zenko extends World<ZenkoWorldParameters> {
      * @param {object} result - contains both stderr and stdout from the CLI command.
      * @returns {boolean} - if the result is a success or a failure
      */
-    checkResult(result: Utils.Command | Utils.Command[]): boolean {
-        const usedResult: Utils.Command[] = Array.isArray(result) ? result : [result];
+    checkResult(result: ApiResult | ApiResult[]): boolean {
+        const usedResult: ApiResult[] = Array.isArray(result) ? result : [result];
```
ghost: Alternatively we could rename it `checkResults` and only accept arrays, so we can only call it with something like `checkResult([myResultObject])` if we only have one

tests/ctst/steps/notifications.ts
```diff
@@ -92,168 +131,182 @@ async function putTag(world: IWorld) {
...
-    this.saved.notificationDestinations.push({
-        destinationName: this.parameters.NotificationDestination,
-        topic: this.parameters.NotificationDestinationTopic,
-        hosts: this.parameters.KafkaHosts,
+Given('two notification destinations', function (this: Zenko) {
+    const notificationDestinations = [];
+    notificationDestinations.push({
+        destinationName: (this.parameters as { NotificationDestination: string }).NotificationDestination,
+        topic: (this.parameters as { NotificationDestinationTopic: string }).NotificationDestinationTopic,
+        hosts: (this.parameters as { KafkaHosts: string }).KafkaHosts,
```
ghost: Can we better type these variables using what is defined in the Zenko world?

ghost (approved):

XinLi0207 (approved):

KazToozs:
ping

PR #1822, opened 2023-03-16
Bugfix/zenko 4527 fix mongodb deployment and upgrade charts

ghost:
- Bump mongodb charts to latest 4.2
- Bump mongodb sharded images to latest 4.2 images from bitnami

See https://github.com/scality/artesca/pull/1641 for the changes in Artesca to handle these new charts.


alexanderchan-scality (approved):

solution-base/mongodb/charts/mongodb-sharded/templates/config-server/config-server-statefulset.yaml
```diff
@@ -208,8 +208,7 @@ spec:
           readinessProbe:
             exec:
               command:
-                - mongo
-                - --disableImplicitSessions
+                - mongosh
```
anurag4DSB, line 211: I am curious as to why did we have this: `--disableImplicitSessions`

solution-base/mongodb/charts/mongodb-sharded/templates/mongos/mongos-service.yaml
```diff
@@ -19,6 +19,9 @@ spec:
   {{- if and (eq .Values.service.type "ClusterIP") .Values.service.clusterIP }}
   clusterIP: {{ .Values.service.clusterIP }}
   {{- end }}
+  {{- if .Values.service.externalIPs }}
+  externalIPs: {{- toYaml .Values.service.externalIPs | nindent 4 }}
+  {{- end }}
```
anurag4DSB, line 24: I might be missing the context, so please help me
When would we need specific external IP addresses

solution-base/mongodb/charts/mongodb-sharded/templates/config-server/config-server-statefulset.yaml
ghost: All of this is coming from the official bitnami charts. We need to change them manually because we do not use helm in the Artesca cluster, so even if the kubernetes resources are "helm-managed", this is something we handle here.

here it's coming from https://github.com/bitnami/charts/blob/9f7ac81da075932c057a406970154f73bbdd33cd/bitnami/mongodb-sharded/templates/config-server/config-server-statefulset.yaml#L211

We only change them when we need specific configuration for our product, but for that we use patch files (from Alex's commits).

solution-base/mongodb/charts/mongodb-sharded/templates/mongos/mongos-service.yaml
ghost: I think we don't need them today, it's a mapper that links to any configuration we might need for this purpose, coming from the Helm charts, but that doesn't mean we use any

anurag4DSB (approved):

PR #1838, opened 2023-04-07
ZENKO-4363 add azure archive tests

welansari:
Issue: ZENKO-4363

### DIFF
--- a/.github/scripts/end2end/configs/zenko.yaml
+++ b/.github/scripts/end2end/configs/zenko.yaml
@@ -94,6 +94,14 @@ spec:
       e2e-cold:
         debug: "true"
         command-timeout: "60s"
+    server:
+      tar:
+        targetSize: 10KB
+        maxObjects: 3
+        maxDelay: "5s"
+      azure:
+        archiveTier: "hot"
+        pollingDelay: "1s"
   ingress:
     workloadPlaneClass: 'nginx'
     controlPlaneClass: 'nginx'
--- a/tests/ctst/features/azureArchive.feature
+++ b/tests/ctst/features/azureArchive.feature
@@ -1,0 +1,124 @@
+Feature: Azure Archive
+
+    @2.7.0
+    @PreMerge
+    @Flaky
+    @AzureArchive
+    Scenario Outline: Archive objects when timeout is reached
+    Given a "<versioningConfiguration>" bucket
+    And a transition workflow to "e2e-azure-archive" location
+    And <objectCount> objects "obj" of size <objectSize> bytes
+    Then the storage class of object "obj-1" must become "e2e-azure-archive"
+    And the storage class of object "obj-2" must become "e2e-azure-archive"
+    And manifest containing object "obj-1" should "contain" object "obj-2"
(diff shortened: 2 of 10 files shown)


tests/ctst/features/azureArchive.feature
```diff
@@ -0,0 +1,129 @@
...
+        # |               Suspended |
+
+    @2.7.0
+    @PreMerge
+    @Flaky
+    @AzureArchive
+    Scenario Outline: Respect maximum number of objects per archived Tar
+    Given a "<versioningConfiguration>" bucket
+    And a transition workflow to "e2e-azure-archive" location
+    And an object "obj1" of size <objectSize> bytes
```
francoisferrand: could we add another parameter to this rule:

> And <count> objects "obj" of size <objectSize> bytes

where we could either run with multiple counts (→ see different behaviors), or at least avoid repeating 5 times the same step/condition ?

tests/ctst/features/azureArchive.feature
```diff
@@ -0,0 +1,129 @@
...
+    And manifest and tar containing object "obj5" should exist
+    And manifest containing object "" should contain <packObjectCount> objects
+    And manifest and tar containing object "obj6" should exist
+    And manifest containing object "" should contain <packObjectCount> objects
+
+    Examples:
+        | versioningConfiguration | objectSize | packObjectCount |
+        |           Non versioned |          1 |               3 |
+        |               Versioned |          1 |               3 |
+        # |               Suspended |          1 |               3 |
```
francoisferrand: should we have another config for "_non-versioned object in a versioned bucket_" ?

welansari: i think we should wait for the follow up test ticket before adding this, as that case will change the semantics of the tests, we might want to make it it's own scenario

.github/scripts/end2end/configs/zenko.yaml
```diff
@@ -94,6 +94,15 @@ spec:
       e2e-cold:
         debug: "true"
         command-timeout: "60s"
+    serverConfiguration:
+      ${AZURE_ARCHIVE_BACKEND_DESTINATION_LOCATION}:
```
francoisferrand: with latest ZKOP :

```
server:
   tar:
      ...
   azure:
      ...
```

tests/ctst/features/azureArchive.feature
francoisferrand: ok, please add this to the followup ticket so we don't forget

welansari: done

solution/deps.yaml
```diff
@@ -94,7 +94,7 @@ vault:
 zenko-operator:
   sourceRegistry: registry.scality.com/zenko-operator
   image: zenko-operator
-  tag: 1.5.21
+  tag: 1.5.24
```
francoisferrand: not needed, just rebase to get latest zkop...

francoisferrand (approved):

KillianG (approved):

PR #1840, opened 2023-04-13
ZENKO-4527: disable mongosh telemetry when starting mongo

ghost:
Following mongodb bump to 4.2, mongosh eventually sends telemetry data to an external server, and causes large timeouts when using it in an offline environment. As Artesca CI is installing Zenko without internet access, the timeouts are causing the install to fail.

Relates to https://github.com/bitnami/charts/issues/10264.

The fix is being tested in Artesca CI here: https://github.com/scality/artesca/actions/runs/4699425668/jobs/8333164490


solution-base/mongodb/charts/mongodb-sharded/templates/mongos/mongos-dep-sts.yaml
```diff
@@ -84,6 +84,9 @@ spec:
             runAsNonRoot: {{ .Values.securityContext.runAsNonRoot }}
             runAsUser: {{ .Values.securityContext.runAsUser }}
           {{- end }}
+          command:
+            - /bin/bash
+            - /entrypoint/disable-telemetry.sh
```
francoisferrand: this looks fishy : we are actually _replacing_ the start script with a wrapper which disable telemetry indeed, but also runs the usual start script...

either we find another name (`mongos-entrypoint.sh` ?), or we find an alternate approach (i imagine it does not work with start scripts ; but maybe it could be done in an initContainer? or updating the mongo image?)

francoisferrand (commented):
should we not also fix the "standard" (non sharded) mongo deployement?

solution-base/mongodb/charts/mongodb-sharded/templates/mongos/mongos-dep-sts.yaml
ghost: I vote for the renaming, because changing the image would require more work everytime we need to upgrade mongodb or the charts...

About the usual mongodb, this one is not (yet) upgraded to 4.2, it still uses `mongo`, not `mongosh` hence not affected by this issue. This is indeed something to work on next, but priority is the mongodb we today use in Artesca. Also, we'll need to rework the `mongo-ci` images we use in the various CIs today (still mongodb 3.x!)

alexanderchan-scality (approved):

solution-base/mongodb/charts/mongodb-sharded/templates/replicaset-entrypoint-configmap.yaml
```diff
@@ -4,6 +4,16 @@ metadata:
   name: {{ include "common.names.fullname" . }}-replicaset-entrypoint
   labels: {{- include "common.labels.standard" . | nindent 4 }}
 data:
+  mongos-entrypoint.sh: |-
+    #!/bin/bash
+
+    . /liblog.sh
+
+    # Disable MongoSH telemtry to support offline deployments
```
francoisferrand:
```suggestion
    # Disable MongoSH telemetry to support offline deployments
```

solution-base/mongodb/charts/mongodb-sharded/templates/replicaset-entrypoint-configmap.yaml
```diff
@@ -27,4 +37,8 @@ data:
       {{- end }}
     fi
 
+    # Disable MongoSH telemtry to support offline deployments
```
francoisferrand:
```suggestion
    # Disable MongoSH telemetry to support offline deployments
```

francoisferrand (approved):

PR #1845, opened 2023-04-14
Render 3-shards mongodb manifests

rahulreddy:
This does not change the current default of deploying 1 shard by default. The provided manifests allow enabling 3 shards at some customers especially who will be migrating from Zenko 1.2


alexanderchan-scality (approved):

rachedbenmustapha (approved):

solution-base/mongodb/charts/mongodb-sharded/templates/shard/shard-data-statefulset.yaml
```diff
@@ -374,7 +374,9 @@ spec:
         {{- end }}
         {{- if $.Values.shardsvr.persistence.selector }}
         selector:
-{{ toYaml $.Values.shardsvr.persistence.selector | indent 10 }}
+          matchLabels:
+            shard: '{{ $i }}'
+{{ toYaml $.Values.shardsvr.persistence.selector.matchLabels | indent 12 }}
```
ghost, line 379: Should this goes into a patch file as wel?

solution-base/build.sh
```diff
@@ -37,6 +37,8 @@ MONGODB_SINGLE_NODE_PATH=${ISO_ROOT}/deploy/mongodb-1-node.yaml
 MONGODB_THREE_NODE_PATH=${ISO_ROOT}/deploy/mongodb-3-nodes.yaml
 MONGODB_SHARDED_SINGLE_NODE_PATH=${ISO_ROOT}/deploy/mongodb-sharded-1-node.yaml
 MONGODB_SHARDED_THREE_NODE_PATH=${ISO_ROOT}/deploy/mongodb-sharded-3-nodes.yaml
+MONGODB_SHARDED_SINGLE_NODE_THREE_SHARDS_PATH=${ISO_ROOT}/deploy/mongodb-sharded-1-node-3-shards.yaml
+MONGODB_SHARDED_THREE_NODE_THREE_SHARDS_PATH=${ISO_ROOT}/deploy/mongodb-sharded-3-nodes-3-shards.yaml
```
ghost, line 41: I am interested in having a bit more context, this will create two new yaml files for MongoDB, each containing multiple resources, right? But I am not able to see where it is used in https://github.com/scality/artesca/pull/1721/files#diff-9f3a8c4af71c5060627586aa78119e38ab963af820e4b554b5d204c369f8534cR28 (more specifically, how the nShards is linked at install and upgrade time).

I am asking because my PR to handle mongodb upgrade must also handle this case, but I do see logic in the Zenko.py file still relying on only one shard (like in https://github.com/scality/artesca/blob/e16707350054fbfa6980131206358ff57d4075c6/installer/artesca_installer/salt/_modules/zenko.py#L158), am I missing something?

PR #1848, opened 2023-04-18
Improvement/zenko 4564 improve ci

ghost:
Various tries to improve the robustness of the Zenko CI

It:

- improves the robustness of some init parts that could fail because of race conditions
- improves the ctst test suite with a better runner, allowing to use more runner, considerably reducing the build duration
- uses another runner for the backbeat tests, as it was seen that the cpu had a great impact on the tests results
- uses two replicas of vault/cloudserver pods to allow for rollout restarts that could be related to location addition, and would eventually cause test failure as the services were initially SPOFs
- retries vault tests once in case of errors.

One vault test is known to be "unstable", and is being reworked in https://github.com/scality/vault2/pull/124

Overall, the builds are much more stable now, but might still fail; I will try to keep an eye on any failure and work on them to find solutions.

Open discussion: can we benefit from more 8core runners to allow for more than 1 build at a time in the CI? To be checked with platform engineering.

### DIFF
--- a/.github/actions/install-doc-dependencies/action.yaml
+++ b/.github/actions/install-doc-dependencies/action.yaml
@@ -4,37 +4,33 @@ description: "Prepare required packages before building the documentation"
 runs:
   using: composite
   steps:
+    - name: Cache deb packages
+      id: cache-deb
+      uses: actions/cache@v3
+      with:
+        path: ~/deb_cache
+        key: ${{ runner.os }}-deb-${{ hashFiles('.github/actions/install-doc-dependencies/action.yaml') }}
+        restore-keys: |
+          ${{ runner.os }}-deb-
     - name: Install required dependencies
       shell: bash
       run: |
         set -e
 
+        PACKAGES="xindy enchant git graphviz curl latexmk make ghostscript inkscape plantuml python3 python3-buildbot-worker texlive-font-utils python3-pip texlive-fonts-extra texlive-fonts-recommended texlive-latex-extra texlive-luatex texlive-xetex texlive-latex-recommended tox emacs less"
+
         install_packages() {
+          if [ -d "$HOME/deb_cache" ]; then
+            sudo cp -R "$HOME/deb_cache/." /var/cache/apt/archives/
+          fi
+
           sudo apt-get update
-          sudo apt-get install --no-install-recommends -y \
-            xindy \
-            enchant \
-            git \
-            graphviz \
-            curl \
-            latexmk \
-            make \
-            ghostscript \
-            inkscape \
-            plantuml \
-            python3 \
-            python3-buildbot-worker \
-            texlive-font-utils \
-            python3-pip \
-            texlive-fonts-extra \
-            texlive-fonts-recommended \
-            texlive-latex-extra \
-            texlive-luatex \
-            texlive-xetex \
-            texlive-latex-recommended \
-            tox \
-            emacs \
-            less
+          sudo apt-get install --no-install-recommends -y ${PACKAGES}
+
+          if [ ! -d "$HOME/deb_cache" ]; then
+            mkdir -p "$HOME/deb_cache"
+            sudo cp -R /var/cache/apt/archives/*.deb "$HOME/deb_cache"
+          fi
         }
 
         if ! install_packages; then
--- a/.github/actions/install-doc-dependencies/action.yaml
+++ b/.github/actions/install-doc-dependencies/action.yaml
@@ -1,4 +1,3 @@
----
 name: "Install doc dependencies"
 description: "Prepare required packages before building the documentation"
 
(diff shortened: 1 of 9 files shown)


.github/actions/install-doc-dependencies/action.yaml
```diff
@@ -8,31 +8,49 @@ runs:
     - name: Install required dependencies
       shell: bash
       run: |
+        set -e
+
+        install_packages() {
+          sudo apt-get install --no-install-recommends -y \
```
alexanderchan-scality: it may be safer to add `apt-get update` here such that  it is executed per `install_packages`.

.github/actions/install-doc-dependencies/action.yaml
```diff
@@ -8,31 +8,49 @@ runs:
...
-          texlive-fonts-extra \
-          texlive-fonts-recommended \
-          texlive-latex-extra \
-          texlive-luatex \
-          texlive-xetex \
-          texlive-latex-recommended \
-          tox \
-          emacs \
-          less
+        for i in {1..3}; do
```
alexanderchan-scality: minor: is it necessary to run the install multiple times? would it not be sufficient to run once using the default 'azure.*' mirror, and if that fails, modify the source list then run the install  again?

.github/actions/install-end2end-dependencies/action.yaml
```diff
@@ -8,9 +8,33 @@ runs:
...
+
+          function retry_apt_get() {
+            retry 5 5 sudo apt-get "$@"
+          }
+
           set -exu;
-          sudo apt-get update
-          sudo apt-get install -y \
+          retry_apt_get update
+          retry_apt_get install -y \
```
alexanderchan-scality: would  these dependencies not also encounter the 'azure.*' mirror issue as well?

.github/actions/install-doc-dependencies/action.yaml
```diff
@@ -8,31 +8,49 @@ runs:
     - name: Install required dependencies
       shell: bash
       run: |
+        set -e
```
francoisferrand, line 18: for optimizing the build (speed + stability), it may be worth using github's `cache` : if we cache the content of `/var/cache/apt`, then will be more tolerent to network issues and install will be much faster (and possibly we don't need complicated retry script)

.github/actions/wait-docker-ready/action.yaml
```diff
@@ -8,11 +8,14 @@ runs:
...
-            sleep 2
-          done
-          echo "Could not reach Docker daemon from buildbot worker" >&2
-          exit 1'
+        bash -c '
+        for i in {1..150}
+        do
+          docker info &> /dev/null && exit
+          if [ $i -eq 50 ]; then
+            sudo systemctl restart docker
```
francoisferrand, line 16: that seems really fishy : docker daemon in CI is stable as it gets, there is no reason we would need to restart it....
(even waiting for it is weird : we have many CI jobs which build images, and I don't think we need this kind of retry/restart of docker)

→ seems like this whole step is a workaround for another issue, would be better to investigate/fix that issue instead of burying it

.github/scripts/end2end/run-e2e-ctst.sh
```diff
@@ -3,7 +3,7 @@ set -exu
 
 ZENKO_NAME=${1:-end2end}
 COMMAND=${2:-"premerge"}
-PARALLEL_RUNS=${3:-2}
+PARALLEL_RUNS=${3:-6}
```
francoisferrand: can this be computed based on `nproc` ?
e.g. maybe `$(( $(nproc) - 2 ))`

francoisferrand:
> Open discussion: can we benefit from more 8core runners to allow for more than 1 build at a time in the CI? To be checked with platform engineering

What do you mean?
These 8-core workers are github-managed, and thus created elastically on demand, so there is no limit here: we can run as many zenko build in parallel, each would get its own runner(s)...

.github/actions/wait-docker-ready/action.yaml
ghost: You may be right indeed, but looking at this, for example: https://github.com/scality/Zenko/actions/runs/4741155549/jobs/8417897238#step:4:163 does not show anything bad beyond the fact that we timeout trying to contact the docker daemon (usually it takes "0s" to complete). Hence this try to restart the daemon, in case something was preventing it from starting...

.github/scripts/end2end/run-e2e-ctst.sh
ghost: Yes definitely, I will try today to optimize how many vCPUs CTST can use safely (I think nproc - 3 considering what was observed on 4core runners)

ghost:
> These 8-core workers are github-managed, and thus created elastically on demand, so there is no limit here: we can run as many zenko build in parallel, each would get its own runner(s)...

Yes and no, we still have a limit (50, to be confirmed, after discussing with Thomas C.), but far higher than what I thought, I faced a case yesterday where some runners based on 8core where not starting right away, but it was related to a github issue.

.github/actions/install-doc-dependencies/action.yaml
```diff
@@ -1,38 +1,36 @@
...
+
+        install_packages() {
+          sudo apt-get update
+          sudo apt-get install --no-install-recommends -y ${PACKAGES}
+        }
+
+        if ! install_packages; then
+          echo "Switching to archive.ubuntu.com"
+          sudo sed -i 's/azure\.//' /etc/apt/sources.list
+          sudo apt-get update
```
alexanderchan-scality, line 31: minor: this line can be removed

alexanderchan-scality (approved):

XinLi0207 (approved):

.github/actions/wait-docker-ready/action.yaml
francoisferrand: never saw that in other workflows/jobs (neither the problem nor the workaround), so I am wondering why we need it at all...

.github/scripts/end2end/vault-e2e-test.sh
```diff
@@ -49,33 +47,46 @@ if [ ! -z "$VAULT_TEST_IMAGE_TAG" ] ; then
...
-  --env="KEYCLOAK_TEST_PORT=${KEYCLOAK_TEST_PORT}" \
-  --env="KEYCLOAK_TEST_REALM_NAME=${KEYCLOAK_TEST_REALM_NAME}" \
-  --env="KEYCLOAK_TEST_CLIENT_ID=${KEYCLOAK_TEST_CLIENT_ID}" \
-  --env="KEYCLOAK_TEST_GRANT_TYPE=${KEYCLOAK_TEST_GRANT_TYPE}" \
-  --env="VAULT_TEST_CONFIG=${VAULT_TEST_CONFIG}" \
-  --env="VAULT_OIDC_TEST=${VAULT_OIDC_TEST}" \
-  --command -- yarn ft_test
+# Run the tests up to two times if they fail
+exit_code=1
+for i in 1 2; do
```
francoisferrand, line 52: could we instead retry the *individual* tests? (i.e. possibly using a 'retry' flag in the js tests)

that would be faster, and since we would retry less tests we would less chance to fail on retry as well...

PR #1850, opened 2023-04-19
improvement: ZENKO-4546 storage consumption metrics bump PR

XinLi0207:

fixes #<ISSUE>


solution/deps.yaml
```diff
@@ -92,9 +91,9 @@ vault:
   tag: 8.7.4
   envsubst: VAULT_TAG
 zenko-operator:
-  sourceRegistry: registry.scality.com/zenko-operator
+  sourceRegistry: registry.scality.com/sf-eng
   image: zenko-operator
-  tag: 1.5.13
+  tag: 72390ca1afbf1aaf71b4017495b5d53b14fff33e
```
XinLi0207: will change here post release zkop

KillianG (approved):

solution/deps.yaml
ghost: cc @francoisferrand as we have [ongoing discussions](https://scality.slack.com/archives/C031XT2FDA9/p1681830380769799) on this, maybe bumping ZKOP in dev/2.6 should be avoided, can you confirm?

XinLi0207: thanks for letting me know the concern about zkop in zenko 2.6, if don't bump zkop, can we introduce[ this env](https://github.com/scality/zenko-operator/pull/373/files#diff-e6542f41494ebe04c99b56959487c4a3fcec511efddb016773806f2da7896b66R143) for pensieve-api in Zenko instead of zkop?

ghost: I don't think so... Let's wait for a confirmation first, as we'll need a solution anyway to unblock this branch... Maybe it is fine for this bump, but we need confirmation :smile:

francoisferrand: ZKOP cannot be bumped in zenko 2.6 (yet), we have a ticket for this : https://scality.atlassian.net/browse/ZKOP-248
But it will probably be done next week.

Generally, if possible i think it is best to wait until CNES release (e.g. artesca 1.7.5) is complete to merge anything in zenko 2.6, to avoid adding more stuff to that release while in RC.

ghost (changes requested):
Putting a lock here till we fix the 2.6 branch (should be fine after the next Aretsca 1.7.5 release)

ankit-gautam23 (commented):
Is it bumped??

ghost (approved):

XinLi0207:
ping

PR #1870, opened 2023-05-15
Bugfix/zenko 4583 reconfigure mongodb rollback time

ghost:
## This PR also bumps some projects to mutualize PRs

The goal of this change is to improve the user experience with mongodb, more than providing a real fix.

If one mongodb instance is down for too long today (1day by default), and some operations were done on the cluster more than 1d before the down mongodb instance comes back, an error will be thrown, where mongodb refuses to start, as the first oplog entry to recover is too old.

We must distinguish 2 scenarios here:

- All the cluster was down for more than 1 day OR only a subset of the mongodb instances were down, but no operations were completed during the downtime duration: in this case, the fix will help a bit being more flexible, allowing for a full "weekend" to complete before restarting the service, that will be able to come back without issue. In this scenario the oplog was not overriden, so having a higher rollback duration here is fine.

- A subset of the mongodb instances were down, but the oplog was overriden during the downtime duration. In this case, we have 2 more cases. A too high rollback duration could cause dead loops, with a mongodb not able to recover before the oplog is overriden, or alternatively, the oplog could be already overriden, and this change has no effect as a full init sync will be required.

To sum up, the main motivation of this change is UX, to be flexible if something went wrong during a weekend. The choice of 3 days vs 1 day is the most efficient value to avoid as much as possible the dead loops, while improving UX.

To complete this work, we document several procedures to:
- Properly shutdown Artesca (by putting mongodb in readonly mode)
- Manually change the rollback duration, if needed
- How to trigger an automated full init sync (should not be needed in Mongodb 4.4+)
(see https://github.com/scality/artesca/pull/1668)

Note 1: Tests showed that under load, the oplog is overriden in less than 1h.

Note 2: I saw strange cases in AWS with snapshots where the issue would arise or not, using the same snapshot... Even after seeting mongodb in readonly. Let's hope this behavior got fixed in 4.4, that brings some improvements on this side.


solution-base/mongodb/charts/mongodb-sharded/values.yaml
```diff
@@ -353,7 +353,9 @@ configsvr:
   ## mongodbExtraFlags:
   ##  - "--wiredTigerCacheSizeGB=2"
   ##
-  mongodbExtraFlags: []
+  mongodbExtraFlags:
+    - "--setParameter rollbackTimeLimitSecs=259200"
+
```
francoisferrand: I think we set many parameters through the makefile as well : any specific reason we set this here instead?

ghost: Because the Makefile for now does not work anymore. Charts are not stored at the same place and they put in place a 6 month retention policy AFAIK. I am planning to work on it with the 4.4 upgrade.

I can add it to a patch file though, so that we don't loose this change later.

ghost: Done here: https://github.com/scality/Zenko/pull/1870/commits/2345e954c53fad2019d10da53eae167bc5c49d2e

francoisferrand: > Because the Makefile for now does not work anymore. Charts are not stored at the same place and they put in place a 6 month retention policy AFAIK. I am planning to work on it with the 4.4 upgrade.

i meant the makefile used to render the chart and generate the Zenko-Base ISO, which does not depend on the url or 6-month retention, but just renders the chart with some specific values :

* https://github.com/scality/Zenko/blob/40ea563f558a4f75f3c50ea1415784c91091c12a/solution-base/build.sh#L128
* https://github.com/scality/Zenko/blob/40ea563f558a4f75f3c50ea1415784c91091c12a/solution-base/build.sh#L166

(looks like it is a build.sh script and not a makefile, sorry for the confusion)

ghost: Oh ok I see, then I will make the appropriate changes :smile: 

I also foudn the workaround for the Makefile, putting it here but I'll do it later:

`CHART_REPO:="https://raw.githubusercontent.com/bitnami/charts/archive-full-index/bitnami"`

Source: https://github.com/bitnami/charts/issues/10539

ghost: Done (and rebased) in https://github.com/scality/Zenko/pull/1870/commits/456d0a1acec6975ac62f58ab7ba9fdbcb6b471dc

ghost: Update after some rebases: done in https://github.com/scality/Zenko/pull/1870/commits/08c20c404a621522aaa13bfd8f7083cbefbd229d for the sharded mode only. I do not want to ointroduce more entropy in this PR, and the fix is only for an Artesca use case, so sharded.

KazToozs (approved):

francoisferrand (approved):

PR #1884, opened 2023-05-30
ZENKO-4359 azure archive suspend tests

welansari:
Add Azure Archive E2E suspend / resume tests

Issue: ZENKO-4359

### DIFF
--- a/.github/scripts/end2end/run-e2e-ctst.sh
+++ b/.github/scripts/end2end/run-e2e-ctst.sh
@@ -69,6 +69,10 @@ INSTANCE_ID=$(kubectl get zenko ${ZENKO_NAME} -o jsonpath='{.status.instanceID}'
 AZURE_ARCHIVE_ACCESS_TIER="Hot"
 AZURE_ARCHIVE_MANIFEST_ACCESS_TIER="Hot"
 
+BACKBEAT_API_HOST=$(kubectl get secret -l app.kubernetes.io/name=connector-cloudserver-config,app.kubernetes.io/instance=end2end -o jsonpath='{.items[0].data.config\.json}' | base64 -di | jq .backbeat.host)
+BACKBEAT_API_HOST=${BACKBEAT_API_HOST:1:-1}
+BACKBEAT_API_PORT=$(kubectl get secret -l app.kubernetes.io/name=connector-cloudserver-config,app.kubernetes.io/instance=end2end -o jsonpath='{.items[0].data.config\.json}' | base64 -di | jq .backbeat.port)
+
 # Setting CTST world params
 WORLD_PARAMETERS="$(jq -c <<EOF
 {
@@ -105,7 +109,9 @@ WORLD_PARAMETERS="$(jq -c <<EOF
   "TimeProgressionFactor":"${TIME_PROGRESSION_FACTOR}",
   "KafkaObjectTaskTopic":"${KAFKA_OBJECT_TASK_TOPIC}",
   "KafkaDeadLetterQueueTopic":"${KAFKA_DEAD_LETTER_TOPIC}",
-  "InstanceID":"${INSTANCE_ID}"
+  "InstanceID":"${INSTANCE_ID}",
+  "BackbeatApiHost":"${BACKBEAT_API_HOST}",
+  "BackbeatApiPort":"${BACKBEAT_API_PORT}"
 }
 EOF
 )"
--- a/tests/ctst/features/azureArchive.feature
+++ b/tests/ctst/features/azureArchive.feature
@@ -162,6 +162,7 @@ Feature: Azure Archive
         |               Versioned |           2 |        100 |           15 |
         |               Suspended |           2 |        100 |           15 |
 
+
     @2.7.0
     @PreMerge
     @Flaky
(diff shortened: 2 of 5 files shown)


tests/ctst/features/azureArchive.feature
```diff
@@ -122,3 +122,24 @@ Feature: Azure Archive
...
+    Given a "<versioningConfiguration>" bucket
+    And a transition workflow to "e2e-azure-archive" location
+    And that lifecycle is "paused" for the "e2e-azure-archive" location
+    And <objectCount> objects "obj" of size <objectSize> bytes
+    Then the storage class of object "obj-1" should be "" for <timeout> seconds
+    Then the storage class of object "obj-2" should be "" for <timeout> seconds
+    Given that lifecycle is "resumed" for the "e2e-azure-archive" location
+    Then the storage class of object "obj-1" must become "e2e-azure-archive"
+    And the storage class of object "obj-2" must become "e2e-azure-archive"
+
```
francoisferrand, line 207: these tests check the case where objects are added while the location is suspended.

can we add tests also for the objects which were already transitioning while suspend is initiated?
e.g. they will get aborted & requeued, should 1. stay 'transitioning' while suspended; 2. perform the transition as expected on resume

tests/ctst/steps/azureArchive.ts
```diff
@@ -390,6 +391,37 @@ Then('object {string} should expire in {int} days', async function (this: Zenko,
...
+Given('that lifecycle is {string} for the {string} location',
+    async function (this: Zenko, status: string, locatiom: string) {
+        let path: string;
+        if (status === 'paused') {
+            path = `/_/lifecycle/pause/${locatiom}`;
+        } else {
+            path = `/_/lifecycle/resume/${locatiom}`;
+        }
+        await axios.post(`http://${this.parameters.backbeatApiHost}:${this.parameters.backbeatApiPort}${path}`);
+    });
```
KillianG: nit
```suggestion
    async function (this: Zenko, status: string, location: string) {
        let path: string;
        if (status === 'paused') {
            path = `/_/lifecycle/pause/${location}`;
        } else {
            path = `/_/lifecycle/resume/${location}`;
        }
        await axios.post(`http://${this.parameters.backbeatApiHost}:${this.parameters.backbeatApiPort}${path}`);
    });
```

tests/ctst/steps/azureArchive.ts
```diff
@@ -390,6 +391,37 @@ Then('object {string} should expire in {int} days', async function (this: Zenko,
...
+
+Then('the storage class of object {string} should be {string} for {int} seconds',
```
KillianG: Not sure if this function really does what you want it to do

"should be XXXX for YYYY seconds" implies that the storage classe remains the same during the time in parameter, but here if you sleep directly at the beginning of the function you just check that **after** the timeout it is of the right storage class...

Maybe you should checkk the storage class every second instead ?
Or rename the quote to "should be {string} after {int} seconds"

KillianG: That said, I already implemented such a function in my PR so maybe it's worth waiting for mine to be merged..
https://github.com/scality/Zenko/pull/1903/files#diff-985e55b52be55a8b2b36dd0baa510782e4c12d8bacb4037c0997aaac84414a82R369

.github/scripts/end2end/run-e2e-ctst.sh
```diff
@@ -65,8 +65,12 @@ KAFKA_HOST_PORT=${KAFKA_HOST_PORT:1:-1}
...
...}'","NotificationDestinationAlt":"'${NOTIF_ALT_DEST_NAME}'","NotificationDestinationTopicAlt":"'${NOTIF_ALT_DEST_TOPIC}'","KafkaHosts":"'${KAFKA_HOST_PORT}'","KeycloakPassword":"'${KEYCLOAK_TEST_PASSWORD}'","KeycloakHost":"'${KEYCLOAK_TEST_HOST}'","KeycloakPort":"'${KEYCLOAK_TEST_PORT}'","keycloakRealm":"'${KEYCLOAK_TEST_REALM_NAME}'","keycloakClientId":"'${KEYCLOAK_TEST_CLIENT_ID}'","keycloakGrantType":"'${KEYCLOAK_TEST_GRANT_TYPE}'","StorageManagerUsername":"'${STORAGE_MANAGER_USER_NAME}'","StorageAccountOwnerUsername":"'${STORAGE_ACCOUNT_OWNER_USER_NAME}'","DataConsumerUsername":"'${DATA_CONSUMER_USER_NAME}'","ServiceUsersCredentials":'${SERVICE_USERS_CREDENTIALS}',"azureAccountName":"'${AZURE_ACCOUNT_NAME}'","azureAccountKey":"'${AZURE_SECRET_KEY}'","azureArchiveContainer":"'${AZURE_ARCHIVE_BUCKET_NAME}'","AzureArchiveAccessTier":"'${AZURE_ARCHIVE_ACCESS_TIER}'","azureArchiveManifestTier":"'${AZURE_ARCHIVE_MANIFEST_ACCESS_TIER}'","azureArchiveQueue":"'${AZURE_ARCHIVE_QUEUE_NAME}'","kafkaObjectTaskTopic":"'${KAFKA_OBJECT_TASK_TOPIC}'","kafkaDeadLetterQueueTopic":"'${KAFKA_DEAD_LETTER_TOPIC}'","BackbeatApiHost":"'${BACKBEAT_API_HOST}'","BackbeatApiPort":"'${BACKBEAT_API_PORT}'"}'
```
francoisferrand: please rebase, this has changed...

tests/ctst/steps/azureArchive.ts
```diff
@@ -471,6 +472,17 @@ Then('object {string} should expire in {int} days', async function (this: Zenko,
...
 
+Given('that lifecycle is {string} for the {string} location',
+    async function (this: Zenko, status: string, location: string) {
+        let path: string;
+        if (status === 'paused') {
+            path = `/_/lifecycle/pause/${location}`;
+        } else {
+            path = `/_/lifecycle/resume/${location}`;
+        }
+        await axios.post(`http://${this.parameters.BackbeatApiHost}:${this.parameters.BackbeatApiPort}${path}`);
```
francoisferrand: why are we using `axios` here: aren't we usually using simple http request?
(I can see axios used indeed, but seems to be related to hooking aws/s3 requests in some way....)

.github/scripts/end2end/run-e2e-ctst.sh
welansari: The change was introduced only on development/2.8, should i target that instead ?

francoisferrand: i guess you can "backport" the wrapping of this structure (but not the rest, i.e. change of field names & all);
or (to be confirmed with Florent) maybe it's fine to just target 2.8, as (thanks to Bert-e) the tests will be executed anyway...

KillianG (approved):

tests/ctst/features/azureArchive.feature
```diff
@@ -156,4 +157,25 @@ Feature: Azure Archive
...
+        |               Suspended |           2 |        100 |           2 | 
+
+    @2.7.0
+    @PreMerge
+    @Flaky
+    @AzureArchive
+    Scenario Outline: Pause and resume archiving to azure
+    Given a "<versioningConfiguration>" bucket
+    And a transition workflow to "e2e-azure-archive" location
+    And that lifecycle is "paused" for the "e2e-azure-archive" location
```
KillianG, line 200: We might face an issue where this makes the `e2e-azure-archive` location paused while the other tests are still running, which can cause a failure of the other tests
We might think of a way to make this test able to run not in parrallels of other tests

KillianG (changes requested):

tests/ctst/features/azureArchive.feature
KillianG: AzureArchive scenarios are already running separately due to [that](https://github.com/scality/Zenko/blob/edeb827df7f518716ec45050a45063751549240b/tests/ctst/common/hooks.ts#L13)

KillianG (approved):

tests/ctst/features/azureArchive.feature
francoisferrand: how is it resolved? i don't see additional test case?

welansari: added test case where i push objects before suspending the workflow

francoisferrand (approved):

PR #1903, opened 2023-06-20
Add CTST test for restore requeuing

KillianG:
Will need to release backbet and put the right version instead of using custom image

### DIFF
--- a/.github/scripts/end2end/run-e2e-ctst.sh
+++ b/.github/scripts/end2end/run-e2e-ctst.sh
@@ -39,7 +39,6 @@ KEYCLOAK_TEST_PORT="80"
 KEYCLOAK_TEST_REALM_NAME=${OIDC_REALM}
 KEYCLOAK_TEST_CLIENT_ID=${OIDC_CLIENT_ID}
 KEYCLOAK_TEST_GRANT_TYPE="password"
-SORBET_DEPLOYMENT_NAME="${ZENKO_NAME}-cold-sorbet-azure-e2e-azure-archive"
 
 # get Zenko service users credentials
 BACKBEAT_LCBP_1_CREDS=$(kubectl get secret -l app.kubernetes.io/name=backbeat-lcbp-user-creds,app.kubernetes.io/instance=end2end -o jsonpath='{.items[0].data.backbeat-lifecycle-bp-1\.json}' | base64 -d)
@@ -51,10 +50,10 @@ SORBET_FWD_2_SECRETKEY=$(kubectl get secret -l app.kubernetes.io/name=sorbet-fwd
 SERVICE_USERS_CREDENTIALS=$(echo '{"backbeat-lifecycle-bp-1":'${BACKBEAT_LCBP_1_CREDS}',"backbeat-lifecycle-conductor-1":'${BACKBEAT_LCC_1_CREDS}',"backbeat-lifecycle-op-1":'${BACKBEAT_LCOP_1_CREDS}',"backbeat-qp-1":'${BACKBEAT_QP_1_CREDS}',"sorbet-fwd-2":{"accessKey":"'${SORBET_FWD_2_ACCESSKEY}'","secretKey":"'${SORBET_FWD_2_SECRETKEY}'"}}' | jq -R)
 
 # Get backbeat UUID
-UUID=$(kubectl get secret -l app.kubernetes.io/name=backbeat-config,app.kubernetes.io/instance=end2end \
+BACKBEAT_UUID=$(kubectl get secret -l app.kubernetes.io/name=backbeat-config,app.kubernetes.io/instance=end2end \
     -o jsonpath='{.items[0].data.config\.json}' | base64 -di | jq .extensions.replication.topic)
-UUID=${UUID%.*}
-UUID=${UUID:1}
+BACKBEAT_UUID=${BACKBEAT_UUID%.*}
+BACKBEAT_UUID=${BACKBEAT_UUID:1}
 
 # Extracting kafka host from bacbeat's config
 KAFKA_HOST_PORT=$(kubectl get secret -l app.kubernetes.io/name=backbeat-config,app.kubernetes.io/instance=end2end \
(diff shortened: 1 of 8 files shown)


.github/scripts/end2end/configs/zenkoversion.yaml
```diff
@@ -58,8 +58,8 @@ spec:
       image: '${CLOUDSERVER_IMAGE}' 
       tag: '${CLOUDSERVER_TAG}'
     backbeat:
-      image: '${BACKBEAT_IMAGE}' 
-      tag: '${BACKBEAT_TAG}'
+      image: 'registry.scality.com/playground/kgardahaut/backbeat' 
+      tag: '8.6.25'
```
ghost: Leaving a note to not forget reverting this and bump backbeat instead when it's released

.github/scripts/end2end/run-e2e-ctst.sh
```diff
@@ -84,5 +93,6 @@ kubectl run $POD_NAME \
         --env=TARGET_VERSION=$VERSION  \
         --env=AZURE_BLOB_URL=$AZURE_BACKEND_ENDPOINT  \
         --env=AZURE_QUEUE_URL=$AZURE_BACKEND_QUEUE_ENDPOINT \
+        --env=UUID=$UUID \
         --env=VERBOSE=1 \
-        -- ./run "$COMMAND" $WORLD_PARAMETERS "--parallel $PARALLEL_RUNS --retry $RETRIES --retry-tag-filter @Flaky"
+        -- ./run "$COMMAND" $WORLD_PARAMETERS "--parallel $PARALLEL_RUNS --retry $RETRIES --retry-tag-filter @Flaky"
```
ghost: We can keep the last line here, no need to remove it

.github/scripts/end2end/run-e2e-ctst.sh
```diff
@@ -72,6 +79,8 @@ docker run \
     "${E2E_IMAGE}" /bin/bash \
     -c "SUBDOMAIN=${SUBDOMAIN} CONTROL_PLANE_INGRESS_ENDPOINT=${OIDC_ENDPOINT} ACCOUNT=${ZENKO_ACCOUNT_NAME} KEYCLOAK_REALM=${KEYCLOAK_TEST_REALM_NAME} STORAGE_MANAGER=${STORAGE_MANAGER_USER_NAME} STORAGE_ACCOUNT_OWNER=${STORAGE_ACCOUNT_OWNER_USER_NAME} DATA_CONSUMER=${DATA_CONSUMER_USER_NAME} /ctst/bin/seedKeycloak.sh"; [[ $? -eq 1 ]] && exit 1 || echo 'Keycloak Configured!'
 
+# Patch sorbet env
+kubectl set env deployment/${SORBET_DEPLOYMENT_NAME} SORBETD_AZURE_RESTORETIMEOUT=3s
```
ghost: I would argue that this should be done in the preparation steps in the CI, not here, as running CTST should be idempotent from the cluster state

tests/ctst/features/azureArchive.feature
```diff
@@ -122,3 +122,28 @@ Feature: Azure Archive
         |               Versioned |           2 |        100 |           2 |
         # |               Suspended |           2 |        100 |           2 |
 
+    @2.7.0
+    @PreMerge
+    @Flaky
```
ghost, line 133: I know other tests are are also flagged as flaky, but do we know why they are flaky? can we make them more robust?

tests/ctst/steps/azureArchive.ts
```diff
@@ -4,9 +4,9 @@ import assert from 'assert';
 import { safeJsonParse } from '../common/utils';
 import { Given, Then, When, After, setDefaultTimeout } from '@cucumber/cucumber';
 import { AzureHelper, S3, Constants, Utils } from 'cli-testing';
+import { exec } from 'child_process';
```
ghost, line 8: Do you think a work should be done on CTST to make it simpler to run other cli tools like sorbetclt here? With a set of methods, from CTST, to execute a tool with command line parameters and then handle the errors for you?

tests/ctst/steps/azureArchive.ts
```diff
@@ -368,6 +401,20 @@ When('i restore object {string} for {int} days', async function (this: Zenko, ob
     await S3.restoreObject(this.getCommandParameters());
 });
 
+When('i run sorbetctl to retry failed restore for {string} location', function (this: Zenko, location: string) {
+    const backbeatUUID = process.env.UUID;
```
ghost: Do we really need that as an environment variable? The world parameters can be extended with this uuid (and maybe with a more meaningful name), what do you think?

tests/ctst/features/azureArchive.feature
KillianG: The transition to cold is flaky.. I think we can make them more robust but I thinks this better goes in a separate ticket

Though, it would be nice to set a step as flaky instead of the full test, actually only the part where we wait for the storage class to change here is flaky...
It would be clearer to have only this step as flaky

tests/ctst/steps/azureArchive.ts
KillianG: Yep, as we discussed in the office on friday, I would say that an implementation of kube api would be very nice here

tests/ctst/steps/azureArchive.ts
```diff
@@ -330,30 +330,63 @@ Then('manifest containing object {string} should contain {int} objects',
...
-    }
-    await AzureHelper.sendBlobCreatedEventToQueue(
-        this.parameters.azureArchiveQueue,
-        this.parameters.azureArchiveContainer,
-        `rehydrate/${tarName}`,
-        getAzureCreds(this),
-    );
-});
+        assert(tarName);
+        while (!found) {
```
francoisferrand: shouldn't we have a timeout here?
or is it handled by CTST already?

tests/ctst/steps/azureArchive.ts
```diff
@@ -368,6 +401,20 @@ When('i restore object {string} for {int} days', async function (this: Zenko, ob
     await S3.restoreObject(this.getCommandParameters());
 });
 
+When('i run sorbetctl to retry failed restore for {string} location', function (this: Zenko, location: string) {
+    const backbeatUUID = this.parameters.backbeatUUID;
+    assert(backbeatUUID);
+    const command = `/ctst/sorbetctl forward list failed --trigger-retry --skip-invalid \
+    --kafka-dead-letter-topic=${backbeatUUID}.cold-dead-letter-${location} \
+    --kafka-object-task-topic=${backbeatUUID}.backbeat-lifecycle-object-tasks \
```
francoisferrand: extra indent for readability

.github/scripts/end2end/run-e2e-ctst.sh
```diff
@@ -49,6 +49,12 @@ SORBET_FWD_2_ACCESSKEY=$(kubectl get secret -l app.kubernetes.io/name=sorbet-fwd
 SORBET_FWD_2_SECRETKEY=$(kubectl get secret -l app.kubernetes.io/name=sorbet-fwd-creds,app.kubernetes.io/instance=end2end -o jsonpath='{.items[0].data.secretKey}' | base64 -d)
 SERVICE_USERS_CREDENTIALS=$(echo '{"backbeat-lifecycle-bp-1":'${BACKBEAT_LCBP_1_CREDS}',"backbeat-lifecycle-conductor-1":'${BACKBEAT_LCC_1_CREDS}',"backbeat-lifecycle-op-1":'${BACKBEAT_LCOP_1_CREDS}',"backbeat-qp-1":'${BACKBEAT_QP_1_CREDS}',"sorbet-fwd-2":{"accessKey":"'${SORBET_FWD_2_ACCESSKEY}'","secretKey":"'${SORBET_FWD_2_SECRETKEY}'"}}' | jq -R)
 
+# Get backbeat UUID
+BACKBEAT_UUID=$(kubectl get secret -l app.kubernetes.io/name=backbeat-config,app.kubernetes.io/instance=end2end \
```
francoisferrand: `BACKBEAT_UUID` : this is the prefix of all topics, right?
→ if so, this is not `backbeat UUID`, but actually just the 'InstanceID' of zenko, as used also in mongo...

 Instead of relying on the "shared" prefix in all topics (which could probably be found in Zenko CR instead) and hard-coded extra prefixes, it would be better to simply check the `coldStorageArchiveTopicPrefix`, `coldStorageRestoreTopicPrefix` and `coldStorageStatusTopicPrefix` fields in the config : which will contain both instance id and '-cold-archive/....' prefix.

francoisferrand (approved):
LGTM once comments are adressed

tests/ctst/steps/azureArchive.ts
KillianG: Yes it is already handled by CTST

.github/workflows/end2end.yaml
```diff
@@ -355,12 +355,22 @@ jobs:
...
         run: |-
           echo "CTST_TAG=$(sed 's/.*"cli-testing": ".*#\(.*\)".*/\1/;t;d' ./tests/ctst/package.json)" >> $GITHUB_ENV
+      - name: Extract environment
+        run: |-
+          sudo curl --fail -L https://github.com/mikefarah/yq/releases/download/${{ env.YQ_VERSION}}/${{ env.YQ_BINARY }} -o /usr/bin/yq
+          sudo chmod +x /usr/bin/yq
+          cd solution/
+          cat <<EOF >> $GITHUB_ENV
+          SORBET_TAG=$(yq eval '.sorbet.tag' deps.yaml)
+          EOF
```
ghost: As this duplicates the kafka step when installing yq, I would suggest creating a dedicated action file for that, that can be shared with `build-kafka`.

ghost: You can also avoid the `cd` by setting the `working-directory` parameter to the solution folder

tests/ctst/steps/azureArchive.ts
```diff
@@ -330,30 +330,63 @@ Then('manifest containing object {string} should contain {int} objects',
...
+                this.parameters.azureArchiveContainer,
+                `rehydrate/${tarName}`,
+                getAzureCreds(this),
+            );
+        }
+    });
+
+Then('the storage class of object {string} must stay {string} for {int} seconds',
+    async function (this: Zenko, objectName: string, storageClass: string, seconds: number) {
+        const objName = objectName ||  this.getSaved<string>('objectName');
```
ghost: nit
```suggestion
        const objName = objectName || this.getSaved<string>('objectName');
```

tests/ctst/steps/azureArchive.ts
```diff
@@ -368,6 +401,20 @@ When('i restore object {string} for {int} days', async function (this: Zenko, ob
...
+    assert(backbeatUUID);
+    const command = `/ctst/sorbetctl forward list failed --trigger-retry --skip-invalid \
+        --kafka-dead-letter-topic=${this.parameters.kafkaDeadLetterQueueTopic} \
+        --kafka-object-task-topic=${this.parameters.kafkaObjectTaskTopic} \
+        --kafka-brokers ${this.parameters.KafkaHosts}`;
+    exec(command, (error) => {
+        if (error) {
+            throw error;
+        }
+    });
```
ghost: don't we need some kind of await here, before continuing?

tests/ctst/world/Zenko.ts
```diff
@@ -90,6 +90,8 @@ export interface ZenkoWorldParameters {
     azureArchiveManifestTier: string;
     azureArchiveQueue: string;
     [key: string]: unknown;
+    kafkaObjectTaskTopic: string;
+    kafkaDeadLetterQueueTopic: string;
```
ghost, line 93: This can be set above the generic line for consistency

.github/workflows/end2end.yaml
KillianG: Adding a specific file just for those 2 lines of installing yq feels a bit overkill to me
I now use `working-directory` though

tests/ctst/features/azureArchive.feature
```diff
@@ -128,3 +128,28 @@ Feature: Azure Archive
...
+    @AzureArchive
+    Scenario Outline: Failed restore objects from tar must be retried and restored
+    Given a "<versioningConfiguration>" bucket
+    And a transition workflow to "e2e-azure-archive" location
+    And <objectCount> objects "obj" of size <objectSize> bytes
+    Then object "obj-1" should be "transitioning" and have the storage class "e2e-azure-archive"
+    And object "obj-2" should be "transitioning" and have the storage class "e2e-azure-archive"
+    And manifest containing object "obj-1" should "contain" object "obj-2"
+    When i restore object "obj-1" for <restoreDays> days
+    Then blob for object "obj-1" must be rehydrated "without" notifying the queue
```
francoisferrand: > blob for object "obj-1" must be rehydrated "without" notifying the queue

This does not really describe the scenario.
"without notifying" is actually not an observable outcome of this scenario, but more a detail of the implementation of the "failure".

Since the scenario is "faile restore must be retried", it should look something the following, where the scenario describes the _intent_ and code deals with the gory details of "how do we actually do/simulate this":

```
When i restore object "obj-1" for <restoreDays> days
    And object "obj-1" fails to rehydrate
    Then object "obj-1" should be "transitioning" and have the storage class "e2e-azure-archive"
    And an entry should be added to dead-letter-queue

When i run sorbetctl to retry failed restore for "e2e-azure-archive" location
    Then object "obj-1" should be "restored" and have the storage class "e2e-azure-archive"
    And object "obj-1" should expire in <restoreDays> days
    And object "obj-1" should have the same data
```

(this does not really affect the actual implementation of the steps/..., but more how to they are called)

KillianG: >     And object "obj-1" fails to rehydrate
I am not sure about this step, as this will just do nothing at all..

>     And an entry should be added to dead-letter-queue
This step seems a bit to "technical" to me, I think in such scenarios we don't want to know how it works internally, just the fact that it fails and it can be retried... no ?


I'm thinking of a way to make in clearer

tests/ctst/steps/azureArchive.ts
```diff
@@ -338,30 +339,81 @@ Then('manifest containing object {string} should contain {int} objects',
...
-        this.parameters.azureArchiveQueue,
-        this.parameters.azureArchiveContainer,
-        `rehydrate/${tarName}`,
-        getAzureCreds(this),
-    );
-});
+    });
+
+Then('blob for object {string} fails to rehydrate',
+    async function (this: Zenko, objectName: string) {
```
francoisferrand, line 392: please introduce helper function (avoid duplicating this code!), and add comment to explain what/why this function does what it does (i.e. how it simulates a failure to rehydrate)

.github/scripts/end2end/configs/zenkoversion.yaml
```diff
@@ -58,7 +58,7 @@ spec:
       image: '${CLOUDSERVER_IMAGE}' 
       tag: '${CLOUDSERVER_TAG}'
     backbeat:
-      image: '${BACKBEAT_IMAGE}' 
+      image: '${BACKBEAT_IMAGE}'
```
francoisferrand: and should not be changed (not even be in history, it uses introduces noise after the PR is merged : so remove the "change" + "revert")

KazToozs (commented):
LGTM

KazToozs (approved):

PR #1913, opened 2023-07-07
Improvement/zenko 4616 mongodb 5.0

ghost:
To be merged when we have dedicated branch / pensieve API mongodb driver migration merged and released.

The PR is about upgrading the mongodb charts to support v5.0. There is also a need to update the init scripts for the shard server https://github.com/scality/Zenko/pull/1913/commits/3703a948ba7658950544ac2bf542f1791da44b4d


VERSION
```diff
@@ -1,4 +1,4 @@
-VERSION="2.8.4"
+VERSION="2.9.0"
```
francoisferrand, line 1: should not target 2.9 branch, do not forget to create it and retarget PR before merging.

(though we should not release zenko 2.9.0 right after : if you need to create a release afterwards for bumping in Artesca, should be a pre-release like `2.9.0-preview.1` or `2.9.0-alpha.1`)

ghost: Yes, this is only for testing.

solution/deps.yaml
```diff
@@ -47,9 +47,9 @@ kafka-lag-exporter:
   tag: 0.7.1
   envsubst: KAFKA_LAGEXPORTER_TAG
 pensieve-api:
-  sourceRegistry: registry.scality.com/pensieve-api
+  sourceRegistry: registry.scality.com/playground/wlardier
```
nicolas2bert: Reminder: don't forget to update it :)

nicolas2bert (approved):

solution/deps.yaml
ghost: Yes! Thank you :slightly_smiling_face:

KillianG (approved):

solution/deps.yaml
ghost: Updated and branch rebased to dev/2.9

VERSION
ghost: PR is now rebased on the latest dev/2.9 branch.

PR #1930, opened 2023-08-02
4179: add E2E cold delete tests

welansari:
Issue: ZENKO-4179

### DIFF
--- a/tests/ctst/common/common.ts
+++ b/tests/ctst/common/common.ts
@@ -1,6 +1,6 @@
 import fs from 'fs';
 import { tmpNameSync } from 'tmp';
-import { Given, setDefaultTimeout, Then } from '@cucumber/cucumber';
+import { Given, setDefaultTimeout, Then, When } from '@cucumber/cucumber';
 import { Constants, S3, Utils } from 'cli-testing';
 import Zenko from 'world/Zenko';
 import { extractPropertyFromResults, safeJsonParse } from './utils';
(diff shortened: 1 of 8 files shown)


.github/workflows/end2end.yaml
```diff
@@ -221,6 +221,9 @@ jobs:
   build-kafka:
     runs-on: ubuntu-20.04
     steps:
+      - name: Setup tmate session
+        uses: mxschmitt/action-tmate@v3
+
```
francoisferrand: (just a reminder for later)  this should not be merged like this : either remove it, or activate it conditionally on 
"re-run with debug logs" (though i can't find how to detect this condition, i know it is possible...)

tests/ctst/features/dmf.feature
```diff
@@ -0,0 +1,46 @@
...
+    @2.7.0
+    @PreMerge
+    @Dmf
+    Scenario Outline: Deletion of a restored object
+    Given a "<versioningConfiguration>" bucket
+    And a transition workflow to "e2e-cold" location
+    And <objectCount> objects "obj" of size <objectSize> bytes
+    Then object "obj-1" should be "transitioned" and have the storage class "e2e-cold"
+    And object "obj-2" should be "transitioned" and have the storage class "e2e-cold"
+    Then dmf volume should contain <objectCount> objects
```
francoisferrand: should be "And" since we already have a "Then" before.

tests/ctst/features/dmf.feature
```diff
@@ -0,0 +1,46 @@
...
+    When i restore object "obj-1" for <restoreDays> days
+    Then object "obj-1" should be "restored" and have the storage class "e2e-cold"
+    When i restore object "obj-2" for <restoreDays> days
+    Then object "obj-2" should be "restored" and have the storage class "e2e-cold"
+    Then dmf volume should contain <objectCount> objects
+    When i delete object "obj-1"
+    And i delete object "obj-2"
```
francoisferrand: not sure we want to restore obj2 : so we can see both the "delete a cold object" and "delete a restore object"

tests/ctst/common/common.ts
```diff
@@ -151,3 +151,89 @@ Then('object {string} should have the user metadata with key {string} and value
...
+            assert(res.stdout);
+            const parsed = safeJsonParse(res.stdout);
+            assert(parsed.ok);
+            const head = parsed.result as {
+                StorageClass: string | undefined,
+                Restore: string | undefined,
+            };
+            const expectedClass = storageClass !== '' ? storageClass : undefined;
+            if (head?.StorageClass === expectedClass) {
+                conditionOk = true;
```
francoisferrand, line 217: * should also verify transition is not in progress
* ideally, we should also verify the intermediate state : e.g. when transitioning/transitioned, we should wait until we indeed reach that "transitioned" state, but also verify that we are in the "transition in progress" state until then (and fail otherwise) [and similarly for restoring/restored]

welansari: This might turn the tests more flaky as we might miss the intermediate state cause it generally is very brief

francoisferrand: i mean there other way around : if the "final" state is not reached (e.g. conditionOk), then we should be in the matching "transitioning/restoring" state

welansari: Transitioning state is not visible externally, it's an internal flag in the objecMD.
Added the check for the restore as we can see if an object is currently being restored via a simple `headObject`

francoisferrand (approved):

KillianG (approved):

PR #1946, opened 2023-08-29
Release zenko 2.7.27 with settings for improved lifecycle performance

francoisferrand:
- Bump backbeat 8.6.27
- Bump cloudserver 8.7.28
- Bump zkop 1.5.30 and update defaults for lifecycle performance (esp. object & bucket processor concurrency, and internal cloudserveri replicas)
- Bump zenko 2.7.27

Issue: ZENKO-4643


fmonjalet (approved):

anurag4DSB (approved):

.github/scripts/end2end/configs/zenkoversion.yaml
```diff
@@ -112,6 +112,14 @@ spec:
...
         tag: '${REDIS_EXPORTER_TAG}'
       kubedb: '${REDIS_KUBEDB_TAG}'
+  defaults:
+    backbeatConcurrency:
+      lifecycleBucketProcessor: 30
+      lifecycleObjectProcessor: 20
+    replicaMultipliers:
+      cloudserver: 16
+      internalCloudserver: 4
+      vault: 4
```
anurag4DSB, line 122: Have we documented these?

francoisferrand: what do you mean "documented" ?
* There will be an entry in the release notes, though the current status is we won't share the details of the parameters, which do not make sense for Artesca users. As for the replicaMultipliers, these are actually the same values as Artesca for both cloudserver & vault: the only _product_ change is scaling up internal cloudserver.
* These are not user-configurable fields, just moving the default _values_ out of ZKOP up into ZenkoVersion CR : so they must not be added to documentation and should never be tweaked by customers/end-users (who can use the Zenko CR to manually tweak his setup)

anurag4DSB: Cool, I was just wondering about the release notes for replicaMultipliers, if they are the same, we are good.

PR #1953, opened 2023-09-07
create/edit/delete locations CTST

KillianG:


.github/scripts/end2end/run-e2e-ctst.sh
```diff
@@ -95,12 +96,14 @@ WORLD_PARAMETERS="$(jq -c <<EOF
   "AzureAccountName":"${AZURE_ACCOUNT_NAME}",
   "AzureAccountKey":"${AZURE_SECRET_KEY}",
   "AzureArchiveContainer":"${AZURE_ARCHIVE_BUCKET_NAME}",
+  "AzureArchiveContainer2":"${AZURE_ARCHIVE_BUCKET_NAME_2}",
```
ghost, line 101: How about an array to support as many buckets as needed?

tests/ctst/common/common.ts
```diff
@@ -189,7 +195,7 @@ When('i restore object {string} for {int} days', async function (this: Zenko, ob
 });
 
 // wait for object to transition to a location or get restored from it
-Then('object {string} should be {string} and have the storage class {string}',
+Then('object {string} should be {string} and have the storage class {string}', { timeout: 100000 + 20 * 1000 },
```
ghost: Can't we put the real value directly? Or maybe we want to dynamically change the second part in the future?

tests/ctst/features/azureArchive.feature
```diff
@@ -42,6 +44,32 @@ Feature: Azure Archive
...
     @AzureArchive
+    @ColdStorage
+    Scenario Outline: Create, read, update and delete azure archive location
+    Given an azure archive location "<locationName>"
+    And a "<versioningConfiguration>" bucket
+    And a transition workflow to "<locationName>" location
+    And <objectCount> objects "obj" of size <objectSize> bytes
+    Then object "obj-1" should be "transitioning" and have the storage class "<locationName>"
+    And object "obj-2" should be "transitioning" and have the storage class "<locationName>"
+    And object "obj-3" should be "transitioning" and have the storage class "<locationName>"
```
ghost, line 55: Alternatively you can have a single `And` that can handle any number of object as saved from the previous step; if at some point we want to support more than 3 objects, it will be easier to manage

tests/ctst/world/Zenko.ts
```diff
@@ -844,6 +846,58 @@ export default class Zenko extends World<ZenkoWorldParameters> {
...
+     */
+    async managementAPIRequest(
+        method: Method,
+        path: string,
+        headers: AxiosRequestHeaders = {},
+        payload: object = {}
+    ): Promise<{ statusCode: number; data: object } | { statusCode: number; err: unknown }> {
+        const token = await this.getWebIdentityToken(
+            'zenko-end2end',
+            '123',
```
ghost: Should be set from the world parameters, as we have this information in
`OIDC_USERNAME`  and `OIDC_PASSWORD`  in the github workflow

tests/ctst/steps/azureArchive.ts
```diff
@@ -148,25 +151,50 @@ async function cleanZenkoBucket(
     world: Zenko,
     bucketName: string,
 ): Promise<void> {
+    if (!bucketName) {
+        return;
+    }
```
ghost, line 156: Do we have situations where we try to delete buckets without providing a name? For tests, I would suggest letting the code throw in this case, if something wrong is happening, we may affect other tests results by ignoring one bucket deletion

tests/ctst/steps/azureArchive.ts
```diff
@@ -148,25 +151,50 @@ async function cleanZenkoBucket(
...
+        const results = await S3.listObjectVersions(world.getCommandParameters());
+        const res = safeJsonParse(results.stdout);
+        assert(res.ok);
+        const parsedResults = res.result as listingResult;
+        const versions = parsedResults.Versions || [];
+        const deleteMarkers = parsedResults.DeleteMarkers || [];
+        await Promise.all(versions.concat(deleteMarkers).map(obj => {
+            world.addCommandParameter({ key: obj.Key });
+            world.addCommandParameter({ versionId: obj.VersionId });
+            return S3.deleteObject(world.getCommandParameters());
```
ghost, line 170: Will this work without flakiness in parallel, if we reuse the same object that we update concurrently?

tests/ctst/steps/azureArchive.ts
```diff
@@ -334,6 +366,7 @@ Then('blob for object {string} fails to rehydrate',
     async function (this: Zenko, objectName: string) {
         const tarName = await isObjectRehydrated(this, objectName);
         assert(tarName);
+        await Utils.sleep(60000);
```
ghost: Maybe worth a comment about why we set this specific value, as it slows down the tests (or some kind of check logic to reduce the duration?)

tests/ctst/steps/azureArchive.ts
```diff
@@ -372,10 +405,13 @@ When('i run sorbetctl to retry failed restore for {string} location', async func
         --kafka-object-task-topic=${this.parameters.KafkaObjectTaskTopic} \
         --kafka-brokers ${this.parameters.KafkaHosts}`;
     try {
-        await util.promisify(exec)(command);
+        const result = await util.promisify(exec)(command);
+        process.stdout.write(`Running command: ${command}\n`);
+        process.stdout.write(`Sorbetctl command result: ${result.stdout}\n`);
```
ghost: Maybe you want to put the first log before running the command, in case of error, it will ease the debugging to have access to it

tests/ctst/steps/azureArchive.ts
```diff
@@ -372,10 +405,13 @@ When('i run sorbetctl to retry failed restore for {string} location', async func
...
     } catch (err) {
         assert.ifError(err);
     }
+    await Utils.sleep(60000);
```
ghost, line 416: Same comment about the duration

tests/ctst/steps/azureArchive.ts
```diff
@@ -407,11 +443,80 @@ Then('object {string} should expire in {int} days', async function (this: Zenko,
...
+                accountName: this.parameters.AzureAccountName,
+                accountKey: this.parameters.AzureAccountKey,
+            },
+        },
+    };
+    const result = await this.managementAPIRequest('POST', `/config/${this.parameters.InstanceID}/location`, {},
+        locationConfig);
+    assert.strictEqual(result.statusCode, 201);
+    this.addToSaved('locationName', locationName);
+    await Utils.sleep(60000); // Wait for location to be updated
```
ghost: Don't we have an async way to test it, so that we don't wait for nothing, or not wait enough in case of slow CI?
Otherwise, maybe worth defining this duration as a const as it seems to be used multiple times

tests/ctst/steps/azureArchive.ts
```diff
@@ -407,11 +443,80 @@ Then('object {string} should expire in {int} days', async function (this: Zenko,
...
+});
+
+When('i change azure archive location {string} container target', async function (this: Zenko, locationName: string) {
+    const result = await this.managementAPIRequest('GET', `/config/overlay/view/${this.parameters.InstanceID}`);
+    if ('err' in result) {
+        assert.ifError(result.err);
```
ghost, line 477: This looks strange, if we detect an error, then shouldn't just we call the assert, that will prevent the rest of the test from running, allowing us to remove the root if condition?

tests/ctst/steps/azureArchive.ts
```diff
@@ -407,11 +443,80 @@ Then('object {string} should expire in {int} days', async function (this: Zenko,
...
+    } else {
+        const { locations } = result.data as { locations: Record<string, unknown> };
+        assert(locations[locationName]);
+        const locationConfig = locations[locationName] as Record<string, unknown>;
+        const details = locationConfig.details as { bucketName: string, auth?: { accountKey?: string } };
+        const auth = details.auth as { accountKey: string };
```
ghost: the `as` here should not be needed as it can be known from `details`' type

tests/ctst/steps/azureArchive.ts
```diff
@@ -407,11 +443,80 @@ Then('object {string} should expire in {int} days', async function (this: Zenko,
...
+        details.bucketName = this.parameters.AzureArchiveContainer2;
+        auth.accountKey = this.parameters.AzureAccountKey ;
```
ghost:
```suggestion
        auth.accountKey = this.parameters.AzureAccountKey;
```

ghost (commented):
Overall the changes and tests look good, I just have various comments on the code

tests/ctst/common/common.ts
KillianG: wdym by real value ?

tests/ctst/steps/azureArchive.ts
KillianG: There is no parallelism with `@ColdStorage` tests

tests/ctst/steps/azureArchive.ts
KillianG: 'err' field is not always in the result object, and `data` field is not here if `err` field is, so we need to check that to be able to use data as

tests/ctst/steps/azureArchive.ts
KillianG: This could be possible if we had a simple access to kube to check the status of the pods, unfortunately this is not the case yet so we "just" wait..

KillianG: And this is not the purpose of this ticket to add a way to check pod's status

tests/ctst/steps/azureArchive.ts
ghost: I mean, within the same test here, we issue S3 API calls in a Promise.all, while changing the command; but we use the same object to issue the command; we might end up with race conditions, if a delete is being run with another object key?

tests/ctst/steps/azureArchive.ts
ghost: Here is the ticket, if you want to add a `// TODO` : https://scality.atlassian.net/browse/CTST-35

tests/ctst/features/azureArchive.feature
KillianG: If at some point we need to handle more than 3 object we'll add this function at that time, but for now I don't think we need it for only 3 objects..

tests/ctst/steps/azureArchive.ts
KillianG: The function `cleanZenkoBucket` is called for all scenarios, this is an after each
I faced the issue locally where i just don't create a bucket and then it fails here.. I can remove it but I feel like this situation can only happend when we just dont create a bucket and try to delete one at the end..

tests/ctst/steps/azureArchive.ts
KillianG: I'm not sure I understand your point, `obj` is mapped from versions and deleteMarkers, here we just list all objects versions including delete markers and delete them all

.github/scripts/end2end/run-e2e-ctst.sh
KillianG: I'm not sure this is really needed for 2 containers yet.. might be if in the future we need more, but I'm not sure this is likely gonna happen

tests/ctst/steps/azureArchive.ts
KillianG: good catch ^^

tests/ctst/common/common.ts
ghost: 120000 in this case

KillianG: Ah ok, I did this to make it clear that we add 20s to the default 100k but I can put 120k directly yeah

tests/ctst/steps/azureArchive.ts
ghost: Fine by me, at least we are aware of it in case it happens one day

tests/ctst/steps/azureArchive.ts
ghost: The point was more about concurrently editing the world object that we pass to the s3 API, I think it's fine like this as we perform a deep copy of it in the lower layers of the code, but this may be causing issues one day if we change the world implementation, lgtm for now

ghost (approved):

ghost (commented):
Just a comment, in case you are not 100% confident about the changes, especially if they can still be flaky, it could be a good time to track it with a techdebt ticket, if we cannot do better for now (e.g. because of the lack of kubectl support in CTST)

tests/ctst/steps/azureArchive.ts
```diff
@@ -372,10 +406,14 @@ When('i run sorbetctl to retry failed restore for {string} location', async func
...
-        await util.promisify(exec)(command);
+        process.stdout.write(`Running command: ${command}\n`);
+        const result = await util.promisify(exec)(command);
+        process.stdout.write(`Sorbetctl command result: ${result.stdout}\n`);
     } catch (err) {
         assert.ifError(err);
     }
+    // Wait for backbeat to process the retry and update the object MDs
+    await Utils.sleep(60000);
 });
```
welansari, line 418: can we not use the metadata `x-amz-meta-scal-s3-restore-attempt` to wait for backebat ?
(I'm not sure if it's visible via a headObject)

KillianG: We could, but this is set only on retry
https://github.com/scality/backbeat/blob/718f67c67a2ceecdeada55780fcf4acfa6be033d/extensions/lifecycle/tasks/LifecycleRetriggerRestoreTask.js#L20

welansari: isn't that what we are doing here ?

tests/ctst/steps/azureArchive.ts
```diff
@@ -148,25 +151,50 @@ async function cleanZenkoBucket(
...
-    await Promise.all(versions.concat(deleteMarkers).map(obj => {
-        world.addCommandParameter({ key: obj.Key });
-        world.addCommandParameter({ versionId: obj.VersionId });
-        return S3.deleteObject(world.getCommandParameters());
-    }));
-    world.deleteKeyFromCommand('key');
-    world.deleteKeyFromCommand('versionId');
+    const createdObjects = world.getSaved<Map<string, string>>('createdObjects');
+    if (createdObjects !== undefined) {
+        const results = await S3.listObjectVersions(world.getCommandParameters());
```
francoisferrand, line 161: why are we _listing_ here ?

if we are using `createdObjects` to track the object created, we should be able to directly delete these objects with no listing.

tests/ctst/steps/azureArchive.ts
KillianG: You are right, MB..

That said, isnt this MD deleted when the restore succeed ?

tests/ctst/steps/azureArchive.ts
KillianG: We need to list objects to delete all objects versions including deleteMarkers, because we add versionId to savedObject when putting the object the first time, but not when adding user MD to it nor do we save the deleteMarker when deleting an object... That said we can probably just list objects instead of checking `world.getSaved` which is not a big deal i think..

tests/ctst/steps/azureArchive.ts
welansari: I think so yes, but that's not an issue here.
Here we add a delay to make sure backbeat processed the object, we can do the same by waiting for the object to have `x-amz-meta-scal-s3-restore-attempt: 1` in its metadata.

KillianG: I think it's actually an issue, when we retry, the restore succeed then the counter is deleted ? Or am i missing something ?

welansari: The restore is triggered only once the object metadata got updated with `x-amz-meta-scal-s3-restore-attempt`.
Maybe i understood you comment wrong, are we waiting for the restore to be fully done ? or just for backbeat to process the initial retry order (so before sorbet restores the object) ?

KillianG: Here we are retrying for real, so there maybe a little time where `x-amz-meta-scal-s3-restore-attempt` is set and right after the MDs are updated, so i dont think we can rely on that..
Here the failure is not a real failure, it's just that we are not notified of the success of the restore (since it is azurite the restore is done instantly), so the retry is really quick as well..

welansari (approved):

PR #1957, opened 2023-09-25
MongoDB resource limits

ghost:
Introducing new variables for:

- K8s memory limit (mongod and mongos)
- K8s memory requests (mongod and mongos)
- K8s wiredtiger cache size

These values are set dynamically in the CI based on the available capacity of the runner. In Artesca, these values are set in the installer, in the same way as the storage class. However, the values are computed dynamically using the reported k8s memory capacity, not the system, as kubernetes might reduce the amount of memory that the pods can access.

This was tested end to end in https://github.com/scality/artesca/pull/2242

For these changes, passing the CI mean that the logic is properly working, no test implementation is required or possible.


solution-base/build.sh
```diff
@@ -194,7 +199,11 @@ function render_mongodb_sharded_yamls()
         --set 'configsvr.persistence.selector.matchLabels.app\.kubernetes\.io/name=mongodb-sharded-config' \
         --set 'configsvr.persistence.selector.matchLabels.app\.kubernetes\.io/part-of=zenko' \
         --set "configsvr.mongodbExtraFlags={--setParameter rollbackTimeLimitSecs=259200}" \
-        --set "shardsvr.dataNode.mongodbExtraFlags={--setParameter rollbackTimeLimitSecs=259200}" \
+        --set mongos.resources.requests.memory=${MONGODB_MONGOS_RAM_REQUEST} \
+        --set mongos.resources.limits.memory=${MONGODB_MONGOS_RAM_LIMIT} \
```
francoisferrand, line 203: i have been wondering if we should introduce these parameters (RAM_REQUEST, RAM_LIMIT...), or just "patch" the CR. The problem i see are:
* Requires lots of coupling for every update (introduce new variable in Zenko, set value in Artesca...) when zenko does not have any logic around it
* Possibly lots of variables over time
* Is there a value which can be passed to these variables to "ignore" a specific field (e.g. make it similar to not having the request/limit spec at all)?

An alternative solution is to keep the resources management as before here (e.g. unset), and let the installer patch the resources spec... Coupling with mongo charts would be relatively low, as it only depends on the kind/name of STS/deployments which are used, and gives more flexibility.

solution-base/build.sh
```diff
@@ -194,7 +199,11 @@ function render_mongodb_sharded_yamls()
...
+        --set "shardsvr.dataNode.mongodbExtraFlags={--wiredTigerCacheSizeGB=${MONGODB_WIREDTIGER_CACHE_SIZE_GB}}" \
```
francoisferrand: for these flags we need a variable indeed, but i wonder if it should be specifically `MONGODB_WIREDTIGER_CACHE_SIZE_GB`, or a more general variable like `MONGODB_EXTRA_FLAGS` ?

solution-base/build.sh
```diff
@@ -194,7 +199,11 @@ function render_mongodb_sharded_yamls()
...
+        --set shardsvr.dataNode.resources.limits.memory=${MONGODB_SHARDSERVER_RAM_SAFE_LIMIT} \
```
francoisferrand: why "_SAFE_" ?

there is nothing safe here (or even explicit relationship with `MONGODB_WIREDTIGER_CACHE_SIZE_GB` in zenko), it is really just the k8s memory limit : so best keep if consistent with the other variables...

solution-base/build.sh
ghost: The issue I have with patching the CR is that we introduce risks of having changes in the bitnami generated yaml files leading to incompatibilities, while having such variables is something we do already, and also let us test such configuration in Zenko CI, not only Artesca, as I believe MongoDB is Zenko-specific. For example, having multiple shards or more MongoDB services (e.g., arbitrer), will require deep changes in the "filling" logic of the Artesca Installer, while template variables are easy to replace and control...

However I agree with the fact that we might end up with many variables (although I am not sure if we need more as of today)

Do you have any preference, or is it a strong request for change?

solution-base/build.sh
ghost: The safe means that the k8s limit is not formally required as long as the wiredtiger cache is limited, it only protects against memory leak (we never seen that with mongod yet).
I will remove it and make the name clearer

.github/scripts/end2end/install-kind-dependencies.sh
```diff
@@ -95,6 +95,23 @@ build_solution_base_manifests() {
...
+    MONGODB_SHARDSERVER_RAM_SAFE_LIMIT=$(awk "BEGIN {printf \"%.0fGi\", 2 * ${MONGODB_WIREDTIGER_CACHE_SIZE_GB}}")
+    MONGODB_SHARDSERVER_RAM_REQUEST=${MONGODB_WIREDTIGER_CACHE_SIZE_GB}Gi
+    MONGODB_MONGOS_RAM_REQUEST=$(awk "BEGIN {printf \"%.0fGi\", ${TOTAL_RAM_GB} * 0.033}")
+
+    # Replace values before deploying
+    sed -i "s/MONGODB_WIREDTIGER_CACHE_SIZE_GB/${MONGODB_WIREDTIGER_CACHE_SIZE_GB}/g" $DIR/_build/root/deploy/*
+    sed -i "s/MONGODB_MONGOS_RAM_LIMIT/${MONGODB_MONGOS_RAM_LIMIT}/g" $DIR/_build/root/deploy/*
+    sed -i "s/MONGODB_SHARDSERVER_RAM_SAFE_LIMIT/${MONGODB_SHARDSERVER_RAM_SAFE_LIMIT}/g" $DIR/_build/root/deploy/*
+    sed -i "s/MONGODB_SHARDSERVER_RAM_REQUEST/${MONGODB_SHARDSERVER_RAM_REQUEST}/g" $DIR/_build/root/deploy/*
+    sed -i "s/MONGODB_MONGOS_RAM_REQUEST/${MONGODB_MONGOS_RAM_REQUEST}/g" $DIR/_build/root/deploy/*
```
francoisferrand: you can actually use bash's math computation syntax : `$(( 0.335 * ${TOTAL_RAM_GB} ))`

we already rely on this, c.f. `count=$(($count + 1))` line 160, so we already must use bash (and not a generic /bin/sh as specified in the shebang)

solution-base/build.sh
ghost: About the "no value" thing, indeed it is not possible directly (without additional logic in the installer) with this approach, but the whole point is to enforce this change everywhere and for everyone.

francoisferrand: > Do you have any preference, or is it a strong request for change?

I think i may have a slight preference for patching the CR, just sharing my doubts here for your consideration.

IMHO patching the resource limits would be fine: the dependency is more on k8s than on the details of the chart... and we introduce more resources (artbiters...) we will need also corresponding patch in artesca installer to set the values, so we don't actually reduce the coupling, but just reduce the freedom...

That said, this is indeed not a request for change, and we probably don't need more variables _today_ (the point is more about maintenance in the future than fixing today, as is generally the case with engineering).

> About the "no value" thing, indeed it is not possible directly (without additional logic in the installer) with this approach, but the whole point is to enforce this change everywhere and for everyone.

That may be more concerning. While the "max memory limit" question has been debated, there may be other question/impacts of memory request, where some tweaks may be needed... And i would really want to avoid having to change zenko again just because we need to set a different value or remove one of the limits...

Can you try with an empty string (`""`) or 0, see if that allows reverting to the "unlimited" behavior while specifying the field?

ghost: As discussed over slack, we will go with the variables for now, but we must ensure the ability to disable this behavior (i.e., set 0 as this was confirmed to mean "unlimited"). This work will be done in Artesca, as an optional parameter to be documented.
In such case, the wiredtiger cache size will be removed as well.

.github/scripts/end2end/install-kind-dependencies.sh
```diff
@@ -99,17 +99,18 @@ build_solution_base_manifests() {
...
     # Detect total system RAM in GiB
     TOTAL_RAM_GB=$(awk '/MemTotal/ {printf "%.0f", $2/1024/1024}' /proc/meminfo)
+  
     # Compute MongoDB settings based on the total RAM
-    MONGODB_WIREDTIGER_CACHE_SIZE_GB=$(awk "BEGIN {printf \"%.0f\", ${TOTAL_RAM_GB} * 0.335}")
-    MONGODB_MONGOS_RAM_LIMIT=$(awk "BEGIN {printf \"%.0fGi\", ${TOTAL_RAM_GB} * 0.165}")
-    MONGODB_SHARDSERVER_RAM_SAFE_LIMIT=$(awk "BEGIN {printf \"%.0fGi\", 2 * ${MONGODB_WIREDTIGER_CACHE_SIZE_GB}}")
-    MONGODB_SHARDSERVER_RAM_REQUEST=${MONGODB_WIREDTIGER_CACHE_SIZE_GB}Gi
-    MONGODB_MONGOS_RAM_REQUEST=$(awk "BEGIN {printf \"%.0fGi\", ${TOTAL_RAM_GB} * 0.033}")
+    MONGODB_SHARDSERVER_EXTRA_FLAGS=$((TOTAL_RAM_GB * 335 / 1000))
```
francoisferrand: did you mean MONGODB_WIRETIGER_CACHE_SIZE_GB?

.github/scripts/end2end/install-kind-dependencies.sh
```diff
@@ -99,17 +99,18 @@ build_solution_base_manifests() {
...
+    MONGODB_MONGOS_RAM_LIMIT=$((TOTAL_RAM_GB * 165 / 1000))Gi
+    MONGODB_SHARDSERVER_RAM_LIMIT=$((2 * MONGODB_SHARDSERVER_EXTRA_FLAGS))Gi
+    MONGODB_SHARDSERVER_RAM_REQUEST=${MONGODB_SHARDSERVER_EXTRA_FLAGS}Gi
+    MONGODB_MONGOS_RAM_REQUEST=$((TOTAL_RAM_GB * 33 / 1000))Gi
 
     # Replace values before deploying
-    sed -i "s/MONGODB_WIREDTIGER_CACHE_SIZE_GB/${MONGODB_WIREDTIGER_CACHE_SIZE_GB}/g" $DIR/_build/root/deploy/*
+    sed -i "s/MONGODB_SHARDSERVER_EXTRA_FLAGS/--wiredTigerCacheSizeGB=${MONGODB_SHARDSERVER_EXTRA_FLAGS}/g" $DIR/_build/root/deploy/*
```
francoisferrand: here as well:

```
sed -i "s/MONGODB_SHARDSERVER_EXTRA_FLAGS/--wiredTigerCacheSizeGB=${MONGODB_WIRETIGER_CACHE_SIZE_GB}/g" $DIR/_build/root/deploy/*
```

.github/scripts/end2end/install-kind-dependencies.sh
ghost: Yes

.github/scripts/end2end/install-kind-dependencies.sh
ghost: Thanks, done in https://github.com/scality/Zenko/pull/1957/commits/45ea576761c41ec4c5d2ecb4610cb6ef0f27c5be

benzekrimaha (approved):
LGTM

francoisferrand (approved):

PR #1972, opened 2023-11-01
Zenko-4628 - put user MD on transitioned object

KillianG:
Issue: ZENKO-4628

### DIFF
--- a/tests/ctst/features/azureArchive.feature
+++ b/tests/ctst/features/azureArchive.feature
@@ -243,8 +243,18 @@ Feature: Azure Archive
     And a transition workflow to "e2e-azure-archive" location
     And <objectCount> objects "obj" of size <objectSize> bytes
     Then object "obj-1" should be "transitioned" and have the storage class "e2e-azure-archive"
-    And object "obj-2" should be "transitioned" and have the storage class "e2e-azure-archive"
-    
+    Then i "should not" be able to add user metadata to object "obj-1"
+    When i restore object "obj-1" for 10 days
+    When i restore object "obj-2" for 10 days
+    Then blob for object "obj-1" must be rehydrated
+    Then blob for object "obj-2" must be rehydrated
+    Then object "obj-1" should be "restored" and have the storage class "e2e-azure-archive"
+    Then object "obj-2" should be "restored" and have the storage class "e2e-azure-archive"
+    Then i "should" be able to add user metadata to object "obj-1"
+    Then i "should" be able to add user metadata to object "obj-2"
+    Then object "obj-1" should be "transitioned" and have the storage class "e2e-azure-archive"
+    Then object "obj-2" should be "transitioned" and have the storage class "e2e-azure-archive"
+
     Examples:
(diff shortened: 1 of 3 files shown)


KazToozs (approved):

tests/ctst/features/azureArchive.feature
```diff
@@ -187,4 +187,26 @@ Feature: Azure Archive
...
+    Scenario Outline: Cannot add object MD to a transitioned object
+    Given a "<versioningConfiguration>" bucket
+    And a transition workflow to "e2e-azure-archive" location
+    And <objectCount> objects "obj" of size <objectSize> bytes
+    Then object "obj-1" should be "transitioned" and have the storage class "e2e-azure-archive"
+    Then i "should not" be able to add user metadata to object "obj-1"
+    When i restore object "obj-1" for 10 days
+    Then blob for object "obj-1" must be rehydrated
+    Then object "obj-1" should be "restored" and have the storage class "e2e-azure-archive"
+    Then i "should" be able to add user metadata to object "obj-1"
```
welansari, line 253: Here we are replacing the original object, should we test if the object still expires ? and if it still keeps it's original metadata (if this is the intended behaviour i'm not sure)

tests/ctst/features/azureArchive.feature
```diff
@@ -187,4 +187,34 @@ Feature: Azure Archive
...
+    When i restore object "obj-1" for 10 days
+    When i restore object "obj-2" for 10 days
+    Then blob for object "obj-1" must be rehydrated
+    Then blob for object "obj-2" must be rehydrated
+    Then object "obj-1" should be "restored" and have the storage class "e2e-azure-archive"
+    Then object "obj-2" should be "restored" and have the storage class "e2e-azure-archive"
+    Then i "should" be able to add user metadata to object "obj-1"
+    Then i "should" be able to add user metadata to object "obj-2"
+    Then object "obj-1" should be "transitioned" and have the storage class "e2e-azure-archive"
+    Then object "obj-2" should be "transitioned" and have the storage class "e2e-azure-archive"
```
welansari, line 256: I'm afraid this test will pass even without having the same object, we should probably check before that it has the restoration fields (using a headObject).

And maybe we should adapt this step to make it say `... should be "expired ..."` to be clearer about what we expect.

KillianG: We are already checking for the restoration field here: https://github.com/scality/Zenko/pull/1972/files#diff-c8e830ea281891b15b656a69c9fdfbd92030bcb87f31efc8f5173a8e7584374bR207-R208

That said I can change the "transitioned" to "expired" for phrasing only because it will not check anything more.. or do you have an idea what i can check specifically ?

welansari (approved):

tests/ctst/features/azureArchive.feature
welansari: Since it's a new version we can leave it to transitioned. (I misunderstood what the copy object does and thought it was supposed to keep the same metadata and not create a new version)

PR #1975, opened 2023-11-14
Add Azure Archive feature with restore notifications

KillianG:
This PR is not testinf or `s3:LifecycleTransition` notification because it's currently not working, this will be check with another ticket: ZENKO-4706

### DIFF
--- a/tests/ctst/common/common.ts
+++ b/tests/ctst/common/common.ts
@@ -8,6 +8,51 @@ import assert from 'assert';
 
 setDefaultTimeout(Constants.DEFAULT_TIMEOUT);
 
+type listingObject = {
+    Key: string,
+    VersionId: string,
+}
+
+type listingResult = {
+    Versions: listingObject[],
+    DeleteMarkers: listingObject[],
+}
+
+/**
+ * Cleans the created test bucket
+ * @param {Zenko} world world object
+ * @param {string} bucketName bucket name
+ * @returns {void}
+ */
+export async function cleanZenkoBucket(
+    world: Zenko,
+    bucketName: string,
+): Promise<void> {
+    if (!bucketName) {
+        return;
+    }
+    world.resetCommand();
+    world.addCommandParameter({ bucket: bucketName });
+    const createdObjects = world.getSaved<Map<string, string>>('createdObjects');
+    if (createdObjects !== undefined) {
+        const results = await S3.listObjectVersions(world.getCommandParameters());
+        const res = safeJsonParse(results.stdout);
+        assert(res.ok);
+        const parsedResults = res.result as listingResult;
+        const versions = parsedResults.Versions || [];
+        const deleteMarkers = parsedResults.DeleteMarkers || [];
+        await Promise.all(versions.concat(deleteMarkers).map(obj => {
+            world.addCommandParameter({ key: obj.Key });
+            world.addCommandParameter({ versionId: obj.VersionId });
+            return S3.deleteObject(world.getCommandParameters());
+        }));
+        world.deleteKeyFromCommand('key');
+        world.deleteKeyFromCommand('versionId');
+    }
+    await S3.deleteBucketLifecycle(world.getCommandParameters());
+    await S3.deleteBucket(world.getCommandParameters());
+}
+
 async function addMultipleObjects(this: Zenko, numberObjects: number,
     objectName: string, sizeBytes: number, userMD?: string) {
     for (let i = 1; i <= numberObjects; i++) {
--- a/tests/ctst/common/common.ts
+++ b/tests/ctst/common/common.ts
@@ -24,7 +24,7 @@ type listingResult = {
  * @param {string} bucketName bucket name
  * @returns {void}
  */
-export async function cleanZenkoBucket(
+export async function cleanS3Bucket(
     world: Zenko,
     bucketName: string,
 ): Promise<void> {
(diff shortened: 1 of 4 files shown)


tests/ctst/steps/notifications.ts
```diff
@@ -375,12 +375,4 @@ Then('i should {string} a notification for {string} event in destination {int}',
...
         const expected = receive === 'receive';
         assert.strictEqual(receivedNotification, expected);
-        if (this.getSaved<string>('bucketVersioning') === 'Non versioned') {
-            await emptyNonVersionedBucket(this);
-        } else {
-            await emptyVersionedBucket(this);
-        }
-        this.resetCommand();
-        this.addCommandParameter({ bucket: this.getSaved<string>('bucketName') });
-        await S3.deleteBucket(this.getCommandParameters());
```
welansari, line 385: We'll need to clean and remove the buckets created in `BucketNotification` tests.

We can use an `After` hook just like we do in `AzureArchive` tests.

Since your test has both `BucketNotification` and `AzureArchive` tags this might behave weirdly

tests/ctst/features/azureArchive.feature
```diff
@@ -232,6 +232,35 @@ Feature: Azure Archive
...
+    Then object "obj-1" should be "transitioned" and have the storage class "e2e-azure-archive"
+    And object "obj-2" should be "transitioned" and have the storage class "e2e-azure-archive"
+    When i restore object "obj-1" for <restoreDays> days
+    When i restore object "obj-2" for <restoreDays> days
+    Then i should "receive" a notification for "s3:ObjectRestore:Post" event in destination 0
+    And blob for object "obj-2" must be rehydrated
+    Then i should "receive" a notification for "s3:ObjectRestore:Completed" event in destination 0
+    When i wait for <restoreDays> days
+    Then i should "receive" a notification for "s3:ObjectRestore:Delete" event in destination 0
+
```
francoisferrand: nit: extra blank line

tests/ctst/features/azureArchive.feature
```diff
@@ -232,6 +232,35 @@ Feature: Azure Archive
...
+    Scenario Outline: Restore notifications are triggered
+    Given a "<versioningConfiguration>" bucket
+    And one notification destination
+    And i subscribe to "s3:ObjectRestore:*" notifications for destination 0
+    And a transition workflow to "e2e-azure-archive" location
+    And <objectCount> objects "obj" of size <objectSize> bytes
+    Then object "obj-1" should be "transitioned" and have the storage class "e2e-azure-archive"
+    And object "obj-2" should be "transitioned" and have the storage class "e2e-azure-archive"
+    When i restore object "obj-1" for <restoreDays> days
+    When i restore object "obj-2" for <restoreDays> days
```
francoisferrand: we restore 2 objects, but check notifcation for one only → one restore should be enough?

tests/ctst/features/azureArchive.feature
```diff
@@ -232,6 +232,35 @@ Feature: Azure Archive
...
+    Then i should "receive" a notification for "s3:ObjectRestore:Post" event in destination 0
+    And blob for object "obj-2" must be rehydrated
+    Then i should "receive" a notification for "s3:ObjectRestore:Completed" event in destination 0
+    When i wait for <restoreDays> days
+    Then i should "receive" a notification for "s3:ObjectRestore:Delete" event in destination 0
```
francoisferrand, line 253: can we check the notification refers to `obj-2` ?

KillianG: That is already the case, in the code of the step, we check for the last object created

tests/ctst/common/common.ts
```diff
@@ -8,6 +8,51 @@ import assert from 'assert';
...
+    if (createdObjects !== undefined) {
+        const results = await S3.listObjectVersions(world.getCommandParameters());
+        const res = safeJsonParse(results.stdout);
+        assert(res.ok);
+        const parsedResults = res.result as listingResult;
+        const versions = parsedResults.Versions || [];
+        const deleteMarkers = parsedResults.DeleteMarkers || [];
+        await Promise.all(versions.concat(deleteMarkers).map(obj => {
+            world.addCommandParameter({ key: obj.Key });
+            world.addCommandParameter({ versionId: obj.VersionId });
```
francoisferrand, line 46: * does `addCommandParameter` have the semantic to replace an existing parameter with the same "key" ? Otherwise we may end up with multiple "keys" in the parameters.... (which may be ok, as aws cli to take into account the last one; but seems flimsy anyway)
* are we able to both params in a single call to `addCommandParameter`, as would be natural in nodejs? (i.e. `world.addCommandParameter({ key: obj.Key, versionId: obj.VersionId });`)

KillianG: > does addCommandParameter have the semantic to replace an existing parameter with the same "key" ?

It is replacing yes, so we won't face this issue

> are we able to both params in a single call to addCommandParameter, as would be natural in nodejs? (i.e. world.addCommandParameter({ key: obj.Key, versionId: obj.VersionId });)

We are not yet able to do that.. i created ZENKO-4708 to address that

welansari (approved):

tests/ctst/common/common.ts
```diff
@@ -8,6 +8,51 @@ import assert from 'assert';
...
+    DeleteMarkers: listingObject[],
+}
+
+/**
+ * Cleans the created test bucket
+ * @param {Zenko} world world object
+ * @param {string} bucketName bucket name
+ * @returns {void}
+ */
+export async function cleanZenkoBucket(
```
francoisferrand: not really a "Zenko" bucket: this would actually work for any S3 bucket...
the only link to Zenko is the type of `world`, and I am pretty sure this may be replaced with a more generic type when needed.

francoisferrand (approved):

benzekrimaha (approved):

PR #1990, opened 2023-12-15
Fix ReplicationLagWarning alert

francoisferrand:
* Fix unit, metrics is in millisecond
* In sharded deployment, the same `job` is used for members of a replicaset,
so this rule must be filtered by pod or instance.

Issue: ZENKO-4715


monitoring/kafka/alerts.yaml
```diff
@@ -111,6 +111,7 @@ groups:
           > ${maxConsumerLagMessagesWarningThreshold}
       or
       kafka_consumergroup_group_max_lag_seconds{namespace="${namespace}",cluster_name="${cluster}",group!=""}
+          < (1/0)
```
welansari, line 114: not sure what this is supposed to do ?

francoisferrand: it is needed to filter-out `NaN` / `+Inf` values
(unfortunately, I did not find a way to add a comment in there...)

welansari (approved):

rachedbenmustapha (approved):

PR #2023, opened 2024-02-21
Improvement/zenko 4685 bucket policies and conditions

ghost:
> Will be rebased with creation of integration branches after the next Artesca 1.7 release.

image

## Bucket policies for ARTESCA:

- Bump of Cloudserver with latest changes around the feature
- Bump of Vault (enabling the implicit denies by default)
- Bump of ZKOP (adds a CR option to disable Implicit Denies from Vault)

> Note: will be rebased with the real tags once we have them.

## Bump of CTST

CTST is bumped to use the SDK, that greatly speed up the new tests. The changes are backported from https://github.com/scality/Zenko/pull/1995

## Tests

We test all S3 APIs against 20 authorization scenarios with all combinations of Bucket Policies and IAM Policies states:
* No policy
* Policy does not apply
* Policy applies with DENY
* Policy applies with ALLOW
* Policy applies with both ALLOW and DENY (IAM case only)

All APIs are tested for IAM Users, Assumed Role users, and Cross Account users.

> Note: the scenarios are generated with a js script as a helper, as we must list all the scenarios in the `.feature` files, which is too much to be managed by hand.

Additional tests:
* Customer use case, with a Bucket Policy granting read access to buckets and write accesses to objects in the bucket. All S3 APIs are tested.
* Conditions: both retention periods and IP addresses checks.
* Web Identities: a single sanity check to ensure internal roles permission can be extended with Bucket Policies. The main logic is already tested by the "Assumed Role" scenarios above.

## Test logic

All tests try to reuse steps as much as possible. All S3 APIs are defined in an object where we define when we need specific setup or error codes as output.

PR #2024, opened 2024-02-23
Add test for restore adjust and re-restore

KillianG:
add a test for reducing restore duration and a test for increasing restore duration

Issue: ZENKO-4727

### DIFF
--- a/tests/ctst/common/common.ts
+++ b/tests/ctst/common/common.ts
@@ -327,11 +327,16 @@ Then('kafka consumed messages should not take too much place on disk',
             for (let j = 0; j < newOffsets[i].partitions.length; j++) {
                 // Checking that the min offset has increased due to kafkacleaner
                 // or that it didn't need to change because there was no new messages
-                assert.ok(newOffsets[i].partitions[j].low > previousOffsets[i].partitions[j].low ||
-                    newOffsets[i].partitions[j].high === newOffsets[i].partitions[j].low,
-                `Topic ${topics[i]} partition ${j} offset has not increased,
-                previousOffsets: ${previousOffsets[i].partitions[j].low} / ${previousOffsets[i].partitions[j].high},
-                newOffsets: ${newOffsets[i].partitions[j].low} / ${newOffsets[i].partitions[j].high}`);
+                // or new messages came after a clean but still cleaned to previous high offset
+                const lowOffsetIncreased = newOffsets[i].partitions[j].low > previousOffsets[i].partitions[j].low;
+                const noNewMessages = newOffsets[i].partitions[j].high === newOffsets[i].partitions[j].low;
+                const newMessagesAfterClean =
+                    newOffsets[i].partitions[j].low === previousOffsets[i].partitions[j].high &&
+                    previousOffsets[i].partitions[j].high !== '0';
+                assert.ok(lowOffsetIncreased || noNewMessages || newMessagesAfterClean,
+                    `Topic ${topics[i]} partition ${j} offset has not increased,
+                    previousOffsets: ${previousOffsets[i].partitions[j].low} / ${previousOffsets[i].partitions[j].high},
+                    newOffsets: ${newOffsets[i].partitions[j].low} / ${newOffsets[i].partitions[j].high}`);
             }
         }
     });
--- a/tests/ctst/common/common.ts
+++ b/tests/ctst/common/common.ts
@@ -304,7 +304,7 @@ Then('kafka consumed messages should not take too much place on disk', { timeout
     async function (this: Zenko) {
         const kfkcIntervalSeconds = parseInt(this.parameters.KafkaCleanerInterval);
 
-        setTimeout(() => {
+        const timeoutID = setTimeout(() => {
             assert.fail('Kafka cleaner did not clean the topics');
         }, (kfkcIntervalSeconds * 1000 + 3000) * 5); // Timeout after 5 kafkacleaner intervals
 
@@ -358,6 +358,7 @@ Then('kafka consumed messages should not take too much place on disk', { timeout
                 }
             }
         }
+        clearTimeout(timeoutID);
 
         // If a topic remains in this array, it means it has not been cleaned
         assert(topics.length === 0, `Topics ${topics.join(', ')} still have not been cleaned`);
--- a/tests/ctst/common/common.ts
+++ b/tests/ctst/common/common.ts
@@ -300,18 +300,22 @@ Then('i {string} be able to add user metadata to object {string}',
         }
     });
 
-Then('kafka consumed messages should not take too much place on disk', { timeout: 130000 },
+Then('kafka consumed messages should not take too much place on disk', { timeout: -1 },
     async function (this: Zenko) {
+        const kfkcIntervalSeconds = parseInt(this.parameters.KafkaCleanerInterval);
+
+        setTimeout(() => {
+            assert.fail('Kafka cleaner did not clean the topics');
+        }, (kfkcIntervalSeconds * 1000 + 3000) * 5); // Timeout after 5 kafkacleaner intervals
+
         const ignoredTopics = ['dead-letter'];
         const kafkaAdmin = new Kafka({ brokers: [this.parameters.KafkaHosts] }).admin();
         const topics: string[] = (await kafkaAdmin.listTopics())
             .filter(t => (t.includes(this.parameters.InstanceID) &&
             !ignoredTopics.some(e => t.includes(e))));
 
-        const timeStart = new Date().getTime();
-        const seconds = parseInt(this.parameters.KafkaCleanerInterval);
 
-        while (topics.length > 0 && new Date().getTime() - timeStart < 120000) {
+        while (topics.length > 0) {
             const previousOffsets = await getTopicsOffsets(topics, kafkaAdmin);
             // Checking topics offsets before kafkacleaner passes to be sure kafkacleaner works
             // This function can be improved by consuming messages and
@@ -319,8 +323,8 @@ Then('kafka consumed messages should not take too much place on disk', { timeout
             // Instead of waiting for a fixed amount of time,
             // we could also check for metrics to see last kafkacleaner run
 
-            // 10 seconds added to be sure kafkacleaner had time to process
-            await Utils.sleep(seconds * 1000 + 10000);
+            // 3 seconds added to be sure kafkacleaner had time to process
+            await Utils.sleep(kfkcIntervalSeconds * 1000 + 3000);
 
             const newOffsets = await getTopicsOffsets(topics, kafkaAdmin);
 
--- a/tests/ctst/common/common.ts
+++ b/tests/ctst/common/common.ts
@@ -308,13 +308,13 @@ Then('kafka consumed messages should not take too much place on disk', { timeout
             assert.fail('Kafka cleaner did not clean the topics');
         }, (kfkcIntervalSeconds * 1000 + 3000) * 5); // Timeout after 5 kafkacleaner intervals
 
-        const ignoredTopics = ['dead-letter'];
-        const kafkaAdmin = new Kafka({ brokers: [this.parameters.KafkaHosts] }).admin();
-        const topics: string[] = (await kafkaAdmin.listTopics())
-            .filter(t => (t.includes(this.parameters.InstanceID) &&
-            !ignoredTopics.some(e => t.includes(e))));
-
         try {
+            const ignoredTopics = ['dead-letter'];
+            const kafkaAdmin = new Kafka({ brokers: [this.parameters.KafkaHosts] }).admin();
+            const topics: string[] = (await kafkaAdmin.listTopics())
+                .filter(t => (t.includes(this.parameters.InstanceID) &&
+                !ignoredTopics.some(e => t.includes(e))));
+
             while (topics.length > 0) {
                 const previousOffsets = await getTopicsOffsets(topics, kafkaAdmin);
                 // Checking topics offsets before kafkacleaner passes to be sure kafkacleaner works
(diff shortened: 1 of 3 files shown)


tests/ctst/features/azureArchive.feature
```diff
@@ -121,6 +121,38 @@ Feature: Azure Archive
...
+    Then object "obj-2" should be "cold" and have the storage class "e2e-azure-archive"
+    And object "obj1" should expire in 45 days
+    When i wait for 45 days
+    Then object "obj-1" should be "cold" and have the storage class "e2e-azure-archive"
+
+    Examples:
+        | versioningConfiguration | objectCount | objectSize | restoreDays |
+        |           Non versioned |           2 |        100 |           15 |
+        |               Versioned |           2 |        100 |           15 |
+        |               Suspended |           2 |        100 |           15 |
```
francoisferrand, line 185: as per the ticket, should also add a test case `restore→expiration→re-restore` (and verify it expires a second time as expected)

tests/ctst/common/common.ts
```diff
@@ -258,18 +271,14 @@ Given('a transition workflow to {string} location', async function (this: Zenko,
...
-    const versionId = this.getSaved<Map<string, string>>('createdObjects')?.get(objName);
-    if (versionId) {
-        this.addCommandParameter({ versionId });
-    }
-    this.addCommandParameter({ restoreRequest: `Days=${days}` });
-    await S3.restoreObject(this.getCommandParameters());
+    await restoreObject.call(this, days, objectName);
 });
 
+When('i adjust restore duration of object {string} for {int} days',
```
francoisferrand: "adjust restore" is not a separate step : this is not intrinsically an adjust, but only because of the condition of the object when this happens...

→ the steps should represent what we actually do (e.g. "issue an s3:restore command"), to make it easier to understand what goes on. I.e. we may have a "adjust restore" **_scenario_**, but that scenario would be simply "archive → restore → wait until restored → restore → check the restore duration has been updated"

tests/ctst/steps/azureArchive.ts
```diff
@@ -373,7 +373,7 @@ When('i run sorbetctl to retry failed restore for {string} location', async func
 When('i wait for {int} days', { timeout: 10 * 60 * 1000 }, async function (this: Zenko, days: number) {
     const realTimeDay = days * 24 * 60 * 60 * 1000 /
         (this.parameters.TimeProgressionFactor > 1 ? this.parameters.TimeProgressionFactor : 1);
-    await Utils.sleep(realTimeDay + 200 * 1000); // We also wait for sorbet forwarder to see the object has expired
+    await Utils.sleep(realTimeDay); // We also wait for sorbet forwarder to see the object has expired
```
francoisferrand: seems unrelated : if needed, should go in a separate commit to explain what happens here...

(and comment should probably be updated, since we don't have the "extra' delay?)

tests/ctst/features/iam-policies/AssumeRoleWithWebIdentity.feature
```diff
@@ -125,52 +125,52 @@ Feature: Assume Role with Web Identity
...
+      | action                          | type                  | withVersioning | objectExists   | error                                           |
+      | RestoreObject                   | STORAGE_MANAGER       | with           | exists         | InvalidObjectState                              |
+      | GetBucketCors                   | STORAGE_MANAGER       | without        | does not exist | NoSuchCORSConfiguration                         |
+      | GetObjectLockConfiguration      | STORAGE_MANAGER       | without        | exists         | ObjectLockConfigurationNotFoundError            |
+      | GetObjectRetention              | STORAGE_MANAGER       | with           | exists         | Bucket is missing Object Lock Configuration     |
+      | GetObjectLegalHold              | STORAGE_MANAGER       | with           | exists         | Bucket is missing Object Lock Configuration     |
+      | PutObjectRetention              | STORAGE_MANAGER       | with           | exists         | MalformedXML: request xml is undefined or empty |
```
welansari: nit: for consistency with the other expected errors, should we only specify the error type `MalformedXML` without the description ?

tests/ctst/features/azureArchive.feature
```diff
@@ -121,6 +121,70 @@ Feature: Azure Archive
         |               Versioned |           2 |      30000 |               1 |
         |               Suspended |           2 |      30000 |               1 |
 
+    @2.7.0
+    @PreMerge
+    @Flaky
+    @AzureArchive
+    @ColdStorage
+    @WIP
```
welansari: leftover from debugging ?

tests/ctst/features/azureArchive.feature
```diff
@@ -121,6 +121,70 @@ Feature: Azure Archive
...
+        |           Non versioned |           2 |        100 |           15 |
+        |               Versioned |           2 |        100 |           15 |
+        |               Suspended |           2 |        100 |           15 |
+
+    @2.7.0
+    @PreMerge
+    @Flaky
+    @AzureArchive
+    @ColdStorage
+    @WIP
```
welansari: same here

welansari (approved):

solution/deps.yaml
```diff
@@ -16,7 +16,7 @@ cloudserver:
   sourceRegistry: registry.scality.com/cloudserver
   dashboard: cloudserver-dashboards
   image: cloudserver
-  tag: 8.8.7
+  tag: 8.8.20
```
francoisferrand: already in #2023 I guess : will need to rebase after it is merged

francoisferrand (approved):

solution/deps.yaml
```diff
@@ -99,7 +99,7 @@ vault:
 zenko-operator:
   sourceRegistry: registry.scality.com/zenko-operator
   image: zenko-operator
-  tag: 1.5.43
+  tag: 1.5.44
```
francoisferrand: not needed anymore, already bump'ed

tests/ctst/common/common.ts
```diff
@@ -300,40 +300,58 @@ Then('i {string} be able to add user metadata to object {string}',
         }
     });
 
-Then('kafka consumed messages should not take too much place on disk',
+Then('kafka consumed messages should not take too much place on disk', { timeout: 130000 },
```
francoisferrand: timeout should should be based on `KafkaCleanerInterval`, i.e. a number of time the interval?

tests/ctst/common/common.ts
```diff
@@ -300,40 +300,58 @@ Then('i {string} be able to add user metadata to object {string}',
...
 
-        // Checking topics offsets before kafkacleaner passes to be sure kafkacleaner works
-        // This function can be improved by consuming messages and
-        // verify that the timestamp is not older than last kafkacleaner run
-        // Instead of waiting for a fixed amount of time,
-        // we could also check for metrics to see last kafkacleaner run
-        
-        // 10 seconds added to be sure kafkacleaner had time to process
-        await Utils.sleep(seconds * 1000 + 10000);
+        while (topics.length > 0 && new Date().getTime() - timeStart < 120000) {
```
francoisferrand: what is the point of checking the duration?
* the timeout on the test is designed to catch test taking too long to complete
* here we simply want to try "a few times" (or even not limit, and just fail without timeout if it never settles...)

tests/ctst/common/common.ts
```diff
@@ -300,40 +300,58 @@ Then('i {string} be able to add user metadata to object {string}',
...
+            const previousOffsets = await getTopicsOffsets(topics, kafkaAdmin);
 
-        const newOffsets = await getTopicsOffsets(topics, kafkaAdmin);
+            const seconds = parseInt(this.parameters.KafkaCleanerInterval);
```
francoisferrand: no point doing this in the loop, just compute once before the loop

tests/ctst/common/common.ts
```diff
@@ -300,40 +300,58 @@ Then('i {string} be able to add user metadata to object {string}',
...
+                        `Topic ${topics[i]} partition ${j} offset has not increased,
+                        previousOffsets: ${previousOffsets[i].partitions[j].low} /\
+                         ${previousOffsets[i].partitions[j].high},
+                        newOffsets: ${newOffsets[i].partitions[j].low} / ${newOffsets[i].partitions[j].high}`);
+                    topics.splice(i, 1);
+                }
             }
         }
+
+        assert(topics.length === 0, `Topics ${topics.join(', ')} still have not been cleaned`);
```
francoisferrand: here we should fail if we exited the loop with the "`newMessagesAfterClean`" case, i.e. if it never converged

tests/ctst/common/common.ts
```diff
@@ -300,40 +300,58 @@ Then('i {string} be able to add user metadata to object {string}',
...
+            await Utils.sleep(seconds * 1000 + 10000);
+
+            const newOffsets = await getTopicsOffsets(topics, kafkaAdmin);
+
+            for (let i = 0; i < topics.length; i++) {
+                process.stdout.write(`\nChecking topic ${topics[i]}\n`);
+                for (let j = 0; j < newOffsets[i].partitions.length; j++) {
+                    // Checking that the min offset has increased due to kafkacleaner
+                    // or that it didn't need to change because there was no new messages
+                    // or new messages came after a clean but still cleaned to previous high offset
```
francoisferrand: the comment is hard to relate with the code now...

tests/ctst/common/common.ts
```diff
@@ -300,40 +300,58 @@ Then('i {string} be able to add user metadata to object {string}',
...
+                    const lowOffsetIncreased = newOffsets[i].partitions[j].low > previousOffsets[i].partitions[j].low;
+                    const noNewMessages = newOffsets[i].partitions[j].high === newOffsets[i].partitions[j].low;
```
francoisferrand: not used here → move after the `if` for clarify, where they will be used

tests/ctst/common/common.ts
KillianG: This is to end properly and not getting killed by the timeout so at the end we can log each topic that is not cleaned in time, if we just wait for the timeout, we'll end up with no readable logs.. just "test timed out" with no info..

tests/ctst/common/common.ts
KillianG: Should be great but unfortuneatly we can't access `Zenko` object here :(
And I am not sure if this is great anyway, because if we set kafkacleaner interval to 10mn for some reason, we'll end up with 30mn timeout ?

tests/ctst/common/common.ts
KillianG: Maybe i missed something, but this is already what it does, while there is `newMessagesAfterClean` we continue (i.e we don't remove the topic from the list of topics to be checked) and here, if there is topics that remains in the array, it's that they haven't been cleaned and we print them to be able to debug in case it happens

tests/ctst/common/common.ts
```diff
@@ -300,40 +300,68 @@ Then('i {string} be able to add user metadata to object {string}',
...
-            for (let j = 0; j < newOffsets[i].partitions.length; j++) {
-                // Checking that the min offset has increased due to kafkacleaner
-                // or that it didn't need to change because there was no new messages
-                assert.ok(newOffsets[i].partitions[j].low > previousOffsets[i].partitions[j].low ||
-                    newOffsets[i].partitions[j].high === newOffsets[i].partitions[j].low,
-                `Topic ${topics[i]} partition ${j} offset has not increased,
-                previousOffsets: ${previousOffsets[i].partitions[j].low} / ${previousOffsets[i].partitions[j].high},
-                newOffsets: ${newOffsets[i].partitions[j].low} / ${newOffsets[i].partitions[j].high}`);
+
+
```
francoisferrand: nit: extra empty line

tests/ctst/common/common.ts
```diff
@@ -300,40 +300,68 @@ Then('i {string} be able to add user metadata to object {string}',
...
+                        previousOffsets: ${previousOffsets[i].partitions[j].low} /\
+                         ${previousOffsets[i].partitions[j].high},
+                        newOffsets: ${newOffsets[i].partitions[j].low} / ${newOffsets[i].partitions[j].high}`);
+
+                    // Topic is cleaned, we don't need to check it anymore
+                    topics.splice(i, 1);
+                }
             }
         }
+        clearTimeout(timeoutID);
```
francoisferrand: in case of `assert` before, this will not be executed : which could cause strange behavior.

a safer approach would be to use `try{}finally{}` block:

```suggestion
        const timeoutID = setTimeout(....);
        try {
           //run the check
        } finally {
           clearTimeout(timeoutID);
        }
```

(atlernatively, we could maybe [`async.timeout`](https://caolan.github.io/async/v3/docs.html#timeout) instead)

tests/ctst/common/common.ts
```diff
@@ -300,40 +300,72 @@ Then('i {string} be able to add user metadata to object {string}',
...
-            for (let j = 0; j < newOffsets[i].partitions.length; j++) {
-                // Checking that the min offset has increased due to kafkacleaner
-                // or that it didn't need to change because there was no new messages
-                assert.ok(newOffsets[i].partitions[j].low > previousOffsets[i].partitions[j].low ||
-                    newOffsets[i].partitions[j].high === newOffsets[i].partitions[j].low,
-                `Topic ${topics[i]} partition ${j} offset has not increased,
-                previousOffsets: ${previousOffsets[i].partitions[j].low} / ${previousOffsets[i].partitions[j].high},
-                newOffsets: ${newOffsets[i].partitions[j].low} / ${newOffsets[i].partitions[j].high}`);
+
+        try {
```
francoisferrand, line 311: should encompass listing of kafka topic as well, i.e. start immediately after creating the timer

francoisferrand (approved):

PR #2028, opened 2024-02-27
Improvement/zenko 4730 mongodb v6.0

benzekrimaha:
This PRs aims to bump the mongo dependency from 5 to 6 

Linked Issue : https://scality.atlassian.net/browse/ZENKO-4730


.github/scripts/end2end/deploy-zenko.sh
ghost, line 1: No need to keep these changes (only there for testing)

.github/workflows/end2end.yaml
ghost, line 1: This can be kept with sleeps, but should only run when we specify a "debug" flag, like in Integration

solution-base/mongodb/patches/mongodb-sharded-add-configsvr-service.patch
```diff
@@ -14,64 +14,22 @@ index 04b183a2..0fe370f2 100644
    {{- if .Values.configsvr.external.replicasetName -}}
      {{- .Values.configsvr.external.replicasetName }}
 diff --git a/solution-base/mongodb/charts/mongodb-sharded/templates/config-server/config-server-service.yaml b/solution-base/mongodb/charts/mongodb-sharded/templates/config-server/config-server-service.yaml
-new file mode 100644
-index 00000000..fff5cea7
---- /dev/null
+index fad94092..825b0161 100644
+--- a/solution-base/mongodb/charts/mongodb-sharded/templates/config-server/config-server-service.yaml
```
ghost: Actually this file is not part of the charts, so if someone deletes the associated yaml, then fetch all the charts, this patch file won't work, I think it's fine, for this one, to have the patch actually creating the whole file, like previously

.github/scripts/end2end/deploy-zenko.sh
```diff
@@ -142,4 +142,4 @@ for i in $(seq 1 120); do
     fi
     # Debug log to ease understanding of failures in the CI
     kubectl get pods -A
-done
+done
```
francoisferrand: nit: EOL should not be removed (and generally should not change files when not needed)

.github/scripts/end2end/kustomization.yaml
```diff
@@ -46,13 +46,13 @@ images:
   newTag: 0.34.0-debian-11-r24
 - name: metalk8s-registry-from-config.invalid/zenko-base-2.3.11/mongodb-sharded
   newName: docker.io/bitnami/mongodb-sharded
-  newTag: 5.0.18-debian-11-r15
+  newTag: 6.0.10-debian-11-r8
 - name: metalk8s-registry-from-config.invalid/zenko-base-2.3.18/bitnami-shell
   newName: docker.io/bitnami/bitnami-shell
   newTag: 10-debian-10-r404
```
francoisferrand: * bitnami shell changed as well
* it would be better to automatically "pick up" these values from solution-base, somehow...

.github/workflows/end2end.yaml
```diff
@@ -428,6 +428,19 @@ jobs:
         env:
           STAGE: end2end-http
         if: always()
+      - name: "Debug: SSH to runner"
```
francoisferrand: not part of this PR, should be in https://scality.atlassian.net/browse/ZENKO-4749

solution-base/mongodb/scripts-sharded/create-app-user-sharded.sh
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash
```
francoisferrand: * these scripts are named "-sharded" : so we can put them in the same folder as the other scripts
* there is lot of redundancy here between the sharded and non-sharded scripts:
   - for create-app-user, the diff seems to only be the `mongosh` (vs `mongo`) command --> we should align (check version to use either mongo or mongosh? find from some other env variable? automatically 'detect' which one is available? ...)
    - for set-default-write-concern, it does not depend on sharding but on the version of mongo... and furtunately we don't use it in non-sharded mongo deployment

solution-base/mongodb/how_to_upgrade.md
```diff
@@ -0,0 +1,18 @@
...
+
+## Upgrade steps
+
+
+1. Check for all available versions of the chart using `helm search repo bitnami/mongodb-sharded --versions` to check the configuration values for a specified version : `helm show values --version 6.6.7  bitnami/mongodb-sharded`
+2. Bump mongo version in Kustomization.yaml and deps.yaml files 
+3. Upgrade mongodb sharded version in charts => `CHART_MONGO_SHARDED_VERSION` in the solution-base/mongodb/Makefile (e.g CHART_MONGO_SHARDED_VERSION:="6.6.7")
+4. Upgrade charts to the version targetted => `make fetch-mongodb-sharded`
+5. Manually update the patches by applying the changes manually in the new upgraded charts (note that this can be applied only to patches that target mongo sharded) =>  `git diff -- solution-base/mongodb/charts/mongodb-sharded/values.yaml > solution-base/mongodb/patches/secret-name.patch` (this operation needs to be done for every patch).
+6. Once the patches updated apply them to the charts with the command `make patch`
```
francoisferrand: to ensure this step is not missed, we should add a pre-merge step where we run step 4 (`make fetch-mongodb-sharded`), and check the resulting git diff is empty:

--> this will ensure that every modification we perform on the charts is tracked in the patches as well, giving us confidence in future upgrades

.github/workflows/end2end.yaml
benzekrimaha: Yes , all of the debug related commits will be dropped

solution-base/mongodb/how_to_upgrade.md
benzekrimaha: Done here : https://github.com/scality/Zenko/pull/2028/commits/5dad6206707553b6d64405df7144d62915f491b7#diff-cf60998e2851400a5ff7faeaf9677e20a46bab4b90a3b4934293e480ca335d2fR657

solution-base/mongodb/scripts-sharded/create-app-user-sharded.sh
benzekrimaha: => Done here : https://github.com/scality/Zenko/pull/2028/commits/c54d619978cc1070b11672c258e0b88d38cbbf15#diff-f3e6344f5181df0001c53dac90d9291cb2ff6105589ab8a3d277e4f379f3cc5bR3
Keeping only one file and switching commands

.github/scripts/end2end/deploy-zenko.sh
```diff
@@ -141,11 +141,17 @@ env $(dependencies_env) envsubst < ${ZENKOVERSION_PATH} | kubectl -n ${NAMESPACE
 env $(dependencies_env) envsubst < ${ZENKO_CR_PATH} | kubectl -n ${NAMESPACE} apply -f -
 
 k_cmd="kubectl -n ${NAMESPACE} get zenko/${ZENKO_NAME}"
-for i in $(seq 1 120); do
+for i in $(seq 1 200); do
```
ghost: Is mongodb 6 slower to start?

.github/scripts/end2end/deploy-zenko.sh
```diff
@@ -141,11 +141,17 @@ env $(dependencies_env) envsubst < ${ZENKOVERSION_PATH} | kubectl -n ${NAMESPACE
...
     fi
     # Debug log to ease understanding of failures in the CI
     kubectl get pods -A
 done
+# auto describe any pod that is still failing at this point
+# include a describe and logs from this pod
+# kubectl get pods -A -o jsonpath='{range .items[*]}{.metadata.namespace}{"\t"}{.metadata.name}{"\n"}{end}' | while read ns pod; do
+#     kubectl describe pod -n $ns $pod
+#     kubectl logs -n $ns $pod
+# done
```
ghost: Maybe something to drop (or have it on all branches)

solution-base/mongodb/scripts/create-app-user.sh
ghost, line 1: I saw also the prepare-pra.sh file that still uses the `mongo` bin, we could use the same approach as here (or just use mongosh, as we always use sharded mongo)

solution-base/yarn.lock
ghost, line 1: To be removed

solution-base/mongodb/Makefile
```diff
@@ -25,7 +25,12 @@ fetch-mongodb-sharded:
 		--version ${CHART_MONGO_SHARDED_VERSION} \
 		--untar \
 		--untardir ${CHART_DIR}
-
 patch:
-	@git apply --check ${PATCH_FILES}
-	@git apply ${PATCH_FILES}
+	@for patch_file in $(PATCH_FILES); do \
+		if [ "$$patch_file" != "${PATCH_DIR}/create-app-creds.patch" ] && [ "$$patch_file" != "${PATCH_DIR}/credentials-from-secret.patch" ]; then \
```
ghost: Maybe we can have a comment here to explain why these files are excluded?

.github/workflows/end2end.yaml
```diff
@@ -742,6 +742,37 @@ jobs:
           source: /tmp/artifacts
         if: always()
 
+  mongo-patches-check:
```
ghost: I would suggest having this as a condition before running build-iso, as we already check dashboards version in there, and the iso will be wrong anyway, if the patches were not properly applied

.github/workflows/end2end.yaml
```diff
@@ -742,6 +742,37 @@ jobs:
...
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+      - name: run patches
+        run: make -C solution-base/mongodb patch
+      - name: Check if patches were applied right
```
ghost:
```suggestion
      - name: Ensure mongodb charts are patched
```

.github/workflows/end2end.yaml
```diff
@@ -742,6 +742,37 @@ jobs:
...
+        run: |
+          mkdir -p /tmp/artifacts/build_status/${{ github.job }}
+          echo 'SUCCESSFUL' > /tmp/artifacts/build_status/${{ github.job }}/.final_status
+      - name: Set status in ARTIFACTS to enable re-run only failed jobs
+        if: failure()
+        run: |
+          mkdir -p /tmp/artifacts/build_status/${{ github.job }}
+          echo 'FAILED' > /tmp/artifacts/build_status/${{ github.job }}/.final_status
+      - name: Upload artifacts
+        uses: scality/action-artifacts@v3
```
ghost:
```suggestion
        uses: scality/action-artifacts@v4
```

.github/workflows/end2end.yaml
```diff
@@ -742,6 +742,37 @@ jobs:
           source: /tmp/artifacts
         if: always()
 
+  mongo-patches-check:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
```
ghost:
```suggestion
        uses: actions/checkout@v4
```

.github/workflows/end2end.yaml
```diff
@@ -742,6 +742,37 @@ jobs:
...
+      - name: Set status in ARTIFACTS to enable re-run only failed jobs
+        if: success()
+        run: |
+          mkdir -p /tmp/artifacts/build_status/${{ github.job }}
+          echo 'SUCCESSFUL' > /tmp/artifacts/build_status/${{ github.job }}/.final_status
+      - name: Set status in ARTIFACTS to enable re-run only failed jobs
+        if: failure()
+        run: |
+          mkdir -p /tmp/artifacts/build_status/${{ github.job }}
+          echo 'FAILED' > /tmp/artifacts/build_status/${{ github.job }}/.final_status
```
ghost: Is this really needed? We already have `write-final-success-status` and `write-final-failed-status` to handle that at the end of the build, if anything failed

solution-base/mongodb/how_to_upgrade.md
```diff
@@ -0,0 +1,18 @@
+# MongoDB Sharded Upgrade Process
+
+## Overview
+
+Upgrading MongoDB sharded involves several steps to ensure a smooth transition to the new version. Below is a structured guide to follow for a successful upgrade (note that all along
+I'll be taking as an example the upgrade from 5 to 6) :
```
ghost:
```suggestion
I will be taking as an example the upgrade from 5 to 6):
```

solution-base/mongodb/how_to_upgrade.md
```diff
@@ -0,0 +1,18 @@
...
+
+## Upgrade steps
+
+
+1. Check for all available versions of the chart using `helm search repo bitnami/mongodb-sharded --versions` to check the configuration values for a specified version : `helm show values --version 6.6.7  bitnami/mongodb-sharded`
```
ghost:
```suggestion
1. Check for all available versions of the chart using `helm search repo bitnami/mongodb-sharded --versions` to check the configuration values for a specified version: `helm show values --version 6.6.7  bitnami/mongodb-sharded`
```

solution-base/mongodb/how_to_upgrade.md
```diff
@@ -0,0 +1,18 @@
...
+2. Bump mongo version in Kustomization.yaml and deps.yaml files
```
ghost:
```suggestion
2. Bump mongo version in `.github/scripts/end2end/kustomization.yaml` and `solution-base/deps.yaml` files 
```

solution-base/mongodb/how_to_upgrade.md
```diff
@@ -0,0 +1,18 @@
...
+
+Upgrading MongoDB sharded involves several steps to ensure a smooth transition to the new version. Below is a structured guide to follow for a successful upgrade (note that all along
+I'll be taking as an example the upgrade from 5 to 6) :
+
+## Upgrade steps
+
+
+1. Check for all available versions of the chart using `helm search repo bitnami/mongodb-sharded --versions` to check the configuration values for a specified version : `helm show values --version 6.6.7  bitnami/mongodb-sharded`
+2. Bump mongo version in Kustomization.yaml and deps.yaml files 
+3. Upgrade mongodb sharded version in charts => `CHART_MONGO_SHARDED_VERSION` in the solution-base/mongodb/Makefile (e.g CHART_MONGO_SHARDED_VERSION:="6.6.7")
```
ghost:
```suggestion
3. Upgrade mongodb sharded version in charts => `CHART_MONGO_SHARDED_VERSION` in the `solution-base/mongodb/Makefile` file (e.g. `CHART_MONGO_SHARDED_VERSION:="6.6.7"`)
```

solution-base/mongodb/how_to_upgrade.md
ghost:
```suggestion
6. Once the patches are updated, apply them to the charts with the command `make patch`
```

solution-base/mongodb/how_to_upgrade.md
```diff
@@ -0,0 +1,18 @@
...
+4. Upgrade charts to the version targetted => `make fetch-mongodb-sharded`
+5. Manually update the patches by applying the changes manually in the new upgraded charts (note that this can be applied only to patches that target mongo sharded) =>  `git diff -- solution-base/mongodb/charts/mongodb-sharded/values.yaml > solution-base/mongodb/patches/secret-name.patch` (this operation needs to be done for every patch).
+6. Once the patches updated apply them to the charts with the command `make patch`
+7. Applying the changes introduced by the upgrade can then be done (by checking the release notes for the new version e.g `https://www.mongodb.com/docs/v6.0/release-notes/6.0/`)
+8. The CI has to pass for the update to be considered effective.
```
ghost:
```suggestion
8. The CI has to pass for the update to be considered effective.
9. Perform an upgrade check with the product(s) using this Zenko version, before merging the upgrade PR. Some changes may be required at upper-levels (e.g., update mongosh commands, or add new logic in the upgrade).
```

ghost (approved):

solution-base/mongodb/charts/mongodb-sharded/values.yaml
```diff
@@ -350,7 +350,8 @@ configsvr:
   ## mongodbExtraFlags:
   ##  - "--wiredTigerCacheSizeGB=2"
   ##
-  mongodbExtraFlags: []
+  mongodbExtraFlags:
+    - "--oplogSize=1024"
```
benzekrimaha: not needed here as the patch was removed...

.github/workflows/end2end.yaml
```diff
@@ -720,6 +736,35 @@ jobs:
           source: /tmp/artifacts
         if: always()
 
+  write-final-failed-status:
```
francoisferrand: this is not correctly merged, `write-final-failed-status` has been removed (both the "failed" and "success" jobs have been replaced with a single `write-final-status` job)

solution/deps.yaml
```diff
@@ -64,6 +64,9 @@ kafka-connect:
   image: kafka-connect
   tag: 2.13-3.1.2-1.13.0
   envsubst: KAFKA_CONNECT_TAG
+mongodb-connector:
```
francoisferrand: bad merge, this dependency is already present somewhere else in the file

solution-base/yarn.lock
francoisferrand, line 1: no reason to have `yarn.lock` file, should not be in this PR

solution-base/mongodb/patches/deps.patch
```diff
@@ -0,0 +1,31 @@
+diff --git a/solution-base/mongodb/charts/mongodb-sharded/values.yaml b/solution-base/mongodb/charts/mongodb-sharded/values.yaml
+index b55ebd40..9cb2131f 100644
+--- a/solution-base/mongodb/charts/mongodb-sharded/values.yaml
++++ b/solution-base/mongodb/charts/mongodb-sharded/values.yaml
+@@ -79,7 +79,7 @@ diagnosticMode:
+ image:
+   registry: docker.io
+   repository: bitnami/mongodb-sharded
+-  tag: 6.0.10-debian-11-r8
++  tag: 6.0.12-debian-11-r5
```
francoisferrand: why do we need to patch these `values` ?
we specify/override them anyway when we render the chart (reading them from our one source of truth, `deps.yaml`) : no point duplicating these values

--> can't we remove this patch ?

solution-base/mongodb/patches/statefulset-permissions.patch
francoisferrand, line 1: is this aptch still needed?

there seems to be a ` .Values.configsvr.containerSecurityContext.runAsUser` value now, we shoudl be able to use it without patching the chart

solution-base/mongodb/patches/mongodb-sharded-fix-podmonitor.patch
```diff
@@ -1,20 +1,21 @@
...
  apiVersion: monitoring.coreos.com/v1
  kind: PodMonitor
  metadata:
-@@ -32,4 +31,3 @@ spec:
-     matchLabels: {{- include "common.labels.matchLabels" $ | nindent 6 }}
+@@ -37,6 +36,4 @@ spec:
+     matchLabels: {{- include "common.labels.matchLabels" ( dict "customLabels" $podLabels "context" $ ) | nindent 6 }}
        app.kubernetes.io/component: shardsvr
- {{- end }}
+       shard: {{ $i | quote }}
```
francoisferrand: should not add a `shard` label filter, since we create a single podMonitor which matches every shard's pod : either we keep the loop (and need to filter on the shard), or we keep a single one and must not filter on the shard.

(later solution prefered, as it was before)

francoisferrand (changes requested):

solution-base/mongodb/patches/mongodb-sharded-fix-podmonitor.patch
benzekrimaha: this was not added manually , it's coming from the charts update ( using `make fetch-mongodb-sharded ` ) , the patch was just updated to apply the already existing changes , should we patch this as well ?

.github/workflows/end2end.yaml
```diff
@@ -187,9 +187,25 @@ jobs:
...
+          git diff --exit-code || { echo "Error: Changes detected after running 'make fetch-mongodb-sharded'. Please ensure step is completed."; exit 1; }
+        if: always()
+
   build-iso:
     runs-on: ubuntu-22.04
-    needs: [build-kafka, check-dashboard-versions]
+    needs:
+      - build-kafka
+      - check-dashboard-versions
+      - mongo-patches-check
```
francoisferrand: `mongo-patches-check` should not block `build-iso`, there is no dependency

.github/workflows/end2end.yaml
```diff
@@ -187,9 +187,25 @@ jobs:
...
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+      - name: run patches
+        run: make -C solution-base/mongodb patch
+      - name: Ensure mongodb charts are patched
+        run: |
+          git diff --exit-code || { echo "Error: Changes detected after running 'make fetch-mongodb-sharded'. Please ensure step is completed."; exit 1; }
+        if: always()
```
francoisferrand: why `always()` ?
- if checkout fails, no point doing anything more
- `make patch` should not fail unless there are changes to apply, and these fail to apply --> if it fails, not point doing git diff

solution-base/mongodb/patches/mongodb-sharded-fix-podmonitor.patch
francoisferrand: Yes, since we have a single podMonitor, it **must not** filter on the shard...

Or, maybe the podMonitors are now properly generated (by the chart), with one podMonitor per shard & proper filtering; and we could thus remove this patch altogether...

solution-base/mongodb/scripts/set-default-write-concern-majority.sh
```diff
@@ -1,7 +1,13 @@
 #!/bin/bash
 
+if which mongo &> /dev/null; then
+    mongoCmd="mongo"
+else
+    mongoCmd="mongosh"
+fi
+
```
francoisferrand: can now be clean'ed up: this is always `mongosh`

solution-base/mongodb/scripts/create-app-user.sh
```diff
@@ -1,7 +1,13 @@
 #!/bin/bash
 
-create_user() {
-    mongo --host "mongodb://127.0.0.1/?replicaSet=${MONGODB_REPLICA_SET_NAME}"  -u 'root' -p "$MONGODB_ROOT_PASSWORD" <<EOF
+if which mongo &> /dev/null; then
+    mongoCmd="mongo"
+else
+    mongoCmd="mongosh"
+fi
+
```
francoisferrand: can now be clean'ed up: this is always mongosh

solution-base/mongodb/how_to_upgrade.md
```diff
@@ -0,0 +1,19 @@
...
+
+Upgrading MongoDB sharded involves several steps to ensure a smooth transition to the new version. Below is a structured guide to follow for a successful upgrade (note that all along
+I will be taking as an example the upgrade from 5 to 6):
+
+## Upgrade steps
+
+
+1. Check for all available versions of the chart using `helm search repo bitnami/mongodb-sharded --versions` to check the configuration values for a specified version: `helm show values --version 6.6.7  bitnami/mongodb-sharded`
+2. Bump mongo version in `.github/scripts/end2end/kustomization.yaml` and `solution-base/deps.yaml` files.
+3. Upgrade mongodb sharded version in charts => `CHART_MONGO_SHARDED_VERSION` in the `solution-base/mongodb/Makefile` file (e.g. `CHART_MONGO_SHARDED_VERSION:="6.6.7"`)
```
francoisferrand, line 13: makefile should be updated to pick up the values from `solution-base/deps.yaml`

solution-base/mongodb/how_to_upgrade.md
```diff
@@ -0,0 +1,19 @@
...
+4. Upgrade charts to the version targetted => `make fetch-mongodb-sharded`
+5. Manually update the patches by applying the changes manually in the new upgraded charts (note that this can be applied only to patches that target mongo sharded) =>  `git diff -- solution-base/mongodb/charts/mongodb-sharded/values.yaml > solution-base/mongodb/patches/secret-name.patch` (this operation needs to be done for every patch).
+6. Once the patches are updated, apply them to the charts with the command `make patch`.
+7. Applying the changes introduced by the upgrade can then be done (by checking the release notes for the new version e.g `https://www.mongodb.com/docs/v6.0/release-notes/6.0/`)
```
francoisferrand: i don't understand what you mean withi this step. You already "applied the changes" earlier...

solution-base/mongodb/how_to_upgrade.md
```diff
@@ -0,0 +1,19 @@
...
+8. The CI has to pass for the update to be considered effective.
+9. Perform an upgrade check with the product(s) using this Zenko version, before merging the upgrade PR. Some changes may be required at upper-levels (e.g., update mongosh commands, or add new logic in the upgrade).
```
francoisferrand: nothing specific to mongo upgrades here. This applies to every code change: the ci has to pass, and you need to test, considering upgrades when appropriate.

solution-base/mongodb/how_to_upgrade.md
```diff
@@ -0,0 +1,19 @@
...
+## Overview
+
+Upgrading MongoDB sharded involves several steps to ensure a smooth transition to the new version. Below is a structured guide to follow for a successful upgrade (note that all along
+I will be taking as an example the upgrade from 5 to 6):
+
+## Upgrade steps
+
+
+1. Check for all available versions of the chart using `helm search repo bitnami/mongodb-sharded --versions` to check the configuration values for a specified version: `helm show values --version 6.6.7  bitnami/mongodb-sharded`
+2. Bump mongo version in `.github/scripts/end2end/kustomization.yaml` and `solution-base/deps.yaml` files.
```
francoisferrand: `.github/scripts/end2end/kustomization.yaml` does nto include the mongo version

.github/workflows/end2end.yaml
francoisferrand: maybe the patch were not applied, or some patches were missing or not commited.

there is no dependency between these jobs, the `mongo-patches-check` job will be (quickly) visible as failed when it happens, and the final build status will be failed in any of the job fails : so no reason to introduce dependency, better to keep everything independent.

solution-base/mongodb/how_to_upgrade.md
benzekrimaha: i am talking about the additional changes ( like in the case of mongo 6 the switch from mongo to mongosh that is not done directly on the charts/patches)

solution-base/mongodb/how_to_upgrade.md
benzekrimaha: IMHO as this is a guide , we can keep it because for mongo the upgrade is mandatory not optional to make sure everything is working fine

solution-base/mongodb/Makefile
```diff
@@ -3,10 +3,10 @@ CHART_DIR:="${ROOT_DIR}/charts"
 
 CHART_REPO:="https://charts.bitnami.com/bitnami"
 CHART_MONGO_REPLICASET_VERSION:="7.8.0"
```
francoisferrand, line 5: `CHART_MONGO_REPLICASET_VERSION:="7.8.0"` not used anymore, should be removed...
(left-over from non-sharded)

solution-base/mongodb/how_to_upgrade.md
francoisferrand: the previous statement undermines this:
> The CI has to pass for the update to be considered effective.

As for any change, it must first be tested: the dev must deploy his change, and indeed in case of mongo also try an upgrade (though the upgrade part is probably not so much more critical than other changes)

Only after testing should the PR is open for review, and CI looked at (CI is a gatekeeper, not a first stage of work)

ghost: > nothing specific to mongo upgrades here. This applies to every code change: the ci has to pass, and you need to test, considering upgrades when appropriate.
...
> As for any change, it must first be tested: the dev must deploy his change, and indeed in case of mongo also try an upgrade (though the upgrade part is probably not so much more critical than other changes)

To me this is debatable, today we test nothing in Artesca CI that comes from Zenko, beyond the UI and its deployment. It should ideally not be required to test every change we introduce in Zenko in Artesca. And actually, we do not test everything we integrate. Also, this process may be done by someone else in the future, without ocntext of all the problems we faced till now.

For MongoDB, as per my experience, it may pass Zenko CI, Artesca CI but fail the nightly tests, or we could find issues with the new k8s version, or face upgrade issues not visible in the CI (everything passes, but in fact it is not working properly). So the main point here is really to stress that upgrading mongodb is not trivial and complete testing must be done before.

Maybe we can rephrase it better, but I strongly suggest that we keep this, and don't hide it under the "good practises", as it is not trivial...

.github/workflows/end2end.yaml
ghost: IMHO it's better to have it as a condition; if the patches are not applied, we should block any ISO generation, as the resulting base iso will not work as expected - and the whole build would fail anyway. It's not slowing down the build either.

solution-base/mongodb/how_to_upgrade.md
francoisferrand: please reword to make this easier to understand

solution-base/mongodb/how_to_upgrade.md
francoisferrand: my point is not necessarily to remove it, but that either we need to give an "exhautive" test plan, or none at all : but saying "ci has to pass ; you must perform an upgrade" is opening the door to only perform an upgrade...

i.e. upgrading without checking that all _features_ are still working (esp. oplog management is notoriously picky, and we have quite a few other use cases: custom listing & index management in lifecycle, indexes in sorbet, monitoring & dashboards...)

→ if the point is about listing all the test that need to be performed, then the list must be extended significantly
→ if the point is only about the need to perform an upgrade, then it should be reworded indeed. Maybe just have a generic "_test all features to verify they are not affected by the new version; check the logs to verify there is no "silent errors", and don't forget to check that upgrade is working as well_"
→ the line about the CI is really out of place, if you really want to mention CI it should be the other way : "_Passing CI tests is not enough, you must also ...._"

.github/workflows/end2end.yaml
```diff
@@ -187,9 +187,25 @@ jobs:
...
+          git diff --exit-code || { echo "Error: Changes detected after running 'make fetch-mongodb-sharded'. Please ensure step is completed."; exit 1; }
+        if: success()
+
   build-iso:
     runs-on: ubuntu-22.04
-    needs: [build-kafka, check-dashboard-versions]
+    needs:
+      - build-kafka
+      - check-dashboard-versions
+
```
francoisferrand: nit: extra empty line (or keep the way it was written already)

solution-base/mongodb/Makefile
francoisferrand: why do we exclude these patches?
- if we exclude them, we might as well remove the patches altogether
- instead of re-doing what `git apply` does, would be simpler to to just filter the files in `PATCH_FILES`, something like:
   ```
   PATCH_FILES := $(shell ls -d ${PATCH_DIR}/* | grep -v -e create-app-creds.patch -e credentials-from-secret.patch)
   ```

solution-base/mongodb/Makefile
```diff
@@ -2,30 +2,27 @@ ROOT_DIR:=$(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
...
-		--version ${CHART_MONGO_REPLICASET_VERSION} \
-		--untar \
-		--untardir ${CHART_DIR}
-
 fetch-mongodb-sharded:
 	@${HELM} fetch mongodb-sharded \
 		--repo ${CHART_REPO} \
 		--version ${CHART_MONGO_SHARDED_VERSION} \
 		--untar \
 		--untardir ${CHART_DIR}
```
francoisferrand: nit: keep empty line between makefile rules

solution-base/mongodb/how_to_upgrade.md
```diff
@@ -0,0 +1,19 @@
...
+
+Upgrading MongoDB sharded involves several steps to ensure a smooth transition to the new version. Below is a structured guide to follow for a successful upgrade (note that all along
+I will be taking as an example the upgrade from 5 to 6):
+
+## Upgrade steps
+
+
+1. Check for all available versions of the chart using `helm search repo bitnami/mongodb-sharded --versions` to check the configuration values for a specified version: `helm show values --version 6.6.7  bitnami/mongodb-sharded`
+2. Bump mongo version in `solution-base/deps.yaml` files.
+3. Upgrade mongodb sharded version in charts => `CHART_MONGO_SHARDED_VERSION` in the `solution-base/mongodb/Makefile` file (e.g. `CHART_MONGO_SHARDED_VERSION:="6.6.7"`)
```
francoisferrand: nit:
```suggestion
3. Bump mongodb sharded version in charts => `CHART_MONGO_SHARDED_VERSION` in the `solution-base/mongodb/Makefile` file (e.g. `CHART_MONGO_SHARDED_VERSION:="6.6.7"`)
```

.github/workflows/end2end.yaml
```diff
@@ -187,9 +187,24 @@ jobs:
...
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+      - name: run patches
+        run: make -C solution-base/mongodb patch
+      - name: Ensure mongodb charts are patched
+        run: |
+          git diff --exit-code || { echo "Error: Changes detected after running 'make fetch-mongodb-sharded'. Please ensure step is completed."; exit 1; }
+        if: success()
```
francoisferrand: `if: success()` is the (implicit) default behavior --> not needed

solution-base/mongodb/how_to_upgrade.md
benzekrimaha: https://github.com/scality/Zenko/pull/2028/commits/4809ce468c3808ca839a6924c79b0931dcd09e09#diff-a2232853b265dbe8c46e493c507a5b212f5e815cd22ab84e8a526ad808d3dab8R17

solution-base/mongodb/Makefile
benzekrimaha: as the files were removed , the if was removed as well

solution-base/mongodb/patches/mongodb-sharded-fix-podmonitor.patch
benzekrimaha: patch updated here : https://github.com/scality/Zenko/pull/2028/commits/278f25041e230c660f3642820ad413f93a4e16fd#diff-bb1ea55e85cfff3a0fc51b35e678b12f7421adf03e1f4842710a3fe7f80f50a0R1-R24 
Indeed the updated chart creates one podmonitor/shard , went back to the previous solution : keeping a single one and not filtering on the shard

solution-base/mongodb/patches/statefulset-permissions.patch
benzekrimaha: The patch has been removing , checking to see if there are no side effects will update soon

solution-base/mongodb/Makefile
francoisferrand: why not apply as before?
both chunks can be removed, no reason to make code more complex than it was....

solution-base/mongodb/patches/statefulset-permissions.patch
benzekrimaha: we still need the patch , creating the repo then changing ownership instead of just running the command results in the following error => `2024-09-24T11:34:43.597879041Z stdout F {"t":{"$date":"2024-09-24T11:34:43.597+00:00"},"s":"W",  "c":"QUERY",    "id":20478,   "ctx":"conn903","msg":"getMore command executor error","attr":{"error":{"code":237,"codeName":"CursorKilled","errmsg":"operation was interrupted"},"stats":{},"cmd":{"getMore":4981207424808190151,"collection":"shards","maxTimeMS":1000}}} `

solution-base/mongodb/Makefile
```diff
@@ -3,10 +3,10 @@ CHART_DIR:="${ROOT_DIR}/charts"
 
 CHART_REPO:="https://charts.bitnami.com/bitnami"
 CHART_MONGO_REPLICASET_VERSION:="7.8.0"
-CHART_MONGO_SHARDED_VERSION:="5.1.4"
+CHART_MONGO_SHARDED_VERSION:="6.6.7"
 
 PATCH_DIR:="${ROOT_DIR}/patches"
-PATCH_FILES:="$(shell ls -d ${PATCH_DIR}/*)"
+PATCH_FILES := $(shell ls -d ${PATCH_DIR}/* | tr '\n' ' ')
```
francoisferrand: change not needed

solution-base/mongodb/Makefile
```diff
@@ -25,7 +25,10 @@ fetch-mongodb-sharded:
 		--version ${CHART_MONGO_SHARDED_VERSION} \
 		--untar \
 		--untardir ${CHART_DIR}
-
 patch:
-	@git apply --check ${PATCH_FILES}
-	@git apply ${PATCH_FILES}
+	@for patch_file in $(PATCH_FILES); do \
```
francoisferrand, line 29: change not needed

solution-base/mongodb/Makefile
```diff
@@ -25,7 +25,10 @@ fetch-mongodb-sharded:
...
+		echo "Applying patch: $$patch_file"; \
+		cd $(PATCH_DIR) && git apply --3way --check "$$patch_file" || exit 1; \
+		cd $(PATCH_DIR) && git apply --3way "$$patch_file"; \
+	done
+
```
francoisferrand: change not needed

solution-base/mongodb/how_to_upgrade.md
```diff
@@ -0,0 +1,18 @@
...
+I will be taking as an example the upgrade from 5 to 6):
+
+## Upgrade steps
+
+
+1. Check for all available versions of the chart using `helm search repo bitnami/mongodb-sharded --versions` to check the configuration values for a specified version: `helm show values --version 6.6.7  bitnami/mongodb-sharded`
+2. Bump mongo version in `solution-base/deps.yaml` files.
+3. Bump mongodb sharded version in charts => `CHART_MONGO_SHARDED_VERSION` in the `solution-base/mongodb/Makefile` file (e.g. `CHART_MONGO_SHARDED_VERSION:="6.6.7"`)
+4. Upgrade charts to the version targetted => `make fetch-mongodb-sharded`
+5. Manually update the patches by applying the changes manually in the new upgraded charts (note that this can be applied only to patches that target mongo sharded) =>  `git diff -- solution-base/mongodb/charts/mongodb-sharded/values.yaml > solution-base/mongodb/patches/secret-name.patch` (this operation needs to be done for every patch).
```
francoisferrand: nit: the part about "mongo sharded" is not relevant anymore?

```suggestion
5. Manually update the patches by applying the changes manually in the new upgraded charts =>  `git diff -- solution-base/mongodb/charts/mongodb-sharded/values.yaml > solution-base/mongodb/patches/secret-name.patch` (this operation needs to be done for every patch).
```

Also please note that this must be done only when the patch does not apply : as long as git able to apply the patch (and it works at runtime obviously), there is no need to proactively edit/re-export every patch.

solution-base/mongodb/how_to_upgrade.md
```diff
@@ -0,0 +1,18 @@
...
+6. Once the patches are updated, apply them to the charts with the command `make patch`.
+7. After upgrading, you may need to apply additional changes that are not directly handled by the charts or patches. For example, when upgrading to MongoDB 6.0, you need to switch from using `mongo` to `mongosh`. Refer to the release notes for the new version (e.g., `https://www.mongodb.com/docs/v6.0/release-notes/6.0/`) for details on these additional changes.
+9. Passing CI tests is not enough, you must also perform an upgrade check with the product(s) using this Zenko version, before merging the upgrade PR. Some changes may be required at upper-levels (e.g., update mongosh commands, or add new logic in the upgrade).
```
francoisferrand:
```suggestion
8. Passing CI tests is not enough, you must also perform an upgrade check with the product(s) using this Zenko version, before merging the upgrade PR. Some changes may be required at upper-levels (e.g., update mongosh commands, or add new logic in the upgrade).
```

solution-base/mongodb/Makefile
benzekrimaha: actually without this , the patch files that are retrieved cannot be applied one by one resulting in an error ( I believe path is too long) when we try to apply the patches later on , this was also the main reason why I did the change to go for a loop

solution-base/mongodb/Makefile
benzekrimaha: https://github.com/scality/Zenko/pull/2028#discussion_r1774876326

solution-base/mongodb/Makefile
francoisferrand: that is weird...
what is the error?

- if the error is really the line of the path/command line, we can just remove the dir from the variable: `PATCH_FILES := $(shell cd ${PATCH_DIR}/ && ls -d * | tr '\n' ' ')` and change path before apply patches: `@git -C ${PATCH_DIR} apply --check ${PATCH_FILES}` and `@git -C ${PATCH_DIR} apply ${PATCH_FILES}`
- you also changing the patch commands to add the `--3way` param, could this have an effect?

francoisferrand (approved):

PR #2037, opened 2024-03-15
Retry to avoid flaky e2e-sharded test failure

francoisferrand:
- Retry "transition a 0 byte object test"
- gha: use ubuntu runner to build iso
- gha: remove job's `.final_status` uploads
- debug: add log & requestId
- gha: dump kafka topics in artifacts

This does not solve the underlying issue yet, but adds a workaround which will benefit all, with extra logs to make investigation easier and solving some debt in github action workflows. 

Issue: ZENKO-4776

### DIFF
--- a/.github/actions/archive-artifacts/action.yaml
+++ b/.github/actions/archive-artifacts/action.yaml
@@ -14,3 +14,29 @@ runs:
         kubectl get events -A -o yaml > /tmp/artifacts/data/${STAGE}/kind-logs/all-events.log;
         kind export logs /tmp/artifacts/data/${STAGE}/kind-logs/kind-export;
         tar zcvf /tmp/artifacts/${{ github.sha }}-${STAGE}-logs-volumes.tgz /tmp/artifacts/data/${STAGE}/kind-logs;
+    - name: Dump kafka
+      shell: bash
+      run: |-
+        set -exu;
+
+        NAMESPACE=${NAMESPACE:-default}
+        KAFKA=$(kubectl get pods -n ${NAMESPACE} -lkafka_cr=${ZENKO_NAME:-end2end}-base-queue -o jsonpath='{.items[0].metadata.name}');
+
+        kubectl exec -itn ${NAMESPACE} ${KAFKA} -c kafka -- \
+            env KAFKA_OPTS= kafka-topics.sh --bootstrap-server :9092 --list \
+            > /tmp/artifacts/data/${STAGE}/kafka-topics.log;
+        kubectl exec -itn ${NAMESPACE} ${KAFKA} -c kafka -- \
+            env KAFKA_OPTS= kafka-consumer-groups.sh --bootstrap-server :9092 --list \
+            | grep -v '^__' \
+            > /tmp/artifacts/data/${STAGE}/kafka-consumer-groups.log;
+        for topic in $(cat /tmp/artifacts/data/${STAGE}/kafka-topics.log); do
+            kubectl exec -itn ${NAMESPACE} ${KAFKA} -c kafka -- \
+                env KAFKA_OPTS= kafka-console-consumer.sh --bootstrap-server :9092 --topic $topic \
+                                                          --from-beginning --timeout-ms 1000 \
+                > /tmp/artifacts/data/${STAGE}/kafka-messages-$topic.log;
+        done;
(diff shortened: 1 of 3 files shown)


.github/workflows/end2end.yaml
```diff
@@ -140,16 +140,6 @@ jobs:
           mkdir -p /tmp/artifacts/data/${{ github.job }}/act-logs/
           mv tests/workflows/act-*.log /tmp/artifacts/data/check-workflows/act-logs/
           tar zcvf /tmp/artifacts/${{ github.sha }}-${{ github.job }}-act-logs.tgz /tmp/artifacts/data/${{ github.job }}/act-logs;
-      - name: Set status in ARTIFACTS to enable re-run only failed jobs
```
ghost, line 143: IIRC, in the release workflow, `scality/action-artifacts@v3`, we rely on this before creating the promotion of the artifacts. Will it work if we remove it from the build?

francoisferrand: The status is still written at the end of the build (either the `write-final-failed-status` or `write-final-success-status`), so we should be fine.

(this was discussed in https://scality.slack.com/archives/CFDJAPQ6S/p1709823455660169)

ghost (approved):

.github/actions/archive-artifacts/action.yaml
```diff
@@ -14,3 +14,29 @@ runs:
...
+            kubectl exec -itn ${NAMESPACE} ${KAFKA} -c kafka -- \
+                env KAFKA_OPTS= kafka-console-consumer.sh --bootstrap-server :9092 --topic $topic \
+                                                          --from-beginning --timeout-ms 1000 \
+                > /tmp/artifacts/data/${STAGE}/kafka-messages-$topic.log;
+        done;
+        for cg in $(cat /tmp/artifacts/data/${STAGE}/kafka-consumer-groups.log); do
+            kubectl exec -itn ${NAMESPACE} ${KAFKA} -c kafka  -- \
+                env KAFKA_OPTS= kafka-consumer-groups.sh --bootstrap-server :9092 --describe --group $cg \
+                > /tmp/artifacts/data/${STAGE}/kafka-offsets-$cg.log;
+        done;
```
KillianG, line 42: Shouldn't this be better in a seperate script file ? Would be easier to maintain and to read..

tests/zenko_tests/node_tests/backbeat/tests/lifecycle/transition.js
```diff
@@ -116,29 +117,39 @@ testsToRun.forEach(test => {
...
+
                 cloudServer.setKey(key);
                 cloud.setKey(`${srcBucket}/${key}`);
                 series([
                     next => cloudServer.putObject(Buffer.from(''), next),
                     next => checkTransition(toLoc, cloudServer, cloud, null, next),
                     next => checkRestoration(toLoc, cloudServer, null, next),
                 ], done);
-            });
+            }).retries(3);
```
KillianG, line 152: Do we know why this is flaky ? Isn't there a real fix that we could do ? If this is just a temp workaround, should we create a ticket to investigate the underlying issue ?

KillianG (approved):
2 small comments, otherwise LGTM

tests/zenko_tests/node_tests/backbeat/tests/lifecycle/transition.js
francoisferrand: not yet, as indicated in the PR comment:

> This does not solve the underlying issue yet, but adds a workaround which will benefit all, with extra logs to make investigation easier and solving some debt in github action workflows.

we have the ticket already, either it will not be closed by this PR or we will create a followup indeed.

.github/actions/archive-artifacts/action.yaml
francoisferrand: This is a stop-gap measure only. Eventually this (or whatever else we can/need to get from kafka) should be capitalized upon in a real script : this one is not to be reused, and moving to a separate script would lead to extending it, little by little.... until we have created an adhoc solution without thinking of what is needed.

Like we may want to put this in an SosReport plugin ; and/or maybe need to smarter, since we also now have kafka-cleaner which removes messages (which we may need, to troubleshoot "missed" first messages)

So keeping it here for now, to avoid introducing any bias; and we should re-assess if/when we we do followup updates.

PR #2040, opened 2024-03-28
Improvement/zenko 4737 add e2e test for dmf retries

KillianG:
Add test e2e test for dmf retry jobs and commands

Issue: ZENKO-4737

### DIFF
--- a/tests/ctst/common/common.ts
+++ b/tests/ctst/common/common.ts
@@ -20,7 +20,7 @@ function getObjectNameWithBackendFlakiness(this: Zenko, objectName: string) {
     const backendFlakynessRetryNumber = this.getSaved<string>('backendFlakynessRetryNumber');
     const backendFlakyness = this.getSaved<string>('backendFlakyness');
 
-    if (!backendFlakyness || !backendFlakynessRetryNumber) {
+    if (!backendFlakyness || !backendFlakynessRetryNumber || !objectName) {
         return objectName;
     }
 
@@ -203,8 +203,8 @@ When('i restore object {string} for {int} days', async function (this: Zenko, ob
 
 // wait for object to transition to a location or get restored from it
 Then('object {string} should be {string} and have the storage class {string}', { timeout: 130000 },
-    async function (this: Zenko, objectName: string, objectTransitionStatus: string, storageClass: string) {
-        const objName =
+    async function (this: Zenko, objectName: string, operation: string, storageClass: string) {
+    const objName =
             getObjectNameWithBackendFlakiness.call(this, objectName) || this.getSaved<string>('objectName');
         this.resetCommand();
         this.addCommandParameter({ bucket: this.getSaved<string>('bucketName') });
--- a/tests/ctst/common/common.ts
+++ b/tests/ctst/common/common.ts
@@ -10,10 +10,72 @@ import { createBucketWithConfiguration, putObject } from 'steps/utils/utils';
 
 setDefaultTimeout(Constants.DEFAULT_TIMEOUT);
 
+type retryDmf = {
+    commandRetryNumber?: number,
+    jobRetryNumber?: number,
+    jobRetryOP?: string,
+}
+
+type listingObject = {
+    Key: string,
+    VersionId: string,
+}
+
+type listingResult = {
+    Versions: listingObject[],
+    DeleteMarkers: listingObject[],
+}
+
+/**
+ * Cleans the created test bucket
+ * @param {Zenko} world world object
+ * @param {string} bucketName bucket name
+ * @returns {void}
+ */
+export async function cleanS3Bucket(
+    world: Zenko,
+    bucketName: string,
+): Promise<void> {
+    if (!bucketName) {
+        return;
+    }
+    world.resetCommand();
+    world.addCommandParameter({ bucket: bucketName });
+    const createdObjects = world.getSaved<Map<string, string>>('createdObjects');
+    if (createdObjects !== undefined) {
+        const results = await S3.listObjectVersions(world.getCommandParameters());
+        const res = safeJsonParse(results.stdout);
+        assert(res.ok);
+        const parsedResults = res.result as listingResult;
+        const versions = parsedResults.Versions || [];
+        const deleteMarkers = parsedResults.DeleteMarkers || [];
+        await Promise.all(versions.concat(deleteMarkers).map(obj => {
+            world.addCommandParameter({ key: obj.Key });
+            world.addCommandParameter({ versionId: obj.VersionId });
+            return S3.deleteObject(world.getCommandParameters());
+        }));
+        world.deleteKeyFromCommand('key');
+        world.deleteKeyFromCommand('versionId');
+    }
+    await S3.deleteBucketLifecycle(world.getCommandParameters());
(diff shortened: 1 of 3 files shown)


tests/ctst/features/dmf.feature
```diff
@@ -21,6 +21,36 @@ Feature: DMF
     |               Versioned |           2 |        100 |
     |               Suspended |           2 |        100 |
 
+    @2.7.0
```
ghost, line 23: Should be 2.8.0 if we introduce those tests for zenko 2.8

tests/ctst/common/common.ts
```diff
@@ -55,10 +55,41 @@ export async function cleanS3Bucket(
     await S3.deleteBucket(world.getCommandParameters());
 }
 
+/**
+ * @param {Zenko} this world object
+ * @param {string} objectName object name
+ * @returns {string} the object name based on the backend flakyness
+ */
+function getObjectNameWithBackendFlakyness(this: Zenko, objectName: string) {
```
ghost: nit: flakiness

tests/ctst/steps/dmf.ts
```diff
@@ -16,3 +17,12 @@ Then('dmf volume should contain {int} objects', async (objectCount: number) => {
     }
     assert(conditionOk);
 });
+
+Given('a flaky backend that will require {int} retries for {string}',
```
ghost, line 21: So we want to retry during the same test, and not rerun it fully, right?

I would say, no need to mention "flaky", as its misleading and we might later think of actual bugs as just flakiness

tests/ctst/common/common.ts
```diff
@@ -55,10 +55,41 @@ export async function cleanS3Bucket(
...
+    let objectNameFinal;
+    const backendFlakynessRetryNumber = this.getSaved<string>('backendFlakynessRetryNumber');
```
ghost, line 20: shouldn't it a number that we increment if we pass a retry number > 1?

tests/ctst/steps/dmf.ts
KillianG: This is to test a feature, which is the retry of a failed job in DMF
I think flaky can be the right word, actually this is not supposed to happen unless there is a problem on DMF side, other than flaky it could be buggy, out of service, busy etc.. what do you think ?

tests/ctst/common/common.ts
KillianG: I am not sure I understand why we should increment it ?

ghost: From what I understand you set `retryNumber` as a test parameter, which is 1 here, that you eventually use as the object name to work with. If you set a number > 1, we never retry twice, and the object name will have `-2` in its name... Maybe I misunderstood but to me what you are trying to implement is a way to retry a scenario step and the ability to track those objects from the logs, right?
That's why I think it should be incremented every time we retry, otherwise there is no need to have a test parameter if the value is static

tests/ctst/steps/dmf.ts
ghost: So by flaky you mean we timeout, right? How about 
```suggestion
Given('a backend that may require {int} retries for {string} due to timeouts',
```

or something like this

.github/scripts/end2end/run-e2e-ctst.sh
```diff
@@ -156,4 +156,4 @@ kubectl run $POD_NAME \
       }
     ]
   }
-}' -- ./run "$COMMAND" $WORLD_PARAMETERS --parallel $PARALLEL_RUNS --retry $RETRIES --retry-tag-filter @Flaky
+}' -- ./run "$COMMAND" $WORLD_PARAMETERS --tags @WIP --parallel $PARALLEL_RUNS --retry $RETRIES --retry-tag-filter @Flaky
```
welansari: left over from testing ?

tests/ctst/features/dmf.feature
```diff
@@ -21,6 +21,36 @@ Feature: DMF
...
+    @PreMerge
+    @Dmf
+    @ColdStorage
+    @WIP
```
welansari: same as before ?

tests/ctst/common/common.ts
```diff
@@ -237,7 +269,8 @@ When('i restore object {string} for {int} days', async function (this: Zenko, ob
 // wait for object to transition to a location or get restored from it
 Then('object {string} should be {string} and have the storage class {string}', { timeout: 130000 },
     async function (this: Zenko, objectName: string, objectTransitionStatus: string, storageClass: string) {
-        const objName = objectName ||  this.getSaved<string>('objectName');
+        const objName =
+            getObjectNameWithBackendFlakyness.call(this, objectName) || this.getSaved<string>('objectName');
```
welansari:
```suggestion
        const objName =
            getObjectNameWithBackendFlakyness.call(this, objectName);
```

the `||` is useless as `getObjectNameWithBackendFlakyness` always returns a value ?
Btw we don't handle the case where `objectName` is not specified in `getObjectNameWithBackendFlakyness`

tests/ctst/common/common.ts
```diff
@@ -279,7 +312,7 @@ Then('object {string} should be {string} and have the storage class {string}', {
     });
 
 When('i delete object {string}', async function (this: Zenko, objectName: string) {
-    const objName = objectName ||  this.getSaved<string>('objectName');
+    const objName = getObjectNameWithBackendFlakyness.call(this, objectName) ||  this.getSaved<string>('objectName');
```
welansari: same

tests/ctst/common/common.ts
```diff
@@ -222,7 +254,7 @@ Given('a transition workflow to {string} location', async function (this: Zenko,
 });
 
 When('i restore object {string} for {int} days', async function (this: Zenko, objectName: string, days: number) {
-    const objName = objectName ||  this.getSaved<string>('objectName');
+    const objName = getObjectNameWithBackendFlakyness.call(this, objectName) ||  this.getSaved<string>('objectName');
```
welansari: same

tests/ctst/common/common.ts
KillianG: It returns undefined in case we dont specify the object name, that's why there was ` this.getSaved<string>('objectName');` before my modification

.github/scripts/end2end/run-e2e-ctst.sh
KillianG: Needs to be remove but I knew it yeah I was just testing it

tests/ctst/common/common.ts
welansari: It returns undefined only when `backendFlakyness` and `backendFlakynessRetryNumber` are not defined right ?
Should we handle the other case ?

KillianG: We might want to handle the case where objectName is undefined not to make object name mandatory in the scenario
So i added a check for objectName in `getObjectNameWithBackendFlakyness` to return undefined if it is undefined

KillianG:
Due to a mistake in the name of the branch, switch to a new PR [here](https://github.com/scality/Zenko/pull/2041)... my apologies

PR #2053, opened 2024-04-05
ZENKO-4796: pull zenko UI image from ghcr.io

ChengYanJin:

fixes #<ISSUE>

### DIFF
--- a/.github/workflows/end2end.yaml
+++ b/.github/workflows/end2end.yaml
@@ -211,6 +211,12 @@ jobs:
           username: "${{ secrets.REGISTRY_LOGIN }}"
           password: "${{ secrets.REGISTRY_PASSWORD }}"
           registry: registry.scality.com
+      - name: Login to ghcr.io
+        uses: docker/login-action@v2
+        with:
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+          registry: ghcr.io
       - name: Login to Docker Hub
         uses: docker/login-action@v2
         with:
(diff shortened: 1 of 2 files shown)


JBWatenbergScality (approved):
LGTM !

francoisferrand (changes requested):
We are already doing the change on Zenko, accross all components : will actually be done there, as part https://scality.atlassian.net/browse/PTFE-1376 (esp. it needs to be done accross all branches)

On the other hand,
* all tags must be migrated from registry.scality.com to ghcr (allow building older versions), as documented in platform engineering's blog
* you are bumping to newer release, should be a separate & explicit bump [ghcr migration should be NOOP]

ChengYanJin:
> We are already doing the change on Zenko, accross all components : will actually be done there, as part https://scality.atlassian.net/browse/PTFE-1376 (esp. it needs to be done accross all branches)
> 
> On the other hand,
> 
> * all tags must be migrated from registry.scality.com to ghcr (allow building older versions), as documented in platform engineering's blog
> * you are bumping to newer release, should be a separate & explicit bump [ghcr migration should be NOOP]

Hello @francoisferrand, it should be good now. We migrated all the existing tags to ghcr and separated the bump zenko-ui to a dedicate commit ;)

francoisferrand (changes requested):
this ticket is about migration, which is handled separately (in Zenko4800)
→ please create a separate ticket & PR (possibly after #2061 is merged) to track the bump, if you need it

(Jira is not just to have a number and please bert-e, it is really to document the work we need to do: you don't need the registry migration to bump to 2.2.9, but you problably have many other reasons: like to deliver fixes already merged in Zenko-UI and which "block" product issues)

ChengYanJin:
close because duplicate with Zenko4800

PR #2058, opened 2024-04-12
Run zenko and ctst locally

KillianG:
ZENKO-4799


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,126 @@
+{
+    "name": "Codespace",
+    "build": {
+        "context": ".",
+        "dockerfile": "Dockerfile"
+    },
```
tcarmet, line 6:
```suggestion
    "build": {
        "context": ".",
        "dockerfile": "Dockerfile"
    },
    "prebuildHashPaths": [
		"Dockerfile"
	],
```

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,126 @@
...
+        "TIME_PROGRESSION_FACTOR": "1",
+        "ZENKO_MONGODB_DATABASE": "zenko-database",
+        "ZENKO_MONGODB_SHARDED": "true" 
+        },
+
+    // Use 'forwardPorts' to make a list of ports inside the container available locally.
+    // "forwardPorts": [],
+
+    // Use 'postCreateCommand' to run commands after the container is created.
+    "postCreateCommand": "bash .devcontainer/setup.sh"
```
tcarmet, line 81: While this may become useless in the near future, let's keep it while we are doing the `docker login` in the `setup.sh`

```suggestion
    "secrets": {
        "REGISTRY_LOGIN": {
            "description": "Login registry."
        },
        "REGISTRY_PASSWORD": {
            "description": "Password registry."
        }
    },
    // Use 'postCreateCommand' to run commands after the container is created.
    "postCreateCommand": "bash .devcontainer/setup.sh"
```

.github/scripts/end2end/configure-e2e.sh
```diff
@@ -54,7 +54,7 @@ kubectl run ${POD_NAME} \
   --restart=Never \
   --pod-running-timeout=5m \
   --namespace=${NAMESPACE} \
-  --image-pull-policy=Always \
+  --image-pull-policy=IfNotPresent \
```
tcarmet, line 78: good call

tcarmet (commented):
Added some suggestions that we previously discussed, I may be flying over and add more as we evolve.

.devcontainer/setup.sh
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash
+
+env_variables=$(yq eval '.env | to_entries | .[] | .key + "=" + .value' .github/workflows/end2end.yaml | sed 's/\${{[^}]*}}//g') && export $env_variables
```
tcarmet, line 3: nice trick

.devcontainer/setup.sh
```diff
@@ -0,0 +1,56 @@
...
+    kind load docker-image  ${E2E_IMAGE_NAME}:${E2E_IMAGE_TAG}
+    docker rmi ${E2E_IMAGE_NAME}:${E2E_IMAGE_TAG}
+
+    bash configure-e2e-ctst.sh
+)
+
+docker image prune -af
+
+CTST_TAG=$(sed 's/.*"cli-testing": ".*#\(.*\)".*/\1/;t;d' ./tests/ctst/package.json)
+SORBET_TAG=$(yq eval '.sorbet.tag' solution/deps.yaml)
```
tcarmet, line 52: I don't know if you want to export those, leaving it here just in case:

```suggestion
export CTST_TAG=$(sed 's/.*"cli-testing": ".*#\(.*\)".*/\1/;t;d' ./tests/ctst/package.json)
export SORBET_TAG=$(yq eval '.sorbet.tag' solution/deps.yaml)
```

tcarmet (approved):
👌

.devcontainer/devcontainer.json
tcarmet: let's ignore this one for now.

.devcontainer/setup.sh
KillianG: I don't think I need it, better not export it if we don't need no ?

tcarmet: yes, better leave it off.

.devcontainer/setup.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash
+
+env_variables=$(yq eval '.env | to_entries | .[] | .key + "=" + .value' .github/workflows/end2end.yaml | sed 's/\${{[^}]*}}//g') && export $env_variables
+export GIT_ACCESS_TOKEN=${GITHUB_TOKEN}
```
welansari, line 4: This doesn't export `ZENKO_MONGODB_SHARDED` which is why mongo was not deployed in sharded mode the other day.
Some env vars like `ZENKO_MONGODB_SHARDED` and others are not defined in the env section, maybe we should add them.

welansari: Oh never mind you manually added it to `containerEnv`

welansari (approved):

anurag4DSB (commented):
Tried this out in codespaces. The issue is discussed and showed to Killian already. Somehow my registry env is not getting set.

will review again once resolved.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,60 @@
...
+        }
+    },
+    // Use 'postCreateCommand' to run commands after the container is created.
+    "postCreateCommand": "bash .devcontainer/setup.sh"
+    // "postStartCommand": "bash .devcontainer/startup.sh"
+    // Configure tool-specific properties.
+    // "customizations": {},
+
+    // Uncomment to connect as root instead. More info: https://aka.ms/dev-containers-non-root.
+    // "remoteUser": "root"
```
francoisferrand: should be removed

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,60 @@
...
+            }
+        }
+    },
+    "containerEnv": {
+        "ZENKO_MONGODB_DATABASE": "zenko-database",
+        "ZENKO_MONGODB_SHARDED": "true" 
+        },
+
+    // Use 'forwardPorts' to make a list of ports inside the container available locally.
+    // "forwardPorts": [],
```
francoisferrand: should be removed

.devcontainer/setup.sh
```diff
@@ -0,0 +1,55 @@
...
+    docker rmi ${E2E_IMAGE_NAME}:${E2E_IMAGE_TAG}
+)
+
+(
+    cd .github/scripts/end2end
+
+    bash configure-e2e-ctst.sh
+)
+
+docker image prune -af
```
francoisferrand, line 49: why do we prune images?

.devcontainer/setup.sh
```diff
@@ -0,0 +1,55 @@
...
+(
+    cd tests/zenko_tests
+
+    envsubst < 'e2e-config.yaml.template' > 'e2e-config.yaml'
+    if [[ "${ENABLE_RING_TESTS}" == "false" ]]; then
+    yq -i 'del(.locations[] | select(.locationType == "location-scality-ring-s3-v1"))' e2e-config.yaml
+    fi
+    docker build -t $E2E_IMAGE_NAME:$E2E_IMAGE_TAG .
+    kind load docker-image  ${E2E_IMAGE_NAME}:${E2E_IMAGE_TAG}
+    docker rmi ${E2E_IMAGE_NAME}:${E2E_IMAGE_TAG}
```
francoisferrand, line 40: why remove the image?

.devcontainer/setup.sh
```diff
@@ -0,0 +1,55 @@
...
+    done
+
+(
+    cd tests/zenko_tests
+
+    envsubst < 'e2e-config.yaml.template' > 'e2e-config.yaml'
+    if [[ "${ENABLE_RING_TESTS}" == "false" ]]; then
+    yq -i 'del(.locations[] | select(.locationType == "location-scality-ring-s3-v1"))' e2e-config.yaml
+    fi
+    docker build -t $E2E_IMAGE_NAME:$E2E_IMAGE_TAG .
```
francoisferrand, line 38: we may want to add `--cache-from` params, to speed up the build:

```suggestion
    docker build --cache-from $E2E_IMAGE_NAME:$E2E_IMAGE_TAG --cache-from cache-from: type=gha,scope=end2end-test -t $E2E_IMAGE_NAME:$E2E_IMAGE_TAG .
```

(try to read cache both from gha, same scope as in workflow, and from the image which may have been built already...)

francoisferrand (commented):
LGTM with a few comments, but I think this should probably target 2.6...

.devcontainer/setup.sh
KillianG: We prune images because 8 core codespaces are limited to 64GB, which is just just just enough, if we keep all images we end up with warnings in vscode etc... so I prefer remove images that we don't use as they are already in kind

.devcontainer/setup.sh
KillianG: same as before

.devcontainer/setup.sh
tcarmet: I don't think the codespace token will have access to use the github actions cache.

anurag4DSB (approved):
I was able to spin up a cluster, great work :muscle: 
would be worth adding a readme or confluence page for guidance.

anurag4DSB (approved):
Thanks for document

.devcontainer/setup.sh
francoisferrand: @KillianG could you try this?

KillianG: This unfortuneatly does not work :(

PR #2060, opened 2024-04-19
ZENKO-4804 upload test result to trunk

tcarmet:


.github/workflows/end2end.yaml
```diff
@@ -445,6 +447,16 @@ jobs:
       - name: Debug wait
         uses: ./.github/actions/debug-wait
         timeout-minutes: 60
+      - name: Upload results
+        if: "!cancelled() && env.TRUNK_TOKEN"
+        uses: trunk-io/analytics-uploader@main
+        with:
+          junit-paths: /artifacts/data/artifacts/*.xml
+          org-slug: ${{ github.repository_owner }}
+          token: ${{ env.TRUNK_TOKEN }}
```
francoisferrand, line 433: don't we need to set `repo-head-branch` ?

tcarmet: Only if we want to override the repository head, and as we don't know yet how everything will work exactly, I think it's best we leave as is.

KillianG (approved):

KillianG:
Can we see how it looks on Trunk.io ?

.github/workflows/end2end.yaml
```diff
@@ -420,6 +420,8 @@ jobs:
         uses: ./.github/actions/install-end2end-dependencies
       - name: Wait for Docker daemon to be ready
         uses: ./.github/actions/wait-docker-ready
+      - name: Kubectl tool installer
+        uses: Azure/setup-kubectl@v3
```
francoisferrand, line 424: kubectl is already installed at the end of `install-end2end-dependencies` :

```
    - name: Kubectl tool installer
      uses: Azure/setup-kubectl@v4
      with:
        version: 'v${{ env.KUBECTL_VERSION }}'
```

francoisferrand: kubectl is already installed at the end of `install-end2end-dependencies` :

```
    - name: Kubectl tool installer
      uses: Azure/setup-kubectl@v4
      with:
        version: 'v${{ env.KUBECTL_VERSION }}'
```

```suggestion
```

.github/workflows/end2end.yaml
```diff
@@ -481,6 +493,8 @@ jobs:
...
```
francoisferrand: not be needed either

```suggestion
```

.github/scripts/end2end/run-e2e-test.sh
```diff
@@ -120,7 +120,35 @@ run_e2e_test() {
...
+            "mountPath": "/artifacts"
+          }
+        ]
+      }
+    ],
+    "volumes": [
+      {
+        "name": "artifacts",
+        "hostPath": {
+          "path": "/data/artifacts",
```
francoisferrand: `/data` works in CI, but most likely not if the test were to run locally: should instead rely on $(pwd) or something similar (which may possibly be overridden...)

```suggestion
          "path": "${ZENKO_E2E_ARTIFACTS:-$(pwd)}/artifacts",
```

.github/workflows/end2end.yaml
```diff
@@ -445,6 +447,16 @@ jobs:
       - name: Debug wait
         uses: ./.github/actions/debug-wait
         timeout-minutes: 60
+      - name: Upload results
+        if: "!cancelled() && env.TRUNK_TOKEN"
```
francoisferrand, line 428: should we `[continue-on-error](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error)`, to avoid any unexpected failures ?

```suggestion
        if: "!cancelled() && env.TRUNK_TOKEN"
        continue-on-error: true
```

Then ideally we need to have something to track when this is broken... but maybe it is safer to keep it possibly failing for now, as it is not even in beta yet?

.github/workflows/end2end.yaml
```diff
@@ -533,6 +556,8 @@ jobs:
         uses: ./.github/actions/install-end2end-dependencies
       - name: Wait for Docker daemon to be ready
         uses: ./.github/actions/wait-docker-ready
+      - name: Kubectl tool installer
+        uses: Azure/setup-kubectl@v3
```
francoisferrand: should not be needed either
```suggestion
```

tests/zenko_tests/node_tests/config.json
```diff
@@ -1,6 +1,6 @@
 {
     "reporterEnabled": "spec, mocha-junit-reporter",
     "mochaJunitReporterReporterOptions": {
-        "mochaFile": "junit-node-[hash].xml"
+        "mochaFile": "/artifacts/junit-node-[hash].xml"
```
francoisferrand: should be configurable, these tests are also run manually/locally....

.github/workflows/end2end.yaml
tcarmet: I think the point is you just don't want your pipeline to fail if trunk happens to be facing a brief service interruption.

If you want to track when it's broken, you can [track each step failure rate](https://app.trunk.io/scality/scality/Zenko/ci-analytics/1_01db756e-9927-4cfb-b98d-dc56e2952ed2/1_dc061645-8e2f-4d67-9087-7630167d9ee3) with trunk

tests/zenko_tests/node_tests/config.json
tcarmet: If they are ran, because of the way the tests works, by creating a pod inside the cluster and running the tests from within I assume they are run with the Dockerfile that is available, in which the volume `/artifacts/` has been created.

In this particular spot, I see no need of making this configurable and I don't even think those config would support it.

.github/scripts/end2end/run-e2e-test.sh
tcarmet: Ok but I cannot use this var, as we need to be consistent with [the one defined inside `bootstrap-kind.sh`](https://github.com/scality/Zenko/blob/feature/ZENKO-4804-upload-test-results-trunk/.github/scripts/end2end/bootstrap-kind.sh#L6). I also can't use `$pwd` as a path prefix.

tcarmet: Actually I had to revert that change because `/data/artifacts` is a reference to a already hardcoded variable, we are referring the host path, in this case it is the container kind, and when bootstrapping kind we refer the host path there as well, which is the machine that will execute the tests, CI or local. So in the container context, `/data/` is bind to where you will specify `VOLUME_ROOT` when bootstrapping kind, which is wherever the user decides to do so.

Too many layers 😅

francoisferrand: 🤯

.github/workflows/end2end.yaml
francoisferrand: that is nice! but it cannot track failures of trunk upload step... (inception)

.github/workflows/end2end.yaml
francoisferrand: yes, as discussed with the guys, best to leave it unless we are in a situation where we see it is not computed properly

.github/workflows/end2end.yaml
```diff
@@ -445,6 +445,16 @@ jobs:
       - name: Debug wait
         uses: ./.github/actions/debug-wait
         timeout-minutes: 60
+      - name: Upload results
+        if: "!cancelled() && env.TRUNK_TOKEN"
+        uses: trunk-io/analytics-uploader@main
```
francoisferrand, line 429: should not be quoted:

```suggestion
        if: !cancelled() && env.TRUNK_TOKEN
        uses: trunk-io/analytics-uploader@main
```

francoisferrand: (same for each)

tests/zenko_tests/Dockerfile
```diff
@@ -1,5 +1,7 @@
 FROM ubuntu:18.04
 
+VOLUME /artifacts
```
francoisferrand: i am not sure we _want_ to create a volume here: which would create lead to automatically creating "anonymous" volume if run without binding/... We can create a volume bind even if the image does not have such a volume, so IMHO we should just leave the image as before...

tests/ctst/Dockerfile
```diff
@@ -3,6 +3,8 @@ ARG CTST_TAG
 FROM registry.scality.com/sorbet/sorbet:$SORBET_TAG as sorbet
 FROM registry.scality.com/cli-testing/cli-testing:$CTST_TAG
 
+VOLUME /artifacts
```
francoisferrand: i am not sure we _want_ to create a volume here: which would create lead to automatically creating "anonymous" volume if run without binding/... We can create a volume bind even if the image does not have such a volume, so IMHO we should just leave the image as ebfore (esp. since the `/artifacts` path is a parameter we specify when running the tests)

tests/zenko_tests/node_tests/config.json
francoisferrand: The main issue we face for running the tests "locally" is zenko install, setup of locations/buckets/..., and all the required env variables: the test themself should be able to run outside of cluster. This should be addressed with the .devcontainer / codespace.

So we must not add to the debt by requiring the test to run in a pod or container...

francoisferrand: the patht to the junit report can be overriden by specifying another path in [the `MOCHA_FILE` env variable](https://github.com/michaelleeallen/mocha-junit-reporter/blob/master/README.md#usage)

.github/workflows/end2end.yaml
tcarmet: This will trigger a syntax error:
Screenshot 2024-04-23 at 1 39 44 PM

.github/scripts/end2end/run-e2e-test.sh
```diff
@@ -120,7 +121,36 @@ run_e2e_test() {
...
+  "apiVersion": "v1",
+  "kind": "Pod",
+  "spec": {
+    "containers": [
+      {
+        "name": "'$POD_NAME'",
+        "volumeMounts": [
+          {
+            "name": "artifacts",
+            "mountPath": "/artifacts"
```
francoisferrand: for readability, it may be better to use a different name for this "inner" path : maybe `reports`, so that we can easily see the different between the host `/artifacts` and the inner mount point `/reports`

→ in the runner, the reports would thus be in `/artifacts/data/reports`

francoisferrand (commented):
Looks good, but can you please rebase this PR on 2.6 branch, so we get more information across the branches?

francoisferrand (approved):

PR #2066, opened 2024-05-07
Scuba and Quotas

ghost:
- Bump project version
- Introduce Scuba in Zenko
- CTST tests for quotas

See [2.8+ integration branches](https://github.com/scality/Zenko/compare/w/2.8/feature/ZENKO-4789-scuba-and-quotas?expand=1) for added tests on Restores.


.github/scripts/end2end/run-e2e-ctst.sh
```diff
@@ -61,6 +61,11 @@ KAFKA_CLEANER_INTERVAL=$(kubectl get zenko ${ZENKO_NAME} -o jsonpath='{.spec.kaf
...
 
+# Set up environment variables for testing
+kubectl set env deployment end2end-connector-cloudserver SCUBA_HEALTHCHECK_FREQUENCY=100
+kubectl rollout status deployment end2end-connector-cloudserver
+kubectl patch cronjob end2end-ops-count-items -p '{"spec":{"suspend":true}}'
```
ghost: Note for reviewers:

- Setting the healthcheck frequency to something fast will ensure the tests can be executed without issue; as scuba is disabled till count items ran at least once (if no bucket, no account in the metrics, no quota checks), and periodic checks are every 60s, but tests may start directly
- The count items might affect the existing tests, as it runs once per hour, but the tests are expecting "realtime" metrics (obtained today with the inflights).

solution/deps.yaml
```diff
@@ -80,8 +80,14 @@ s3utils:
...
   dashboard: s3utils/s3utils-dashboards
   image: s3utils
-  tag: 1.14.5
+  tag: 1.14.6
   envsubst: S3UTILS_TAG
+scuba:
+  sourceRegistry: ghcr.io/scality
+  dashboard: scuba/scuba-dashboards
+  image: scuba
+  tag: 1.0.0-preview.2
```
ghost: Note for reviewers: do we want a 1.0.0 tag? Not sure if it makes sense for a Scuba point of view, as it's still under dev...

.github/scripts/end2end/configs/zenkoversion.yaml
```diff
@@ -111,6 +111,9 @@ spec:
     vault:
       image: '${VAULT_IMAGE}'
       tag: '${VAULT_TAG}'
+    scuba:
+      image: '${SCUBA_IMAGE}'
+      tag: '${SCUBA_TAG}'
```
francoisferrand, line 119: should also add matching dashboards image

.github/scripts/end2end/run-e2e-ctst.sh
```diff
@@ -61,6 +61,11 @@ KAFKA_CLEANER_INTERVAL=$(kubectl get zenko ${ZENKO_NAME} -o jsonpath='{.spec.kaf
...
 
+# Set up environment variables for testing
+kubectl set env deployment end2end-connector-cloudserver SCUBA_HEALTHCHECK_FREQUENCY=100
```
francoisferrand, line 65: can we do this instead via zenko CR ?
we have tests which add locations, and will thus re-deploy cloudserver...

.github/scripts/end2end/run-e2e-ctst.sh
francoisferrand: what is this "suspend" flag?
* will this be stable (vs adding locations...) ?
* if the cronjob is suspended, how will be get the initial metrics?

Can't we keep the cronjob (mostly useless, probably), and just run an extra job in parallel if needeD?

.github/workflows/end2end.yaml
```diff
@@ -562,6 +402,12 @@ jobs:
           username: "${{ github.repository_owner }}"
           password: "${{ github.token }}"
           registry: ghcr.io
+      - name: Login to Registry (scality, testing)
+        uses: docker/login-action@v3
+        with:
+          username: "${{ secrets.REGISTRY_LOGIN }}"
+          password: "${{ secrets.REGISTRY_PASSWORD }}"
+          registry: registry.scality.com
```
francoisferrand: must not be added → left-over to be removed

.github/workflows/end2end.yaml
```diff
@@ -383,166 +383,6 @@ jobs:
           cache-from: type=gha,scope=end2end-ctst
           cache-to: type=gha,mode=max,scope=end2end-ctst
```
francoisferrand: why are you removing the e2e (non ctst) tests?
these tests are not currently covered by CTST, must be preserved

.github/scripts/end2end/run-e2e-ctst.sh
ghost: We do not have this option in the CR, as this was the decision from the review

solution/deps.yaml
francoisferrand: I would rather not have a preview release in production...
if we don't want to make a 1.0 release, I guess a `0.x.y` release or something like this may be acceptable, should be discussed with Scuba team as well

.github/workflows/end2end.yaml
ghost: The PR is not finalized, I want to pass the CIs first with the latest images, and get reviews, then I will remove the left overs to reduce number of CI builds, so yes, they won't be removed

francoisferrand: you should re-enable them already, now that your tests are somewhat finalized : so we may detect already if there are some unexpected impact on backbeat...

ghost: (I am looking for reviews on the tests first)

.github/scripts/end2end/configs/zenkoversion.yaml
```diff
@@ -111,6 +111,9 @@ spec:
     vault:
       image: '${VAULT_IMAGE}'
       tag: '${VAULT_TAG}'
+    scuba:
```
francoisferrand, line 117: also need to update the "real" `zenkoversion.yaml`

tests/ctst/common/common.ts
```diff
@@ -85,3 +90,91 @@ Given('an object {string} that {string}', async function (this: Zenko, objectNam
...
+            }
+            await Utils.sleep(delay);
+        }
+    });
+
+Then('the API should {string} with {string}', function (this: Zenko, result: string, expected: string) {
+    this.cleanupEntity();
+    const action = this.getSaved<ActionPermissionsType>('currentAction');
+    switch (result) {
+    case 'success':
```
KillianG: Should probably be `succeed` as the full sentence would be `The API should succeed...`

tests/ctst/common/common.ts
```diff
@@ -85,3 +90,91 @@ Given('an object {string} that {string}', async function (this: Zenko, objectNam
...
+                    this.getResult().stdout?.includes(action.expectedResultOnAllowTest) ||
+                    this.getResult().err === null, true);
+        } else {
+            assert.strictEqual(!!this.getResult().err, false);
+        }
+        break;
+    case 'fail':
+        assert.strictEqual(this.getResult().err?.includes(expected), true);
+        break;
+    }
```
KillianG, line 191: Should we not add a default and fail directly if yhe result is not succeed or fail ? to avoid any misspelling problem..

tests/ctst/common/common.ts
```diff
@@ -85,3 +90,91 @@ Given('an object {string} that {string}', async function (this: Zenko, objectNam
...
+        if (objectName) {
+            this.addToSaved('objectName', objectName);
+        } else {
+            this.addToSaved('objectName', `object-${Utils.randomString()}`);
+        }
+        await putObject(this, this.getSaved<string>('objectName'));
+    }
+});
+
+When('I PUT an object with size {int}', async function (this: Zenko, size: number) {
```
KillianG, line 215:
```suggestion
When('i PUT an object with size {int}', async function (this: Zenko, size: number) {
```

KillianG: That said, we already have this function in 2.7: https://github.com/scality/Zenko/blob/310e350703b91e07a2bda01ce76aeb7c4269377a/tests/ctst/common/common.ts#L103
We might want to backport it to avoid duplication after waterfall ?

tests/ctst/features/quotas/CountItems.feature
```diff
@@ -0,0 +1,12 @@
+Feature: CountItems measures the utilization metrics
+  The utilization metrics are computed for accounts, buckets and locations
+
+@2.6.0
+@PreMerge
+@CronJob
+@CountItems
+Scenario Outline: Countitems runs without error and compute utilization metrics
+    Given an existing bucket "" "without" versioning, "without" ObjectLock "" retention mode
```
KillianG: I am not sure about that `"" retention` what is set for retention mode here ?

tests/ctst/features/quotas/CountItems.feature
```diff
@@ -0,0 +1,12 @@
...
+    And an object "" that "exists"
+    When the CountItems cronjobs completes without error
```
KillianG: can "CountItems" be parameterized ? to be reused for another job if we want ?

tests/ctst/features/quotas/Quotas.feature
```diff
@@ -0,0 +1,96 @@
...
+    This feature ensures that quotas are correctly set and honored
+    for different APIs.
+
+    @2.6.0
+    @PreMerge
+    @Quotas
+    @CronJob
+    @DataWrite
+    Scenario Outline: Quotas are evaluated during write operations
+        Given an action "<action>"
```
KillianG, line 11: We usually don't change indentation after `Scenario Outline` either we do it everywhere, either nowhere.. What do you think ?

tests/ctst/features/quotas/Quotas.feature
```diff
@@ -0,0 +1,96 @@
...
+        And an upload size of <uploadSize> B for the object ""
+        And a STORAGE_MANAGER type
```
KillianG, line 13: This sentence seems unclear, we might want to make more explanative on what is happening here

tests/ctst/steps/quotas/quotas.ts
```diff
@@ -0,0 +1,150 @@
...
+
+function hashStringAndKeepFirst20Characters(input: string) {
+    return createHash('sha256').update(input).digest('hex').slice(0, 20);
+}
+
+/**
+ * The objective of this hook is to prepare all the buckets and accounts
+ * we use during quota checks, so that we avoid running the job multiple
+ * times, which affects the performance of the tests.
+ * The steps arer: create an account, then create a simple bucket
```
KillianG:
```suggestion
 * The steps are: create an account, then create a simple bucket
```

.github/scripts/end2end/run-e2e-ctst.sh
ghost: Suspended avoids the cronjob from running, unless we trigger it from the tests

The risk of keeping it is flakiness if it runs exactly when we execute some quota tests... In this case, it can change the metrics and erase some inflights, and this might cause some tests to fail.
The risk is quite low, so we could choose to keep the cronjob, I will put it back and we'll disable it later if it causes issues

solution/deps.yaml
ghost: Asked Scuba team, I would rather prefer having a 1.0.0 as the project had this package version

tests/ctst/steps/quotas/quotas.ts
```diff
@@ -0,0 +1,150 @@
...
+                    output[scenarioWithExampleID] = {
+                        AccessKey: CacheHelper.parameters?.AccessKey || Constants.DEFAULT_ACCESS_KEY,
+                        SecretKey: CacheHelper.parameters?.SecretKey || Constants.DEFAULT_SECRET_KEY,
+                    };
+                }
+            }
+        }
+    
+        await createJobAndWaitForCompletion(world, 'end2end-ops-count-items', 'quotas-setup');
+        await Utils.sleep(2000);
```
KillianG, line 70: Why are we waiting 2 seconds here ? Might be worth a comment to explain if really needed, and is it parameterizable / calculable instead of hardcoded

tests/ctst/steps/quotas/quotas.ts
```diff
@@ -0,0 +1,150 @@
...
+    world.resetGlobalType();
+    // Save the bucket name for the scenario
+    world.addToSaved('bucketName', key);
+    // Save the account name for the scenario
+    Zenko.saveAccountAccessKeys(config.AccessKey, config.SecretKey);
+    world.addToSaved('accountName', key);
+});
+
+Given('a bucket quota set to {int} B', async function (this: Zenko, quota: number) {
+    if (quota === 0) {
```
KillianG, line 102:
```suggestion
    if (quota <= 0) {
```

KillianG: Or can you set "uint" in the sentence definition ?

tests/ctst/common/common.ts
ghost: Any reason to have a lowercase `i`?

ghost: And for the other one, I will try, but the function is not standard with the internal putObject we have, that already implements file creation, and we'll need to add code related to flakiness

tests/ctst/features/quotas/CountItems.feature
ghost: When there is no value, the string is empty, so the bucket has no retention mode
We can set without instead, will do the same

tests/ctst/features/quotas/CountItems.feature
ghost: If we have only 1 cronjob in the step, it's the same as having one implementation for each (and eventually call a generic function), but let's parameterize it

tests/ctst/features/quotas/Quotas.feature
ghost: We change the indentation everywhere, but sometimes it's only 2 spaces
All bucket policies, IAM policies, cloudserverAuth follow this (standard) format

tests/ctst/features/quotas/Quotas.feature
ghost: I will not change it, it's used in depth in almost every test suite, not the objective of this PR to change all the tests

ghost: But in short: it runs the test as a specific identity type, which is the standard way of using Zenko 2/Artesca

tests/ctst/steps/quotas/quotas.ts
ghost: I don't want to add this condition, as <= should throw an error from the API: this is something we might want to test someday. 0 is for no quota, which is exactly what this step shouldn't do

There is also no uint support

tests/ctst/common/common.ts
ghost: I chose to keep the sentence, but call `addMultipleObjects` inside: the step must be a `When` as we retrieve the result, and expect things on it

tests/ctst/steps/quotas/quotas.ts
ghost: Added a comment

tests/ctst/common/common.ts
KillianG: > Any reason to have a lowercase `i`?

Usually we have the first letter in lowercase because the sentence starts with When or And etc... but you are right, the I should be in uppercase mb...

Would you be able to modify the functions that starts with a lowercase one when integrating to other branches ? I see some of them like `When i restore..`, `When i run sorbetctl` but there might be some others

tests/ctst/steps/quotas/quotas.ts
KillianG: >  this is something we might want to test someday. 

I feel like this is the same for 0 quotas in that case, we might want to test that, setting 0 quotas actually removes the quotas limit, what do you think ?

tests/ctst/features/quotas/CountItems.feature
KillianG: `without` seems better, as it is used for versioning and ObjectLock..

tests/ctst/common/common.ts
ghost: I will try to do it yes

tests/ctst/features/quotas/CountItems.feature
ghost: Yes, added it!

tests/ctst/steps/quotas/quotas.ts
ghost: Oh, no, because we have dedicated functions for that, as we have separate permissions: deleteAccountQuota (or bucket), 0 is not a valid value for the update API. We could decide to call the delete dynamically if needed, but for now it's not used...

tests/ctst/features/quotas/CountItems.feature
ghost: It's done

tests/ctst/common/common.ts
ghost: (I won't guarantee the above sentence, to save time though :laughing: )

tests/ctst/common/common.ts
ghost: Yes good catch, tests were not properly checking results in the happy scenarios, now it passes (got some issues for the deletes, because of bad values in the feature file, let's see the results now)

tests/ctst/common/common.ts
ghost: Done

.github/workflows/end2end.yaml
ghost: Done

solution/deps.yaml
ghost: We are aligned, I created a `1.0.0` release (now used in the PR)

tests/ctst/steps/quotas/quotas.ts
KillianG: OK 👌

KillianG (approved):

francoisferrand (approved):
LGTM

PR #2078, opened 2024-05-28
Compute prev tag to generate release notes

francoisferrand:
Issue: ZENKO-4795

### DIFF
--- a/.github/scripts/get-previous-tag.js
+++ b/.github/scripts/get-previous-tag.js
@@ -1,0 +1,22 @@
+
+function getPreviousTag(version, allReleases) {
+    while (version) {
+        // Strip trailing number from release to build the "prefix" for the previous version
+        prefix = version.replace(/\.\d+(-\w+)?$/, "");
+        if (prefix === version) {
+            return undefined;
+        }
+
+        version = prefix;
+
+        // Find the latest release with the same prefix
+        previous_tags = allReleases.filter(r => r.tag_name.startsWith(version));
+        if (previous_tags.length > 0) {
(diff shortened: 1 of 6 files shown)


.github/scripts/get-previous-tag.js
```diff
@@ -0,0 +1,22 @@
+
+function getPreviousTag(version, allReleases) {
```
ghost, line 2: (picking random line)
Is this issue related to what I got on another repo (cloudserver), where the action would generate the release notes from the last released tag **of the release tagged as "latest**? Typically, if we release last the 2.6.X, then a 2.9.X, the latter will have all the changelogs since the 2.6.X if it's marked as the "latest".

So my question is: if this is indeed the same problem, it affects all repos, and eventually we can have the same error again, where the release fails as the release notes are too long (e.g., a 7.10.X release in cloudserver, and then a 8.8.X).

ghost: In such case, shouldn't we use a scality action based on the current one, so it benefits all repos?

francoisferrand: yes, it could affect all repos, though not sure as it does not happen all the time or on every repo: and (even on repo) it sometimes works as expected.... maybe a combination of where the release is made, what the default branch is and on which branch the "latest" release is (like having the latest release on the default branch or not).

let's proceed incrementally, take the fix here first, and see how we generalize (we may want to generalize more than just the release notes) as it is a separate increment.

ghost (approved):

welansari (approved):

PR #2079, opened 2024-05-28
Flaky transition test fix

KillianG:


welansari (approved):

.github/scripts/end2end/configure-e2e-ctst.sh
```diff
@@ -22,8 +22,13 @@ kubectl wait --for condition=DeploymentFailure=false --timeout 10m zenko/${ZENKO
...
 # Patch sorbet env
-SORBET_DEPLOYMENT_NAME="${ZENKO_NAME}-cold-sorbet-azure-e2e-azure-archive"
-kubectl set env deployment/${SORBET_DEPLOYMENT_NAME} SORBETD_AZURE_RESTORETIMEOUT=3s
+kubectl patch zenko $ZENKO_NAME --type merge -p '{"spec":{"sorbet":{"configurationOverrides":{"e2e-azure-archive":{"pending-job-poll-after-age":"10s"}}}}}'
+kubectl patch zenko $ZENKO_NAME --type merge -p '{"spec":{"sorbet":{"configurationOverrides":{"e2e-azure-archive":{"pending-job-poll-check-interval":"10s"}}}}}'
+
+kubectl patch zenko $ZENKO_NAME --type merge -p '{"spec":{"sorbet":{"configurationOverrides":{"e2e-cold":{"pending-job-poll-after-age":"10s"}}}}}'
+kubectl patch zenko $ZENKO_NAME --type merge -p '{"spec":{"sorbet":{"configurationOverrides":{"e2e-cold":{"pending-job-poll-check-interval":"10s"}}}}}'
+
+kubectl patch zenko $ZENKO_NAME --type merge -p '{"spec":{"sorbet":{"server":{"azure":{"restoreTimeout":"15s"}}}}}'
```
francoisferrand: we should do this in a single patch, to avoid repeated re-deploy by ZKOP...

or even add this to the `zenko.yaml` file used by the tests, which will be much more readable...

.github/scripts/end2end/run-e2e-ctst.sh
```diff
@@ -69,9 +69,10 @@ AZURE_ARCHIVE_MANIFEST_ACCESS_TIER="Hot"
 INSTANCE_ID=$(kubectl get zenko ${ZENKO_NAME} -o jsonpath='{.status.instanceID}')
 
 KAFKA_CLEANER_INTERVAL=$(kubectl get zenko ${ZENKO_NAME} -o jsonpath='{.spec.kafkaCleaner.interval}')
+SORBETD_RESTORE_TIMEOUT=$(kubectl get zenko ${ZENKO_NAME} -o jsonpath='{.spec.sorbet.server.azure.restoreTimeout}')
```
francoisferrand, line 72: should we similarly pick up the `pending-job-poll-after-age` and `pending-job-poll-check-interval` ?
so timeouts can be adjusted in the tests...

KillianG: We could, but it will also depends on backbeat conrule interval and this will not make jobs going faster because we already make succeed when an object is transitioned
Eventually we can use it to change the interval in which we check if the object is transitioned or not..

But polling delay will depend on the location name which makes it hard to use properly, and I am not sure if it is worth the effort though..

ghost (approved):

francoisferrand (approved):

PR #2085, opened 2024-05-31
CTST improvements

ghost:
**Main goal of this PR**: to better manage identities in the tests, allowing to rename the test folders properly, with no risks having conflicts between tests. This change required deep updates of CTST and Zenko test suite:

- Bump CTST to v1.0.0
- Better logging.
- Better identity management.
- Reduce by 1s the duration of each test, with a total gain of ~17min, by better using the identities cache.
- Various code fixes (e.g., `execShellCommand` during dmf tests would both reject and resolve in 2.7+ branches...).
- Decreasing total builds duration times, by running @Azure and @Dmf scenarios in parallel, we gain 1h per build in 2.8+ branches. 
- Correct existing tests that were false positives (no bug detected, it was just the tests).
- Backport some code changes affecting the same tests from upper branches, to reduce the integration branches mess. Also include some functions we have in upper branches, so not used today but still useful, and better for maintenance between branches.

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - 

Example of passing builds (only difference is CTST tag, from hash to 1.0.0):
* 2.7 https://github.com/scality/Zenko/actions/runs/9373290348
* 2.8 https://github.com/scality/Zenko/actions/runs/9373381220
* 2.9 https://github.com/scality/Zenko/actions/runs/9373381227


williamlardier:
ping

KillianG (approved):

.github/scripts/end2end/run-e2e-ctst.sh
```diff
@@ -60,7 +45,7 @@ INSTANCE_ID=$(kubectl get zenko ${ZENKO_NAME} -o jsonpath='{.status.instanceID}'
...
+WORLD_PARAMETERS='{"subdomain":"'${SUBDOMAIN}'","ssl":false,"port":"'${ZENKO_PORT}'","AccountName":"'${ZENKO_ACCOUNT_NAME}'","AdminAccessKey":"'${ADMIN_ACCESS_KEY_ID}'","AdminSecretKey":"'${ADMIN_SECRET_ACCESS_KEY}'","VaultAuthHost":"'${VAULT_AUTH_HOST}'","NotificationDestination":"'${NOTIF_DEST_NAME}'","NotificationDestinationTopic":"'${NOTIF_DEST_TOPIC}'","NotificationDestinationAlt":"'${NOTIF_ALT_DEST_NAME}'","NotificationDestinationTopicAlt":"'${NOTIF_ALT_DEST_TOPIC}'","KafkaHosts":"'${KAFKA_HOST_PORT}'","KeycloakPassword":"'${KEYCLOAK_TEST_PASSWORD}'","KeycloakHost":"'${KEYCLOAK_TEST_HOST}'","KeycloakPort":"'${KEYCLOAK_TEST_PORT}'","KeycloakRealm":"'${KEYCLOAK_TEST_REALM_NAME}'","KeycloakClientId":"'${KEYCLOAK_TEST_CLIENT_ID}'","KeycloakGrantType":"'${KEYCLOAK_TEST_GRANT_TYPE}'","StorageManagerUsername":"'${STORAGE_MANAGER_USER_NAME}'","StorageAccountOwnerUsername":"'${STORAGE_ACCOUNT_OWNER_USER_NAME}'","DataConsumerUsername":"'${DATA_CONSUMER_USER_NAME}'","ServiceUsersCredentials":'${SERVICE_USERS_CREDENTIALS}',"InstanceID":"'${INSTANCE_ID}'","KafkaCleanerInterval":"'${KAFKA_CLEANER_INTERVAL}'"}'
```
francoisferrand: can we use multi-line, to make this readable (and maintainable) ?

```suggestion
WORLD_PARAMETERS="$(jq -c <<EOF
{
  "subdomain":"${SUBDOMAIN}",
  "ssl":false,"port":"${ZENKO_PORT}",
  "AccountName":"${ZENKO_ACCOUNT_NAME}",
  "AdminAccessKey":"${ADMIN_ACCESS_KEY_ID}",
  "AdminSecretKey":"${ADMIN_SECRET_ACCESS_KEY}",
  "VaultAuthHost":"${VAULT_AUTH_HOST}",
  "NotificationDestination":"${NOTIF_DEST_NAME}",
  "NotificationDestinationTopic":"${NOTIF_DEST_TOPIC}",
  "NotificationDestinationAlt":"${NOTIF_ALT_DEST_NAME}",
  "NotificationDestinationTopicAlt":"${NOTIF_ALT_DEST_TOPIC}",
  "KafkaHosts":"${KAFKA_HOST_PORT}",
  "KeycloakPassword":"${KEYCLOAK_TEST_PASSWORD}",
  "KeycloakHost":"${KEYCLOAK_TEST_HOST}",
  "KeycloakPort":"${KEYCLOAK_TEST_PORT}",
  "KeycloakRealm":"${KEYCLOAK_TEST_REALM_NAME}",
  "KeycloakClientId":"${KEYCLOAK_TEST_CLIENT_ID}",
  "KeycloakGrantType":"${KEYCLOAK_TEST_GRANT_TYPE}",
  "StorageManagerUsername":"${STORAGE_MANAGER_USER_NAME}",
  "StorageAccountOwnerUsername":"${STORAGE_ACCOUNT_OWNER_USER_NAME}",
  "DataConsumerUsername":"${DATA_CONSUMER_USER_NAME}",
  "ServiceUsersCredentials":${SERVICE_USERS_CREDENTIALS},
  "InstanceID":"${INSTANCE_ID}",
  "KafkaCleanerInterval":"${KAFKA_CLEANER_INTERVAL}"
}
EOF
)"
```

tests/ctst/common/hooks.ts
```diff
@@ -17,11 +21,88 @@ const noParallelRun = atMostOnePicklePerTag(['@AfterAll']);
...
 
-After(function (this: Zenko) {
-    this.resetSaved();
+/**
+ * The objective of this hook is to prepare all the buckets and accounts
+ * we use during quota checks, so that we avoid running the job multiple
+ * times, which affects the performance of the tests.
+ * The steps are: create an account, then create a simple bucket
+ */
+Before({ tags: '@Quotas', timeout: 1200000 }, async function ({ gherkinDocument, pickle }) {
```
francoisferrand: this before hook is specific to `Quotas` : can we define it in `quotas.ts` instead?
this improve "locality", making the logic easier to follow...

ghost: Not really, the hooks in cucumber are evaluated in the same order they are defined, and this hook must be placed here ... Having it in a separate file will cause issues as it changes the order of tests. That's why we have a dedicated file for hooks.
There is no "ordering" feature yet in cucumber

ghost: The reason we need to have it here, is that the main `Before` hook is defined just above, and is required before anything starts, as it is the one that init the Zenko world class, and reset the identity from a previous test.

.github/scripts/end2end/run-e2e-ctst.sh
ghost: Done in 2.6 and 2.7 branches (the multi line was already in place in 2.8+)

williamlardier:
ping

tests/ctst/common/hooks.ts
francoisferrand: :-(

Could we keep the `Before()` here, but move the body (i.e. the async function) in the `quota.ts` ?
it would address the locality aspect, but maybe introduce a (blocking?) dependency loop?

.github/scripts/end2end/run-e2e-ctst.sh
```diff
@@ -3,22 +3,7 @@ set -exu
...
-PARALLEL_RUNS=$(awk -v vcpu=$VCPU_COUNT 'BEGIN {
-  if (vcpu <= 0) {
-    print 1
-  } else if (vcpu >= 1 && vcpu <= 4) {
-    print int(vcpu * 0.5)
-  } else {
-    print (4 * 0.5) + ((vcpu - 4) * 1)
-  }
-}')
+PARALLEL_RUNS=4
```
francoisferrand: why hard-code to 4, instead of using `nproc` as before?

ghost: We have a static 4 in the upper branches, it's to have everything consistent. I saw little benefits of having more anyway, since we added more tests, it has a cpu impact that can increase the "flakies" rate

tests/ctst/common/hooks.ts
ghost: I will try that, I think it should work...

ghost: I made the change, it passes the local build, so we should be good, all branches are updated, and I also moved the teardown hook

.github/scripts/end2end/run-e2e-ctst.sh
francoisferrand: i am fine with a max of `4` (or simplifying the computation to `max(1, ncore/2)`) : my concern is more that we may use 4 cores even when running on a smaller vm (possibly locally)

ghost: I see, I updated the logic to keep the dynamic aspect, but simplify it: https://github.com/scality/Zenko/pull/2085/commits/64fce00face4d8d85ebbae0159e3d589d031c60f

No need to have the 0.5 vcpu per vcpu above 4, as the old computation was relevant when we used to run CTST in CLI mode. SDK is lighter.

tests/ctst/steps/quotas/quotas.ts
```diff
@@ -88,15 +81,40 @@ Before({ tags: '@Quotas', timeout: 1200000 }, async function ({ gherkinDocument,
...
     world.addToSaved('bucketName', key);
-    // Save the account name for the scenario
-    Zenko.saveAccountAccessKeys(config.AccessKey, config.SecretKey);
     world.addToSaved('accountName', key);
-});
+    // Save the account name for the scenario
+    Identity.addIdentity(IdentityEnum.ACCOUNT, key, configuration[key], undefined, true, true);
+}
+
+export async function quotaScenarioteardown(world: Zenko) {
```
francoisferrand, line 92: nit: `teardownQuotaScenarios` (matching `prepareQuotaScenarios`)

francoisferrand (approved):

tests/ctst/steps/quotas/quotas.ts
ghost: I will do it in the next PR (for kafkacleaner flakiness & azure locations), thanks

PR #2087, opened 2024-06-04
Deploy PRA

KillianG:
Issue: ZENKO-4772


.github/actions/deploy/action.yaml
```diff
@@ -51,9 +51,11 @@ runs:
     - name: Install operator
       shell: bash
       run: |-
-        set -exu;
+        set -ex;
```
francoisferrand: should keep the `u` flag, if it causes issue it mean you are using a non-referenced variable at some point: and code can be fixed...

.github/actions/deploy/action.yaml
```diff
@@ -51,9 +51,11 @@ runs:
...
         export OPERATOR_IMAGE_NAME=$(yq eval '."zenko-operator" | .sourceRegistry + "/" + .image' solution/deps.yaml)
-        export OPERATOR_IMAGE_TAG=$(yq eval '."zenko-operator".tag' solution/deps.yaml)
+        if [[ -z "${OPERATOR_IMAGE_TAG}" ]]; then
```
francoisferrand, line 56: where could this be set? in which use case do we need this?

.github/workflows/end2end.yaml
```diff
@@ -451,6 +451,89 @@ jobs:
...
+        uses: docker/login-action@v3
+        with:
+          username: "${{ github.repository_owner }}"
+          password: "${{ github.token }}"
+          registry: ghcr.io
+      - name: Deploy Zenko
+        uses: ./.github/actions/deploy
+        env:
+          ZENKO_MONGODB_SHARDED: "true"
+          OPERATOR_IMAGE_TAG: "784163337244df4110ae6c551ffe8ba2fba9a91e"
```
francoisferrand: to be changed before merge (we can create a `1.6.0-preview.1` release)

.github/actions/deploy/action.yaml
francoisferrand: should this not be specified by a "parameter" of the composite step instead?

.github/scripts/end2end/configs/zenko_dr_sink.yaml
```diff
@@ -0,0 +1,46 @@
...
+        tag:   '${KAFKA_CONNECT_TAG}'
+      cruiseControl:
+        image: '${KAFKA_CRUISECONTROL_IMAGE}'
+        tag: '${KAFKA_CRUISECONTROL_TAG}'
+      lagExporter:
+        image: '${KAFKA_LAGEXPORTER_IMAGE}'
+        tag: '${KAFKA_LAGEXPORTER_TAG}'
+      monitoring:
+        image: '${JMX_JAVAAGENT_IMAGE}'
+        tag: '${JMX_JAVAAGENT_TAG}'
```
francoisferrand: missing EOL at EOF

.github/scripts/end2end/configs/zenko_dr_source.yaml
```diff
@@ -0,0 +1,43 @@
...
+        tag:   '${KAFKA_CONNECT_TAG}'
+      cruiseControl:
+        image: '${KAFKA_CRUISECONTROL_IMAGE}'
+        tag: '${KAFKA_CRUISECONTROL_TAG}'
+      lagExporter:
+        image: '${KAFKA_LAGEXPORTER_IMAGE}'
+        tag: '${KAFKA_LAGEXPORTER_TAG}'
+      monitoring:
+        image: '${JMX_JAVAAGENT_IMAGE}'
+        tag: '${JMX_JAVAAGENT_TAG}'
```
francoisferrand: missing EOL at EOF

.github/scripts/end2end/deploy-dr.sh
```diff
@@ -0,0 +1,24 @@
...
+
+cat configs/zenko_dr_source.yaml | envsubst | kubectl apply -f -
+
+k_cmd="kubectl -n ${NAMESPACE} get zenkodrsource/${DR_SOURCE_NAME}"
+for i in $(seq 1 120); do
+    conditions=$($k_cmd -o "jsonpath={.status.conditions}")
+    if kubectl wait --for condition=Available --timeout 5s --namespace ${NAMESPACE} zenkodrsource/${DR_SOURCE_NAME}; then
+        break;
+    fi
+done
```
francoisferrand: missing EOL at EOF

.github/scripts/end2end/prepare-pra.sh
```diff
@@ -0,0 +1,35 @@
...
+kind: Secret
+metadata:
+  name: ${ZENKO_MONGODB_SECRET_NAME}
+stringData:
+  mongodb-root-username: $MONGODB_ROOT_USERNAME
+  mongodb-root-password: $MONGODB_ROOT_PASSWORD 
+  mongodb-username: $MONGODB_PRA_USERNAME
+  mongodb-password: $MONGODB_PRA_PASSWORD 
+  mongodb-database: $MONGODB_PRA_DATABASE
+EOF
```
francoisferrand: missing EOL at EOF

francoisferrand (approved):

.github/actions/deploy/action.yaml
KillianG: As this is get by `yq` from `deps.yaml` I think we can keep it here, because anyway I will get it the same way in the composite step..
I prefer keep it here instead of having to use the same bash script in multiple places...

This is also why I removed the `u` from `set -exu` to be able to check if `OPERATOR_IMAGE_TAG` is not set, and if that's the case, get it from `deps.yaml`
If i keep the `u` it will just say that `OPERATOR_IMAGE_TAG` is not set and will exit here...

welansari (approved):

.github/actions/deploy/action.yaml
francoisferrand: > As this is get by yq from deps.yaml I think we can keep it here, because anyway I will get it the same way in the composite step..

the point of having it as an explicit parameter is to make it clear that it is an input that may be provided : instead of passing it through some env variable (which may be missed), making it a parameter makes it obvious & documented that it can be set.
it does not really change the way the script is used : either the parameter is set (and its value can be used), or a default behavior can get the value from the `deps.yaml` file.

> This is also why I removed the u from set -exu to be able to check if OPERATOR_IMAGE_TAG is not set, and if that's the case, get it from deps.yaml. If i keep the u it will just say that OPERATOR_IMAGE_TAG is not set and will exit here...

`if [[ -z ${VARIABLE} ]]` is strange, `-z` is a basic test operator (used with -z).
But you have multiple option to get things to work:

* testing if the variable is set with advanced test syntax `if ! [[ -v VAR ]]`
* testing if the variable is set with the isset operator `if ! isset VAR`
* providing a default value in the variable expension `${MY_VARIABLE:-defaultValue}`, or possibly `if [ -z ${VAR:-} ]`

PR #2097, opened 2024-06-24
Fix Kafka cleaner & Azure Location creation flakies

ghost:
- Fix Kafkacleaner test, with dynamic sleep based on the partitions, and comparing the `low` property evolution to be more robust.
- Add kubernetes clients for observing zenko status and data service deployments (cloudserver & backbeat)
- In 2.8+ branches, fix the azure tests creating locations by using these new functions.

Change in 2.8+: https://github.com/scality/Zenko/commit/032cb7ed5349e7b3e89e37e046556799c2a7aef7

### DIFF
--- a/tests/ctst/common/common.ts
+++ b/tests/ctst/common/common.ts
@@ -329,11 +329,11 @@ Then('i {string} be able to add user metadata to object {string}',
 Then('kafka consumed messages should not take too much place on disk', { timeout: -1 },
     async function (this: Zenko) {
         const kfkcIntervalSeconds = parseInt(this.parameters.KafkaCleanerInterval);
-        const checkInterval = kfkcIntervalSeconds * 1000 + 5000;
+        const checkInterval = kfkcIntervalSeconds * (1000 + 5000);
 
         const timeoutID = setTimeout(() => {
             assert.fail('Kafka cleaner did not clean the topics within the expected time');
-        }, checkInterval * 5); // Timeout after 5 Kafka cleaner intervals
+        }, checkInterval * 10); // Timeout after 10 Kafka cleaner intervals
 
         try {
             const ignoredTopics = ['dead-letter'];
--- a/tests/ctst/common/common.ts
+++ b/tests/ctst/common/common.ts
@@ -305,7 +305,7 @@ Then('object {string} should be {string} and have the storage class {string}', {
     });
 
 When('i delete object {string}', async function (this: Zenko, objectName: string) {
-    const objName = getObjectNameWithBackendFlakiness.call(this, objectName) ||  this.getSaved<string>('objectName');
+    const objName = getObjectNameWithBackendFlakiness.call(this, objectName) || this.getSaved<string>('objectName');
     this.resetCommand();
     this.addCommandParameter({ bucket: this.getSaved<string>('bucketName') });
     this.addCommandParameter({ key: objName });
@@ -328,12 +328,9 @@ Then('i {string} be able to add user metadata to object {string}',
 
 Then('kafka consumed messages should not take too much place on disk', { timeout: -1 },
     async function (this: Zenko) {
+        let timeoutID;
         const kfkcIntervalSeconds = parseInt(this.parameters.KafkaCleanerInterval);
-        const checkInterval = kfkcIntervalSeconds * (1000 + 3000);
-
-        const timeoutID = setTimeout(() => {
-            assert.fail('Kafka cleaner did not clean the topics');
-        }, checkInterval * 5); // Timeout after 5 kafkacleaner intervals
+        const checkInterval = kfkcIntervalSeconds * 1000 + 5000;
 
         try {
             const ignoredTopics = ['dead-letter'];
@@ -342,8 +339,13 @@ Then('kafka consumed messages should not take too much place on disk', { timeout
                 .filter(t => (t.includes(this.parameters.InstanceID) &&
                 !ignoredTopics.some(e => t.includes(e))));
 
+            timeoutID = setTimeout(() => {
+                assert.fail('Kafka cleaner did not clean the topics within the expected time');
+            }, (topics.length || 1) * checkInterval * 5); // Timeout after 5 Kafka cleaner intervals
+
+            const previousOffsets = await getTopicsOffsets(topics, kafkaAdmin);
+
             while (topics.length > 0) {
-                const previousOffsets = await getTopicsOffsets(topics, kafkaAdmin);
                 // Checking topics offsets before kafkacleaner passes to be sure kafkacleaner works
                 // This function can be improved by consuming messages and
                 // verify that the timestamp is not older than last kafkacleaner run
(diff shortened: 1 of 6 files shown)


williamlardier:
ping

tests/ctst/common/common.ts
```diff
@@ -357,32 +358,23 @@ Then('kafka consumed messages should not take too much place on disk', { timeout
...
-                        // it means the topic is completly cleaned even though lowOffset didnt increased
-                        assert.ok(lowOffsetIncreased || allMessagesCleaned,
-                            `Topic ${topics[i]} partition ${j} offset has not increased,
-                            previousOffsets: ${previousOffsets[i].partitions[j].low} /\
-                            ${previousOffsets[i].partitions[j].high},
-                            newOffsets: ${newOffsets[i].partitions[j].low} / ${newOffsets[i].partitions[j].high}`);
-
-                        // Topic is cleaned, we don't need to check it anymore
+                        // Ensure we're accessing the correct partition details
+                        const lowOffsetIncreased = parseInt(partition.low) >
```
KillianG: why do we need to `parseInt` here and not below?

tests/ctst/common/common.ts
```diff
@@ -357,32 +358,23 @@ Then('kafka consumed messages should not take too much place on disk', { timeout
...
-                        const newMessagesAfterClean =
-                            newOffsets[i].partitions[j].low === previousOffsets[i].partitions[j].high &&
-                            newOffsets[i].partitions[j].high > previousOffsets[i].partitions[j].high;
-
-                        if (newMessagesAfterClean) {
-                            // If new messages appeared after we gathered the offsets, we need to recheck after
-                            this.logger.warn('New messages after clean', { topic: topics[i] });
-                            continue;
-                        }
+                        const partition = newOffsets[i].partitions[j];
```
KillianG: Should be renamed newOffsetPartition so that we can create a variable as well for previousOffset

tests/ctst/common/common.ts
```diff
@@ -357,32 +358,23 @@ Then('kafka consumed messages should not take too much place on disk', { timeout
...
-
-                        // Topic is cleaned, we don't need to check it anymore
+                        // Ensure we're accessing the correct partition details
+                        const lowOffsetIncreased = parseInt(partition.low) >
+                            parseInt(previousOffsets[i].partitions[j].low);
+                        const allMessagesCleaned = partition.high === partition.low;
+
+                        if (!(lowOffsetIncreased || allMessagesCleaned)) {
+                            // Log warning if the condition is not met for this partition
+                            this.logger.warn(`Partition ${j} of topic ${topics[i]} not cleaned as expected`);
```
KillianG: I think we can keep the verbose log

tests/ctst/steps/utils/kubernetes.ts
```diff
@@ -84,3 +123,174 @@ export async function createJobAndWaitForCompletion(world: Zenko, jobName: strin
         throw err;
     }
 }
+
+export async function waitForZenkoToStabilize(world: Zenko, needsReconciliation = false, namespace = 'default') {
+    // ZKOP pulls the overlay configuration from Pensieve every 5 seconds
+    // So the status might not be updated immediately after the overlay is applied.
+    // So, this function will first wait till we detect a reconciliation
+    // (deploymentInProgress = true), and then wait for the status to be available
+    const timeout = 15 * 60 * 1000;
```
KillianG: why 15 minutes ? can we make it configurable ? The default TO in cucumber is 100 seconds, should we make it 100 seconds as well by default ?

tests/ctst/common/common.ts
```diff
@@ -357,32 +358,23 @@ Then('kafka consumed messages should not take too much place on disk', { timeout
 
                 for (let i = 0; i < topics.length; i++) {
                     this.logger.debug('Checking topic', { topic: topics[i] });
+                    let topicCleaned = true;
```
KillianG: topicsCleaned should probably be false by default I think

tests/ctst/steps/utils/kubernetes.ts
ghost: I think we can access the timeout from the test by default, by passing the test object in the function, so that we wait just what's needed

tests/ctst/common/common.ts
ghost: I am not done with the pr yet, jira transitioned it in review because I created the PR (not in draft to have the integration branches), but I still have some parts to improve, especially on the 2.8 branch
I will let you know when it's done

tests/ctst/steps/utils/kubernetes.ts
ghost: So, to answer the question: the functions will have the timeout as a parameter, but the default remains 15min; the reason is that the service restart can take up to 3~4min as per my observations, and sometimes longer, if some tests are running at high frequency, blocking kube from restarting the pods quickly (as they are still handling requests).
So, the steps reyling on these functions will need specific timeouts.

tests/ctst/common/common.ts
```diff
@@ -357,32 +359,37 @@ Then('kafka consumed messages should not take too much place on disk', { timeout
...
-                        // Topic is cleaned, we don't need to check it anymore
+                        // Ensure we're accessing the correct partition details
+                        const lowOffsetIncreased = parseInt(newOffsetPartition.low) >
+                            parseInt(oldOffsetPartition.low);
+                        // We tolerate one message not being cleaned, as it can be due to the
+                        // message being consumed during the check
+                        const allMessagesCleaned = parseInt(newOffsetPartition.low) + 1 >=
+                            parseInt(newOffsetPartition.high);
+
+                        // We consider one topic as xleaned if kafkacleaner affected the
```
ghost: Will be done before merging, keeping the PR as-is for now, for the review.

```suggestion
                        // We consider one topic as cleaned if kafkacleaner affected the
```

tests/ctst/common/common.ts
```diff
@@ -342,8 +339,13 @@ Then('kafka consumed messages should not take too much place on disk', { timeout
                 .filter(t => (t.includes(this.parameters.InstanceID) &&
                 !ignoredTopics.some(e => t.includes(e))));
 
+            timeoutID = setTimeout(() => {
+                assert.fail('Kafka cleaner did not clean the topics within the expected time');
+            }, (topics.length || 1) * checkInterval * 5); // Timeout after 5 Kafka cleaner intervals
```
francoisferrand: why do we increase the timeout by the number of topics?
- the timeout is based on kafka-cleaner period, and we check every (remaining) topic on each iteration: so number of topics should not increase the timeout...
- move the timeout here prevents catching issues with kafka connection (`listTopics()` call above) which could get stuck...

ghost: I adapted the logic, it's true we do not need to wait for each topic, but for sure we needed to wait more, as the cleaning might be slower in the CI due to other tests jeopardizing resources temporarily...

tests/ctst/common/common.ts
francoisferrand: if the PR is in draft state, you should keep it in draft ; and you can request integration branches by explicitely sending the `/create_integration_branches` command to bert-e

francoisferrand (approved):

benzekrimaha (approved):
LGTM

PR #2114, opened 2024-08-01
E2E tests for archives in PRA env

KillianG:
Missing e2e tests for restore on dr site, uninstall and pause/resume which are blocked by other tickets


.github/scripts/end2end/configs/zenkoversion-pra.yaml
francoisferrand, line 1: what is the different with the "standard" zenkoversion?
we generally don't want to have a different zenkoversion, and esp. here this file is filled with template variables: so we should just use the same...

(in the short/mid term, I want to remove this duplicate zenkoversion, and have a single definition : i.e. use the same thing as in prod...)

.github/scripts/end2end/configure-e2e-pra.sh
francoisferrand, line 1: this just creates a location e2e-cold, right?
can't we just use the "standard" location creation path we already have everywhere?

any reason not to go throught the same setup?

.github/scripts/end2end/configure-e2e-pra.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/sh
+
+set -exu
+
+. "$(dirname $0)/common.sh"
+
+OIDC_USERNAME=${OIDC_USERNAME:-zenko-end2end-pra}
+OIDC_EMAIL=${OIDC_EMAIL:-e2e-pra@zenko.local}
```
francoisferrand: these are not used

.github/scripts/end2end/configure-e2e-pra.sh
```diff
@@ -0,0 +1,20 @@
...
+
+INSTANCE_ID=$(kubectl get zenko end2end-pra -o jsonpath='{.status.instanceID}')
```
francoisferrand: `end2end-pra` should be a parameter of the script...

.github/scripts/end2end/prepare-pra.sh
```diff
@@ -6,19 +6,22 @@ export MONGODB_PRA_DATABASE="${MONGODB_PRA_DATABASE:-'pradb'}"
...
+echo 'ZENKO_UI_INGRESS="ui.dr.zenko.local"' >> $GITHUB_ENV
 
 MONGODB_ROOT_USERNAME="${MONGODB_ROOT_USERNAME:-'root'}"
 MONGODB_ROOT_PASSWORD="${MONGODB_ROOT_PASSWORD:-'rootpass'}"
 
 kubectl exec -it data-db-mongodb-sharded-mongos-0 -- mongo "admin" \
     -u "root" \
     -p "rootpass" \
-    --eval "db.createUser({user:$MONGODB_PRA_USERNAME,pwd:$MONGODB_PRA_PASSWORD,roles:[{role:'enableSharding',db:$MONGODB_PRA_DATABASE },{role:'readWrite',db:$MONGODB_PRA_DATABASE },{role:'read',db:'local'}]})"
+    --eval "db.createUser({user:$MONGODB_PRA_USERNAME,pwd:$MONGODB_PRA_PASSWORD,roles:[{role:'enableSharding',db:$MONGODB_PRA_DATABASE },{role:'readWriteAnyDatabase',db:'admin'}]})"
```
francoisferrand: why do we need PRA user to read every database?
this is supposed to be a "fake" standard mongo install : which has access to a single DB (zenko's DB)...

.github/scripts/end2end/run-e2e-pra-ctst.sh
```diff
@@ -0,0 +1,144 @@
...
+  "AdminAccessKey":"${ADMIN_ACCESS_KEY_ID}",
+  "AdminSecretKey":"${ADMIN_SECRET_ACCESS_KEY}",
+  "DRAdminAccessKey":"${DR_ADMIN_ACCESS_KEY_ID}",
+  "DRAdminSecretKey":"${DR_ADMIN_SECRET_ACCESS_KEY}",
+  "VaultAuthHost":"${VAULT_AUTH_HOST}",
+  "KafkaHosts":"${KAFKA_HOST_PORT}",
+  "StorageManagerUsername":"${STORAGE_MANAGER_USER_NAME}",
+  "StorageAccountOwnerUsername":"${STORAGE_ACCOUNT_OWNER_USER_NAME}",
+  "DataConsumerUsername":"${DATA_CONSUMER_USER_NAME}",
+  "ServiceUsersCredentials":${SERVICE_USERS_CREDENTIALS},
```
francoisferrand: why do we need to pass service user credentials?
these are internal to the system, while e2e test run from the perspective of a user....

.github/scripts/end2end/run-e2e-pra-ctst.sh
```diff
@@ -0,0 +1,144 @@
...
+EOF
+)"
+
+# Set up environment variables for testing
+kubectl set env deployment end2end-connector-cloudserver SCUBA_HEALTHCHECK_FREQUENCY=100
+kubectl rollout status deployment end2end-connector-cloudserver
+
+# disable moniroting of count-items
+kubectl set env cronjob end2end-ops-count-items PROMETHEUS_POLLING_ATTEMPTS=1
+kubectl set env cronjob end2end-ops-count-items PROMETHEUS_POLLING_PERIOD=1
```
francoisferrand: should be merged

```suggestion
kubectl set env cronjob end2end-ops-count-items PROMETHEUS_POLLING_ATTEMPTS=1 PROMETHEUS_POLLING_PERIOD=1
```

.github/scripts/end2end/run-e2e-pra-ctst.sh
```diff
@@ -0,0 +1,144 @@
...
+
+E2E_IMAGE=$E2E_CTST_IMAGE_NAME:$E2E_IMAGE_TAG
+POD_NAME="${ZENKO_NAME}-ctst-tests"
+
+CTST_VERSION=$(sed 's/.*"cli-testing": ".*#\(.*\)".*/\1/;t;d' ../../../tests/ctst/package.json)
+# Configure keycloak
```
francoisferrand: nit: missing line-break

.github/scripts/end2end/run-e2e-ctst.sh
```diff
@@ -146,4 +148,4 @@ kubectl run $POD_NAME \
       }
     ]
   }
-}' -- ./run "$COMMAND" $WORLD_PARAMETERS --parallel $PARALLEL_RUNS --retry $RETRIES --retry-tag-filter @Flaky --format junit:$JUNIT_REPORT_PATH
+}' -- ./run "$COMMAND" $WORLD_PARAMETERS --parallel $PARALLEL_RUNS --tags "not @PRA" --retry $RETRIES --retry-tag-filter @Flaky --format junit:$JUNIT_REPORT_PATH
```
francoisferrand: apart from the extra 3 "DR" parameters and the tags `@PRA` / `not @PRA`, this is the same script in both cases...
→ it probably would make more sense to keep a single script which gracefully ignores missing DR site, and takes a parameter with the tags (or just adds "$@" at the end of the run command, so that any extra param can be passed...)

```suggestion
}' -- ./run "$COMMAND" $WORLD_PARAMETERS --parallel $PARALLEL_RUNS --retry $RETRIES --retry-tag-filter @Flaky --format junit:$JUNIT_REPORT_PATH "$@"
```

(with an additional `shift 4` at the beginning of the file, to avoid adding the 4 first params: zenko name, command, <unused> and retries...)

.github/scripts/end2end/run-e2e-pra-ctst.sh
```diff
@@ -0,0 +1,144 @@
+#!/bin/bash
+set -exu
+
+ZENKO_NAME=${1:-end2end}
+COMMAND=${2:-"premerge"}
+PARALLEL_RUNS=${PARALLEL_RUNS:-$(( ( $(nproc) + 1 ) / 2 ))}
+RETRIES=${4:-3}
+JUNIT_REPORT_PATH=${JUNIT_REPORT_PATH:-"ctst-junit.xml"}
```
francoisferrand: may need different file names for PRA and the other tests, not sure we will "merge" artifacts properly otherwise...

tests/ctst/Dockerfile
```diff
@@ -1,6 +1,11 @@
 ARG CTST_TAG
+ARG DRCTL_TAG
+ARG KUBECONFIG_PATH
```
francoisferrand: not used...

tests/ctst/Dockerfile
```diff
@@ -1,6 +1,11 @@
...
+
+FROM ghcr.io/scality/zenko-drctl:$DRCTL_TAG as drctl
 FROM ghcr.io/scality/cli-testing:$CTST_TAG
 
+
```
francoisferrand: extraneous empty line

.github/scripts/end2end/configs/zenkoversion-pra.yaml
KillianG: We need to use a specific zenkoversion for PRA because we need to add drctl field which is only known in zkop 1.6.x versions, and we don't want to bump zkop in other tests yet..

.github/scripts/end2end/configure-e2e-pra.sh
KillianG: They are used in common.sh for get_token

.github/scripts/end2end/configs/zenkoversion-pra.yaml
francoisferrand: zkop should not be affected if there are more fields, they will just be ignored.
but k8s will indeed refuse to validate (and create) the zenkoversion CR if it has fields not in the manifest.

Instead of duplicating the file (and making somewhat wide-ranging changes, which will need to be reverted afterwards), we could instead somewhat _patch_ the zenkoversion file in the CI : either using yq to add the extra fields, by introducing a variable (like we do for annotations in the zenko CR), or possibly even using sed...

francoisferrand: btw it seems to work when Drctl is added to the "regular" zenkoversion : as it is there already, in this PR.. :/

.github/scripts/end2end/common.sh
```diff
@@ -8,4 +8,3 @@ get_token() {
         https://localhost/auth/realms/${OIDC_REALM}/protocol/openid-connect/token | \
         jq -cr '.id_token'
 }
-
```
francoisferrand: unrelated change?

.github/scripts/end2end/configs/zenkoversion-pra.yaml
KillianG: It works for end2end-pra but not for the others, or am i missing something ?

.github/scripts/end2end/configure-e2e.sh
```diff
@@ -63,10 +63,10 @@ kubectl run kafka-topics \
...
     --attach=True \
     --command -- bash -c \
-    "kafka-topics.sh --create --topic $UUID.backbeat-replication-replay-0 --partitions 5 --bootstrap-server $KAFKA_HOST_PORT ; \
-    kafka-topics.sh --create --topic $UUID.backbeat-data-mover --partitions 5 --bootstrap-server $KAFKA_HOST_PORT ; \
-    kafka-topics.sh --create --topic $NOTIF_DEST_TOPIC --bootstrap-server $KAFKA_HOST_PORT ; \
-    kafka-topics.sh --create --topic $NOTIF_ALT_DEST_TOPIC --bootstrap-server $KAFKA_HOST_PORT"
+    "kafka-topics.sh --create --topic $UUID.backbeat-replication-replay-0 --partitions 5 --bootstrap-server $KAFKA_HOST_PORT --if-not-exists ; \
+    kafka-topics.sh --create --topic $UUID.backbeat-data-mover --partitions 5 --bootstrap-server $KAFKA_HOST_PORT --if-not-exists ; \
+    kafka-topics.sh --create --topic $NOTIF_DEST_TOPIC --bootstrap-server $KAFKA_HOST_PORT --if-not-exists ; \
+    kafka-topics.sh --create --topic $NOTIF_ALT_DEST_TOPIC --bootstrap-server $KAFKA_HOST_PORT --if-not-exists"
```
francoisferrand: why would these topic already exist?

.github/scripts/end2end/run-e2e-ctst.sh
```diff
@@ -146,4 +157,4 @@ kubectl run $POD_NAME \
       }
     ]
   }
-}' -- ./run "$COMMAND" $WORLD_PARAMETERS --parallel $PARALLEL_RUNS --retry $RETRIES --retry-tag-filter @Flaky --format junit:$JUNIT_REPORT_PATH
+}' -- ./run "$COMMAND" $WORLD_PARAMETERS --parallel $PARALLEL_RUNS --retry $RETRIES "$@" --retry-tag-filter @Flaky --format junit:$JUNIT_REPORT_PATH
```
francoisferrand: nit: may be best to put `"$@"` at the end, so it may override params already provided

.github/workflows/end2end.yaml
```diff
@@ -381,7 +385,9 @@ jobs:
         with:
           push: true
           context: ./tests/ctst
-          build-args: CTST_TAG=${{ env.CTST_TAG }}
+          build-args: |
+            CTST_TAG=${{ env.CTST_TAG }}
+            DRCTL_TAG=v1.0.2
```
francoisferrand: nit: we may want to add workflow variable to hold the DRCTL version (`v1.0.2`) which is used in multiple places

.github/workflows/end2end.yaml
```diff
@@ -685,11 +683,14 @@ jobs:
...
           ZENKO_ENABLE_SOSAPI: true
+          TIME_PROGRESSION_FACTOR: 86400
+          TRANSITION_ONE_DAY_EARLIER: false
+          EXPIRE_ONE_DAY_EARLIER: false
       - name: Configure E2E test environment
         run: bash configure-e2e-ctst.sh
         working-directory: ./.github/scripts/end2end
       - name: Run CTST end to end tests
-        run: bash run-e2e-ctst.sh
+        run: bash run-e2e-ctst.sh "" "" "" "" --tags 'not @PRA'
```
francoisferrand, line 694: FYI, the "cleaner" way to avoid this would be to improve the parsing of args in `run-e2e-ctst.sh`:
* parse the args before `--` as before ($1, $2, ...)
* then instead of shifting by 4, shift until the `--`

this would allow using it simply like this: `bash run-e2e-ctst.sh -- --tags 'not @PRA'`

can be improved later, no need to fix right now

.github/scripts/end2end/prepare-pra.sh
```diff
@@ -6,11 +6,14 @@ export MONGODB_PRA_DATABASE="${MONGODB_PRA_DATABASE:-'pradb'}"
...
-export ZENKO_S3_INGRESS="s3.zenko-pra.local"
-export ZENKO_UI_INGRESS="ui.zenko-pra.local"
+echo 'ZENKO_MONGODB_DATABASE="pradb"' >> $GITHUB_ENV
+echo 'ZENKO_MONGODB_SECRET_NAME="mongodb-db-creds-pra"' >> $GITHUB_ENV
+
+echo 'ZENKO_IAM_INGRESS="iam.dr.zenko.local"' >> $GITHUB_ENV
+echo 'ZENKO_STS_INGRESS="sts.dr.zenko.local"' >> $GITHUB_ENV
+echo 'ZENKO_MANAGEMENT_INGRESS="management.dr.zenko.local"' >> $GITHUB_ENV
+echo 'ZENKO_S3_INGRESS="s3.dr.zenko.local"' >> $GITHUB_ENV
+echo 'ZENKO_UI_INGRESS="ui.dr.zenko.local"' >> $GITHUB_ENV
```
francoisferrand, line 16: seems to me this would be better set directly in the workflow, in the env for the `Deploy second Zenko for PRA`

francoisferrand (approved):
please also rebase (and reset followup branches) : the build will happen anyway, best to do it before the queue...

.github/scripts/end2end/deploy-zenko.sh
```diff
@@ -38,7 +38,7 @@ fi
 
 # TODO: use kustomize
 ZENKO_MONGODB_SHARDED=${ZENKO_MONGODB_SHARDED:-'false'}
-if [ ${ZENKO_MONGODB_SHARDED} == 'true' ]; then
+if [ "${ZENKO_MONGODB_SHARDED}" = 'true' ]; then
```
benzekrimaha, line 41:
```suggestion
if [ "${ZENKO_MONGODB_SHARDED}" == 'true' ]; then
```

.github/workflows/end2end.yaml
KillianG: Does this require some sort of parsing or is there a "native" way to do it ?

welansari (approved):

.github/scripts/end2end/configure-e2e.sh
francoisferrand: I see indeed a build error related to this, in "Configure E2E CTST environment" (https://github.com/scality/Zenko/actions/runs/10314993644/job/28554450831)
```
Error while executing topic command : Topic 'destination-topic-1' already exists.
```

This means that somehow we have already created them (unexpectedly) in previous steps... e.g. when running configuring e2e pra.
Do we really need to run configure-e2e twice? Should it not be enough to just run the "Configure E2E PRA environment" step?

.github/workflows/end2end.yaml
francoisferrand: no sure (looping on bash argument is quite native, maybe there could be a smart trick with getopt)... but again : no need to fix right now, we don't really have the time...

PR #2124, opened 2024-08-12
Support multiple kafkaClusters in kafka alerts

KillianG:
Issue: ZENKO-4857

### DIFF
--- a/monitoring/kafka/alerts.test.yaml
+++ b/monitoring/kafka/alerts.test.yaml
@@ -35,7 +35,7 @@ tests:
               description: 'Kafka: Broker count is down'
             exp_labels:
               severity: warning
-              service: ${service}
+              kafkaCluster: ${cluster}
       - alertname: BrokersCountCritical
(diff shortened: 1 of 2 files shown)


monitoring/kafka/alerts.test.yaml
```diff
@@ -35,6 +35,7 @@ tests:
               description: 'Kafka: Broker count is down'
             exp_labels:
               severity: warning
+              service: ${service}
```
francoisferrand: in the dashboard, we use `Cluster` as the label to identify specific instances → would be more appropriate than `service` (which is an implementation detail in k8s context)

to avoid confusion, since the alerts are 'global' and not a specifically chosen context (like a dashboard: user chooses to open kafka dashboard) : maybe we should prefix that with the name of the app → `kakfaCluster`

```suggestion
              kafkaCluster: ${service}
```

francoisferrand (approved):

welansari (approved):

PR #2126, opened 2024-08-14
End2End PRA add uninstall re-install test

KillianG:
Issue: ZENKO-4775

### DIFF
--- a/tests/ctst/steps/pra.ts
+++ b/tests/ctst/steps/pra.ts
@@ -48,7 +48,7 @@ interface DrState {
     };
 }
 
-async function installPRA(world: Zenko, sinkS3Endpoint = 'http://s3.zenko.local') {
+async function installPRA(world: Zenko, sinkS3Endpoint = 'http://s3.zenko.local', timeout: string = '30m') {
     const kafkaExternalIpOption = world.parameters.KafkaExternalIps ?
         { kafkaExternalIps: world.parameters.KafkaExternalIps } :
         { kafkaExternalIpsDiscovery: true };
@@ -162,14 +162,15 @@ async function waitForPhase(
     return false;
 }
 
-Given('a DR installed', { timeout: 330000 }, async function (this: Zenko) {
+const installTimeout = 360000;
+Given('a DR installed', { timeout: installTimeout }, async function (this: Zenko) {
     Identity.useIdentity(IdentityEnum.ACCOUNT, Zenko.sites['source'].accountName);
     const credentials = Identity.getCurrentCredentials();
     await createSecret(this, 'drctl-s3-creds', {
         accessKey: Buffer.from(credentials.accessKeyId).toString('base64'),
         secretAccessKey: Buffer.from(credentials.secretAccessKey).toString('base64'),
     });
-    await installPRA(this);
+    await installPRA(this, undefined, installTimeout.toString() + 'ms');
     return;
 });
 
--- a/tests/ctst/steps/pra.ts
+++ b/tests/ctst/steps/pra.ts
@@ -48,7 +48,7 @@ interface DrState {
     };
 }
 
-async function installPRA(world: Zenko, sinkS3Endpoint = 'http://s3.zenko.local', timeout: string = '30m') {
+async function installPRA(world: Zenko, sinkS3Endpoint = 'http://s3.zenko.local', timeout = '30m') {
     const kafkaExternalIpOption = world.parameters.KafkaExternalIps ?
         { kafkaExternalIps: world.parameters.KafkaExternalIps } :
         { kafkaExternalIpsDiscovery: true };
@@ -71,7 +71,7 @@ async function installPRA(world: Zenko, sinkS3Endpoint = 'http://s3.zenko.local'
         // prometheusHostname: 'prom.dr.zenko.local', // could be any name, cert will be auto-generated
         prometheusExternalIpsDiscovery: true,
         prometheusDisableTls: true,
-        forceRotateServiceCredentials: true,
+        forceRotateServiceCredentials: world.hasDRAlreadyBeenInstalled,
         ...kafkaExternalIpOption,
     });
 }
@@ -170,7 +170,8 @@ Given('a DR installed', { timeout: installTimeout }, async function (this: Zenko
         accessKey: Buffer.from(credentials.accessKeyId).toString('base64'),
         secretAccessKey: Buffer.from(credentials.secretAccessKey).toString('base64'),
     });
-    await installPRA(this, undefined, installTimeout.toString() + 'ms');
+    await installPRA(this, undefined, `${installTimeout.toString()}ms`);
+    this.hasDRAlreadyBeenInstalled = true;
     return;
 });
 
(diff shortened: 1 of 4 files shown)


.github/scripts/end2end/configs/zenko.yaml
```diff
@@ -52,7 +52,7 @@ spec:
         size: 20Gi
         storageClassName: "standard"
   kafkaCleaner:
-    interval: 30s
+    interval: 5m
```
francoisferrand: 5m will make the kafka-cleaner test muuuchhh longer...
is KFKC period really an issue, now that you fixed everything else?
do we have an idea on the mecanism by which it breaks the re-install tests? (e.g. cpu measurement if load-related, race-condition in the pra, ...)

KillianG: This fixes the re-install part, before that change, the second install took way too much time...
I don't know exactly why, but passing it to 5mn fixed the reinstall...
I can probably shorten it to 2/3 minutes but 30s is way too fast i guess...

KillianG: Made it back to 30s and it works

tests/ctst/steps/pra.ts
```diff
@@ -161,7 +162,7 @@ async function waitForPhase(
     return false;
 }
 
-Given('a DR installed', { timeout: 360000 }, async function (this: Zenko) {
+Given('a DR installed', { timeout: 330000 }, async function (this: Zenko) {
```
ghost: Should we specify the timeout for the install command, to match this one? By default it's 30min (so 1800000).

tests/ctst/steps/pra.ts
```diff
@@ -48,7 +48,7 @@ interface DrState {
     };
 }
 
-async function installPRA(world: Zenko, sinkS3Endpoint = 'http://s3.zenko.local') {
+async function installPRA(world: Zenko, sinkS3Endpoint = 'http://s3.zenko.local', timeout: string = '30m') {
```
francoisferrand: please fix the lint issues

tests/ctst/steps/pra.ts
```diff
@@ -161,14 +162,15 @@ async function waitForPhase(
...
+const installTimeout = 360000;
+Given('a DR installed', { timeout: installTimeout }, async function (this: Zenko) {
     Identity.useIdentity(IdentityEnum.ACCOUNT, Zenko.sites['source'].accountName);
     const credentials = Identity.getCurrentCredentials();
     await createSecret(this, 'drctl-s3-creds', {
         accessKey: Buffer.from(credentials.accessKeyId).toString('base64'),
         secretAccessKey: Buffer.from(credentials.secretAccessKey).toString('base64'),
     });
-    await installPRA(this);
+    await installPRA(this, undefined, installTimeout.toString() + 'ms');
```
francoisferrand: please fix lint issue

tests/ctst/steps/pra.ts
```diff
@@ -71,6 +71,7 @@ async function installPRA(world: Zenko, sinkS3Endpoint = 'http://s3.zenko.local'
         // prometheusHostname: 'prom.dr.zenko.local', // could be any name, cert will be auto-generated
         prometheusExternalIpsDiscovery: true,
         prometheusDisableTls: true,
+        forceRotateServiceCredentials: true,
```
francoisferrand: this "force" option should be passed only the `second` install : it shoudl be done autoamtically on the first install (allowing seamless operation), so that is what we want to test in general.

but indeed, we need to pass it on the _second_ install (and only on followup installs)

tests/ctst/steps/pra.ts
```diff
@@ -161,14 +162,16 @@ async function waitForPhase(
...
+Given('a DR installed', { timeout: installTimeout }, async function (this: Zenko) {
     Identity.useIdentity(IdentityEnum.ACCOUNT, Zenko.sites['source'].accountName);
     const credentials = Identity.getCurrentCredentials();
     await createSecret(this, 'drctl-s3-creds', {
         accessKey: Buffer.from(credentials.accessKeyId).toString('base64'),
         secretAccessKey: Buffer.from(credentials.secretAccessKey).toString('base64'),
     });
-    await installPRA(this);
+    await installPRA(this, undefined, `${installTimeout.toString()}ms`);
+    this.hasDRAlreadyBeenInstalled = true;
```
francoisferrand: nit: `wasPraInstalled` (or maybe better: `praInstallsCount` ?)

francoisferrand (approved):

tests/ctst/steps/pra.ts
```diff
@@ -48,7 +48,7 @@ interface DrState {
     };
 }
 
-async function installPRA(world: Zenko, sinkS3Endpoint = 'http://s3.zenko.local') {
+async function installPRA(world: Zenko, sinkS3Endpoint = 'http://s3.zenko.local', timeout = '30m') {
```
ghost, line 51: I think we are not using the timeout here? As per the linter, so it is not propagated to the drctl install command

tests/ctst/steps/pra.ts
```diff
@@ -71,6 +71,7 @@ async function installPRA(world: Zenko, sinkS3Endpoint = 'http://s3.zenko.local'
         // prometheusHostname: 'prom.dr.zenko.local', // could be any name, cert will be auto-generated
         prometheusExternalIpsDiscovery: true,
         prometheusDisableTls: true,
+        forceRotateServiceCredentials: world.praInstallCount > 0,
```
ghost:
```suggestion
        forceRotateServiceCredentials: world.praInstallCount > 0,
        timeout,
```

tests/ctst/steps/pra.ts
```diff
@@ -161,14 +162,16 @@ async function waitForPhase(
...
+const installTimeout = 360000;
+Given('a DR installed', { timeout: installTimeout }, async function (this: Zenko) {
     Identity.useIdentity(IdentityEnum.ACCOUNT, Zenko.sites['source'].accountName);
     const credentials = Identity.getCurrentCredentials();
     await createSecret(this, 'drctl-s3-creds', {
         accessKey: Buffer.from(credentials.accessKeyId).toString('base64'),
         secretAccessKey: Buffer.from(credentials.secretAccessKey).toString('base64'),
     });
-    await installPRA(this);
+    await installPRA(this, undefined, `${installTimeout.toString()}ms`);
```
ghost: I would suggest that the command timeout is slightly lower than the step timeout as it will likely take maybe 1s more, and it's better that we first see the command doing a timeout, and not the step?

tests/ctst/world/Zenko.ts
```diff
@@ -114,6 +114,8 @@ export default class Zenko extends World<ZenkoWorldParameters> {
     static readonly PRIMARY_SITE_NAME = 'admin';
     static readonly SECONDARY_SITE_NAME = 'dradmin';
 
+    public praInstallCount = 0;
```
ghost: For the future in case we have other scenarios, we can set it in the cache of CTST instead...

ghost (approved):

francoisferrand:
Merged in #2136 (+last commit added to improvement/ZENKO-4876 branch)

PR #2132, opened 2024-08-28
Improvement/zenko 4871 readonly for pra

KillianG:


tests/ctst/steps/pra.ts
```diff
@@ -276,6 +277,26 @@ Then('object {string} should {string} be {string} and have the storage class {st
         }
     });
 
+When('the user tries to perform PutObject on {string} site', async function (this: Zenko, site: string) {
```
ghost:
```suggestion
When('the DATA_ACCESSOR user tries to perform PutObject on {string} site', async function (this: Zenko, site: string) {
```

tests/ctst/steps/pra.ts
```diff
@@ -276,6 +277,26 @@ Then('object {string} should {string} be {string} and have the storage class {st
...
+    if (site === 'DR') {
+        Identity.useIdentity(IdentityEnum.ACCOUNT, `${Zenko.sites['source'].accountName}-replicated`);
+    } else {
+        Identity.useIdentity(IdentityEnum.ACCOUNT, Zenko.sites['source'].accountName);
+    }
+    this.resetCommand();
+    this.addToSaved('accountName', Zenko.sites['source'].accountName);
+
+    await this.setupEntity(EntityType.DATA_ACCESSOR);
```
ghost: I see a CI error here, I believe it's a limitation from CTST core: we use the current identity information to get the subdomain to use, and so we expect to see credentials for the current identity. Usually it works because you will always have an account AK/SK set, but here it's different, as we use an identity with no credentials...

To sum up: to make it work here you need to create an access key for the replicated account

Or, you can review this :relieved: https://github.com/scality/cli-testing/pull/82

tests/ctst/features/pra.feature
```diff
@@ -22,6 +22,10 @@ Feature: PRA operations
 
     # Check that objects are transitioned in the DR site
     Given access keys for the replicated account
+
+    When the DATA_ACCESSOR user tries to perform PutObject on "DR" site
+    Then it "should not" pass Vault authentication
```
ghost: Are we sure this won't be flaky? As we just deployed DR at this stage, the "access keys for the replicated account" may complete quickly with no guarantee that the data accessor was already replicated
While having this check after the checks for the transitioned objects, we benefit from the waiting times

tests/ctst/steps/pra.ts
```diff
@@ -274,6 +275,26 @@ Then('object {string} should {string} be {string} and have the storage class {st
...
+    
+    const preName = this.getSaved<string>('accountName') ||
+        this.parameters.AccountName || Constants.ACCOUNT_NAME;
```
ghost: Code not used

```suggestion
```

tests/ctst/steps/pra.ts
```diff
@@ -274,6 +275,26 @@ Then('object {string} should {string} be {string} and have the storage class {st
...
     });
 
+When('the DATA_ACCESSOR user tries to perform PutObject on {string} site', async function (this: Zenko, site: string) {
+    if (site === 'DR') {
+        Identity.useIdentity(IdentityEnum.ACCOUNT, `${Zenko.sites['source'].accountName}-replicated`);
+    } else {
+        Identity.useIdentity(IdentityEnum.ACCOUNT, Zenko.sites['source'].accountName);
+    }
+    this.resetCommand();
+    this.addToSaved('accountName', Zenko.sites['source'].accountName);
```
ghost: shouldn't this be in the if/else above, so we use the -replicated one if we are using the data_accessor role from this account?
If we want to use this for the later tests, we should put this line at the end of the step

tests/ctst/features/pra.feature
KillianG: If drsink and drsource are running everything is replicated

tests/ctst/steps/pra.ts
KillianG: The account name in zenkos is always the same, the `-replicated` is only used in the test to determine the account we are using, but in the end the account name is the same

KillianG: that said im not sure i understand this
> If we want to use this for the later tests, we should put this line at the end of the step

tests/ctst/features/pra.feature
ghost: I mean, are we 100% sure it will be replicated at the time we do the check? if it takes some time (even several seconds) it might not be true?

KillianG: we do the check after the drsink and drsource are in running phase, which means everything is replicated, if something is not then its a bug and we want the test to fail

.github/scripts/end2end/deploy-zenko.sh
```diff
@@ -37,7 +37,8 @@ else
 fi
 
 # TODO: use kustomize
-export ZENKO_ANNOTATIONS="annotations:"
+export ZENKO_ANNOTATIONS="annotations:
```
francoisferrand: should not be added manually : it will be added automatically by the DrCtl

tests/ctst/steps/pra.ts
```diff
@@ -163,7 +164,7 @@ async function waitForPhase(
     return false;
 }
 
-const installTimeout = 360000;
+const installTimeout = 600000;
```
francoisferrand: why is it increasing?
is it there really an issue with the duration here?

ghost: I got the same issue on my build, reviewing the logs showed no error, it was just slower than 6min

tests/ctst/features/pra.feature
ghost: yes ok replicated through the initial dump, but not sure this is enough if the AK was created after (it may need time to replicated)
But didn't you get an issue with this test at first, and then it worked?

tests/ctst/steps/pra.ts
KillianG: I am testing some things for the flaky tests, this is not related to this ticket exactly, just trying to make things less flaky..

tests/ctst/features/pra.feature
KillianG: AK are created on the dr site itself, i am not sure i understand what you want to be replicated exactly ?
Never had such an issue AFAIR

ghost: Sorry I meant the role, not AK; the role will need to be replicated at the time this step runs, and having the step "running" is not enough iff the role was created after the dump, but that shouldn't be the case
We just need to be careful if later we have more resources created on the source site that needs to be replicated in the dr site

tests/ctst/steps/pra.ts
francoisferrand: > I am testing some things for the flaky tests, this is not related to this ticket exactly, just trying to make things less flaky..

please do that in another branch (even if based on top of this one) : try to keep things separate, and one ticket = one PR (esp. when "testing the CI", it creates lots of noise to keep pushing to the PR : prefer to use a branch which is not in an open PR, PR are for _review_ before _merging_, not testing...)

.github/scripts/end2end/configs/zenko.yaml
```diff
@@ -52,7 +52,7 @@ spec:
         size: 20Gi
         storageClassName: "standard"
   kafkaCleaner:
-    interval: 30s
+    interval: 2m
```
francoisferrand: is it really needed?

.github/workflows/end2end.yaml
```diff
@@ -487,6 +487,9 @@ jobs:
       - name: Configure E2E CTST test environment
         run: bash configure-e2e-ctst.sh
         working-directory: ./.github/scripts/end2end
+      - name: Debug wait
+        uses: ./.github/actions/debug-wait
+        timeout-minutes: 60
```
francoisferrand: not to be commited

solution/deps.yaml
```diff
@@ -21,7 +21,7 @@ cloudserver:
 drctl:
   sourceRegistry: ghcr.io/scality
   image: zenko-drctl
-  tag: v1.0.4
+  tag: a38d1f8848d8422317584effcbcf0fa2b1de04d9
```
francoisferrand: not to be commited

tests/ctst/steps/pra.ts
KillianG: ok

ghost (approved):
lgtm (I know we have flakies, but unrelated to these changes...)

tests/ctst/steps/cronjobs/countItems.ts
```diff
@@ -2,12 +2,30 @@ import { When } from '@cucumber/cucumber';
...
+        zenkoName += '-pra';
+    }
+
     switch (job) {
     case 'CountItems':
-        await createJobAndWaitForCompletion(this, 'end2end-ops-count-items');
+        await createJobAndWaitForCompletion(this, `${zenkoName}-ops-count-items`);
+        break;
+    case 'EnsureVaultSeeds':
+        await createJobAndWaitForCompletion(this, `${zenkoName}-ops-vault-check-seeds`);
```
francoisferrand: instead of having a new list of jobs, may be better to just specify the actual name (or suffix) of the job in argument:

```suggestion
    await createJobAndWaitForCompletion(this, `${zenkoName}-ops-${job}`);
```

and call it accordingly: `when the count-items cronjobs completes without error`

(less code, less maintenance, less mistake because we give yet another name to the same thing. the benefit of such an "enum" would be to have compile-time validation, which is not the case here...)

francoisferrand (approved):

PR #2135, opened 2024-08-28
Extend bucket notification timeout and map to the test timeout

ghost:
- Increase timeout for bucket notification tests, to match the current config (ensure the check timeouts before the step)
- Fixes an issue with kafka cleaner, when an error happens, we have no log and we clear the custer timeout, while the step timeout is -1, resulting in builds never completing (6h+), and improve the timeout.
- Add a custom timeout when checking the DMF volume, as the default 100s is too low, and we have this issue very frequently since months.

### DIFF
--- a/tests/ctst/common/common.ts
+++ b/tests/ctst/common/common.ts
@@ -237,14 +237,15 @@ Then('i {string} be able to add user metadata to object {string}',
         }
     });
 
-Then('kafka consumed messages should not take too much place on disk', { timeout: -1 },
+const kafkaCleanerTimeout = 1800000; // 30min
+Then('kafka consumed messages should not take too much place on disk', { timeout: kafkaCleanerTimeout },
     async function (this: Zenko) {
         const kfkcIntervalSeconds = parseInt(this.parameters.KafkaCleanerInterval);
         const checkInterval = kfkcIntervalSeconds * (1000 + 5000);
 
         const timeoutID = setTimeout(() => {
             assert.fail('Kafka cleaner did not clean the topics within the expected time');
-        }, checkInterval * 10); // Timeout after 10 Kafka cleaner intervals
+        }, Math.min(kafkaCleanerTimeout, checkInterval * 10)); // Timeout after 10 Kafka cleaner intervals
 
         try {
             const ignoredTopics = ['dead-letter'];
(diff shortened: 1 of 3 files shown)


tests/ctst/common/common.ts
```diff
@@ -237,14 +237,15 @@ Then('i {string} be able to add user metadata to object {string}',
         }
     });
 
-Then('kafka consumed messages should not take too much place on disk', { timeout: -1 },
+const kafkaCleanerTimeout = 1800000; // 30min
```
francoisferrand: 30 minutes seems verrrry long for kafka cleaner, which is setup to run every 30 second (or maybe 1-2 minutes) in CI...

tests/ctst/common/common.ts
```diff
@@ -307,6 +308,9 @@ Then('kafka consumed messages should not take too much place on disk', { timeout
 
             // If a topic remains in this array, it means it has not been cleaned
             assert(topics.length === 0, `Topics ${topics.join(', ')} still have not been cleaned`);
+        } catch (err: unknown) {
```
francoisferrand, line 311: is this not done in cucumber/ctst?
should it not be?

ghost: We have a "finally" that clears the timeout even if we got an error, which mean we never stop the test as its timeout was -1

ghost: Example: https://github.com/scality/Zenko/actions/runs/10668356308

tests/ctst/common/common.ts
ghost: It's already 30min today, this was the config where we got the best results

ghost: I am trying this now https://github.com/scality/Zenko/pull/2135/commits/0125893e807c72653c7605fa4f0f67178692324f

tests/ctst/common/common.ts
francoisferrand: what I mean is that (even with the catch) the exception should be propagated: and eventually reach cucumber/ctst where it shoudl be caught.... and immediately fail the test (even with no timeout)

ghost: Maybe the timeout was not due to an error... At least this ensures we log the error with a message

Errors not catched by the tests are sent to cucumber directly, it'll mark the test as failed. But in the build above, it didn't work

ghost:
Last try to reproduce the errors with fresh branches, if I cannot, I will close this PR, as the current changes are now already part of other PRs

PR #2141, opened 2024-09-02
Remove all occurences of KubdeDB as it is not used nor supported anymore

KillianG:
Remove kubedb as it is not supported nor used anymore

Went through all integration branch, there is no kubedb occurence in other branchs as well


docs/docsource/installation/install/install_xdm.rst
```diff
@@ -238,13 +235,11 @@ Deploy |product|
         version: |version|
         replicas: 1
         mongodb:
```
francoisferrand, line 252: the whole `mongodb` field is not correct: in the current model, we deploy mongodb beforehand, and add the params of the "external" mongodb in the CR

solution/zenkoversion.yaml
```diff
@@ -132,7 +132,6 @@ spec:
       exporter:
         image: redis_exporter
         tag: '${REDIS_EXPORTER_TAG}'
-      kubedb: '${REDIS_KUBEDB_TAG}'
```
francoisferrand, line 135: must also be done in the other zenkoversion.yaml

KillianG: What do you mean by "the other zenkoversion.yaml" ?

docs/docsource/installation/install/install_xdm.rst
KillianG: I am not sure I understand, should I just remove the mongodb filed ? If not, then what should it be exactly ?

francoisferrand: I mean the `mongo` field configures Zenko for a specific mongo setup.

this doc was giving the whole kubedb config : changing the "provider" does not make the configuration correct, we need to update the doc with the current way to configure mongo in zenko cr (e.g. with external provider and with the other details)

you can check you artesca, you will see that the mongo field is not the same as all.

docs/docsource/installation/install/install_xdm.rst
```diff
@@ -238,13 +235,11 @@ Deploy |product|
...
-          provider: KubeDB
           persistence:
             volumeClaimTemplate:
               size: 300Gi
               storageClassName: sc-300-g
         redis:
-          provider: KubeDB
```
francoisferrand, line 247: same for redis, it is not about removing the provider: a provider still needs to be specified, and parameter adjusted accordingly.

solution/zenkoversion.yaml
francoisferrand: there are 2 files named `zenkoversion.yaml` in the repo : the other one must be updated as well

docs/docsource/installation/install/install_xdm.rst
KillianG: I updated according to a real zenko in my platform
Needed to add a step to create the secret for mongodb credentials as it is required

francoisferrand (approved):

ghost (approved):

KillianG:
ping

KillianG:
ping

PR #2142, opened 2024-09-02
mongo node syn alert implementation

benzekrimaha:
Issue : ZENKO-4881

### DIFF
--- a/monitoring/mongodb/alerts.test.yaml
+++ b/monitoring/mongodb/alerts.test.yaml
@@ -258,5 +258,98 @@ tests:
             exp_annotations:
               description: MongoDb has low disk space
               summary: MongoDb has low disk space
+    
+
+  - name: MongoDbPodRecovering
+    interval: 1m
+    input_series:
+      - series: mongodb_rs_members_state{namespace="zenko", pod="data-db-mongodb-sharded-mongos-0", rs_state="3"}
+        values: 3x1440
+
+    alert_rule_test:
+      - alertname: MongoDbPodRecovering
+        eval_time: 1d
+        exp_alerts:
+          - exp_labels:
+              namespace: zenko
+              severity: critical
+              pod: data-db-mongodb-sharded-mongos-0
+              rs_state: 3
+            exp_annotations:
+              description: "MongoDB pod `data-db-mongodb-sharded-mongos-0` has been in the 'RECOVERING' state more than 24 hours. The instance may be failing to catch up and recover."
+              summary: MongoDB is recovering
+          - exp_labels:
+              namespace: zenko
+              pod: data-db-mongodb-sharded-mongos-0
+              rs_state: 3
+              severity: warning
+            exp_annotations:
+              description: "MongoDB pod `data-db-mongodb-sharded-mongos-0` has been in 'RECOVERING' state for more than 1 hour. This may be expected if the 'Resync a Data Services MongoDB Member' procedure has recently been executed."
+              summary: MongoDB is recovering
+
+  - name: MongoDbInvalidState
+    interval: 1m
+    input_series:
+      - series: mongodb_rs_members_state{namespace="zenko", pod="data-db-mongodb-sharded-mongos-0", rs_state="6"}
+        values: 6x10
+      - series: mongodb_rs_members_state{namespace="zenko", pod="data-db-mongodb-sharded-mongos-0", rs_state="8"}
+        values: 0x10
+      - series: mongodb_rs_members_state{namespace="zenko", pod="data-db-mongodb-sharded-mongos-0", rs_state="10"}
+        values: 0x10
+
+    alert_rule_test:
+      - alertname: MongoDbInvalidState
+        eval_time: 5m
+        exp_alerts:
+          - exp_labels:
+              namespace: zenko
+              severity: critical
+              pod: data-db-mongodb-sharded-mongos-0
+              rs_state: 6
+            exp_annotations:
+              description: "MongoDB pod `data-db-mongodb-sharded-mongos-0` is in an invalid state state (`6`)."
+              summary: "MongoDB node in an invalid state: 6 (UNKNOWN), 8 (DOWN), 10 (REMOVED)"
+
+  - name: MongoDbPodStartup2
+    interval: 1m
+    input_series:
+      - series: mongodb_rs_members_state{namespace="zenko", pod="data-db-mongodb-sharded-mongos-0", rs_state="5"}
+        values: 5x60
+
+    alert_rule_test:
+      - alertname: MongoDbPodStartup2
+        eval_time: 1h
+        exp_alerts:
+          - exp_labels:
+              namespace: zenko
+              severity: warning
+              pod: data-db-mongodb-sharded-mongos-0
+              rs_state: 5
+            exp_annotations:
+              description: "MongoDB pod `data-db-mongodb-sharded-mongos-0` has been in the 'STARTUP2' state for more than 1 hour. Please ensure that the instance is running properly."
+              summary: MongoDB node in STARTUP2 state for too long
+    
+
+  - name: MongoDbRSNotSynced
+    interval: 1m
+    input_series:
+      - series: mongodb_rs_members_state{namespace="zenko", rs_nm="rs0", pod="data-db-mongodb-sharded-mongos-0", member_state="SECONDARY"}
+        values: 0x10
+      - series: mongodb_rs_members_state{namespace="zenko", rs_nm="rs0", pod="data-db-mongodb-sharded-mongos-1", member_state="SECONDARY"}
+        values: 0x10
+      - series: mongodb_rs_members_state{namespace="zenko", rs_nm="rs0", pod="data-db-mongodb-sharded-mongos-2", member_state="SECONDARY"}
+        values: 1x10
+  
+    alert_rule_test:
+      - alertname: MongoDbRSNotSynced
+        eval_time: 10m
+        exp_alerts:
+          - exp_labels:
+              severity: warning
+              rs_nm: rs0
+            exp_annotations:
+              description: "MongoDB replica set `rs0` is not in the expected state. It currently has `1` SECONDARY members instead of the expected number. Please ensure that all instance are running properly."
+              summary: MongoDB replica set out of sync
+
 
 
--- a/monitoring/mongodb/alerts.yaml
+++ b/monitoring/mongodb/alerts.yaml
@@ -139,3 +139,54 @@ groups:
     annotations:
       description: 'MongoDb has low disk space'
       summary: 'MongoDb has low disk space'
+  
+  - alert: MongoDbPodRecovering
+    expr: |
+      mongodb_rs_members_state{namespace="${namespace}", pod=~"${service}.*"} == 3
+    for: 1h
+    labels:
+      severity: warning
+    annotations:
+      description: "MongoDB pod `{{ $labels.pod }}` has been in 'RECOVERING' state for more than 1 hour. This may be expected if the 'Resync a Data Services MongoDB Member' procedure has recently been executed."
+      summary: MongoDB is recovering
+
+  - alert: MongoDbPodRecovering
+    expr: |
+      mongodb_rs_members_state{namespace="${namespace}", pod=~"${service}.*"} == 3
+    for: 1d
+    labels:
+      severity: critical
+    annotations:
+      description: "MongoDB pod `{{ $labels.pod }}` has been in the 'RECOVERING' state more than 24 hours. The instance may be failing to catch up and recover."
+      summary: MongoDB is recovering
+
+  - alert: MongoDbInvalidState
+    expr: |
+      mongodb_rs_members_state{namespace="${namespace}", pod=~"${service}.*", rs_state=~"6|8|10"} > 0
+    for: 5m
+    labels:
+      severity: critical
+    annotations:
+      description: "MongoDB pod `{{ $labels.pod }}` is in an invalid state state (`{{ $labels.rs_state }}`)."
+      summary: "MongoDB node in an invalid state: 6 (UNKNOWN), 8 (DOWN), 10 (REMOVED)"
+
+
+  - alert: MongoDbPodStartup2
+    expr: |
+      mongodb_rs_members_state{namespace="${namespace}", pod=~"${service}.*"} == 5
+    for: 1h
+    labels:
+      severity: warning
+    annotations:
+      description: "MongoDB pod `{{ $labels.pod }}` has been in the 'STARTUP2' state for more than 1 hour. Please ensure that the instance is running properly."
+      summary: MongoDB node in STARTUP2 state for too long
+
+  - alert: MongoDbRSNotSynced
+    expr: |
+      sum by (rs_nm) (mongodb_rs_members_state{namespace="${namespace}", pod=~"${service}.*", member_state="SECONDARY"}) != (${replicas} - 1)
+    for: 10m
+    labels:
+      severity: warning
+    annotations:
+      description: "MongoDB replica set `{{ $labels.rs_nm }}` is not in the expected state. It currently has `{{ $value }}` SECONDARY members instead of the expected number. Please ensure that all instance are running properly."
+      summary: MongoDB replica set out of sync


monitoring/mongodb/alerts.yaml
```diff
@@ -139,3 +139,24 @@ groups:
     annotations:
       description: 'MongoDb has low disk space'
       summary: 'MongoDb has low disk space'
+  
+  - alert: MongoDbNodeRecovering
+    expr: |
+      sum by (pod) (mongodb_rs_members_state{namespace="${namespace}", pod=~"${service}.*", rs_state="3"} >= 1)
```
ghost: We can have all these states (excluding arbitrer)
https://www.mongodb.com/docs/manual/reference/replica-states/

I wonder if we should have alert on all the unwanted states, and not only restrict the alert on the `RECOVERING` state? Like, detecting the state is neither 0; 1 nor 2. We may also need some specific alert depending on the state:

* For 3 (RECOVERING)
  * MongoDB secondaries will periodically enter this state during high load, if they fail to keep up with the incoming data. In general, this state is not harmful.
  * But the issue arises when it stays for too long (the op log is overwritten), that's why I think the `for: 5m` should be changed to something higher, like 24h...
  * When we perform the "full init sync" procedure, the affected mongodb instance will also end up in this state, and depending on the cluster' size, it can take hours, but this is expected, in this case...
  * It's great to have a separate alert, so we can ask the reader to check the logs, and if the instance cannot recover, it will have a specific log (we document it in the troubleshooting procedure - but not sure we want to reference it from Zenko).

* For 5 (STARTUP2) is also fine, but again, if it stays for too long (> several hours), the instance may be stuck.
* For 6 (UNKNOWN), 8 (DOWN) and 10 (REMOVED), these are unwanted states, so we can alert directly without waiting.
* For 9 (ROLLBACK) this is fine as well.

monitoring/mongodb/alerts.yaml
```diff
@@ -139,3 +139,24 @@ groups:
...
+    for: 5m
+    labels:
+      severity: warning
+    annotations:
+      description: "Mongodb node in `{{ $labels.pod }}` is in the 'recovering' state for more than 10 minutes."
```
ghost: (it should be 5 minutes according to your rule)

monitoring/mongodb/alerts.yaml
```diff
@@ -139,3 +139,24 @@ groups:
...
+      summary: MongoDB node recovering for too long
+
+  - alert: MongoDbNodeNotSynced
+    expr: |
+      sum by (pod) (mongodb_mongod_replset_number_of_members{set="data-db-mongodb-sharded-shard-0"}) != ${replicas}
+    for: 10m
+    labels:
+      severity: critical
+    annotations:
+      description: "MongoDB node in `{{ $labels.pod }}` is not present in the replica set anymore."
```
ghost: No need to wait 10min for this one IMHO, this should never happen

ghost:
```suggestion
      description: "The MongoDB instance `{{ $labels.pod }}` is out of the replica set. It does not longer receive any data and must be added back to the cluster to avoid performance and storage problems."
```

Just a note: if we add new alerts, we also want to improve our product doc, so the user can quickly find that, for this example, the right procedure we suggest is to perform a full init sync.

monitoring/mongodb/alerts.yaml
ghost: I would suggest having first a warning after 1h, and then the critical alert (after 24h?).
I also think our alert descriptions should be helpful for the reader in general (someone without deep knowledge about the product/mongodb), but please challenge that if you disagree.

```suggestion
      description: "The Mongodb instance \"`{{ $labels.pod }}`\" is in the 'RECOVERING' state for over an hour. The instance may not be able to join the replica set if the platform ingests a large number of operations during this time. This alert is expected if the \"Resync a Data Services MongoDB Member\" procedure has recently been executed."
```

monitoring/mongodb/alerts.yaml
benzekrimaha: done here :https://github.com/scality/Zenko/pull/2142/commits/6bc518d4047053558ea27dc5a23907ad35493474#

monitoring/mongodb/alerts.yaml
```diff
@@ -139,3 +139,45 @@ groups:
...
       summary: 'MongoDb has low disk space'
+  
+  - alert: MongoDbNodeRecovering
+    expr: |
+      avg_over_time(mongodb_rs_members_state{namespace="${namespace}", pod=~"${service}.*", rs_state="3"}[1h]) == 3
+    for: 1h
+    labels:
+      severity: warning
+    annotations:
+      description: "The Mongodb instance `{{ $labels.pod }}` is in the 'RECOVERING' state for over an hour. The instance may not be able to join the replica set if the platform ingests a large number of operations during this time. This alert is expected if the 'Resync a Data Services MongoDB Member' procedure has recently been executed."
```
francoisferrand: > This alert is expected if the 'Resync a Data Services MongoDB Member' procedure has recently been executed.

is it really ?

we must make sure alerts will not trigger when nothing wrong is happening, to avoid alert fatigue and people either contacting support or ignoring alerts: if recovering is proceeding fine, then we should not have any alert.... it is needed only when recovering fails

monitoring/mongodb/alerts.yaml
```diff
@@ -139,3 +139,45 @@ groups:
     annotations:
       description: 'MongoDb has low disk space'
       summary: 'MongoDb has low disk space'
+  
+  - alert: MongoDbNodeRecovering
+    expr: |
+      avg_over_time(mongodb_rs_members_state{namespace="${namespace}", pod=~"${service}.*", rs_state="3"}[1h]) == 3
```
francoisferrand: what is happening when recovering fails: does the pod restart, or does it stay in recovering forever?

if the pod restarts, then the average may well be below 3...

monitoring/mongodb/alerts.yaml
```diff
@@ -139,3 +139,45 @@ groups:
...
+    for: 1h
+    labels:
+      severity: warning
+    annotations:
+      description: "The Mongodb instance `{{ $labels.pod }}` is in the 'STARTUP2' state for an hour. The instance might be stuck."
+      summary: MongoDB node in STARTUP2 state for too long
+
+  - alert: MongoDbNodeNotSynced
+    expr: |
+      sum by (pod) (mongodb_mongod_replset_number_of_members{set="data-db-mongodb-sharded-shard-0"}) != ${replicas}
```
francoisferrand: this is not precise enough : this alert would not trigger if some pods are RECOVERING...

we already have an alert when there is no PRIMARY:
`absent_over_time(mongodb_rs_members_state{namespace="${namespace}",pod=~"${service}.*",member_state="PRIMARY"}[1m]) == 1`

would it not make sense to check that we have the expected number of SECONDARY ? (i.e. `$replicas - 1`)

francoisferrand: `data-db-mongodb-sharded-shard-0` cannot and should not be hard-coded.

* we should not add a parameter, but make the alert "generic" : i.e. it would trigger for any shard (or config-svr) actually.
* filtering needs to be done, using the "exxisting" parameters (namespace, service)

monitoring/mongodb/alerts.yaml
```diff
@@ -139,3 +139,45 @@ groups:
...
+    labels:
+      severity: warning
+    annotations:
+      description: "The Mongodb instance `{{ $labels.pod }}` is in the 'RECOVERING' state for over an hour. The instance may not be able to join the replica set if the platform ingests a large number of operations during this time. This alert is expected if the 'Resync a Data Services MongoDB Member' procedure has recently been executed."
+      summary: MongoDB node recovering for too long
+
+  - alert: MongoDbInvalidState
+    expr: |
+      avg_over_time(mongodb_rs_members_state{namespace="${namespace}", pod=~"${service}.*", rs_state=~"6|8|10"}[1m]) > 0
+    for: 1m
```
francoisferrand: we can probably tolerate more than 1m, there may be some temporary switch to these states? (on startup or shutdown, maybe) → 5m ?
or are we garanteed this should not happen?

monitoring/mongodb/alerts.yaml
francoisferrand: we have a different way to check for STARTUP2 and RECOVERING states: is it expected? is there a reason not to use the same approach (or possibly alert) ?

monitoring/mongodb/alerts.yaml
benzekrimaha: Indeed we have the `livenessProbe` configured to force the pod restart otherwise it would stay in the recovering state forever

monitoring/mongodb/alerts.yaml
ghost: When we perform a full init sync, if the cluster is still processing operations, there is a risk that we never converge, if the recovering takes longer than the oplog history...
The recovering instance might be in this state for more than one hour on most of the deployed platforms today

benzekrimaha: I went for 2 alerts in order to have a more precise description, as the recovering can be expected in some cases

monitoring/mongodb/alerts.yaml
benzekrimaha: done here : https://github.com/scality/Zenko/pull/2142/commits/5e684e8f37e0033fca8570284fe1c9b2fdd714a3

benzekrimaha: done here : https://github.com/scality/Zenko/pull/2142/commits/5e684e8f37e0033fca8570284fe1c9b2fdd714a3

monitoring/mongodb/alerts.yaml
benzekrimaha: done here : https://github.com/scality/Zenko/pull/2142/commits/5e684e8f37e0033fca8570284fe1c9b2fdd714a3

monitoring/mongodb/alerts.yaml
benzekrimaha: I actually went for a warning if during the last hour it has been in the recovering state for at least once, if we want to have an alert only when it's stuck in that state , we may point that in the MongoDbNodeNotSynced or the alert for the primary as the fact that the node is not present might be due to the fact that it's stuck

monitoring/mongodb/alerts.yaml
```diff
@@ -139,3 +139,45 @@ groups:
...
+
+  - alert: MongoDbNodeNotSynced
+    expr: |
+      sum by (pod) (mongodb_rs_members_state{namespace="${namespace}", pod=~"${service}.*", member_state="SECONDARY"}) != (${replicas} - 1)
+    for: 1m
+    labels:
+      severity: critical
+    annotations:
+      description: "The MongoDB instance `{{ $labels.pod }}` is out of the replica set. It does no longer receive any data and must be added back to the cluster to avoid performance and storage problems."
+      summary: MongoDB node not in replica set
```
francoisferrand: * the `pod` would be different for each time serie (it is `${service}.*`), so the `sum` not compare to ${replicas}. To do this aggregation, we need to aggregate on a label which is hase a different value each "group" of pods (mongo-shard$X, mongo-cfgsvr) but also the same value for every pod in the group
* since we are doing an alert on the aggregate, we will not have a single "pod", and cannot point to the node in the bad state.

monitoring/mongodb/alerts.yaml
```diff
@@ -139,3 +139,45 @@ groups:
...
+
+
+  - alert: MongoDbNodeStartup2
+    expr: |
+      avg_over_time(mongodb_rs_members_state{namespace="${namespace}", pod=~"${service}.*", rs_state="5"}[1h]) == 5
+    for: 1h
+    labels:
+      severity: warning
+    annotations:
+      description: "The Mongodb instance `{{ $labels.pod }}` is in the 'STARTUP2' state for an hour. The instance might be stuck."
```
francoisferrand: `instance` can be misleading, usually we (esp. customers) would think of "instance" as the whole mongodb cluster... → prefer something less ambiguous, like `pod`

```suggestion
      description: "Mongodb pod `{{ $labels.pod }}` is in the 'STARTUP2' state for an hour. The instance might be stuck."
```

francoisferrand: (same for the other alerts)

monitoring/mongodb/alerts.yaml
benzekrimaha: done here : https://github.com/scality/Zenko/pull/2142/commits/3d767a04f769852075e490521342664328265294#diff-237b96ceb855bab52e3f81d4a49e5e17e07fc7c1102d3dec5b30e461fee6fc69R176

monitoring/mongodb/alerts.yaml
benzekrimaha: done here : https://github.com/scality/Zenko/pull/2142/commits/3d767a04f769852075e490521342664328265294#diff-237b96ceb855bab52e3f81d4a49e5e17e07fc7c1102d3dec5b30e461fee6fc69R176

monitoring/mongodb/alerts.yaml
```diff
@@ -139,3 +139,45 @@ groups:
...
+      summary: MongoDB node recovering for too long
+
+  - alert: MongoDbInvalidState
+    expr: |
+      avg_over_time(mongodb_rs_members_state{namespace="${namespace}", pod=~"${service}.*", rs_state=~"6|8|10"}[1m]) > 0
+    for: 5m
+    labels:
+      severity: critical
+    annotations:
+      description: "Mongodb pod `{{ $labels.pod }}` is in the `{{ $labels.rs_state }}` state. Immediate attention is required."
```
ghost: I think we should indicate to the reader what to do in this case:

```suggestion
      description: "MongoDB pod `{{ $labels.pod }}` is in the `{{ $labels.rs_state }}` state. Immediate attention is required. You may collect an sosreport for further analysis."
```

monitoring/mongodb/alerts.yaml
```diff
@@ -139,3 +139,45 @@ groups:
...
+
+
+  - alert: MongoDbNodeStartup2
+    expr: |
+      avg_over_time(mongodb_rs_members_state{namespace="${namespace}", pod=~"${service}.*", rs_state="5"}[1h]) == 5
+    for: 1h
+    labels:
+      severity: warning
+    annotations:
+      description: "Mongodb pod `{{ $labels.pod }}` is in the 'STARTUP2' state for an hour. The instance might be stuck."
```
ghost:
```suggestion
      description: "MongoDB pod `{{ $labels.pod }}` has been in the 'STARTUP2' state for an hour. Please ensure that the instance is running properly. You may follow the 'Resync a Data Services MongoDB Member' procedure to unblock this MongoDB instance instance."
```

monitoring/mongodb/alerts.yaml
```diff
@@ -139,3 +139,45 @@ groups:
     annotations:
       description: 'MongoDb has low disk space'
       summary: 'MongoDb has low disk space'
+  
+  - alert: MongoDbNodeRecovering
```
ghost: Node in a k8s/artesca context means a logical server, where we can have multiple MongoDB instances running (e.g. the mongod, mongos)

```suggestion
  - alert: MongoDBInstanceRecovering
```

(maybe we want to use "Pod" instead of "Instance" as it also has a meaning in mongodb)

monitoring/mongodb/alerts.yaml
```diff
@@ -139,3 +139,45 @@ groups:
...
+    expr: |
+      count_over_time(mongodb_rs_members_state{namespace="${namespace}", pod=~"${service}.*", rs_state="3"}[1h]) > 0
+    for: 1h
+    labels:
+      severity: warning
+    annotations:
+      description: "Mongodb pod `{{ $labels.pod }}` has been in the 'RECOVERING' state at least once in the past hour. The instance may not be able to join the replica set if the platform ingests a large number of operations during this time. This alert is expected if the 'Resync a Data Services MongoDB Member' procedure has recently been executed."
```
ghost:
```suggestion
      description: "MongoDB pod `{{ $labels.pod }}` has been in the 'RECOVERING' state at least once in the past hour. The instance may not be able to join the replica set if the platform ingests a large number of operations during this time. This alert is expected if the 'Resync a Data Services MongoDB Member' procedure has recently been executed."
```

monitoring/mongodb/alerts.yaml
ghost:
```suggestion
      description: "MongoDB pod `{{ $labels.pod }}` is in the `{{ $labels.rs_state }}` state. Immediate attention is required."
```

(applies to everything, let's follow the real naming of MongoDB)

monitoring/mongodb/alerts.yaml
```diff
@@ -139,3 +139,45 @@ groups:
     annotations:
       description: 'MongoDb has low disk space'
       summary: 'MongoDb has low disk space'
+  
+  - alert: MongoDbNodeRecovering
+    expr: |
+      count_over_time(mongodb_rs_members_state{namespace="${namespace}", pod=~"${service}.*", rs_state="3"}[1h]) > 0
+    for: 1h
```
ghost, line 146: Do we want to have the same alert after 1 day but with `critical` severity?

monitoring/mongodb/alerts.yaml
```diff
@@ -139,3 +139,45 @@ groups:
     annotations:
       description: 'MongoDb has low disk space'
       summary: 'MongoDb has low disk space'
+  
+  - alert: MongoDbNodeRecovering
+    expr: |
+      count_over_time(mongodb_rs_members_state{namespace="${namespace}", pod=~"${service}.*", rs_state="3"}[1h]) > 0
```
ghost: Do we really want to detect that only one metric was reporting the "RECOVERING" state in the past hour, and not that the pod was continuously in this state? The potential problems arise when we have this state during the entire last hour (not just once), otherwise I fear that the alert will be triggered very frequently even if there is nothing wrong...

(in this case the description must be updated)

monitoring/mongodb/alerts.yaml
ghost: Also, according to the alert, we can have it if the mongodb instance was briefly in this state repetitively (at least once per minute) over the 5 last minutes, so if we have the alert and we see the right state, it may be misleading:

```suggestion
      description: "MongoDB pod `{{ $labels.pod }}` is repeatedly or continuously in the `{{ $labels.rs_state }}` state. Immediate attention is required. You may collect an sosreport for further analysis."
```

monitoring/mongodb/alerts.yaml
```diff
@@ -139,3 +139,45 @@ groups:
...
+
+  - alert: MongoDbNodeNotSynced
+    expr: |
+      sum by (rs_nm) (mongodb_rs_members_state{namespace="${namespace}", pod=~"${service}.*", member_state="SECONDARY"}) != (${replicas} - 1)
+    for: 1m
+    labels:
+      severity: critical
+    annotations:
+      description: "Mongodb replica set `{{ $labels.rs_nm }}` is out of sync. One or more pods are not receiving data and must be added back to the cluster to avoid performance and storage problems."
+      summary: MongoDb replica set out of sync
```
ghost: Won't this alert always fire if we have one replica that is recovering/other? Here, we fire the alert if during the last minute, we do not see e.g. 2 SECONDAY on a 3-node platform. However, it is totally acceptable to have STARTUP2 or RECOVERING periodically (during high loads, for example). We already detect these cases in the above alerts.

I think what we want with this alert is a longer check duration (it is less sensitive than the primary one, that blocks everything). And then we explain in the description that the instance is missing N MongoDB secondaries, meaning that in case of primary loss we might have metadata loss.

What remains is the duplication of alerts; if one pod is RECOVERING for let's say 5min, we will have both this alert, and later, the MongoDbNodeRecovering one...

monitoring/mongodb/alerts.yaml
benzekrimaha: if it stays for too long (1 day) I think it's appropriate to have it as a critical alert , I believe this is handled by the alert manager

monitoring/mongodb/alerts.yaml
benzekrimaha: when recovering fails  the pod restarts ( as the livenessProbe is configured ) so the average I went for initially will be below 3, I actually went for a warning if during the last hour it has been in the recovering state for at least once, but if we want to have an alert only when it's stuck in that state , we may point that in the `MongoDbRSNotSynced` or the alert for the primary as the fact that the node is not present might be due to the fact that it's stuck

monitoring/mongodb/alerts.yaml
benzekrimaha: Indeed , but I think that if we increase the evaluation duration to 1h we will have the warning ( in case of a recovery or other unusual states ) after 1h and we will have the critical `MongoDbNodeNotSynced` , IMHO keeping both is okay what do you think ? cc : @francoisferrand

monitoring/mongodb/alerts.yaml
ghost: Yes, for that you just need to recreate another alert with the same name but another severity, alertmanager will only display the one with the higher severity, so if we have the critical, it won't duplicate what is showed to the user

monitoring/mongodb/alerts.yaml
ghost: I think the `MongoDbRSNotSynced` alert would make more sense yes, if we cannot make this one less sensitive to recovering states (that can arise at any time if the secondary instance is a bit behind)

monitoring/mongodb/alerts.yaml
ghost: I think 10m is acceptable, as long as we are alerted before the oplog has any risk of being overriden, we are safe (1h can be a bit too long for some platforms today...)
A mongodb instance that is recovering due to usual operations won't stay in this state for too long.

monitoring/mongodb/alerts.yaml
benzekrimaha: After checking with @williamlardier  the node actually stays stuck as it's responding to the probe , so it is not restarted so the average will be 3

monitoring/mongodb/alerts.test.yaml
```diff
@@ -258,5 +258,88 @@ tests:
...
+        eval_time: 1h
+        exp_alerts:
+          - exp_labels:
+              namespace: zenko
+              severity: warning
+              pod: data-db-mongodb-sharded-mongos-0
+              rs_state: 3
+            exp_annotations:
+              description: "MongoDB pod `data-db-mongodb-sharded-mongos-0` has been stuck in the 'RECOVERING' state for the past hour. The instance may not be able to join the replica set if the platform ingests a large number of operations during this time. This alert is expected if the 'Resync a Data Services MongoDB Member' procedure has recently been executed."
+              summary: MongoDB node recovering for too long
```
ghost: Now that we have the new severity, we can test it here

benzekrimaha: The test now is for both the warning and critical alerts https://github.com/scality/Zenko/pull/2142/commits/3adf378e693d439cd90e3fca3291eef1f0c3cee2#diff-94c90083eba829343ee5745d0056d887a02d888847ea33af739a3ef5c3ae92c6R275-R287

monitoring/mongodb/alerts.yaml
ghost: (when the recovering fails, this is another issue IMHO, in general a recovering instance might run for hours without issue; when the recovering fails, we would have other alerts anyway, to notify that not all mongo are up)

monitoring/mongodb/alerts.yaml
```diff
@@ -139,3 +139,55 @@ groups:
...
+      summary: MongoDB node recovering for too long
+
+  - alert: MongoDbPodRecovering
+    expr: |
+      avg_over_time(mongodb_rs_members_state{namespace="${namespace}", pod=~"${service}.*", rs_state="3"}[1d]) == 3
+    for: 1d
+    labels:
+      severity: critical
+    annotations:
+      description: "MongoDB pod `{{ $labels.pod }}` has been stuck in the 'RECOVERING' state for the past day. The instance may not be able to join the replica set if the platform ingests a large number of operations during this time. This alert is expected if the 'Resync a Data Services MongoDB Member' procedure has recently been executed."
```
ghost: For the 1d I think we should remove the end, as 1d really is a lot (we do not guarantee 3 days of oplog history today)

```suggestion
      description: "MongoDB pod `{{ $labels.pod }}` has been stuck in the 'RECOVERING' state for the past day. The instance may not be able to join the replica set if the platform ingests a large number of operations during this time.
```

monitoring/mongodb/alerts.yaml
```diff
@@ -139,3 +139,55 @@ groups:
...
+      summary: MongoDB node in STARTUP2 state for too long
+
+  - alert: MongoDbRSNotSynced
+    expr: |
+      sum by (rs_nm) (mongodb_rs_members_state{namespace="${namespace}", pod=~"${service}.*", member_state="SECONDARY"}) != (${replicas} - 1)
+    for: 10m
+    labels:
+      severity: critical
+    annotations:
+      description: "MongoDB replica set `{{ $labels.rs_nm }}` is missing `{{ $value }}` SECONDARY members for at least 1 hour. This means that in case of primary loss, there is a risk of metadata loss. Immediate action is required to ensure data integrity."
```
ghost: Now that we decided to have an alert for the last 10min, the metadata loss is maybe not appropriate...

So just to not stress the user too much (I also updated the duration in the description):

```suggestion
      description: "MongoDB replica set `{{ $labels.rs_nm }}` is missing `{{ $value }}` SECONDARY members for at least 10 minutes. If the member is not recovering, you may collect an sosreport for further analysis."
```

In such case, the severity can be `warning`.

Now, I think it may make sense to add another alert: if the number of secondary is 0 AND the expected number is >= 1 (to not have this alert for single-node platforms), then we are in a situation where we can really have metadataloss if the primary is lost... Something like:

```
  - alert: MongoDbRSNotProtected
    expr: |
      sum by (rs_nm) (mongodb_rs_members_state{namespace="${namespace}", pod=~"${service}.*", member_state="SECONDARY"}) == 0
      and (${replicas} > 1)
    for: 10m
    labels:
      severity: critical
    annotations:
      description: "MongoDB replica set `{{ $labels.rs_nm }}` is missing all its SECONDARY members for at least 10 minutes. This means that in case of primary loss, there is a risk of metadata loss. Immediate action is required to ensure data integrity."
      summary: MongoDB replica set is not protected
```

benzekrimaha: https://github.com/scality/Zenko/pull/2142/commits/0ec53245375a1574c5b51e7a08bdaa58d8b395d6#diff-237b96ceb855bab52e3f81d4a49e5e17e07fc7c1102d3dec5b30e461fee6fc69R194-R203

ghost (approved):

monitoring/mongodb/alerts.yaml
```diff
@@ -139,3 +139,66 @@ groups:
     annotations:
       description: 'MongoDb has low disk space'
       summary: 'MongoDb has low disk space'
+  
+  - alert: MongoDbPodRecovering
+    expr: |
+      avg_over_time(mongodb_rs_members_state{namespace="${namespace}", pod=~"${service}.*", rs_state="3"}[1h]) == 3
+    for: 1h
```
francoisferrand: this alert triggers after 2 hours : the expression becomes true after 1h, then we "ignore" the error for 1h.

average over time can only match if the value did not change (and not sure how it handles "missing" datapoints), and make computation slower...

The following will do the same more efficiently, just looking at the current state every minute or so ; and alert manager will actually fire the alert only if every point in the `for` duration (1h) matches the alert : which is exactly what you wanted to do, right?

```suggestion
      mongodb_rs_members_state{namespace="${namespace}", pod=~"${service}.*"} == 3
    for: 1h
```

monitoring/mongodb/alerts.yaml
```diff
@@ -139,3 +139,66 @@ groups:
...
+    labels:
+      severity: warning
+    annotations:
+      description: "MongoDB pod `{{ $labels.pod }}` has been stuck in the 'RECOVERING' state for the past hour. The instance may not be able to join the replica set if the platform ingests a large number of operations during this time. This alert is expected if the 'Resync a Data Services MongoDB Member' procedure has recently been executed."
+      summary: MongoDB node recovering for too long
+
+  - alert: MongoDbPodRecovering
+    expr: |
+      avg_over_time(mongodb_rs_members_state{namespace="${namespace}", pod=~"${service}.*", rs_state="3"}[1d]) == 3
+    for: 1d
```
francoisferrand: this will require 2 d to trigger, and require lots of resources to evaluate.

this is especially critical here, since a query over a 1d period will be very slow...

```suggestion
      mongodb_rs_members_state{namespace="${namespace}", pod=~"${service}.*"} == 3
    for: 1d
```

monitoring/mongodb/alerts.yaml
```diff
@@ -139,3 +139,66 @@ groups:
...
+    labels:
+      severity: critical
+    annotations:
+      description: "MongoDB pod `{{ $labels.pod }}` has been stuck in the 'RECOVERING' state for the past day. The instance may not be able to join the replica set if the platform ingests a large number of operations during this time."
+      summary: MongoDB node recovering for too long
+
+  - alert: MongoDbInvalidState
+    expr: |
+      avg_over_time(mongodb_rs_members_state{namespace="${namespace}", pod=~"${service}.*", rs_state=~"6|8|10"}[1m]) > 0
```
francoisferrand: same, why do you want to average over time?
(esp average over 1m will average one or two points...)

```suggestion
      mongodb_rs_members_state{namespace="${namespace}", pod=~"${service}.*", rs_state=~"6|8|10"} > 0
```

monitoring/mongodb/alerts.yaml
```diff
@@ -139,3 +139,66 @@ groups:
...
+    labels:
+      severity: critical
+    annotations:
+      description: "MongoDB pod `{{ $labels.pod }}` is repeatedly or continuously in the `{{ $labels.rs_state }}` state. Immediate attention is required. You may collect an sosreport for further analysis."
+      summary: "MongoDB node in an invalid state: 6 (UNKNOWN), 8 (DOWN), 10 (REMOVED)"
+
+
+  - alert: MongoDbPodStartup2
+    expr: |
+      avg_over_time(mongodb_rs_members_state{namespace="${namespace}", pod=~"${service}.*", rs_state="5"}[1h]) == 5
```
francoisferrand: same

monitoring/mongodb/alerts.yaml
```diff
@@ -139,3 +139,66 @@ groups:
...
+      summary: MongoDB node in STARTUP2 state for too long
+
+  - alert: MongoDbRSNotSynced
+    expr: |
+      sum by (rs_nm) (mongodb_rs_members_state{namespace="${namespace}", pod=~"${service}.*", member_state="SECONDARY"}) != (${replicas} - 1)
+    for: 10m
+    labels:
+      severity: warning
+    annotations:
+      description: "MongoDB replica set `{{ $labels.rs_nm }}` is missing `{{ $value }}` SECONDARY members for the last 10 minutes. If the member is not recovering, you may collect an sosreport for further analysis."
```
francoisferrand: it is not "_is missing `{{ $value }}` SECONDARY members_" : it currently has `{{ $value }}` secondary members, instead of the expected `${replicas} - 1`

overall, I don't think the value is that important : what is important is really that the replica set is not in the expected state (we also have another alert where we check that there is a PRIMARY for each rs, right?)

monitoring/mongodb/alerts.yaml
```diff
@@ -139,3 +139,66 @@ groups:
...
+      summary: MongoDB replica set out of sync
+    
+  - alert: MongoDbRSNotProtected
+    expr: |
+      sum by (rs_nm) (mongodb_rs_members_state{namespace="${namespace}", pod=~"${service}.*", member_state="SECONDARY"}) == 0
+      and on() (vector(${replicas}) > 1)
+    for: 10m
+    labels:
+      severity: critical
```
francoisferrand: not sure this extra level is useful.

In order of criticality:
* having no primary is critical
* not having a quorum is critical, but less. However, it is not so trivial to define when that happens in general case: esp. at 0, we are sure we don't have it... but it may be the case before we reach 0 ; and if there is no quorum this should raise other alerts (because there should be no primary)
* if we have the quorum, missing secondaries is such a warning that we are "less protected"

benzekrimaha: The purpose of this alert is to flag a situation where all secondaries are missing, which, while not immediately catastrophic, significantly increases the risk of data loss in case of primary failure. It complements other alerts, such as quorum loss or primary failure, which would trigger more immediate critical alerts.
We could adjust the severity of this alert to a warning to better reflect the situation's urgency, while keeping the focus on redundancy and resilience of the replica set.

monitoring/mongodb/alerts.yaml
```diff
@@ -139,3 +139,68 @@ groups:
...
+  
+  - alert: MongoDbPodRecovering
+    expr: |
+      mongodb_rs_members_state{namespace="${namespace}", pod=~"${service}.*"} == 3
+    for: 1h
+    labels:
+      severity: warning
+    annotations:
+      description: "MongoDB pod `{{ $labels.pod }}` has been stuck in the 'RECOVERING' state for the past hour. The instance may not be able to join the replica set if the platform ingests a large number of operations during this time. This alert is expected if the 'Resync a Data Services MongoDB Member' procedure has recently been executed."
+      summary: MongoDB node recovering for too long
```
francoisferrand: this is just informative:

```suggestion
      summary: MongoDB is recovering
```

monitoring/mongodb/alerts.yaml
```diff
@@ -139,3 +139,68 @@ groups:
...
       summary: 'MongoDb has low disk space'
+  
+  - alert: MongoDbPodRecovering
+    expr: |
+      mongodb_rs_members_state{namespace="${namespace}", pod=~"${service}.*"} == 3
+    for: 1h
+    labels:
+      severity: warning
+    annotations:
+      description: "MongoDB pod `{{ $labels.pod }}` has been stuck in the 'RECOVERING' state for the past hour. The instance may not be able to join the replica set if the platform ingests a large number of operations during this time. This alert is expected if the 'Resync a Data Services MongoDB Member' procedure has recently been executed."
```
francoisferrand: > The instance may not be able to join the replica set if the platform ingests a large number of operations during this time.

this is alarmist, it can happen but not the most important.
the real issue is that redondancy is reduced.

```suggestion
      description: "MongoDB pod `{{ $labels.pod }}` has been in 'RECOVERING' state for more than 1 hour. This may be expected if the 'Resync a Data Services MongoDB Member' procedure has recently been executed."
```

monitoring/mongodb/alerts.yaml
```diff
@@ -139,3 +139,68 @@ groups:
...
+      summary: MongoDB node recovering for too long
+
+  - alert: MongoDbPodRecovering
+    expr: |
+      mongodb_rs_members_state{namespace="${namespace}", pod=~"${service}.*"} == 3
+    for: 1d
+    labels:
+      severity: critical
+    annotations:
+      description: "MongoDB pod `{{ $labels.pod }}` has been stuck in the 'RECOVERING' state for the past day. The instance may not be able to join the replica set if the platform ingests a large number of operations during this time."
```
francoisferrand:
```suggestion
      description: "MongoDB pod `{{ $labels.pod }}` has been in the 'RECOVERING' state more than 24 hours. The instance may be failing to catch up and recover."
```

monitoring/mongodb/alerts.yaml
```diff
@@ -139,3 +139,68 @@ groups:
...
+      summary: MongoDB node recovering for too long
+
+  - alert: MongoDbInvalidState
+    expr: |
+      mongodb_rs_members_state{namespace="${namespace}", pod=~"${service}.*", rs_state=~"6|8|10"} > 0
+    for: 5m
+    labels:
+      severity: critical
+    annotations:
+      description: "MongoDB pod `{{ $labels.pod }}` is repeatedly or continuously in the `{{ $labels.rs_state }}` state. Immediate attention is required. You may collect an sosreport for further analysis."
```
francoisferrand: > Immediate attention is required
not useful if we don't provide a procedure

> You may collect an sosreport for further analysis
not really useful for the user, only for us if they open a ticket... but filling sosreport is part of the reporting process already

I think this should stay factual, and not try to deduce more than we can. If there is a will to provide more bits and pieces of the support documentation, should be a product-wide effort to update all alerts accross the product, with Techpub intervention.

```suggestion
      description: "MongoDB pod `{{ $labels.pod }}` is in an invalid state state (`{{ $labels.rs_state }}`)."
```

monitoring/mongodb/alerts.yaml
```diff
@@ -139,3 +139,68 @@ groups:
...
+
+
+  - alert: MongoDbPodStartup2
+    expr: |
+      mongodb_rs_members_state{namespace="${namespace}", pod=~"${service}.*"} == 5
+    for: 1h
+    labels:
+      severity: warning
+    annotations:
+      description: "MongoDB pod `{{ $labels.pod }}` has been in the 'STARTUP2' state for an hour. Please ensure that the instance is running properly. You may follow the 'Resync a Data Services MongoDB Member' procedure to unblock this MongoDB instance instance."
```
francoisferrand:
```suggestion
      description: "MongoDB pod `{{ $labels.pod }}` has been in the 'STARTUP2' state for more than 1 hour. Please ensure that the instance is running properly."
```

monitoring/mongodb/alerts.yaml
```diff
@@ -139,3 +139,68 @@ groups:
...
+      summary: MongoDB node in STARTUP2 state for too long
+
+  - alert: MongoDbRSNotSynced
+    expr: |
+      sum by (rs_nm) (mongodb_rs_members_state{namespace="${namespace}", pod=~"${service}.*", member_state="SECONDARY"}) != (${replicas} - 1)
+    for: 10m
+    labels:
+      severity: warning
+    annotations:
+      description: "MongoDB replica set `{{ $labels.rs_nm }}` is not in the expected state. It currently has `{{ $value }}` SECONDARY members instead of the expected number. Immediate action may be required to ensure data integrity."
```
francoisferrand:
```suggestion
      description: "MongoDB replica set `{{ $labels.rs_nm }}` is not in the expected state. It currently has `{{ $value }}` SECONDARY members instead of the expected number. Please ensure that all instance are running properly."
```

monitoring/mongodb/alerts.yaml
francoisferrand: it does not create more risk of data loss : if there is no quorum, write is not possible anymore, and it is only an issue of availability.

the redundancy risk is already covered by the previous alert, and the availability is covered by the alert on missing primary.

ghost: > it does not create more risk of data loss : if there is no quorum, write is not possible anymore, and it is only an issue of availability.

Didn't we got cases where some secondaries were out of the replicaset configuration for some reason? In this case, we still have the quorum...

monitoring/mongodb/alerts.yaml
ghost: Given the alert rule, I think we should really say "repeatedly or continuously": we can be in the correct state when looking at the instance, yet the alert is still firing. This could be a symptom of a pod that is restarting in loop. Or we refine the alert rule to reflect this error if we are in this state continuously

monitoring/mongodb/alerts.yaml
ghost: Not sure this is actionable for the client: how to ensure that the instance are running properly?

monitoring/mongodb/alerts.yaml
benzekrimaha: yes we do have an alert for the primary

monitoring/mongodb/alerts.yaml
francoisferrand: the rule will be evaluated every 30s/1m, and it must match every evaluation for at least 5 minutes : if we are unlucky enough that it matches every time prometheus evaluates it, then the user will also most likely see it....

monitoring/mongodb/alerts.yaml
francoisferrand: yes, and we will see it with the "warning" alert above, not this one.

monitoring/mongodb/alerts.yaml
francoisferrand: that is what is provided in just about every alert.

it does not provide an answer (since we don't know what is happening, only that something is wrong), but will let people know what look for on the platform or in the troubleshooting procedure, maybe check mongo documentation...or eventually reach out to support.

francoisferrand (approved):

PR #2144, opened 2024-09-04
Failover state before initiating restores in pra test

ghost:
Switch to failover before restoring object on the DR site.

### DIFF
--- a/tests/ctst/features/pra.feature
+++ b/tests/ctst/features/pra.feature
@@ -41,10 +41,18 @@ Feature: PRA operations
     And the DATA_ACCESSOR user tries to perform PutObject on "DR" site
     Then it "should not" pass Vault authentication
 
+    # Switch to failover
+    When I request the failover state for the DR
+    Then the DR sink should be in phase "Failover"
+
     # Restore on DR site
     When i restore object "obj2-1" for 2 days on "DR" site
     Then object "obj2-1" should "" be "restored" and have the storage class "e2e-cold" on "DR" site
     And object "obj2-1" should "" be "transitioned" and have the storage class "e2e-cold" on "Primary" site
+
+    # Switch to failback
+    When I resume operations for the DR
+    Then the DR sink should be in phase "Running"
 
     # Pause / Resume DR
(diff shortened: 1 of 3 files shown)


tests/ctst/steps/pra.ts
```diff
@@ -291,6 +291,22 @@ Then('the kafka DR volume exists', { timeout: 60000 }, async function (this: Zen
     assert(volumeParsed.result!['volume phase'] === 'Bound');
 });
 
+When ('I request the failover state for the DR', { timeout: 360000 }, async function (this: Zenko) {
+    await this.zenkoDrCtl?.failover({
+        sinkZenkoNamespace: 'default',
+        wait: true,
+        timeout: '6m',
```
KillianG: Should'nt we make the timeout a bit lower in the command than in the timeout for ctst ? so that we see the command fail before the step and we can have more logs..

ghost: I'll update all functions yes

ghost: Done -> https://github.com/scality/Zenko/pull/2144/commits/9208878fe2b628b1a9546ed9d7cffbce4414e650

KillianG (approved):

tests/ctst/features/pra.feature
```diff
@@ -35,11 +35,19 @@ Feature: PRA operations
...
+    Then the DR sink should be in phase "Failover"
+
     # Restore on DR site
     When i restore object "obj2-1" for 2 days on "DR" site
     Then object "obj2-1" should "" be "restored" and have the storage class "e2e-cold" on "DR" site
     And object "obj2-1" should "" be "transitioned" and have the storage class "e2e-cold" on "Primary" site
 
+    # Switch to failback
+    When I resume operations for the DR
+    Then the DR sink should be in phase "Running"
```
francoisferrand, line 55: we should test that the object is expired (on DR site) during failback (may need to increase the restore duration to ensure we don't succeed thanks to race conditions)

francoisferrand (approved):

PR #2152, opened 2024-09-20
Cleanup created accounts and buckets after tests

ghost:
Without cleanup, the GetRolesForWebIdentity API ends up being very slow, eventually causing Vault to be unresponsive and thus causing timeouts, and tests failures. This API is not supposed to work with more than 100 accounts (as per requirements), and here, we end up with >1000, meaning at least this number of internal (and sequential) DB calls per "GetRolesForWebIdentity" API call.

The logic detects that we created an account (during cross-account tests), and delete all resources, before deleting the account.

Issue: ZENKO-4898


.github/actions/archive-artifacts/action.yaml
```diff
@@ -42,3 +42,22 @@ runs:
...
+        set -exu
+
+        ZENKO_MONGODB_DATABASE="${ZENKO_MONGODB_DATABASE:-zenko-database}"
+        MONGODB_ROOT_USERNAME="${MONGODB_ROOT_USERNAME:-root}"
+        MONGODB_ROOT_PASSWORD="${MONGODB_ROOT_PASSWORD:-rootpass}"
+        NAMESPACE="${NAMESPACE:-default}"
+
+        MONGODB_COLLECTIONS=$(kubectl exec -n ${NAMESPACE} data-db-mongodb-sharded-mongos-0 -- mongo ${ZENKO_MONGODB_DATABASE} -u ${MONGODB_ROOT_USERNAME} -p ${MONGODB_ROOT_PASSWORD} --authenticationDatabase admin --quiet --eval "db.getCollectionNames().join(' ')")
+
+        for collection in ${MONGODB_COLLECTIONS}; do
```
ghost: do we want to keep only the special collections?
For now, I see ~400 collections kept (due to object lock or test failing), they all have less than 10 objects anyway

tests/ctst/common/hooks.ts
```diff
@@ -37,8 +49,129 @@ Before({ tags: '@Quotas', timeout: 1200000 }, async function (scenarioOptions) {
     await prepareQuotaScenarios(this as Zenko, scenarioOptions);
 });
 
+After(async function (this: Zenko, results) {
+    if (results.result?.status === 'FAILED') {
+        return;
```
francoisferrand, line 49: can we add a log here, so it is easy to see that something was not cleand up, and to make an explicit link between a specific test and what bucket/account/... were used (and retained in DB)

tests/ctst/common/hooks.ts
```diff
@@ -37,8 +49,129 @@ Before({ tags: '@Quotas', timeout: 1200000 }, async function (scenarioOptions) {
...
+        this,
+        this.getSaved<string>('bucketName'),
+    );
+});
+
 After({ tags: '@Quotas' }, async function () {
     await teardownQuotaScenarios(this as Zenko);
 });
 
+After(async function (this: Zenko, results) {
```
francoisferrand: this seems mostly related to the iam tests: and we actually don't do anything if there is no saved `crossAccountName`...
- similar to `teardownQuotaScenarios`, would be better to move this cleanup logic to the associated test suite (even if we need to keep the `After` here) : to have better "locality" of the test & associated cleanup code
- since it is specific to these tests, maybe we should use a "tag filter" (like what is done for quotas) to better express this should happen only for these specific tests? (instead of just relying on `crossAccountName`). It would be more explicit.

tests/ctst/world/Zenko.ts
```diff
@@ -943,6 +951,16 @@ export default class Zenko extends World<ZenkoWorldParameters> {
...
             `/config/${this.parameters.InstanceID}/location/${locationName}`);
     }
+
+    saveCreatedObject(objectName: string, versionId: string) {
+        const createdObjects = this.getSaved<Map<string, string>>('createdObjects') || new Map<string, string>();
+        createdObjects.set(objectName, versionId);
+        this.addToSaved('createdObjects', createdObjects);
+    }
+
+    getCreatedObjects() {
```
francoisferrand, line 962: it seems often (or always?) use this to then get an object: `this.getCreatedObjects()?.get(objName)`
so we probably should have a `getCreatedObject(objectName string)` function instead/in addition.

tests/ctst/steps/utils/utils.ts
```diff
@@ -197,17 +197,18 @@ async function createBucketWithConfiguration(
...
-    objectNameArray.push(world.getSaved<string>('objectName'));
-    world.addToSaved('objectNameArray', objectNameArray);
     await uploadSetup(world, 'PutObject');
     world.addCommandParameter({ key: world.getSaved<string>('objectName') });
     world.addCommandParameter({ bucket: world.getSaved<string>('bucketName') });
     const result = await S3.putObject(world.getCommandParameters());
+    const versionId = extractPropertyFromResults<string>(result, 'VersionId');
+    world.saveCreatedObject(world.getSaved<string>('objectName'), versionId || '');
+
     world.addToSaved('versionId', extractPropertyFromResults(
```
francoisferrand: saving `versionId` seems redundant with `world.saveCreatedObject` ; even if we keep it to avoid too wide-ranging changes to the tests, should it not be done inside `saveCreatedObject` ?

tests/ctst/steps/utils/utils.ts
```diff
@@ -197,17 +197,18 @@ async function createBucketWithConfiguration(
...
     world.addToSaved('objectName', objectName || Utils.randomString());
-    const objectNameArray = world.getSaved<string[]>('objectNameArray') || [];
-    objectNameArray.push(world.getSaved<string>('objectName'));
-    world.addToSaved('objectNameArray', objectNameArray);
     await uploadSetup(world, 'PutObject');
     world.addCommandParameter({ key: world.getSaved<string>('objectName') });
     world.addCommandParameter({ bucket: world.getSaved<string>('bucketName') });
     const result = await S3.putObject(world.getCommandParameters());
+    const versionId = extractPropertyFromResults<string>(result, 'VersionId');
+    world.saveCreatedObject(world.getSaved<string>('objectName'), versionId || '');
```
francoisferrand: `objectName` is specified in parameter, should be the "source of truth" in this function...
(it is too easy to forget to read it from saved and use the variable directly... which may lead to issue when refactoring/...)

```suggestion
async function putObject(world: Zenko, objectName: string = Utils.randomString()) {
    world.addToSaved('objectName', objectName);
    await uploadSetup(world, 'PutObject');
    world.addCommandParameter({ key: objectName });
    world.addCommandParameter({ bucket: world.getSaved<string>('bucketName') });
    const result = await S3.putObject(world.getCommandParameters());
    const versionId = extractPropertyFromResults<string>(result, 'VersionId');
    world.saveCreatedObject(objectName, versionId || '');
```

tests/ctst/steps/pra.ts
```diff
@@ -298,10 +299,7 @@ When('the DATA_ACCESSOR user tries to perform PutObject on {string} site', { tim
             }
         }
 
-        this.addCommandParameter({ bucket: this.getSaved<string>('bucketName') });
-        this.addCommandParameter({ key: `${Utils.randomString()}` });
-
-        this.setResult(await S3.putObject(this.getCommandParameters()));
+        await putObject(this, Utils.randomString());
```
francoisferrand:
```suggestion
        await putObject(this);
```

.github/actions/archive-artifacts/action.yaml
```diff
@@ -42,3 +42,22 @@ runs:
...
+            kubectl exec -n ${NAMESPACE} data-db-mongodb-sharded-mongos-0 -- mongo ${ZENKO_MONGODB_DATABASE} -u ${MONGODB_ROOT_USERNAME} -p ${MONGODB_ROOT_PASSWORD} --authenticationDatabase admin --quiet --eval "db.getCollection((\"${collection}\")).find().toArray()" > /tmp/artifacts/data/${STAGE}/mongodb-${collection}.log
```
francoisferrand: can't we use `mongodump` instead?
it allows dumping the whole DB or specific collection, may be faster. However, it exports a BSON: more compact, but maybe less practical to use (for analysis, may need to pass it through `bsondump` or restore it into another mongo)

otherwise, we may use `mongoexport` to dump a collection to JSON, one at a time...

.github/actions/archive-artifacts/action.yaml
francoisferrand: not sure what we want to do...

it takes ~1min20, which is significant (maybe not after 2h of test, but we need to reduce this 2h anyway...)

- we should try to reduce the time. Probably a large time is related to kubectl kube setup & I/O transfers out of the container... May be much faster using mongo dump ; and possibly "copying" the file from the container after the tool has finished, or creating a new container with a host-volume to directly "save" the file to the host...
- if we can't make it significantly faster, we may need some filtering indeed AND/OR do this only if the previous job has failed...

.github/actions/archive-artifacts/action.yaml
ghost: Updated with mongodump & bsondump, let's see if my approach is faster...

tests/ctst/steps/utils/utils.ts
ghost: I removed the lines, and updated the function: it was storing always the last version id, but we may need to store all the known version IDs... We can later add a function in the world to delete the object if removed from the bucket, but I would like to do it in a dedicated ticket, where we centralize all the duplicated logic (so far I see that tests usually require some common functions to manage objects, that should be part of the World and not re-implemented in different places).

tests/ctst/common/hooks.ts
ghost: I moved the logic files in a common file, as they can be used by any scenario using a cross-account identity (can be any identity)
For the tag, I added it, but we must be careful to add a tag if more tests in the future are using this type of identity

tests/ctst/steps/utils/utils.ts
ghost: -> ZENKO-4904

.github/actions/archive-artifacts/action.yaml
ghost: New logic takes 10s to complete, I think we are good...

KillianG (approved):

tests/ctst/common/common.ts
```diff
@@ -31,25 +31,33 @@ export async function cleanS3Bucket(
...
+                world.addCommandParameter({ versionId: obj.VersionId });
+                return S3.deleteObject(world.getCommandParameters());
+            }));
+            world.deleteKeyFromCommand('key');
+            world.deleteKeyFromCommand('versionId');
+        }
+        await S3.deleteBucketLifecycle(world.getCommandParameters());
+        await S3.deleteBucket(world.getCommandParameters());
+    } catch (err) {
+        world.logger.warn('Error cleaning bucket', { bucketName, err });
```
francoisferrand: whu could this happen?
except a bug in the code, it should not happen IMHO.... so having an error is actually a good thing!

Even if it may not be related to the "current" test, this should not be masked (let's be honest: we will never notice a warning, unless we have another issue)... Esp. if it allows to "detect" unexpected system instability (like high cpu usage or pod restarts), it is better to have an error here and fix the root cause of the instability.

ghost: Some tests actually test the deletion of a bucket, or delete a bucket in their logic; in this case and because this can be dynamic, we do not want to fail a successful scenario in this case, because we get a "NoSuchBucket" error. I don't want to filter out this error as well because it may be a real bug in some other scenarios. So having a warning log seem the good trade-off.

tests/ctst/steps/utils/utils.ts
```diff
@@ -196,18 +196,21 @@ async function createBucketWithConfiguration(
...
-    world.addToSaved('objectName', objectName || Utils.randomString());
-    const objectNameArray = world.getSaved<string[]>('objectNameArray') || [];
-    objectNameArray.push(world.getSaved<string>('objectName'));
-    world.addToSaved('objectNameArray', objectNameArray);
+    world.resetCommand();
+    let finalObjectName = objectName;
+    if (!finalObjectName) {
+        finalObjectName = `${Utils.randomString()}`;
+    }
+    world.addToSaved('objectName', finalObjectName);
```
francoisferrand, line 208: is it needed? It this will be done by `putObject()` on line 209...

tests/ctst/common/common.ts
francoisferrand: > I don't want to filter out this error as well because it may be a real bug in some other scenarios. So having a warning log seem the good trade-off.

Correct me if I am wrong, but here it seems to me we are filtering all errors : so we will filter-out `NoSuchBucket` errors as well as all the others, and just adding a log (which we will not look at unless we have another test failure)
→ If there is a bug causing NoSuchBucket we will not _notice_
→ If there is a bug causing other error, we will not _notice_ either

> we do not want to fail a successful scenario in this case, because we get a "NoSuchBucket" error

If this was the test, it would fail systematically ; yet the problem we are facing now is flaky: so I don't understand how it relates..

> Some tests actually test the deletion of a bucket, or delete a bucket in their logic

If one test deletes a bucket (or object), it should delete it from the "world" ; or we may add a tag/label on the test, to skip the cleanup when that tag is present...

ghost: > Correct me if I am wrong, but here it seems to me we are filtering all errors : so we will filter-out NoSuchBucket errors as well as all the others, and just adding a log (which we will not look at unless we have another test failure)
→ If there is a bug causing NoSuchBucket we will not notice
→ If there is a bug causing other error, we will not notice either

Exactly, I do not want any scenario to fail due to unrelated logic during cleanup: we already have scenarios covering the DeleteBucket API. Throwing errors will not bring anything to our flakiness nor help us detecting bugs for this API, as it should be covered by dedicated scenarios. We can also have buckets where compliance locking is enabled and so we cannot remove the objects. It is totally fine to keep some buckets when we cannot naively clean them.

> If this was the test, it would fail systematically ; yet the problem we are facing now is flaky: so I don't understand how it relates..

I think your point is not about flakiness (it is already solved) nor bug detection (we have dedicated scenarios - the cleanup will fail if these scenarios fail as well), but about enforcing that each scenario complies with the cleanup logic; i.e, we do not leave the "saved" variables in a state where we still have a reference to a bucket that was deleted. But we do not want to delete the bucket from the world after we delete it, because we might need it again (e.g., bucket re-creation). We can add a new flag to say "this bucket was deleted", but it imposes a new logic in our tests (logic that is unknown to the dev unless we actually run the scenario and end up with a failure because we tried to delete an already-deleted bucket).
So I think it is better not to alter any existing scenario, and be very flexible. We rely on the dedicated scenarios to detect bugs around bucket cleaning.

tests/ctst/steps/utils/utils.ts
ghost: The line 209 only calls the S3 API wrapper from cli-testing, that will either run a cli or sdk command, but it is already outside of the world's scope, and it won't change the 'objName', only consume the input parameters...

tests/ctst/common/common.ts
francoisferrand: >  fail due to unrelated logic during cleanup: we already have scenarios covering the DeleteBucket API

when I talk about bug, it is not necessarily a bug in the DeleteBucket API (which should be covered explicitely indeed), but more about other bugs... e.g. if the system is unstable or the test itself has an issue, being strict allows to surface these issues.

the point is that we cannot garantee that the tests scenarios will cover any bug : so unless there is a known reason to fail, it should not fail... and it is better to fail (even if hard to detect) than ignore a possible bug (which by definition is not covered by our tests). and leaving the bucket unexpectedly present may still cause more errors later, as we would not be in the "nominal" case of the tests.

> We can add a new flag to say "this bucket was deleted", but it imposes a new logic in our tests (logic that is unknown to the dev unless we actually run the scenario and end up with a failure because we tried to delete an already-deleted bucket)

it is unknown until we tell them it is how it works; and if they don't, they will soon discover it since the test will fail.
the real question here is how many test (or test suites?) would be impacted: do we really have many scenarios which delete their own bucket?

(also I guess deleting a bucket is done is only a few dedicated steps: si it should be handled there, and further limit the changes...)

ghost: I am doing it, but:

- I am not aligned for the reasons above, plus ->
  - If the test itself has an issue, it should fail; if not the test should be re-implemented. We shouldn't rely on hooks to detect that.
  - At first, buckets were not cleaned for a conscious reason: we want to have a "living platform" where multiple operations are affecting multiple resources, to be realistic. That's why each scenario today is fully independent from others (thanks to random bucket names, or different accounts when required), excluding lifecycle. If we have a bucket that is failing to be deleted, it does not mean that the scenario is responsible for this error. So we might spend more time trying to understand links between a failure and some transient state of the system, than letting the right scenarios fail. Sometimes we could have an overload in the CI (or anything, due to very valid reasons) and thus make this cleanup fail. In this case, the goal is to still pass the other scenarios, and not fail the current one. The philosophy behind CTST is that we are testing what customers do: instabilities happen, other operations might exhaust the resources... That's fine: as long as the scenario achieves its objectives, then the platform is robust enough. CTST tests feature scenarios, what the customer will use. We should only fail a scenario when its objectives are not achieved. If the instability has no effect on the scenario results, all good. If we want to remove all instabilities sources, CTST is not the right tool...

Also we should really unify our test suites across branches, and not over-engineer our scenarios we have in 2.7+ branches: they re-implement lots of existing logic, do not follow best practices of being independent and parallelizable (such as lifecycle - we already know that), making it very slow to integrate changes targeting all branches, due to both conflicts and errors during test execution, as it is hard to track dependencies between a step and what it "requires" in order to work. E.g. here, when we modify the "saved" resource, we end up breaking tests that re-implemented the logic in a different place...

ghost: Done with added commits specific for 2.7+ branch: https://github.com/scality/zenko/compare/development/2.7...w/2.7/bugfix/ZENKO-4898 (ctst passing on all branches)

francoisferrand (approved):

ghost:
ping

PR #2154, opened 2024-09-24
Change output to outputformat as param changed in drctl

KillianG:
DRCTL status now fails if it, actually fails

Issue: ZENKO-4903


.github/workflows/end2end.yaml
```diff
@@ -487,6 +487,14 @@ jobs:
...
         run: bash configure-e2e-ctst.sh
         working-directory: ./.github/scripts/end2end
+      - name: "Debug: SSH to runner"
+        uses: scality/actions/action-ssh-to-runner@1.5.0
+        with:
+          tmate-server-host: ${{ secrets.TMATE_SERVER_HOST }}
+          tmate-server-port: ${{ secrets.TMATE_SERVER_PORT }}
+          tmate-server-rsa-fingerprint: ${{ secrets.TMATE_SERVER_RSA_FINGERPRINT }}
+          tmate-server-ed25519-fingerprint: ${{ secrets.TMATE_SERVER_ED25519_FINGERPRINT }}
+        timeout-minutes: 30
```
francoisferrand, line 502: should not be commited!

PR #2157, opened 2024-09-30
Bugfix/zenko 4909

ghost:
Fixes the error in the CI, that blocked the last 2.10 bump: with the previous override on sorbet's s3 endpoint, it was always using the SOURCE endpoint to restore objects, causing errors when we were trying to do it during a failover phase, when the restore is expected to take place in the SINK.

### DIFF
--- a/tests/ctst/steps/pra.ts
+++ b/tests/ctst/steps/pra.ts
@@ -296,6 +296,9 @@ When('the DATA_ACCESSOR user tries to perform PutObject on {string} site', { tim
             } catch (err) {
                 this.logger.error('Failed to setup entity', { err });
             }
+            if (!conditionOk) {
+                await Utils.sleep(1000);
+            }
         }
 
         this.addCommandParameter({ bucket: this.getSaved<string>('bucketName') });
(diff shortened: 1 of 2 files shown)


tests/ctst/steps/pra.ts
```diff
@@ -296,6 +296,9 @@ When('the DATA_ACCESSOR user tries to perform PutObject on {string} site', { tim
             } catch (err) {
                 this.logger.error('Failed to setup entity', { err });
             }
+            if (!conditionOk) {
+                await Utils.sleep(1000);
```
francoisferrand, line 300: Would be nice to explain the delay here, in the code: for people stumbling upon it in the future

francoisferrand (approved):

francoisferrand:
Should be fixed on 2.6, we will have the problem everywhere

benzekrimaha (approved):

tests/ctst/steps/pra.ts
ghost: Will be done in the ticket targeting all branches - but not sure it's needed to have a comment, it's not mandatory, just to avoid calling APIs in a while loop without sleep times...

PR #2160, opened 2024-10-03
Bugfix/zenko 4912

benzekrimaha:
Issue : ZENKO-4912


ghost (approved):

monitoring/mongodb/alerts.test.yaml
```diff
@@ -333,10 +333,6 @@ tests:
   - name: MongoDbRSNotSynced
     interval: 1m
     input_series:
-      - series: mongodb_rs_members_state{namespace="zenko", rs_nm="rs0", pod="data-db-mongodb-sharded-mongos-0", member_state="SECONDARY"}
-        values: 0x10
-      - series: mongodb_rs_members_state{namespace="zenko", rs_nm="rs0", pod="data-db-mongodb-sharded-mongos-1", member_state="SECONDARY"}
-        values: 0x10
```
francoisferrand, line 339: why remove these?

to test the behavior, we need input series which are consistent with the actual production situation : so need to have multiple series, including all the labels you use to make sure we validate it works in exactly the right cases.

→ we aggregate by `rs_nm`, so we should check this aggregation works
→ this typically happens in 3-node cluster, so we should have the 3 series for mongo0, mongo1 and mong2

in this specific case, the fix you introduce is to add an `instance` : so the series should include it as well, possible with extra series to validate we do not unexpectedly "merge" what should be separate

monitoring/mongodb/alerts.yaml
```diff
@@ -183,7 +183,7 @@ groups:
 
   - alert: MongoDbRSNotSynced
     expr: |
-      sum by (rs_nm) (mongodb_rs_members_state{namespace="${namespace}", pod=~"${service}.*", member_state="SECONDARY"}) != (${replicas} - 1)
+      count by (rs_nm, instance) (mongodb_rs_members_state{namespace="${namespace}", pod=~"${service}.*", member_state="SECONDARY"}) != (${replicas} - 1)
```
francoisferrand: what is `instance` ?

benzekrimaha: the instance on the metrics is typically the `IP:PORT => instance="10.233.226.186:9216"`, it is being used as each pod will be reporting the same metric with the same value ( the state which is 2 ) resulting in a 3 node for example in a returned value of 12 without the additionnal filtering

monitoring/mongodb/alerts.test.yaml
benzekrimaha: Done here : https://github.com/scality/Zenko/pull/2160/commits/5bbb61af3a8c86e18d2f5018a175fe9ac03b315e#diff-94c90083eba829343ee5745d0056d887a02d888847ea33af739a3ef5c3ae92c6R336-R345

monitoring/mongodb/alerts.yaml
francoisferrand: if we count by instance, we will only get 1 series for each instance : so the count will mostly by `1`...

monitoring/mongodb/alerts.test.yaml
```diff
@@ -333,12 +333,16 @@ tests:
...
-        values: 1x10
+      - series: mongodb_rs_members_state{namespace="zenko", rs_nm="rs0", pod="data-db-mongodb-sharded-mongos-0", member_state="SECONDARY", instance="10.233.226.137:9216"}
+        values: 2x10
+      - series: mongodb_rs_members_state{namespace="zenko", rs_nm="rs0", pod="data-db-mongodb-sharded-mongos-1", member_state="SECONDARY", instance="10.233.226.138:9216"}
+        values: 2x10
+      - series: mongodb_rs_members_state{namespace="zenko", rs_nm="rs0", pod="data-db-mongodb-sharded-mongos-2", member_state="SECONDARY", instance="10.233.226.139:9216"}
+        values: 2x10
+      - series: mongodb_rs_members_state{namespace="zenko", rs_nm="rs1", pod="data-db-mongodb-sharded-mongos-0", member_state="SECONDARY", instance="10.233.226.140:9216"}
+        values: 2x10
+      - series: mongodb_rs_members_state{namespace="zenko", rs_nm="rs1", pod="data-db-mongodb-sharded-mongos-1", member_state="SECONDARY", instance="10.233.226.140:9216"}
```
francoisferrand: `instance` is the IP of the pod which exposes the metric: so multiple `pod` should not have the same value for `instance`, right?

monitoring/mongodb/alerts.yaml
benzekrimaha: No , when we count by the instance we indeed have the right number cf ( this was done on a 6 nodes ) : 
image

benzekrimaha: No the count by the instance and rs_nm returns indeed the right number of secondary members cf : ( this was done on a 6 nodes artesca ) 
image

francoisferrand: but why does it work? AFAIK, each pod should be a single instance, something fishy here...
(and the exemple in the test does not really match the graph you show here)

can you share the actual metrics: e.g. in explore, check which metrics series we get with

```
mongodb_rs_members_state{namespace="zenko", pod=~"data-db-mongodb_sharded-shard0-data"}
```

monitoring/mongodb/alerts.test.yaml
benzekrimaha: yes , this is why we have an example with multiple IP and one with the same one ( the first fires alerts while the second doesn't )

monitoring/mongodb/alerts.yaml
benzekrimaha: with : 
`mongodb_rs_members_state{namespace="zenko", pod=~"data-db-mongodb-sharded-shard0-data.*"}`\

this is what is returned ( for example in fact we get 3 metrics 2 are secondary members and 1 for the primary ) : 
`{__name__="mongodb_rs_members_state", cl_id="65d9fdecfdcca86bb7cd2b98", container="metrics", endpoint="metrics", instance="10.233.178.136:9216", job="zenko/data-db-mongodb-sharded-shard0-data", member_idx="data-db-mongodb-sharded-shard0-data-1.data-db-mongodb-sharded-headless.zenko.svc.cluster.local:27017", member_state="SECONDARY", namespace="zenko", pod="data-db-mongodb-sharded-shard0-data-0", prometheus="metalk8s-monitoring/prometheus-operator-prometheus", rs_nm="data-db-mongodb-sharded-shard-0", rs_state="1"}`


and the value returned is 2 ( the state as it's a secondary ) => 
image

monitoring/mongodb/alerts.test.yaml
francoisferrand: it should not just be an "example" but also match reality : we should have metrics which are similar to what is in a real cluster, and then introduce changes to simulate various error conditions...

monitoring/mongodb/alerts.test.yaml
```diff
@@ -333,23 +333,30 @@ tests:
...
+        values: 2x10
+      - series: mongodb_rs_members_state{cl_id="65d9fdecfdcca86bb7cd2b98", container="metrics", endpoint="metrics", instance="10.233.178.160:9216", job="zenko/data-db-mongodb-sharded-shard0-data", member_idx="data-db-mongodb-sharded-shard0-data-1.data-db-mongodb-sharded-headless.zenko.svc.cluster.local:27017", member_state="SECONDARY", namespace="zenko", pod="data-db-mongodb-sharded-shard0-data-0", prometheus="metalk8s-monitoring/prometheus-operator-prometheus", rs_nm="data-db-mongodb-sharded-shard-0", rs_state="1"}
+        values: 2x10
+      - series: mongodb_rs_members_state{cl_id="65d9fdecfdcca86bb7cd2b98", container="metrics", endpoint="metrics", instance="10.233.178.160:9216", job="zenko/data-db-mongodb-sharded-shard0-data", member_idx="data-db-mongodb-sharded-shard0-data-2.data-db-mongodb-sharded-headless.zenko.svc.cluster.local:27017", member_state="SECONDARY", namespace="zenko", pod="data-db-mongodb-sharded-shard0-data-0", prometheus="metalk8s-monitoring/prometheus-operator-prometheus", rs_nm="data-db-mongodb-sharded-shard-0", rs_state="1"}
```
francoisferrand: for clarity, remove the labels we do not use (and basically ignore): container, endpoint, cl_id, instance, prometheus, job...
 → focus on the relevant labels

monitoring/mongodb/alerts.yaml
francoisferrand: as we can see in the "full" metrics, each pod reports the status (primary/secondary/...) of every member of the cluster participating in the replicaset, as identified by the `member_idx` label

monitoring/mongodb/alerts.yaml
```diff
@@ -183,10 +183,10 @@ groups:
 
   - alert: MongoDbRSNotSynced
     expr: |
-      sum by (rs_nm) (mongodb_rs_members_state{namespace="${namespace}", pod=~"${service}.*", member_state="SECONDARY"}) != (${replicas} - 1)
+      count by (rs_nm, pod) (mongodb_rs_members_state{namespace="${namespace}", pod=~"${service}.*", member_state="SECONDARY", rs_state="1"}) != (${replicas} - 1)
```
benzekrimaha: The pod was added , as when we don't add this additional filtering we end up with a total of 3 instead of 2 , this is explained by the fact that for each pod we have 2 metrics exported , given the fact that these metrics actually have the same pos, namespace, rs_state and state member it is concidered as 1 metric so gives 1 x 3 which results in 3 , when filtering by the pod as well we make sure that for each pod the exported metrics return indeed 2 secondaries. please don't hesitate to challenge this

monitoring/mongodb/alerts.test.yaml
```diff
@@ -333,23 +333,31 @@ tests:
   - name: MongoDbRSNotSynced
     interval: 1m
     input_series:
-      - series: mongodb_rs_members_state{namespace="zenko", rs_nm="rs0", pod="data-db-mongodb-sharded-mongos-0", member_state="SECONDARY"}
-        values: 0x10
-      - series: mongodb_rs_members_state{namespace="zenko", rs_nm="rs0", pod="data-db-mongodb-sharded-mongos-1", member_state="SECONDARY"}
-        values: 0x10
-      - series: mongodb_rs_members_state{namespace="zenko", rs_nm="rs0", pod="data-db-mongodb-sharded-mongos-2", member_state="SECONDARY"}
+      - series: mongodb_rs_members_state{namespace="zenko", pod="data-db-mongodb-sharded-shard0-data-0", member_state="PRIMARY", rs_nm="data-db-mongodb-sharded-shard-0", rs_state="1",  member_idx="data-db-mongodb-sharded-shard0-data-0.data-db-mongodb-sharded-headless.zenko.svc.cluster.local:27017"}
```
benzekrimaha: member_idx has been added as it influences the calculous , without it we get a result of 1 , but once it is provided ( matching the real life metrics within the plateforms ) the calculous becomes 2 matching the right number of secondaries in the replicaset / pod

francoisferrand: it does not _influence the computation_ : only if you don't include it, you actually provide the exact same time-series multiple times and you end up actually having a single serie.... hence the result of the computation.

monitoring/mongodb/alerts.yaml
francoisferrand: * why do we filter on `rs_state=1` : are the wrong states not covered by another alert, or is this covering extra case?
* if there is an issue, this will however trigger multiple alerts: typically, if a single (mongo) pod is down, each of the remaining pod will generate an alert, and 2 alerts will be raised instead of 1 → should avoid this duplication
* the `pod` is actually not relevant -and possibly misleading- as a parameter of the alert: as it seems to indicate an issue with this pod....when actually the issue is with the "remaining" pod: the one(s) which may not actually trigger the alert... → need to make sure the `pod` label does not end up in the final time serie

maybe something like this would solve the issue:

```suggestion
      min by(rs_nm)(
         count by (rs_nm, pod) (mongodb_rs_members_state{namespace="${namespace}", pod=~"${service}.*", member_state="SECONDARY"}) != (${replicas} - 1)
      )
```

VERSION
```diff
@@ -1,4 +1,4 @@
-VERSION="2.10.1"
+VERSION="2.10.2"
```
francoisferrand, line 1: reminder to rebase: this should be the last commit of the PR (where we will make the release)

monitoring/mongodb/alerts.yaml
```diff
@@ -183,10 +183,10 @@ groups:
 
   - alert: MongoDbRSNotSynced
     expr: |
-      sum by (rs_nm) (mongodb_rs_members_state{namespace="${namespace}", pod=~"${service}.*", member_state="SECONDARY"}) != (${replicas} - 1)
+      floor(avg by(rs_nm)(count by (rs_nm, pod)(mongodb_rs_members_state{namespace="${namespace}", pod=~"${service}.*", member_state="SECONDARY"}))) != (${replicas} - 1)
```
benzekrimaha: went for the floor avg as we want to know when the min or max is different from the expected secondaries number

francoisferrand: to ensure we don't _miss_ any alert, we should do the boolean condition before the final aggregation.
Otherwise, we may be unlucky and different pod metrics may average with the right value....

```
<aggregation operator> by(rs_nm) (
    count by(rs_nm, pod) (...) != ${replicas} - 1
)
```

→ if no `pod` triggers the alert, the aggregation will have nothing to aggregate, and no series will trigger the alert
→ if one (or more) `pod` triggers the alert, then the aggregation operation is used to filter-out the `pod` label, and (possibly, but not mandatory) generate an aggregate `value` for the alert (`min` / `avg` / `max` ... number of replicas : though there could always be corner cases where the resulting value is unexecpted... ; if we don't want a value, just using `group` should work)

monitoring/mongodb/alerts.yaml
```diff
@@ -183,10 +183,10 @@ groups:
 
   - alert: MongoDbRSNotSynced
     expr: |
-      sum by (rs_nm) (mongodb_rs_members_state{namespace="${namespace}", pod=~"${service}.*", member_state="SECONDARY"}) != (${replicas} - 1)
+      group by(rs_nm)(count by (rs_nm, pod)(mongodb_rs_members_state{namespace="${namespace}", pod=~"${service}.*", member_state="SECONDARY"}) != (${replicas} - 1) )
```
francoisferrand: nit: indent to improve readabiltiy...

```suggestion
      group by(rs_nm) (
        count by(rs_nm, pod) (mongodb_rs_members_state{namespace="${namespace}", pod=~"${service}.*", member_state="SECONDARY"})
        != (${replicas} - 1)
      )
```

monitoring/mongodb/alerts.yaml
```diff
@@ -183,10 +183,10 @@ groups:
...
     for: 10m
     labels:
       severity: warning
     annotations:
-      description: "MongoDB replica set `{{ $labels.rs_nm }}` is not in the expected state. It currently has `{{ $value }}` SECONDARY members instead of the expected number. Please ensure that all instance are running properly."
+      description: "MongoDB replica set `{{ $labels.rs_nm }}` is not in the expected state. Please ensure that all instances are running properly."
```
francoisferrand: nit: description should still report on "what" is wrong (i.e. number of SECONDARY)

```suggestion
      description: "MongoDB replica set `{{ $labels.rs_nm }}` is not in the expected state. It does not have the expected number of SECONDARY members. Please ensure that all instances are running properly."
```

monitoring/mongodb/alerts.test.yaml
```diff
@@ -333,23 +333,32 @@ tests:
...
+        values: 2x10
+      - series: mongodb_rs_members_state{namespace="zenko", pod="data-db-mongodb-sharded-shard0-data-2", member_state="PRIMARY", rs_nm="data-db-mongodb-sharded-shard-0", member_idx="data-db-mongodb-sharded-shard0-data-0.data-db-mongodb-sharded-headless.zenko.svc.cluster.local:27017"}
         values: 1x10
-  
+      - series: mongodb_rs_members_state{namespace="zenko", pod="data-db-mongodb-sharded-shard0-data-2", member_state="SECONDARY", rs_nm="data-db-mongodb-sharded-shard-0", member_idx="data-db-mongodb-sharded-shard0-data-1.data-db-mongodb-sharded-headless.zenko.svc.cluster.local:27017"}
+        values: 2x10
+      - series: mongodb_rs_members_state{namespace="zenko", pod="data-db-mongodb-sharded-shard0-data-2", member_state="(not reachable/healthy)", rs_nm="data-db-mongodb-sharded-shard-0", member_idx="data-db-mongodb-sharded-shard0-data-2.data-db-mongodb-sharded-headless.zenko.svc.cluster.local:27017"}
+        values: 2 _ _ _ _ _ _ _ _ _
```
francoisferrand: when this serie disappears, the other series will actually change as well: as the remaining pods will not report on its status anymore... and the state reporte looks strange :/
* On "startup", we should have series : everything is running fine (3 pods having the status of 3 members)
* Then one pod "crashes" : the 3 associated series stop
* Then the other pods update their "vision" of the cluster : they either report a different member state (not reachable?) or drop the member completely (not sure exactly what happens)

benzekrimaha: from what I saw the state is (not reachable/healthy) , the test was updated accordingly

monitoring/mongodb/alerts.test.yaml
```diff
@@ -333,23 +333,32 @@ tests:
...
-        values: 0x10
-      - series: mongodb_rs_members_state{namespace="zenko", rs_nm="rs0", pod="data-db-mongodb-sharded-mongos-2", member_state="SECONDARY"}
+      - series: mongodb_rs_members_state{namespace="zenko", pod="data-db-mongodb-sharded-shard0-data-0", member_state="PRIMARY", rs_nm="data-db-mongodb-sharded-shard-0", member_idx="data-db-mongodb-sharded-shard0-data-0.data-db-mongodb-sharded-headless.zenko.svc.cluster.local:27017"}
+        values: 1x10
+      - series: mongodb_rs_members_state{namespace="zenko", pod="data-db-mongodb-sharded-shard0-data-0", member_state="SECONDARY", rs_nm="data-db-mongodb-sharded-shard-0", member_idx="data-db-mongodb-sharded-shard0-data-1.data-db-mongodb-sharded-headless.zenko.svc.cluster.local:27017"}
+        values: 2x10
+      - series: mongodb_rs_members_state{namespace="zenko", pod="data-db-mongodb-sharded-shard0-data-0", member_state="(not reachable/healthy)", rs_nm="data-db-mongodb-sharded-shard-0", member_idx="data-db-mongodb-sharded-shard0-data-2.data-db-mongodb-sharded-headless.zenko.svc.cluster.local:27017"}
+        values: 2 _ _ _ _ _ _ _ _ _
```
francoisferrand: here the member "2" is initially not avaialble, then the series stops : is this really what you meant?

monitoring/mongodb/alerts.test.yaml
francoisferrand: every metric from `pod="data-db-mongodb-sharded-shard0-data-2"` should be removed when we simulate the pod crashing...

so i guess time series should look something like this:

```
# First pod
- series: member_state{pod="data-0", state="PRIMARY", member_idx="data-0"}
   values: 1x10 1x10
- series: member_state{pod="data-0", state="SECONDARY", member_idx="data-1"}
   values: 2x10 2x10
- series: member_state{pod="data-0", state="SECONDARY", member_idx="data-2"}
   values: 2x10 stale      # `stale` (i.e. not present) after the 10s sample
- series: member_state{pod="data-0", state="(not reachable/healthy)", member_idx="data-2"}
   values: _x10 8x10      # appears from the 11s sample
# Second pod
- series: member_state{pod="data-1", state="PRIMARY", member_idx="data-0"}
   values: 1x10 1x10
- series: member_state{pod="data-1", state="SECONDARY", member_idx="data-1"}
   values: 2x10 2x10
- series: member_state{pod="data-1", state="SECONDARY", member_idx="data-2"}
   values: 2x10 stale      # `stale` (i.e. not present) after the 10s sample
- series: member_state{pod="data-1", state="(not reachable/healthy)", member_idx="data-2"}
   values: _x10 8x10      # appears from the 11s sample
# Third pod : stops responding after 10th sample
- series: member_state{pod="data-2", state="PRIMARY", member_idx="data-0"}
   values: 1x10 stale
- series: member_state{pod="data-2", state="SECONDARY", member_idx="data-1"}
   values: 2x10 stale
- series: member_state{pod="data-2", state="SECONDARY", member_idx="data-2"}
   values: 2x10 stale
```

monitoring/mongodb/alerts.test.yaml
benzekrimaha: I'm trying to keep the same values as in the `pod="data-db-mongodb-sharded-shard0-data-2"`, as it's the one with the secondary down

monitoring/mongodb/alerts.test.yaml
benzekrimaha: Providing this :  series: member_state{pod="data-0", state="SECONDARY", member_idx="data-2"}
   values: 2x10 stale      # `stale` (i.e. not present) after the 10s sample
   
  will cause the expression to return 2 => expected number of secondaries as it's not taking the state="(not reachable/healthy)" metric

francoisferrand (commented):
LGTM, but alert test fail...

monitoring/mongodb/alerts.test.yaml
francoisferrand: AFAIU, `stale` removes the metric series, which should be what really happens:
* the serie with state label equal to `secondary` (and value `2`) stops
* another serie starts with state label equal to `(not reachable/healthy)` and value `8`

(since one label value changes, this is really another serie : even if [as human] we think it is the same one...)

monitoring/mongodb/alerts.test.yaml
```diff
@@ -329,27 +329,39 @@ tests:
...
+      - series: mongodb_rs_members_state{namespace="zenko", pod="data-db-mongodb-sharded-shard0-data-1", member_state="SECONDARY", rs_nm="data-db-mongodb-sharded-shard-0", member_idx="shard0-data-1"}
+        values: 2x20
+      - series: mongodb_rs_members_state{namespace="zenko", pod="data-db-mongodb-sharded-shard0-data-1", member_state="SECONDARY", rs_nm="data-db-mongodb-sharded-shard-0", member_idx="shard0-data-2"}
+        values: 2x8 stale
+      - series: mongodb_rs_members_state{namespace="zenko", pod="data-db-mongodb-sharded-shard0-data-1", member_state="(not reachable/healthy)", rs_nm="data-db-mongodb-sharded-shard-0", member_idx="shard0-data-2"}
+        values: stale _x8 8x10
+      - series: mongodb_rs_members_state{namespace="zenko", pod="data-db-mongodb-sharded-shard0-data-2", member_state="PRIMARY", rs_nm="data-db-mongodb-sharded-shard-0", member_idx="shard0-data-0"}
+        values: 1x10 stale
+      - series: mongodb_rs_members_state{namespace="zenko", pod="data-db-mongodb-sharded-shard0-data-2", member_state="SECONDARY", rs_nm="data-db-mongodb-sharded-shard-0", member_idx="shard0-data-1"}
+        values: 2x10 stale
```
francoisferrand: should be `1x8 stale` and `2x8 stale` for consistency with the third metrics serie (`2x8 stale`)

monitoring/mongodb/alerts.test.yaml
```diff
@@ -329,27 +329,39 @@ tests:
...
+      - series: mongodb_rs_members_state{namespace="zenko", pod="data-db-mongodb-sharded-shard0-data-2", member_state="SECONDARY", rs_nm="data-db-mongodb-sharded-shard-0", member_idx="shard0-data-2"}
+        values: 2x8 stale
+
     alert_rule_test:
       - alertname: MongoDbRSNotSynced
-        eval_time: 10m
+        eval_time: 19m
```
francoisferrand, line 372: should keep a check at different times before (5m → no alert, 10m → no alert, 18m → no alert)

francoisferrand (approved):
LGTM, minor test improvements : please rebase/cleanup before approving

PR #2161, opened 2024-10-03
Bugfix/zenko 4903 fix drctl status flaky

KillianG:

fixes #<ISSUE>

### DIFF
--- a/tests/ctst/steps/dr/drctl.ts
+++ b/tests/ctst/steps/dr/drctl.ts
@@ -122,18 +122,16 @@ type FailbackConfig = {
 };
 
 type UninstallConfig = {
-    sinkZenkoDrInstance?: string;
-    sourceZenkoDrInstance?: string;
     wait?: boolean;
     timeout?: string;
     sourceKubeconfigPath?: string;
     sourceKubeconfigData?: string;
     sinkKubeconfigPath?: string;
     sinkKubeconfigData?: string;
-    sinkZenkoInstance?: string;
-    sinkZenkoNamespace?: string;
-    sourceZenkoInstance?: string;
-    sourceZenkoNamespace?: string;
+    sinkZenkoDrInstance?: string;
+    sinkZenkoDrNamespace?: string;
+    sourceZenkoDrInstance?: string;
+    sourceZenkoDrNamespace?: string;
 };
 
 type StatusConfig = {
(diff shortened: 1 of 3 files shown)


KillianG:
ping

solution/deps.yaml
```diff
@@ -21,7 +21,7 @@ cloudserver:
 drctl:
   sourceRegistry: ghcr.io/scality
   image: zenko-drctl
-  tag: v1.0.4
+  tag: e43bce2bfddd5c667e1488c2a2fa846da53c8c07
```
KillianG: This will be changed after we release drctl

ghost (approved):

benzekrimaha (approved):

.github/workflows/end2end.yaml
```diff
@@ -397,6 +395,10 @@ jobs:
           username: "${{ github.repository_owner }}"
           password: "${{ github.token }}"
           registry: ghcr.io
+      - name: Debug wait
+        uses: ./.github/actions/debug-wait
+        timeout-minutes: 60
+        if: always()
```
francoisferrand: cannot be merged

.github/workflows/end2end.yaml
```diff
@@ -524,9 +526,7 @@ jobs:
   end2end-https:
     needs: [build-kafka, build-test-image, check-dashboard-versions]
     runs-on:
-      - ubuntu
-      - focal
-      - xlarge
+      - ubuntu-22.04-8core
```
francoisferrand: do we actually need the larger workers?
(esp. this is one on e2e-http and e2e-https which are not overloaded?)

charlesprost: This move is about x2 in RAM and CPU, for a doubled price, see https://devdocs.scality.net/ci-engines/github-actions/runners/?h=ubuntu+22.04+8core#github-hosted-larger-pools-of-runners , as xlarge (see https://github.com/scality/devinfra/blob/main/manifests/runner-manager/overlays/prod/runner-manager.yaml#L91-L100 and https://cloud.google.com/compute/vm-instance-pricing#compute-optimized_machine_types for `e2-standard-4` ) is equivalent to ubuntu-22.04-4core
I'm totally fine doing it if this is the right move, but how do we jump to this conclusion exactly? (quickly looked at the ticket and the rest of the PR but I don't see anything obvious)

KillianG: For e2e-http and e2e-https we can probably use `ubuntu-22.04-4core` but we still need to jump to ubuntu 22.04 otherwise k8s bump and dependencies wont work..

charlesprost: @KillianG then I would recommend to use a new runner but stay within the current specs; bumping a runner to a more powerful machine should be done if you experience network flakiness, memory issues, disk limits, ... stuff that should be visible in the logs

PR #2162, opened 2024-10-18
Bump Zenko-ui to 3.0.5

JeanMarcMilletScality:

fixes #<ISSUE>


VERSION
```diff
@@ -1,4 +1,4 @@
-VERSION="2.10.2"
+VERSION="2.10.3"
```
ghost, line 1: It's better to bump Zenko's version in a dedicated commit (that we name accordingly), so we do not mix changes (it's not affecting here because we are at the latest branch, but in previous branch this would mean conflicts, even if the zkui version is the same)

JBWatenbergScality (approved):

ghost (approved):

PR #2165, opened 2024-10-24
Bucket website tests

ghost:
Last commit will be removed after the ongoing zenko release.

### DIFF
--- a/tests/ctst/steps/website/website.ts
+++ b/tests/ctst/steps/website/website.ts
@@ -1,0 +1,86 @@
+import assert from 'assert';
+import { Given, When, Then } from '@cucumber/cucumber';
+import Zenko from '../../world/Zenko';
+import { putObject } from '../utils/utils';
+import { S3, Utils } from 'cli-testing';
+
+const pageMessage = Utils.randomString();
+
+Given('an index html file', async function (this: Zenko) {
+    // push a file with a basic html content named index.html in the bucket
+    const content = `<html><head><title>Index</title></head><body><h1>${pageMessage}</h1></body></html>`;
+    this.addToSaved('objectSize', content.length);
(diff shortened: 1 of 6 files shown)


solution/deps.yaml
francoisferrand, line 1: already done, please rebase on top of latest head...

francoisferrand (approved):

solution/deps.yaml
ghost: Thanks!

tests/ctst/steps/website/website.ts
```diff
@@ -0,0 +1,86 @@
...
+});
+
+Then('the user should be able to load the index.html file from the {string} endpoint',
+    async function (this: Zenko, endpoint: string) {
+        const baseUrl = this.parameters.ssl === false ? 'http://' : 'https://';
+        // The ingress may take some time to be ready (<60s)
+        const uri = `${baseUrl}${this.getSaved<string>('bucketName')}.${endpoint}`;
+        let response;
+        let content;
+        let tries = 60;
```
KillianG, line 66: Would be great to not have that hardcoded, maybe based on the timeout of the step or something like that, what do you think? Or even better, get the ingress status instead, directly in the `the {string} endpoint is added to the overlay` for example?

ghost: We have 100s as the test timeout, and I am expecting the ingress to be here in 60s (and still allow us to get logs, not just "it timeouts"), so for now this is expected

To list the ingresses we would need to have the network client from k8s (requires a CTST PR, and we are not really testing it, but I agree it may help reducing (by 1s) the test duration)

ghost: (so far in the ci it was around 22s)

KillianG (approved):

PR #2168, opened 2024-11-14
Ensure callback is called once in doWhilst with S3 operations

ghost:
The trigger seem to be
https://github.com/aws/aws-sdk-js/issues/1678
... causing the callback to be called twice, which is a problem in these loops.

Issue: ZENKO-4925

**Rationale of the change in the ticket comments**.

### DIFF
--- a/tests/zenko_tests/node_tests/backbeat/ReplicationUtility.js
+++ b/tests/zenko_tests/node_tests/backbeat/ReplicationUtility.js
@@ -1,6 +1,7 @@
 const assert = require('assert');
 const crypto = require('crypto');
 const async = require('async');
+const { jsutil } = require('arsenal');
 
 const { scalityS3Client, awsS3Client } = require('../s3SDK');
 
@@ -596,8 +597,9 @@ class ReplicationUtility {
                 Key: key,
                 VersionId: versionId,
             }, (err, data) => {
+                const cbOnce = jsutil.once(callback);
                 if (err) {
-                    return callback(err);
+                    return cbOnce(err);
                 }
                 status = data.ReplicationStatus;
                 assert.notStrictEqual(
@@ -606,9 +608,9 @@ class ReplicationUtility {
                     `Unexpected CRR failure occurred: ${JSON.stringify(data)}`,
                 );
                 if (status === 'PENDING' || status === 'PROCESSING') {
-                    return setTimeout(callback, 2000);
+                    return setTimeout(cbOnce, 2000);
                 }
-                return callback();
+                return cbOnce();
             }),
             () => (status === 'PENDING' || status === 'PROCESSING'),
             cb,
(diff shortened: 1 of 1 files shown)


welansari (approved):

tests/zenko_tests/node_tests/backbeat/ReplicationUtility.js
```diff
@@ -596,8 +597,9 @@ class ReplicationUtility {
                 Key: key,
                 VersionId: versionId,
             }, (err, data) => {
+                const cbOnce = jsutil.once(callback);
```
francoisferrand, line 599: if I understand correctly, this would happen if/because the headObject callback is invoked twice : both for the actual response and a socket timeout which happens at about the same time.

→ it seems to me a simpler fix would be just:
```
 callback => this.s3.headObject({
                Bucket: bucketName,
                Key: key,
                VersionId: versionId,
            }, jsutil.once((err, data) => {
                [...]
           })
```

However, I am wondering if there may not be a real underlying issue : like the socket timeout (on client side, i.e. test) being shorter than the timeout we have in zenko.... and maybe this could explain the (other) issues we have with these tests, what do you think?

Can we check the timeout (or infirm this hypothesis), and/or should we ignore retryable AWS errors (retry after a timeout) ?

tests/zenko_tests/node_tests/backbeat/ReplicationUtility.js
```diff
@@ -606,9 +608,9 @@ class ReplicationUtility {
                     `Unexpected CRR failure occurred: ${JSON.stringify(data)}`,
                 );
                 if (status === 'PENDING' || status === 'PROCESSING') {
-                    return setTimeout(callback, 2000);
+                    return setTimeout(cbOnce, 2000);
                 }
-                return callback();
+                return cbOnce();
```
francoisferrand: should we take the chance to cleanup the code: and pass the `status` from this callback to the next (instead of using a local variable?

```suggestion
                if (status === 'PENDING' || status === 'PROCESSING') {
                    return setTimeout(() => cbOnce(nil, status), 2000);
                }
                return cbOnce(nil, status);
            }),
            () => (status === 'PENDING' || status === 'PROCESSING'),
```

tests/zenko_tests/node_tests/backbeat/ReplicationUtility.js
ghost: I am not sure it was a timeout but worth reviewing that indeed, let me check

ghost: The current client-side timeout seem to be infinite, so I don't think we are affected by the timeout from client side:

```
const scalityS3Client = new S3({
    ...
    // disable node sdk retries and timeout to prevent InvalidPart
    // and SocketHangUp errors. If retries are allowed, sdk will send
    // another request after first request has already deleted parts,
    // causing InvalidPart. Meanwhile, if request takes too long to finish,
    // sdk will create SocketHangUp error before response.
    maxRetries: 0,
    httpOptions: { timeout: 0 },
});
```

francoisferrand (approved):

ghost:
Will be merged with the upcoming Zenko release - keeping this open till then.

PR #2169, opened 2024-11-15
pra: add test for prometheus federation

francoisferrand:
- Add artesca-root-ca-issuer cluster issuer
- Test that federated metrics are retrieved
- Set federated prometheus name
- Fix prometheus config
- Fix zkop install to support sha1
- gha: Move kafka dump to folder
- gha: Merge alerts workflow

Issue: ZENKO-4876

### DIFF
--- a/tests/ctst/steps/pra.ts
+++ b/tests/ctst/steps/pra.ts
@@ -15,6 +15,7 @@ import {
 } from 'steps/utils/utils';
 import { CacheHelper, Constants, Identity, IdentityEnum, SuperAdmin, Utils } from 'cli-testing';
 import { safeJsonParse } from 'common/utils';
+import { PrometheusDriver } from 'prometheus-query';
 import assert from 'assert';
 import { EntityType } from 'world/Zenko';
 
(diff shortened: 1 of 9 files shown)


tests/ctst/steps/pra.ts
```diff
@@ -70,9 +71,8 @@ async function installPRA(world: Zenko, sinkS3Endpoint = 'http://s3.zenko.local'
         sourceS3Endpoint: 'http://s3.zenko.local',
         sinkS3Endpoint,
         prometheusService: world.parameters.PrometheusService,
-        // prometheusHostname: 'prom.dr.zenko.local', // could be any name, cert will be auto-generated
+        prometheusHostname: 'prom.dr.zenko.local',
```
KillianG, line 74: can it be in world parameters ?

tests/ctst/steps/pra.ts
```diff
@@ -328,8 +328,25 @@ Then('the kafka DR volume exists', { timeout: volumeTimeout + 2000 }, async func
     assert(volumeParsed.result!['volume phase'] === 'Bound');
 });
 
+Then('prometheus should scrap federated metrics from DR sink', { timeout: 70000 }, async function (this: Zenko) {
+    const prom = new PrometheusDriver({
+        endpoint: `http://${this.parameters.PrometheusService}:9090`,
+        baseURL: '/api/v1',
+    });
+
+    for (;;) {
```
KillianG, line 337: Timing out, does not give proper logs actually, it would be great to have a timeout in the code and to properly log that prom does not respond or that metrics are still empty after 70k ms

francoisferrand: i don't think we should litter the code with "manual" timeout: if the logs are not correct, it should be improved in the framework instead (so that we benefit from it everywhere)

ghost: I think the point was more: the test already manually set a 70s timeout, yet we are never calling `this.world.logger.debug` to log what we got when calling prometheus, so in case of timeout, we have 0 clue to understand the problem

ghost (approved):
lgtm beyond Killian's comment

tests/ctst/steps/pra.ts
francoisferrand: if we receive _anything_, we will bail out... so if we timeout, it means we did not receive anything : and the information to troubleshoot will be deep inside prometheus, not in the api call we make here, unfortunately...

so not sure what could be logged...

KillianG: It might be a personal opinion, but I don't think an infinite loop is any better than a "manual" timeout

KillianG (approved):

tests/ctst/steps/pra.ts
francoisferrand: it is not an infinite loop : we are expected to exit eventually, as prometheus should return a metric!

ghost: > so not sure what could be logged...

If we can get some connectivity info then logging will be useful - if we expect nothing, the current approach is good enough...

PR #2170, opened 2024-11-17
Add test for archive object replacement

francoisferrand:
- Add test for archive object replacement
- Bump cloudserver

Issue: ZENKO-4928


.github/scripts/end2end/configs/zenkoversion.yaml
```diff
@@ -9,9 +9,9 @@ spec:
     backbeat:
       image: '${BACKBEAT_DASHBOARD}'
       tag: '${BACKBEAT_TAG}'
-    cloudserver:
-      image: '${CLOUDSERVER_DASHBOARD}'
-      tag: '${CLOUDSERVER_TAG}'
+    # cloudserver:
+    #   image: '${CLOUDSERVER_DASHBOARD}'
+    #   tag: '${CLOUDSERVER_TAG}'
```
ghost: To be re-enabled before merging I guess

tests/ctst/features/dmf.feature
```diff
@@ -92,3 +92,44 @@ Feature: DMF
...
+    @ColdStorage
+    Scenario Outline: Overwriting of a cold mpu object
+    Given a "<versioningConfiguration>" bucket
+     And a transition workflow to "e2e-cold" location
+     And <objectCount> objects "obj" of size <objectSize> bytes
+    Then object "obj-1" should be "transitioned" and have the storage class "e2e-cold"
+     And dmf volume should contain <objectCount> objects
+    Given <objectCount> mpu objects "obj" of size <objectSize> bytes
+    Then object "obj-1" should be "transitioned" and have the storage class "e2e-cold"
+     And dmf volume should contain 1 objects
```
ghost: No need to indent `And` lines in gherkin

```suggestion
    Scenario Outline: Overwriting of a cold mpu object
    Given a "<versioningConfiguration>" bucket
    And a transition workflow to "e2e-cold" location
    And <objectCount> objects "obj" of size <objectSize> bytes
    Then object "obj-1" should be "transitioned" and have the storage class "e2e-cold"
    And dmf volume should contain <objectCount> objects
    Given <objectCount> mpu objects "obj" of size <objectSize> bytes
    Then object "obj-1" should be "transitioned" and have the storage class "e2e-cold"
    And dmf volume should contain 1 objects
```

tests/ctst/steps/utils/utils.ts
```diff
@@ -199,6 +199,84 @@ async function createBucketWithConfiguration(
...
+    const userMetadata = world.getSaved<string>('userMetadata');
+    if (userMetadata) {
+        world.addCommandParameter({ metadata: JSON.stringify(userMetadata) });
+    }
+
+    const result = await S3.copyObject(world.getCommandParameters());
+    const versionId = extractPropertyFromResults<string>(result, 'VersionId');
+    world.saveCreatedObject(key, versionId || '');
+    world.setResult(result);
+    return result;
```
ghost, line 277: Not strictly needed unless you check the result in a subsequent step; we usually use this setResult when we want, e.g., to check the authz result at the end, or get the expected return code, but I don't see such tests in your scenarios

ghost (approved):
Approving because the comments are not blocking & I know they will be handled before merging

tests/ctst/steps/utils/utils.ts
```diff
@@ -199,6 +199,83 @@ async function createBucketWithConfiguration(
     }
 }
 
+async function putMpuObject(world: Zenko, parts: number = 2, objectName: string, content?: string) {
```
KillianG, line 202: Isn't there a minimum of size for parts of an MPU (IIRC its 5MB) ?. If so, should we log an error here if the test is misswritten or should we let S3 respond with an error ?

ghost: We should let the test fail, actually the SDK may be already doing the check for us.

francoisferrand: not sure we have this limitations...
test seems to pass, so we should be fine... unless we silently ignore the s3 error, as there is no assertion after `completeMpuObject` :-/

ghost: We do have this limitation in cloudserver when doing a PutPart, but it does not consider the last part. However, this is only checked during CompleteMPU:

```

    // Note: Part sizes cannot be less than 5MB in size except for the last.
    // However, we do not check this value here because we cannot know which
    // part will be the last until a complete MPU request is made. Thus, we let
    // the completeMultipartUpload API check that all parts except the last are
    // at least 5MB.
```

KillianG (approved):

PR #2174, opened 2024-11-20
Fix flaky prometheus federation check

francoisferrand:
- Increase timeout for prometheus metrics scrapping
- gha: upgrade trunk analytics

Issue: ZENKO-4931

### DIFF
--- a/tests/ctst/steps/pra.ts
+++ b/tests/ctst/steps/pra.ts
@@ -328,7 +328,7 @@ Then('the kafka DR volume exists', { timeout: volumeTimeout + 2000 }, async func
     assert(volumeParsed.result!['volume phase'] === 'Bound');
 });
 
-Then('prometheus should scrap federated metrics from DR sink', { timeout: 70000 }, async function (this: Zenko) {
+Then('prometheus should scrap federated metrics from DR sink', { timeout: 180000 }, async function (this: Zenko) {
     const prom = new PrometheusDriver({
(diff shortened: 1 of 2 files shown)


tests/ctst/steps/pra.ts
```diff
@@ -328,7 +328,7 @@ Then('the kafka DR volume exists', { timeout: volumeTimeout + 2000 }, async func
     assert(volumeParsed.result!['volume phase'] === 'Bound');
 });
 
-Then('prometheus should scrap federated metrics from DR sink', { timeout: 70000 }, async function (this: Zenko) {
+Then('prometheus should scrap federated metrics from DR sink', { timeout: 180000 }, async function (this: Zenko) {
```
ghost, line 331: Do we understand why 70s is not enough with a 30s frequency? is the system too loaded, are we missing metrics?

ghost (approved):

tests/ctst/steps/pra.ts
francoisferrand: it is not enough because there is not just the 30s frequency, there is also the initial "discovery" delay : from the moment the ServiceMonitor is added, then picked'up but prometheus sidecar to update the config, then prometheus to "reload" its config, schedule the actual "discovery" of the endpoints... and only once it si discovered will it be scrapped, every 30 seconds indeed.

rachedbenmustapha (approved):

PR #2177, opened 2024-11-26
Use unique bucket names in scenarios where the name is not random

ghost:
Issue: ZENKO-4932

We had a scenario using the same (fixed) bucket name. In these kind of scenarios, this is required.
But then if they run in parallel we might have the same bucket being deleted / re-created, and eventually the hook at the end, cleaning up the existing buckets, will fail because of this concurrency problem. The solution is to use unique bucket names.

### DIFF
--- a/tests/ctst/features/iam-policies/IAMUser.feature
+++ b/tests/ctst/features/iam-policies/IAMUser.feature
@@ -55,8 +55,8 @@ Feature: IAM Policies for IAM Users
       | GetObject      | Allow  | iu-go-bucket3-1/*        | iu-go-bucket3 | exists         |            |
       | GetObject      | Allow  | iu-go-bucket4/go-object1 | iu-go-bucket4 | exists         | go-object  |
       | GetObject      | Deny   | *                        | iu-go-bucket5 | exists         |            |
-      | GetObject      | Deny   | iu-go-bucket5/*          | iu-go-bucket5 | exists         |            |
-      | GetObject      | Deny   | iu-go-bucket6/go-object  | iu-go-bucket6 | exists         | go-object  |
+      | GetObject      | Deny   | iu-go-bucket6/*          | iu-go-bucket6 | exists         |            |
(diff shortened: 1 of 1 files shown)


tests/ctst/features/iam-policies/IAMUser.feature
```diff
@@ -55,8 +55,8 @@ Feature: IAM Policies for IAM Users
       | GetObject      | Allow  | iu-go-bucket3-1/*        | iu-go-bucket3 | exists         |            |
       | GetObject      | Allow  | iu-go-bucket4/go-object1 | iu-go-bucket4 | exists         | go-object  |
       | GetObject      | Deny   | *                        | iu-go-bucket5 | exists         |            |
-      | GetObject      | Deny   | iu-go-bucket5/*          | iu-go-bucket5 | exists         |            |
-      | GetObject      | Deny   | iu-go-bucket6/go-object  | iu-go-bucket6 | exists         | go-object  |
+      | GetObject      | Deny   | iu-go-bucket6/*          | iu-go-bucket6 | exists         |            |
```
KillianG, line 58: Should be great to be able to pass variables generated during the test in the examples, so that the `given an existing bucket ...` will create an bucket named randomly and in the examples we put something like {bucketName} and it takes the one generated in the beginning
Not sure if it is feasable tho..

KillianG (approved):

tests/ctst/features/iam-policies/IAMUser.feature
ghost: We do that in most scenarios, but here we need to specify the resource name, and we actually want to have control over it because sometimes we alter it, so we want to keep it... Still possible to have it fully automatic but we still need to keep support for static bucket names. For example, the bucket used for website must be named "website".

KillianG: I know we do that in most scenario, what I suggest here is to be able to use it in example, so here for example we could do `${bucketName}/*` in the ressource name
But ofc we need to keep support for creating static bucket names

francoisferrand (approved):

PR #2178, opened 2024-11-28
Improvement/zenko 4941

ghost:
Complete tests of Quota feature in CTST.

-> custom images to be deleted

## Another test to review starting the 2.8 branch:

https://github.com/scality/zenko/compare/development/2.8...w/2.8/improvement/ZENKO-4941#diff-0f9a8cc11d2560e9e9ea8668d53be732997ae7cecb4ec7201c836bc8367fc5c9R182

-> All tests are passing:

```
2024-11-29T09:56:42.1860871Z   @2.8.0 @PreMerge @Quotas @Restore @Dmf @ColdStorage @Only
2024-11-29T09:56:42.1861891Z   Scenario Outline: Restored object expiration updates quotas # features/quotas/Quotas.feature:182
2024-11-29T09:56:42.1880965Z     Given an action "RestoreObject"
2024-11-29T09:56:42.1883582Z     And a STORAGE_MANAGER type
2024-11-29T09:56:42.9398659Z     And a transition workflow to "e2e-cold" location
2024-11-29T09:56:47.9621673Z     And an upload size of 100 B for the object "obj-1"
2024-11-29T09:56:47.9946235Z     Then object "obj-1" should be "transitioned" and have the storage class "e2e-cold"
2024-11-29T09:56:57.1634848Z     Given a bucket quota set to 0 B
2024-11-29T09:56:57.1635502Z     And an account quota set to 0 B
2024-11-29T09:56:57.1638992Z     And a ACCOUNT type
2024-11-29T09:56:57.1639488Z     And an environment setup for the API
2024-11-29T09:56:57.1641680Z     And an "existing" IAM Policy that "applies" with "ALLOW" effect for the current API
2024-11-29T09:56:57.1654784Z     When i restore object "" for 5 days
2024-11-29T09:56:57.1823329Z     Then the API should "succeed" with ""
2024-11-29T09:56:57.1823860Z     And object "obj-1" should be "restored" and have the storage class "e2e-cold"
2024-11-29T09:57:04.2913305Z     Given a STORAGE_MANAGER type
2024-11-29T09:57:04.2913874Z     Then object "obj-1" should expire in 5 days
2024-11-29T09:57:04.3184634Z     When i wait for 5 days
2024-11-29T09:57:09.3217840Z     Then object "obj-1" should be "cold" and have the storage class "e2e-cold"
2024-11-29T09:57:21.6862371Z 
2024-11-29T09:57:21.6891444Z   @2.8.0 @PreMerge @Quotas @Restore @Dmf @ColdStorage @Only
2024-11-29T09:57:21.6892387Z   Scenario Outline: Restored object expiration updates quotas # features/quotas/Quotas.feature:182
2024-11-29T09:57:21.6903644Z     Given an action "RestoreObject"
2024-11-29T09:57:21.6915643Z     And a STORAGE_MANAGER type
2024-11-29T09:57:22.3246302Z     And a transition workflow to "e2e-cold" location
2024-11-29T09:57:27.3445935Z     And an upload size of 100 B for the object "obj-1"
2024-11-29T09:57:27.3756834Z     Then object "obj-1" should be "transitioned" and have the storage class "e2e-cold"
2024-11-29T09:57:37.5580469Z     Given a bucket quota set to 101 B
2024-11-29T09:57:37.5784356Z     And an account quota set to 101 B
2024-11-29T09:57:37.5973259Z     And a ACCOUNT type
2024-11-29T09:57:37.5973875Z     And an environment setup for the API
2024-11-29T09:57:37.5974634Z     And an "existing" IAM Policy that "applies" with "ALLOW" effect for the current API
2024-11-29T09:57:37.5978067Z     When i restore object "" for 5 days
2024-11-29T09:57:37.6263509Z     Then the API should "succeed" with ""
2024-11-29T09:57:37.6264026Z     And object "obj-1" should be "restored" and have the storage class "e2e-cold"
2024-11-29T09:57:44.7328078Z     Given a STORAGE_MANAGER type
2024-11-29T09:57:44.7329148Z     Then object "obj-1" should expire in 5 days
2024-11-29T09:57:44.7473779Z     When i wait for 5 days
2024-11-29T09:57:49.7499272Z     Then object "obj-1" should be "cold" and have the storage class "e2e-cold"
```

```
2024-11-29T09:51:13.8625920Z   @2.6.0 @PreMerge @Quotas @CronJob @DataDeletion @NonVersioned
2024-11-29T09:51:13.8627154Z   Scenario Outline: Quotas are affected by deletion operations between count items runs # features/quotas/Quotas.feature:107
2024-11-29T09:51:13.8628061Z     Given an action "DeleteObject"
2024-11-29T09:51:13.8628617Z     And a permission to perform the "PutObject" action
2024-11-29T09:51:13.8629197Z     And a STORAGE_MANAGER type
2024-11-29T09:51:14.2009069Z     And I PUT an object with size 100
2024-11-29T09:51:14.5443986Z     And a bucket quota set to 10000 B
2024-11-29T09:51:14.5599516Z     And an account quota set to 10000 B
2024-11-29T09:51:14.5727230Z     And an upload size of 1000 B for the object "obj-1"
2024-11-29T09:51:14.6027627Z     And a bucket quota set to 200 B
2024-11-29T09:51:14.6138121Z     And an account quota set to 200 B
2024-11-29T09:51:14.6227240Z     And a IAM_USER type
2024-11-29T09:51:14.6520382Z     And an environment setup for the API
2024-11-29T09:51:14.6521792Z     And an "existing" IAM Policy that "applies" with "ALLOW" effect for the current API
2024-11-29T09:51:14.6751194Z     When I wait 3 seconds
2024-11-29T09:51:14.8870100Z     When I wait 3 seconds
2024-11-29T09:51:14.9774911Z     Then the API should "fail" with "QuotaExceeded"
2024-11-29T09:51:14.9775549Z     When the "count-items" cronjobs completes without error
2024-11-29T09:51:15.1964713Z     When I wait 3 seconds
2024-11-29T09:51:17.6776293Z     And I PUT an object with size 100
2024-11-29T09:51:17.8890763Z     And i delete object "obj-1"
2024-11-29T09:51:17.9091224Z     And I wait 3 seconds
2024-11-29T09:51:18.1982800Z     And i delete object "obj-1"
2024-11-29T09:51:18.2191912Z     And I wait 3 seconds
2024-11-29T09:51:19.0025639Z     Then the API should "fail" with "QuotaExceeded"
2024-11-29T09:51:19.0026380Z     When the "count-items" cronjobs completes without error
2024-11-29T09:51:20.9103545Z     And I PUT an object with size 100
2024-11-29T09:51:20.9672979Z     Then the API should "succeed" with ""
```


ghost:
ping

tests/ctst/features/quotas/Quotas.feature
```diff
@@ -96,3 +96,43 @@ Feature: Quota Management for APIs
...
+        And an upload size of 1000 B for the object "obj-1"
+        And a bucket quota set to <bucketQuota> B
+        And an account quota set to <accountQuota> B
+        And a <userType> type
+        And an environment setup for the API
+        And an "existing" IAM Policy that "applies" with "ALLOW" effect for the current API
+        When I wait 3 seconds
+        And I PUT an object with size <uploadSize>
+        Then the API should "fail" with "QuotaExceeded"
+        When the "count-items" cronjobs completes without error
```
ghost, line 123: No real other way to test this scenario without this step today (which will need removal with the future scuba implementation / drop of count items)

tests/ctst/features/quotas/Quotas.feature
```diff
@@ -96,3 +96,43 @@ Feature: Quota Management for APIs
...
+        When I wait 3 seconds
+        # At this point if negative inflights are not supported, write should
+        # not be possible, as the previous inflights are now part of the current
+        # metrics.
+        And i delete object "obj-1"
+        And I wait 3 seconds
```
francoisferrand, line 131: why do we need this delay as well?
is it not giving the chance to count-items to execute (and thus make the test randomly pass even w/o inflight) ?

tests/ctst/features/quotas/Quotas.feature
```diff
@@ -96,3 +96,43 @@ Feature: Quota Management for APIs
...
+    @Quotas
+    @CronJob
+    @DataDeletion
+    @NonVersioned
+    Scenario Outline: Quotas are affected by deletion operations between count items runs
+        Given an action "DeleteObject"
+        And a permission to perform the "PutObject" action
+        And a STORAGE_MANAGER type
+        And a bucket quota set to 10000 B
+        And an account quota set to 10000 B
```
francoisferrand: what is the point, we override on next lines?

ghost: Increasing the inflights, if no quota is set, they won't increase, and in this test, we don't run count items

tests/ctst/features/quotas/Quotas.feature
ghost: We need to wait for the inflights to be taken into account by SCUBA, this is not atomic in API calls.

tests/ctst/features/quotas/Quotas.feature
```diff
@@ -96,3 +96,43 @@ Feature: Quota Management for APIs
...
+        Then the API should "fail" with "QuotaExceeded"
+        When the "count-items" cronjobs completes without error
+        When I wait 3 seconds
+        # At this point if negative inflights are not supported, write should
+        # not be possible, as the previous inflights are now part of the current
+        # metrics.
+        And i delete object "obj-1"
+        And I wait 3 seconds
+        And I PUT an object with size <uploadSize>
+        Then the API should "succeed" with ""
```
francoisferrand, line 133: we should also check that negative infligts do not allow to bypass the quota (i.e. we enforce `currnetUsage = max(currentReportedUsage + infligts, 0)`), i.e. if we had an pbject of size 1000, set of quota to 200, then delete the object : we must not be able to push another object of size 1000...

tests/ctst/features/quotas/Quotas.feature
francoisferrand: at line 121, it seems we wait for count items:
> When the "count-items" cronjobs completes without error

what i don't understand is really the "setup" part at the beginning:
* And an account quota set to 10000 B
* And an upload size of 1000 B for the object "obj-1"
* And an account quota set to <accountQuota> B

→ setting the second quota only would do the same, nothing happens in between...

tests/ctst/features/quotas/Quotas.feature
francoisferrand: please add a comment...

```suggestion
        And I wait 3 seconds # wait for inflights to be read by SCUBA
```

tests/ctst/features/quotas/Quotas.feature
```diff
@@ -96,3 +96,43 @@ Feature: Quota Management for APIs
...
+        And a STORAGE_MANAGER type
+        And a bucket quota set to 10000 B
+        And an account quota set to 10000 B
+        And an upload size of 1000 B for the object "obj-1"
+        And a bucket quota set to <bucketQuota> B
+        And an account quota set to <accountQuota> B
+        And a <userType> type
+        And an environment setup for the API
+        And an "existing" IAM Policy that "applies" with "ALLOW" effect for the current API
+        When I wait 3 seconds
```
francoisferrand, line 120:
```suggestion
        When I wait 3 seconds # wait for inflights to be read by SCUBA
```

tests/ctst/features/quotas/Quotas.feature
```diff
@@ -96,3 +96,43 @@ Feature: Quota Management for APIs
...
+        Then the API should "succeed" with ""
+
+        Examples:
+            | uploadSize | bucketQuota | accountQuota | userType |
+            | 100        | 200         | 0            | ACCOUNT  |
+            | 100        | 0           | 200          | ACCOUNT  |
+            | 100        | 200         | 200          | ACCOUNT  |
+            | 100        | 200         | 0            | IAM_USER |
+            | 100        | 0           | 200          | IAM_USER |
+            | 100        | 200         | 200          | IAM_USER |
```
francoisferrand: not sure where to put this feedback on [`development/2.8...w/2.8/improvement/ZENKO-4941`#diff-0f9a8cc11d](https://github.com/scality/Zenko/compare/development/2.8...w/2.8/improvement/ZENKO-4941#diff-0f9a8cc11d2560e9e9ea8668d53be732997ae7cecb4ec7201c836bc8367fc5c9R175)

→ should be tagged as `@2.6.0`, since it should work with zenko 2.6.0 as well (the restriction is in our implementation of tests only)

tests/ctst/features/quotas/Quotas.feature
ghost: - quota of 10000 (which enables the inflights)
- send 1000 (inflight at 1000 after x seconds)
- set quota as 200
- try to put more, should be denied because of quota
- run count items, it should update the "current" and set inflight to 0 (so, 200 quota, 1000 current, 0 inflight)
- delete object (-1000 inflight after X seconds)
- should now be able to put an object

We need the large value at the beginning to have the inflight increasing. We could also set to 1000 directly but for sure we need to have a quota before we put the first object (could be a lower value, doesn't matter)

tests/ctst/features/quotas/Quotas.feature
ghost: Keeping this open, will do it when we have the right projects versions (scuba, mostly) to save time

tests/ctst/features/quotas/Quotas.feature
ghost: I updated to have 1000 at the beginning, not a too large value...

tests/ctst/features/quotas/Quotas.feature
ghost: I added a scenario, the same, but at the end, instead of putting 100B (where we have 1000 current and -1000 inflights, which is thus under the 200B quota, 1000-1000+100 < 200), we put 1000B, and expect a quota exceeded error.

tests/ctst/features/quotas/Quotas.feature
francoisferrand: Oh, do you mean `an upload size of 1000 B for the object "obj-1"` is actually the same as `I PUT an object with size <uploadSize>` ?
written like this, it looks like it just sets a parameter (and what is the point of having a duplicated step?)

Also, the huge sequence of "And" makes it look like just simple prerequisitives (it is even followed by policy setup)...

francoisferrand: > quota of 10000
> send 1000 (inflight at 1000 after x seconds)
> set quota as 200
> try to put more, should be denied because of quota

can do the exact same thing with:
> set quota as 1000
> send 1000 (inflight at 1000 after x seconds)
> try to put more, should be denied because of quota

tests/ctst/features/quotas/Quotas.feature
francoisferrand: countItems is used to reset in-flights, right?

should we not simply create another test for the followup? it would achieve the same purpose, right?

tests/ctst/features/quotas/Quotas.feature
francoisferrand: that scenario would only reach 0, not a negative value though...

what I am worried is kind of race conditions where we could have `countItemsMetrics - inflight < 0` for a few instants : i.e. `0 - 1000 + x < 200` would allow up to x-1200... while for additional safety we should have in the code `max(countItemsValue + inflightValue, 0) < quota`. This was not relevant when inflights were >=0, but we should have that extra check now (preventively)

ghost: I am not sure to understand how we can reach such a state: even if the inflights are not applied directly, s3 ensures we cannot delete more objects that what we have, so if we have 0 "current" and -1000 inflights, it means we accepted to delete 1000 bytes from an object that was not part of the metrics, which is not possible?
At the very least, that is not something we can test here.

ghost: At the end, rejecting any negative global current+inflight may block the client in case we really missed some (existing) s3 object in the "current", and we would have to wait for the next count items run. Unless we suspect we can delete more data than what we really have in s3, which is confusing

ghost: For
```
max(countItemsValue + inflightValue, 0) < quota
```

I am not sure it helps? If we have 0 current, -1000 inflight and try to write:
- `0` -> allowed
- Any value between 1 and 1200 -> allowed (we get a value <= quota). E.g. `Math.max(0 -1000 + 1200, 0) = 200 < 200`
- Any value above -> denied

At the end the check does not add any more protection than what we already have, that is, without the `Math.max`?

tests/ctst/features/quotas/Quotas.feature
ghost: > Oh, do you mean an upload size of 1000 B for the object "obj-1" is actually the same as I PUT an object with size <uploadSize> ?

yes

> Also, the huge sequence of "And" makes it look like just simple prerequisitives (it is even followed by policy setup)...

It is a prerequisite indeed, for this scenario

> can do the exact same thing with:

yes, but here we also (indirectly) test changing quotas, and I'm not against more checks...

tests/ctst/features/quotas/Quotas.feature
ghost: If we only want to accept the "200" after the 0, then we are doing the same as before: we don't accept negative inflights. If we just want to accept the 0-1000 range, then the formula should be updated, but to me it does not make sense...

tests/ctst/features/quotas/Quotas.feature
francoisferrand: if we want to test changing quotas, best make it explicit IMHO
* otherwise somebody will eventually remove it... and reduce coverage
* because the purpose of Cucumber is to actually make an "excecutable spec", so it should be very explciit about what it does & tests : it is not just about having the actual scenario, but also about having a scenarios which "user" can relate to (as is, i don't think you could present this scenario to a PO or customer...)

tests/ctst/features/quotas/Quotas.feature
francoisferrand: it is not about rejecting negative infligts, but about ensuring we never think the "current" usage (i.e. last reported countItem size + inflights) should never be < 0.... which could allow accepting more than the quota!

tests/ctst/features/quotas/Quotas.feature
ghost: I am not sure to understand what you mean by "another test"? We are already testing all cases we can in Zenko, this one is mostly to ensure that after resetting the inflights, so we have negative inflight (but positive current + inflight), we cannot bypass the quota

tests/ctst/features/quotas/Quotas.feature
ghost: As discussed I am closing this for now as we won't do it today

ghost:
ping

tests/ctst/common/hooks.ts
```diff
@@ -54,7 +54,13 @@ After(async function (this: Zenko, results) {
...
 });
 
-After({ tags: '@Quotas' }, async function () {
+After({ tags: '@Quotas' }, async function (this: Zenko, results) {
+    if (results.result?.status === 'FAILED') {
+        this.logger.warn('quota was not cleaned for test', {
+            bucket: this.getSaved<string>('bucketName'),
+        });
+        return;
+    }
```
francoisferrand, line 63: should this not be done in `teardownQuotaScenarios` ?

tests/ctst/features/quotas/Quotas.feature
francoisferrand: what i mean is in the same scenario, after "setup",
* we try to write and expect to be rejected by "inflights" → first scenario
* then we run count-items, wait and delete object (generate negative inflight) → second scenario

tests/ctst/features/quotas/Quotas.feature
```diff
@@ -96,3 +97,78 @@ Feature: Quota Management for APIs
...
+    @PreMerge
+    @Quotas
+    @CronJob
+    @DataDeletion
+    @NonVersioned
+    Scenario Outline: Negative inflights do not allow to bypass the quota
+        Given an action "DeleteObject"
+        And a permission to perform the "PutObject" action
+        And a STORAGE_MANAGER type
+        And a bucket quota set to <uploadSize> B
```
francoisferrand:
```suggestion
        And a bucket quota set to <bucketQuota> B
```

tests/ctst/features/quotas/Quotas.feature
```diff
@@ -96,3 +97,78 @@ Feature: Quota Management for APIs
...
+        And an account quota set to <accountQuota> B
+        And an upload size of <bucketQuota> B for the object "obj-1"
```
francoisferrand:
```suggestion
        And an upload size of <uploadSize> B for the object "obj-1"
```

tests/ctst/steps/utils/kubernetes.ts
```diff
@@ -71,12 +74,39 @@ export function createKubeCustomObjectClient(world: Zenko): CustomObjectsApi {
...
+) {
     const batchClient = createKubeBatchClient(world);
     const watchClient = createKubeWatchClient(world);
+
+    const lockFilePath = path.join('/tmp', `${jobName}.lock`);
+    let releaseLock: (() => Promise<void>) | false = false;
+
+    if (!fs.existsSync(lockFilePath)) {
+        fs.writeFileSync(lockFilePath, '');
+    }
```
francoisferrand, line 103: if the lock file exists, it means another test is waiting for this job : in that case, should the function not simply wait (polling) until the lock is removed? (i.e. the "first" test has completed its wait, and the job has indeed completed)

tests/ctst/common/hooks.ts
ghost: I don't think so, we don't need to parse the results from cucumberjs in this function, we want to teardown when applicable, similar to "cleanS3Bucket"

tests/ctst/steps/utils/kubernetes.ts
ghost: The problem with polling like that is that two tests may detect that the file does not exist at the same time, hence the use of the lock file, that internally relies on directory creation (which is atomic) and last modified dates, to ensure the whole approach is safe.

tests/ctst/features/quotas/Quotas.feature
ghost: If I understand correctly, you want to handle the same bucket and objects in a separate scenario, so we also run count items in-between, and have 2 distinct scenario, one focusing on inflights, and one on the actual negative inflights?

If yes then this is against the best practices of cucumberjs test: they should all work with their set of resources, and not rely on one another.

If we just want a scenario specific to the "negative inflight", then we can just delete these lines:
```
        When I wait 3 seconds
        And I PUT an object with size <uploadSize>
        Then the API should "fail" with "QuotaExceeded"
```
However having them also ensure we properly clean the inflights when the request fails, and thus it should not impact what we have after count items. This is the scenario close to what a human would do: put object, hit the quota, then the count items runs, and finally, we should be able to delete data and be allowed to write again.

tests/ctst/steps/utils/kubernetes.ts
francoisferrand: instead of `if (!fs.existsSync(lockFilePath)) { fs.writeFileSync(lockFilePath, '');`, we can do `fs.writeFileSync(path, data, { overwrite: false }` : to atomically write the file without overwriting, so only one will ever succeed.
--> depending on the "outcome" of the creation, the code can decide if it should proceed with reading the cron/... or if it should simply wait because something else is already doing it.

tests/ctst/common/common.ts
```diff
@@ -353,8 +353,7 @@ When('the user tries to perform the current S3 action on the bucket {int} times
             }
             await runActionAgainstBucket(this, this.getSaved<ActionPermissionsType>('currentAction').action);
             if (this.getResult().err) {
```
francoisferrand: maybe better to keep failing on error, except for "retriable" errors : which can be safey ignore as this step is indeed about making multiple attempts

ghost: I updated CTST and now we fail if the error is not retryable

tests/ctst/steps/utils/kubernetes.ts
ghost: I switch to using a file with the `wx` flag, let's see...

tests/ctst/common/hooks.ts
```diff
@@ -17,7 +17,7 @@ import { cleanS3Bucket } from './common';
 process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0';
 
 const { atMostOnePicklePerTag } = parallelCanAssignHelpers;
-const noParallelRun = atMostOnePicklePerTag(['@AfterAll', '@PRA', '@ColdStorage']);
+const noParallelRun = atMostOnePicklePerTag(['@AfterAll', '@PRA', '@ColdStorage', '@Utilization']);
```
KillianG: As per guidelines, you should not be using `atMostOnePicklePerTag` otherwise it means that you test breaks rule #1 of idempotency..

ghost: Yes, not needed since we lock the execution of the jobs, I removed it and tagged all the scenarios appropriately

tests/ctst/steps/utils/kubernetes.ts
ghost: Build is green, so I am resolving this comment.

KillianG (approved):

welansari (approved):

ghost:
ping

PR #2182, opened 2024-12-12
How to write good tests for parallel environments

ghost:

With 10 rules, ensure the new tests in CTST are following a set of rule to ensure their quality.

Fixes ZENKO-4944.

Please feel free to suggest changes or new rules, or start a discussion if you disagree with a rule.

### DIFF
--- a/tests/ctst/HOW_TO_WRITE_TESTS.md
+++ b/tests/ctst/HOW_TO_WRITE_TESTS.md
@@ -91,10 +91,15 @@ during the sleep time.
 
 > Avoid using the @Flaky tag.
 
-If a test is marked as flaky, it is likely not idempotent, or we did not
-do our job correctly to understand the problem. Please fix the test.
-
-If an issue exists in the product, prefer fixing the bug first.
+Whenever there is a flaky, there is an issue : either a bug (corner case or
+race condition) in the product, or simply an issue in the test. So the
+issue should be fixed either way. More often than not, retrying the build
+(i.e. ignoring the flaky) or automating the retry with @Flaky is just
+masking a bug, and leaving it in production instead of fixing it : so it
+must not be done. One frequent problem in flaky tests is that they are not
+idempotent, or that we did not do our job correctly to understand the
+problem. Flaky tests must be investigated, and issues in the tests must be
+fixed to ensure we don't ignore an actual problem in the product.
 
 **Rule #8**
 
@@ -118,5 +123,5 @@ specific tags.
 
 > Validate your test before merging.
 
-If a test pass once, please ensure it passes for the good reasons.
+If a test passes once, please ensure it passes for the good reasons.
 False positives are hard to detect once merged.
--- a/tests/ctst/HOW_TO_WRITE_TESTS.md
+++ b/tests/ctst/HOW_TO_WRITE_TESTS.md
@@ -7,9 +7,7 @@ parallel with other tests.
 *This is specific to Zenko, but the rules are applicable for any suite of*
 *tests in parallel.*
 
-**Rule #1**
-
-> Tests must be idempotent.
+## 1. Tests must be idempotent.
 
 In parallel, the tests can run in any order. It's not because a test works at a
 given time, that it will work if the test order changes after adding new ones
@@ -17,7 +15,7 @@ or renaming files. You must ensure the scenarios are fully idempotent.
 "Flakies" are mostly due to non-idempotent tests, then bugs. It is rarely due
 to the platform.
 
-**Rule #2**
+## 2. Use unique resources names. S3 Bucket, Account and S3 Objects should be unique.
 
 > Use unique resources names. S3 Bucket, Account and S3 Objects should be
 > unique.
@@ -27,26 +25,21 @@ For cross-account scenarios, use unique S3 Policies and S3 Roles.
 
 If data is used, such as lock files, ensure they are named appropriately.
 
-**Rule #3**
-
-> Avoid reconfiguring the environment during test execution.
+## 3. Do not reconfigure the environment if it affects other tests' resources.
 
 If a reconfiguration triggers a reconciliation of the resources, running tests
 might be affected. Also, it restarts the services, which leads to missing logs,
 harder debugging, and longer test execution.
 
-Examples:
+For examples, location creation or bucket Website endpoints, that triggers a
+restart of multiple services, must be created before the test starts, or in a
+dedicated set of tests, that is, a set of test executed before all tests having
+a specific tag used for identifying the feature(s).
 
-- Location creation;
-- Bucket Website endpoints.
+Note: Testing the reconfiguration of the environment is recommended, but it
+should be done carefully to not affect other tests.
 
-... must be created before the test start, or in a dedicated set of tests,
-that is, a set of test executed before all tests having a specific tag
-used for identifying the feature(s).
-
-**Rule #4**
-
-> Do not use `atMostOnePicklePerTag`.
+## 4. Do not use `atMostOnePicklePerTag`.
 
 If a set of scenario requires the use of `atMostOnePicklePerTag`, they won't
 be executed in parallel, which is:
@@ -66,30 +59,28 @@ possible. Solutions exist:
   relative checks.
 - As a last resort, we might have a dedicated test suite.
 
-**Rule #5**
+## 5. Focus on validating features.
 
-> Focus on validating features.
+We only want to assert against externally visible state, as given in the
+feature requirements, and not rely on internal state or implementation
+details.
 
 It is tempting to write scenarios that test more than what is needed
-at the acceptance testing layer: please use unit and functional tests to fully
+at the acceptance testing layer: use unit and functional tests to fully
 validate the behavior of each component; and end-to-end tests only to ensure
 that the product-level feature is working as expected per the requirements.
 Making good use of unit and functional tests makes testing all cases, including
 corner cases or error handling, much easier, leading to higher quality; and
 improves velocity by having a much faster feedback loop.
 
-**Rule #6**
-
-> Ensure assertions are idempotent after sleep times.
+## 6. Ensure assertions are idempotent after sleep times.
 
 Some tests might need to "sleep" for some time, either static or using
 more complex logic. The subsequent assertions must take into account
 any operation on the platform that might be asynchronous and take place
 during the sleep time.
 
-**Rule #7**
-
-> Avoid using the @Flaky tag.
+## 7. Avoid using the @Flaky tag.
 
 Whenever there is a flaky, there is an issue : either a bug (corner case or
 race condition) in the product, or simply an issue in the test. So the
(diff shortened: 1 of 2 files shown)


tests/ctst/HOW_TO_WRITE_TESTS.md
francoisferrand, line 1: should this instead go in a more "natural" location, e.g. `docs/writing-ctst-tests.md` ?

tests/ctst/HOW_TO_WRITE_TESTS.md
```diff
@@ -0,0 +1,121 @@
...
+Writting intergation / behavior tests that run in parallel is not trivial.
+This page lists all the rules to follow when writing a test that runs in
+parallel with other tests.
+
+*This is specific to Zenko, but the rules are applicable for any suite of*
+*tests in parallel.*
+
+**Rule #1**
+
+> Tests must be idempotent.
```
francoisferrand:
```suggestion
> Tests should be idempotent.
```

tests/ctst/HOW_TO_WRITE_TESTS.md
```diff
@@ -0,0 +1,121 @@
...
+If a reconfiguration triggers a reconciliation of the resources, running tests
+might be affected. Also, it restarts the services, which leads to missing logs,
+harder debugging, and longer test execution.
+
+Examples:
+
+- Location creation;
+- Bucket Website endpoints.
+
+... must be created before the test start, or in a dedicated set of test.
```
francoisferrand:
```suggestion
... should be created before the test start, or in a dedicated set of test.
```

francoisferrand: > or in a dedicated set of test

not clear what you mean by "set of test" here : a separate test suite, a separate execution of ctst , ... ?
having separate executions should be avoided, as it makes it more likely that not all tested are executed (when developping locally, once we can do it easily...)

tests/ctst/HOW_TO_WRITE_TESTS.md
```diff
@@ -0,0 +1,121 @@
...
+- Use `Before` / `After` hooks to setup your environment once.
+  Concurrency can be controlled with lock files.
+- Make your assertions more flexible: strict and absolute checks are good
+  for unit or functional testing, but in integration tests, prefer using
+  relative checks.
+- As a last resort, we might have a dedicated test suite.
+
+**Rule #5**
+
+> Avoid scenarios with no value.
```
francoisferrand: i think the last line of the description is a better name for the rule ("no value" is not an appropriate description: doing many tests in zenko still has more value than no tests at all...)

```suggestion
> Focus on validating features
```

tests/ctst/HOW_TO_WRITE_TESTS.md
```diff
@@ -0,0 +1,121 @@
...
+
+It is tempting to write scenarios that test more than what is needed
+at the integration testing layer: please trust your unit and functional
```
francoisferrand: > integration testing layer

this is not really "integration tests" : as far as semantics go, these tests are more akin to "acceptance tests" and "smoke tests".... which we rely on to validate that components are properly integrated.
Best rephrase to avoid the confusion IMHO.

> please trust your unit and functional

"trust" is probably not what we want : if we really trusted, we wouldn't do any test :-)
```
please use unit and functional tests to fully validate the behavior of each component;
and e2e tests only to ensure that the product-level feature is working as expected per
the requirements. Making good use of unit and functional tests makes testing all cases,
including corner cases or error handling, must easier, leading to higher quality; and
improves velocity by having a much faster feedback loop.
```

tests/ctst/HOW_TO_WRITE_TESTS.md
```diff
@@ -0,0 +1,121 @@
...
+> Ensure assertions are idempotent after sleep times.
+
+Some tests might need to "sleep" for some time, either static or using
+more complex logic. The subsequent assertions must take into account
+any operation on the platform that might be asynchronous.
+
+Example: we run a Kubernetes job in tests that look idempotent. However,
+the job itself might not be idempotent and run in parallel, leading to
+some resources being skipped by a previous job that would take longer than
+expected, and the test results will differ from what is expected.
```
francoisferrand: not sure the exemple looks so clear, for an "outside" eye with no content on which job and what they do...

tests/ctst/HOW_TO_WRITE_TESTS.md
```diff
@@ -0,0 +1,121 @@
...
+
+**Rule #7**
+
+> Avoid using the @Flaky tag.
+
+If a test is marked as flaky, it is likely not idempotent, or we did not
+do our job correctly to understand the problem. Please fix the test.
+
+If an issue exists in the product, prefer fixing the bug first.
```
francoisferrand:
```suggestion
Whenever there is a flaky, there is an issue : either a bug (corner case or race condition) in the product,
or simply an issue in the test. So the issue should be fixed either way.

More often than not, retrying the build (i.e. ignoring the flaky) or automating the retry with @Flaky is just
masking a bug, and leaving it in production instead of fixing it : so it must not be done.

One frequent problem in flaky tests is that they are not idempotent, or that we did not do our job correctly to understand the problem. Flaky tests must be investigated, and issues in the tests must be fixed to ensure we
don't ignore an actual problem in the product.
```

tests/ctst/HOW_TO_WRITE_TESTS.md
```diff
@@ -0,0 +1,121 @@
...
+
+Tests requiring to run at the very end of a parallel test suite with a specific
+naming (e.g., `zzz.*`) should be removed. Prefer using `After` hooks with
+specific tags.
+
+**Rule #10**
+
+> Validate your test before merging.
+
+If a test pass once, please ensure it passes for the good reasons.
```
francoisferrand:
```suggestion
If a test passes once, please ensure it passes for the good reasons.
```

francoisferrand (approved):

tests/ctst/HOW_TO_WRITE_TESTS.md
ghost: Why "should" and not "must"? it is not supposed to be a suggestion.

tests/ctst/HOW_TO_WRITE_TESTS.md
ghost: By "set of test" I mean tests with a specific tag, or a single feature file (it can change, but basically, if multiple tests are using the same resource, they should likely have the same tag...)
I'll update

tests/ctst/HOW_TO_WRITE_TESTS.md
ghost: The current docs folder is used with tox, it's a real doc we used to render as html pages, putting it here is logical but it doesn't look good...

tests/ctst/HOW_TO_WRITE_TESTS.md
ghost: I removed the example, it's too specific for an external dev

welansari (approved):

tests/ctst/HOW_TO_WRITE_TESTS.md
```diff
@@ -0,0 +1,127 @@
+# How to write a test that runs in parallel
+
+Writting intergation / behavior tests that run in parallel is not trivial.
+This page lists all the rules to follow when writing a test that runs in
+parallel with other tests.
+
+*This is specific to Zenko, but the rules are applicable for any suite of*
+*tests in parallel.*
+
+**Rule #1**
```
rachedbenmustapha: Just a general formatting remark: it would be more straightforward for readers to just name the rules as titles prefixed by their number. I find the bold rule number followed by the rule itself as a quote a bit distracting.

tests/ctst/HOW_TO_WRITE_TESTS.md
```diff
@@ -0,0 +1,127 @@
...
+harder debugging, and longer test execution.
+
+Examples:
+
+- Location creation;
+- Bucket Website endpoints.
+
+... must be created before the test start, or in a dedicated set of tests,
+that is, a set of test executed before all tests having a specific tag
+used for identifying the feature(s).
```
rachedbenmustapha: Unless we're explicitly testing the reconfiguration itself which we should probably do.

Also, running tests should not be affected, if they are it means that production will, when an upgrade or config change is made.

I agree with the debuggability/log availability part, but they're issues that need to be addressed, not really a reason to avoid reconfiguring altogether (provided of course that the reconfiguration does not touch other tests' resources).

tests/ctst/HOW_TO_WRITE_TESTS.md
```diff
@@ -0,0 +1,127 @@
...
+- As a last resort, we might have a dedicated test suite.
+
+**Rule #5**
+
+> Focus on validating features.
+
+It is tempting to write scenarios that test more than what is needed
+at the acceptance testing layer: please use unit and functional tests to fully
+validate the behavior of each component; and end-to-end tests only to ensure
+that the product-level feature is working as expected per the requirements.
```
rachedbenmustapha, line 71: Maybe it would be beneficial to clearly spell out that we only want to assert against externally visible state (as given in the feature's requirements) and NOT rely on internal state introspection, which end-users don't really care about and subject to change.

tests/ctst/HOW_TO_WRITE_TESTS.md
```diff
@@ -0,0 +1,127 @@
...
+idempotent, or that we did not do our job correctly to understand the
+problem. Flaky tests must be investigated, and issues in the tests must be
+fixed to ensure we don't ignore an actual problem in the product.
+
+**Rule #8**
+
+> Teardown resources.
+
+If a test creates resources, ensure they are deleted at the end of the test.
+Do not delete the resources in case of error, to ease debugging.
```
rachedbenmustapha: I both agree and disagree: tearing down resources should always be done and be reliable, and also we need to be able to debug. A middle ground would be to skip teardown on failure only if in debug mode (the one with tmate active), otherwise always do it.

This would make rule #9 redundant as tests relying on other tests to have run would just not work.

rachedbenmustapha (commented):
General formatting/presentation remarks:
- It would be more straightforward for readers to just name the rules as titles prefixed by their number. I find the bold rule number followed by the rule itself as a quote a bit distracting.
- The word please is not needed in this sort of document (or any documentation or UI really) and sets a tone of asking for a favor vs giving directions that everyone needs to follow. A former tech writer on the team made sure that we removed all pleases from docs (even internal/technical) and UIs for consistency.

benzekrimaha (approved):

tests/ctst/HOW_TO_WRITE_TESTS.md
ghost: Thank you for all the feedback. I updated the page accordingly - I agree with everything but this. I updated as you suggested though.

I think we should not cleanup resources in case of error even if we're not running debug. The reason is that as each test uses a partitioned set of resources (bucket/objects, possibly account), not cleaning in case of error (or flaky) helps us finding the root cause much more easily, by having access to the metadata in the mongodb dumps, as well as the logs of all operations done by CTST for the failing test (identifiable with a single id). The overhead is minimal and it should not affect other tests in any way, even if the bucket is still here.

tests/ctst/HOW_TO_WRITE_TESTS.md
```diff
@@ -0,0 +1,108 @@
...
+
+*This is specific to Zenko, but the rules are applicable for any suite of*
+*tests in parallel.*
+
+## 1. Tests must be idempotent.
+
+In parallel, the tests can run in any order. It's not because a test works at a
+given time, that it will work if the test order changes after adding new ones
+or renaming files. You must ensure the scenarios are fully idempotent.
+"Flakies" are mostly due to non-idempotent tests, then bugs. It is rarely due
```
KillianG, line 15: > it is rarely due to the platform

Even though it's *actually* true, I'm not sure we want to write that in the documentation, as the test are made to verify that the platform works..

tests/ctst/HOW_TO_WRITE_TESTS.md
```diff
@@ -0,0 +1,108 @@
+# How to write a test that runs in parallel
+
+Writting intergation / behavior tests that run in parallel is not trivial.
```
KillianG:
```suggestion
Writing integration / behavior tests that runs in parallel is not trivial.
```

tests/ctst/HOW_TO_WRITE_TESTS.md
ghost: By platform I mean the CI environment, I will make it clearer that we cannot always think it's due to limited resources in the CI, or some sort of slowness we cannot explain

PR #2186, opened 2025-01-21
Deploy metadata in end to end tests

welansari:
- Deploy metadata from chart
- Wait for metadata components to start in order : repd -> bucketd -> cloudserver
- Manually add the proper endpoint in Cloudserver config
- Only deploy metadata in `end2end-sharded`

Issue: ZENKO-4414

### DIFF
--- a/.github/actions/deploy/action.yaml
+++ b/.github/actions/deploy/action.yaml
@@ -83,6 +83,10 @@ runs:
       shell: bash
       run: sh tests/smoke/deploy-sorbet-resources.sh end2end
       working-directory: ./.github/scripts/end2end/operator
+    - name: Deploy metadata
+      shell: bash
+      run: bash deploy-metadata.sh
+      working-directory: ./.github/scripts/end2end
     - name: End-to-end configuration
       shell: bash
       run: bash configure-e2e.sh "end2end" ${E2E_IMAGE_NAME}:${E2E_IMAGE_TAG} "default"
--- a/.github/actions/deploy/action.yaml
+++ b/.github/actions/deploy/action.yaml
@@ -7,6 +7,10 @@ inputs:
     description: "The tag of the Zenko Operator image to use"
     required: false
     default: ""
+  deploy_metadata:
+    description: "Deploy a metadata cluster alongside Zenko"
+    required: false
+    default: "false"
 runs:
   using: composite
   steps:
@@ -85,8 +89,9 @@ runs:
       working-directory: ./.github/scripts/end2end/operator
     - name: Deploy metadata
       shell: bash
-      run: bash deploy-metadata.sh
+      run: ./deploy-metadata.sh
       working-directory: ./.github/scripts/end2end
+      if: ${{ inputs.deploy_metadata == 'true' }}
     - name: End-to-end configuration
       shell: bash
       run: bash configure-e2e.sh "end2end" ${E2E_IMAGE_NAME}:${E2E_IMAGE_TAG} "default"
--- a/.github/scripts/end2end/deploy-metadata.sh
+++ b/.github/scripts/end2end/deploy-metadata.sh
@@ -1,0 +1,36 @@
+#!/bin/sh
+
+set -exu
+
+# create a separate namespace for metadata
+kubectl create namespace metadata
+
+# clone the metadata repository
+git init metadata
+cd metadata
+git fetch --depth 1 --no-tags https://${GIT_ACCESS_TOKEN}@github.com/scality/metadata.git
+git checkout FETCH_HEAD
+
+# install metadata chart in a separate namespace
+cd helm
+helm dependency update cloudserver/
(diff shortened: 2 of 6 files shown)


.github/actions/deploy/action.yaml
```diff
@@ -83,6 +83,11 @@ runs:
       shell: bash
       run: sh tests/smoke/deploy-sorbet-resources.sh end2end
       working-directory: ./.github/scripts/end2end/operator
+    - name: Deploy metadata
+      shell: bash
+      run: bash deploy-metadata.sh
```
francoisferrand: We are already running with bash, so make it executable and execute just:

```suggestion
      run: ./deploy-metadata.sh
```

.github/actions/deploy/action.yaml
```diff
@@ -83,6 +83,11 @@ runs:
...
+      working-directory: ./.github/scripts/end2end
+      if: ${{ env.ENABLE_RING_TESTS == 'true' }}
```
francoisferrand: better to have it as an `input` rather than an env variable
(the input parameter may however takes its "default" value from env)

.github/scripts/end2end/deploy-metadata.sh
```diff
@@ -0,0 +1,47 @@
...
+
+# current chart uses an old version of bucketd that has issues reconnecting to the repd
+# when bucketd is started first. Restarting bucketd after repd is ready.
+kubectl -n metadata rollout restart deployment/s3c-metadata-bucketd
+# wait for the bucketd pods to be created
+kubectl -n metadata rollout status --watch --timeout=300s deploy/s3c-metadata-bucketd
+# wait for all bucketd pods to start serving port 9000
+wait_for_all_pods_behind_services metadata-bucketd metadata 9000 60
+
+# manually add "s3c.local" to the rest endpoints list as it's not configurable in the chart
```
francoisferrand, line 37: why do we need `s3c.local` ?

i did not need such patch on my artesca cluster (in namespace `default` i think), just installed the chart and created the artesca location pointed at `s3c-cloudserver.default.svc.cluster.local`

.github/workflows/end2end.yaml
```diff
@@ -556,8 +553,9 @@ jobs:
     needs: [build-kafka, build-test-image]
     runs-on:
       - ubuntu-22.04-8core
-      # Enable this for Ring-based tests
-      # - scality-cloud
+    env:
+      ENABLE_RING_TESTS: "true"
+      GIT_ACCESS_TOKEN: ${{ secrets.GIT_ACCESS_TOKEN }}
```
francoisferrand: credentials should not be added to env, esp. at such broad "scale" : best to pass them via inputs, in the place where they are needed

.github/scripts/end2end/deploy-metadata.sh
welansari: Cloudserver responds with an error when the host used is not declared in its config file. I've tried with the k8s service endpoint and it doesn't work.

francoisferrand: maybe this comes from the namespace ("metadata") you deploy to, which may not be taken into account?
in my tests this is using the default namespace, with no issue...

francoisferrand (approved):

.github/scripts/end2end/deploy-metadata.sh
welansari: No even with the default namespace i got the same behaviour, which to me, is the expected behaviour. The normal cloudserver we deploy in Zenko has all the k8s endpoints in its config.

.github/scripts/end2end/deploy-metadata.sh
```diff
@@ -0,0 +1,47 @@
...
+. "$(dirname $0)/common.sh"
+
+# create a separate namespace for metadata
+kubectl create namespace metadata
+
+# clone the metadata repository
+git init metadata
+cd metadata
+git fetch --depth 1 --no-tags https://${GIT_ACCESS_TOKEN}@github.com/scality/metadata.git
+git checkout FETCH_HEAD
```
ghost, line 14: Do we want to support a specific metadata version? What if the latest commit of the default branch introduced a breaking change or a bug/limitation to be fixed in another PR before a release is made?

ghost (approved):

.github/scripts/end2end/deploy-metadata.sh
francoisferrand: probably not worth it:
* this chart is mostly not maintained
* it references "unknown" version of metadata
* ...consequently this is not a long term solution, just a stop-gap measure until we switch to using metadata-operator to deploy a "recent" metadata : but at least it allows us to test OOB already...

ghost: No problem, I'm fine if that's not that maintained

PR #2187, opened 2025-01-24
Enable OOB tests

welansari:
- Enable OOB tests
- Increase the timeout of OOB tests as replication of operations between the repds can be slow when the CI vm is saturated (an operation is only visible in the raft log once it got replicated to the majority of the repds)
- Wait for the ingestion consumer group to be in a stable state before starting the tests. Zenko is reconciled just before launching the tests which causes the ingestion pods to restart and the consumer group to rebalance. The consumer group stays in the rebalance state for 30-ish seconds which is enough time for the tests to start. Not waiting for the state to be stable can cause the first messages pushed to the topic to be skipped (consumers start consuming late because of the rebalance) as the consumers are configured to start consuming from the latest offset if no offset was found (which is the case at startup).
- When we wait for the consumer group state to be stable we also check if we have the proper number of consumers connected to the consumer group. When the ingestion pods restart from the reconcile, the previous consumers are kept in the consumer group until their session times out. Not waiting for this can cause messages to be skipped as well, as the new consumer(s) only have some partitions assigned to them and not all, the rest are assigned to the "dead" consumer until future rebalance. At startup no offset is generally stored yet for the partitions assigned to the "dead" consumer, so when the rebalance occurs and the new consumer get assigned to them the new offset becomes the latest and previous messages are skipped.

Issue: ZENKO-4286

### DIFF
--- a/.github/scripts/end2end/configure-e2e.sh
+++ b/.github/scripts/end2end/configure-e2e.sh
@@ -110,6 +110,9 @@ sleep 10
 kubectl wait --for condition=DeploymentFailure=false --timeout 25m -n ${NAMESPACE} zenko/${ZENKO_NAME}
 kubectl wait --for condition=DeploymentInProgress=false --timeout 25m -n ${NAMESPACE} zenko/${ZENKO_NAME}
 
-# wait for ingestion processor to start consuming from Kafka
-ingestion_processor_replicas=$(kubectl -n $NAMESPACE get deploy/end2end-backbeat-ingestion-processor -o jsonpath='{.spec.replicas}')
-wait_for_consumer_group $NAMESPACE $UUID.backbeat-ingestion-group $ingestion_processor_replicas 300
+
+if [ $ENABLE_RING_TESTS = true ]; then
+  # wait for ingestion processor to start consuming from Kafka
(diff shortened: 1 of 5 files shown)


.github/scripts/end2end/configure-e2e.sh
```diff
@@ -109,3 +109,10 @@ sleep 10
 
 kubectl wait --for condition=DeploymentFailure=false --timeout 25m -n ${NAMESPACE} zenko/${ZENKO_NAME}
 kubectl wait --for condition=DeploymentInProgress=false --timeout 25m -n ${NAMESPACE} zenko/${ZENKO_NAME}
+
+
+if [ $ENABLE_RING_TESTS = true ]; then
+  # wait for ingestion processor to start consuming from Kafka
```
francoisferrand, line 115: this should (eventually) get baked into backbeat?
--> so we should probably add a ticket (and we can put it in the EPIC "cleanup kafka management" https://scality.atlassian.net/browse/ARTESCA-9180)

francoisferrand (approved):

ghost (approved):

.github/scripts/end2end/configure-e2e.sh
welansari: https://scality.atlassian.net/browse/BB-655

PR #2190, opened 2025-01-29
Bump UI to 3.1.2

JBWatenbergScality:
Issue: ZENKO-4973


hervedombya (approved):

VERSION
```diff
@@ -1,5 +1,5 @@
-VERSION="2.10.12"
+VERSION="2.11.0"
 
-VERSION_SUFFIX=
+VERSION_SUFFIX=-beta.1
```
francoisferrand: * should be a preview: `VERSION_SUFFIX=-preview.1`
* the "VERSION" commit should be the last one of the PR, and the one with the release : so it is easy to read in changelog (esp. when multiple branches are involved)

francoisferrand (approved):

PR #2194, opened 2025-02-20
Add OOB versioning suspended tests

welansari:
Issue: ZENKO-4330


tests/zenko_tests/node_tests/backbeat/tests/ingestion/ingestionS3C.js
```diff
@@ -9,27 +9,41 @@ const scalityUtils = new IngestionUtility(scalityS3Client, ringS3Client);
...
-            ingestionSrcBucket,
-            null,
-            next,
-        ),
-        next => scalityUtils.waitUntilEmpty(INGESTION_DEST_BUCKET, next),
-        next => scalityUtils.deleteVersionedBucket(INGESTION_DEST_BUCKET, next),
-    ], done));
+    afterEach(function (done) {
+        async.series([
+            next => ringS3CUtils.deleteAllVersions(
```
francoisferrand, line 36: even if the location is created before, since we delete here we should also create the objects here : so the test may be (re)played multiple times...

welansari: Putting the objects is not something we can do in this case. The whole reason behind putting the objects before the location gets created is to have a specific type of null version that doesn't have an internal version id, this type of object can only be created when the bucket is non versioned which cannot happen once versioning gets enabled on it.

I want to avoid having to create the location mid tests for flakiness reasons which is why i went with this approach.

francoisferrand: in that case we should not delete the objects either : so that we can run the test again...

that said, location creation is not flaky as far as I know : this was done to test cold location creation, and the flakiness issue was dealt with (i think)... So we now (in ctst tests) how to do this...

francoisferrand (approved):

.github/workflows/end2end.yaml
```diff
@@ -81,6 +81,9 @@ env:
   RING_S3C_ENDPOINT: http://s3c.local:8000
   RING_S3C_BACKEND_SOURCE_LOCATION: rings3cbackendingestion
   RING_S3C_INGESTION_SRC_BUCKET_NAME: ingestion-test-src-bucket-${{ github.run_id }}
+  RING_S3C_BACKEND_SOURCE_NON_VERSIONED_LOCATION: rings3cbackendingestionnonversioned
+  RING_S3C_INGESTION_SRC_NON_VERSIONED_BUCKET_NAME: ingestion-test-src-non-versioned-bucket-${{ github.run_id }}
```
ghost: the github run_id is an incrementing number (currently 11 char), meaning the bucket name length here is already 52 characters. To be future proof, maybe good to ensure the length cannot exceed 63 chars (the max for an s3 bucket name)

ghost (approved):

PR #2196, opened 2025-02-24
Improvement/zenko 4950 mongodb v7.0

benzekrimaha:
Issue: ZENKO-4950


solution-base/mongodb/charts/mongodb-sharded/values.yaml
```diff
@@ -520,7 +607,7 @@ configsvr:
   serviceAccount:
     ## @param configsvr.serviceAccount.create Specifies whether a ServiceAccount should be created for Config Server
     ##
-    create: false
+    create: true
```
ghost, line 610: Not sure we want to auto create a service account for the config server, as we maintain it ourselves

solution-base/mongodb/charts/mongodb-sharded/values.yaml
```diff
@@ -859,7 +978,7 @@ mongos:
   serviceAccount:
     ## @param mongos.serviceAccount.create Whether to create a Service Account for mongos automatically
     ##
-    create: false
+    create: true
```
ghost, line 981: Same

solution-base/mongodb/charts/mongodb-sharded/values.yaml
```diff
@@ -1135,7 +1287,7 @@ shardsvr:
     serviceAccount:
       ## @param shardsvr.dataNode.serviceAccount.create Specifies whether a ServiceAccount should be created for shardsvr
       ##
-      create: false
+      create: true
```
ghost, line 1290: Same

solution-base/mongodb/charts/mongodb-sharded/values.yaml
```diff
@@ -1430,7 +1613,7 @@ shardsvr:
     serviceAccount:
       ## @param shardsvr.arbiter.serviceAccount.create Specifies whether a ServiceAccount should be created for shardsvr arbiter nodes
       ##
-      create: false
+      create: true
```
ghost, line 1616: Same

solution-base/mongodb/charts/mongodb-sharded/charts/common/README.md
```diff
@@ -24,18 +24,18 @@ data:
...
 Bitnami charts can be used with [Kubeapps](https://kubeapps.dev/) for deployment and management of Helm Charts in clusters.
 
-Looking to use our applications in production? Try [VMware Application Catalog](https://bitnami.com/enterprise), the enterprise edition of Bitnami Application Catalog.
-
 ## Prerequisites
 
-- Kubernetes 1.19+
-- Helm 3.2.0+
+- Kubernetes 1.23+
+- Helm 3.8.0+
```
ghost, line 38: We use helm 3.5 in the dev container and the requirements.sh file, we should bump them to ensure it works

ghost: Same for `KUBECTL_VERSION`

solution-base/mongodb/Makefile
```diff
@@ -3,7 +3,7 @@ CHART_DIR:="${ROOT_DIR}/charts"
 
 CHART_REPO:="https://charts.bitnami.com/bitnami"
 CHART_MONGO_REPLICASET_VERSION:="7.8.0"
```
francoisferrand: not needed/used anymore (since we remove the non-sharded mongo support), right?

solution-base/mongodb/charts/mongodb-sharded/values.yaml
benzekrimaha: done here : https://github.com/scality/Zenko/pull/2196/commits/432b26449084a8afb635c12c535d2e04a9e2ceca

solution-base/mongodb/charts/mongodb-sharded/values.yaml
benzekrimaha: done here: https://github.com/scality/Zenko/pull/2196/commits/432b26449084a8afb635c12c535d2e04a9e2ceca

solution-base/mongodb/charts/mongodb-sharded/values.yaml
benzekrimaha: done here : https://github.com/scality/Zenko/pull/2196/commits/432b26449084a8afb635c12c535d2e04a9e2ceca

solution-base/mongodb/charts/mongodb-sharded/values.yaml
benzekrimaha: done here: https://github.com/scality/Zenko/pull/2196/commits/432b26449084a8afb635c12c535d2e04a9e2ceca

solution-base/mongodb/charts/mongodb-sharded/charts/common/README.md
benzekrimaha: done here : https://github.com/scality/Zenko/pull/2196/commits/3977e3c51de7d518fb2cbb03dcbe227351ea63c4

solution-base/mongodb/Makefile
benzekrimaha: done here : https://github.com/scality/Zenko/pull/2196/commits/7b84f3feb64f3ecfd7c39dc8f2c55d3c701deb1b

solution-base/mongodb/charts/mongodb-sharded/charts/common/README.md
ghost: LGTM so far, I'm waiting for the green build to ensure we are not missing anything

.devcontainer/devcontainer.json
ghost, line 1: (picking random file)

I reviewed the build and I see two things:
- the logging level from MongoDB seem different, as we have no log at all after startup (for integration tests that completed without error), see https://artifacts.scality.net/builds/github:scality:Zenko:staging-d731425ba1.build-iso-and-end2end-test.6699/data/end2end-pra.1/kind-logs/kind-export/kind-worker2/pods/default_data-db-mongodb-sharded-shard0-data-0_bd428108-7ec9-413a-8f14-c8dba9243f28/mongodb/0.log

- I see
```
MongoServerError[Location5624102]: Failed to reconfig: Replica set config contains customized getLastErrorDefaults, which has been deprecated and is now ignored. Use setDefaultRWConcern instead to set a cluster-wide default writeConcern.
```

This is something we set in our custom `solution-base/mongodb/charts/mongodb-sharded/files/docker-entrypoint-initdb.d/set-default-write-concern-majority-sharded.sh`, we should review this, and possibly remove this option if that's deprecated (or replace it if an alaternative exists).

VERSION
```diff
@@ -1,5 +1,5 @@
 VERSION="2.11.0"
```
francoisferrand, line 1: VERSION should not be modified (only if we do the release, which i am sure goes in this PR: since we need other bumps)

solution-base/mongodb/charts/mongodb-sharded/values.yaml
```diff
@@ -229,26 +235,25 @@ common:
     annotations: {}
     ## @param common.serviceAccount.automountServiceAccountToken Automount service account token for the server service account
     ##
-    automountServiceAccountToken: true
-
+    automountServiceAccountToken: false
```
francoisferrand, line 238: this was changed to "stop" auto-mounting service account token : what is the purpose / is there an impact?

esp. considering we do not actually create a `ServiceAccount` (serviceAccount.create == false), so I would think not?

solution-base/mongodb/patches/service-account.patch
```diff
@@ -0,0 +1,49 @@
+diff --git a/solution-base/mongodb/charts/mongodb-sharded/values.yaml b/solution-base/mongodb/charts/mongodb-sharded/values.yaml
+index bb4607cf..38e35c70 100644
+--- a/solution-base/mongodb/charts/mongodb-sharded/values.yaml
++++ b/solution-base/mongodb/charts/mongodb-sharded/values.yaml
+@@ -225,7 +225,7 @@ common:
+   serviceAccount:
+     ## @param common.serviceAccount.create Whether to create a Service Account for all pods automatically
+     ##
+-    create: true
++    create: false
```
francoisferrand: there is actually no need to "patch" values.yaml : this file is a configuration file, and the values can (and should preferably) be specified in `build.sh` when rendering the chart

(similarly the `secret-name.patch` patch should be removed, since we already set/override `existingSecret` in build.sh)

solution-base/mongodb/patches/credentials-helper.patch
```diff
@@ -1,12 +1,11 @@
...
 --- a/solution-base/mongodb/charts/mongodb-sharded/templates/_helpers.tpl
 +++ b/solution-base/mongodb/charts/mongodb-sharded/templates/_helpers.tpl
-@@ -249,3 +249,22 @@ mongodb: .Values.mongos.servicePerReplica.loadBalancerIPs
+@@ -249,3 +249,21 @@ mongodb: .Values.mongos.servicePerReplica.loadBalancerIPs
  {{- include "common.warnings.rollingTag" .Values.metrics.image }}
  {{- include "common.warnings.rollingTag" .Values.volumePermissions.image }}
  {{- end -}}
-+
 +{{/* app credentials environment variables */}}
 +{{- define "mongodb-sharded.appAccountEnvs" -}}
```
francoisferrand, line 10: maybe not worth doing now (no time), but could we not (eventually) specify `extraEnvVarsSecret` config option instead of patch file?

.devcontainer/devcontainer.json
benzekrimaha: concerning the deprecation it has been set since mongo 5 : see here https://www.mongodb.com/docs/manual/release-notes/5.0-compatibility/#removed-customizable-values-for-getlasterrordefaults

solution-base/mongodb/charts/mongodb-sharded/values.yaml
benzekrimaha: I indeed don't think there's an impact this has been automatically changed with the charts bump

solution-base/mongodb/scripts/set-default-write-concern-majority.sh
```diff
@@ -4,15 +4,24 @@ set_default_majority() {
...
+        db.adminCommand({
+            setDefaultRWConcern: 1,
+            defaultWriteConcern: { w: 'majority' }
+        })
+        print("Default write concern set to 'majority'")
+    } else {
+        print("Default write concern already set to 'majority'")
+    }
+} catch (e) {
+    print("Error setting default write concern: " + e)
```
francoisferrand, line 20: error should not be silently ignored : we need to be confident the write concern was set properly...
--> printing the error is fine, but having the script should still fail (and cause set-default-write-concern-majority.sh to fail as well?)

francoisferrand: please check: if the script did not fail before, then this is probably not a blocker...

but eventually we should make sure Mongo does not start if default RW concern is not correct, or at least we have an alert.

solution-base/mongodb/patches/credentials-helper.patch
francoisferrand: if not done now, please create a followup (cleanup) ticket

francoisferrand (approved):

solution-base/mongodb/scripts/set-default-write-concern-majority.sh
benzekrimaha: Indeed fixed here : https://github.com/scality/Zenko/pull/2196/commits/85c4ad780505ddfa725521d1eeda956be09d35fa#diff-d88bfd067bf198f77a23dd1261c53d9d731c812b6f8fcda1735745144e3bcbc9R21

solution-base/mongodb/patches/credentials-helper.patch
benzekrimaha: https://scality.atlassian.net/browse/ZENKO-4982

ghost (approved):
Good to test upgrades with the latest changes, if not done already, as the `solution-base/mongodb/scripts/set-default-write-concern-majority.sh` change is editing a file that was also a reason of failures in the past

ghost:
Note for reviewers: we need to rebase the PR, the latest commit is bumping to the same version already released, so we would end up with no commit after the merge, an we cannot release zenko...

benzekrimaha:
Note 2: the queue was rebuilt, and the rebase done again , now we have the right preview version

PR #2202, opened 2025-03-13
ZENKO-4905-check-patches-on-ci

benzekrimaha:
Issue: ZENKO-4905


solution-base/mongodb/Makefile
```diff
@@ -19,5 +18,26 @@ fetch-mongodb-sharded:
...
+	@for patch in ${PATCH_FILES}; do \
+		echo "Checking $$patch"; \
+		git apply --check "$$patch" || { echo "Patch check failed for $$patch"; exit 1; }; \
+	done
+	@echo "Applying patches..."
+	@for patch in ${PATCH_FILES}; do \
+		echo "Applying $$patch"; \
+		git apply "$$patch" || { echo "Patch apply failed for $$patch"; exit 1; }; \
+	done
+	@echo "Patches applied successfully"
```
francoisferrand: what is the point of these checks?

solution-base/mongodb/Makefile
```diff
@@ -19,5 +18,26 @@ fetch-mongodb-sharded:
...
+
+verify-patch:
+	@echo "Verifying patches are applied..."
+	@for patch in ${PATCH_FILES}; do \
+		echo "Verifying $$patch"; \
+		git apply --reverse --check "$$patch" || { echo "Patch $$patch is not fully applied"; exit 1; }; \
+	done
+	@echo "Patches have been correctly applied"
```
francoisferrand: this only verifies that the patches can be applied : which is not really useful in CI, since the patches have supposedly already been applied...

what we really need is to ensure that every _change_ from the (upstream) charts is present in the patches : i.e. `<upstream charts> + <patches> == <commited charts>`.
The process would thus be:
* Dowload the _current_ version of chart (erasing whatever is stored locally)
* Apply the charts
* Check there is no (local) git diff

solution-base/mongodb/Makefile
benzekrimaha: the old way of applying patches was not working , this fixes it , we first check that the patch is there, then we apply it

solution-base/mongodb/Makefile
benzekrimaha: done here : https://github.com/scality/Zenko/pull/2202/commits/e8da6935e05ca4f579b3e6106bf8e042a6104346

.github/workflows/end2end.yaml
```diff
@@ -159,6 +159,30 @@ jobs:
           password: ${{ secrets.ARTIFACTS_PASSWORD }}
           source: /tmp/artifacts
         if: always()
+    
+  check-mongo-patches:
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: remove existing tar file
```
francoisferrand: no point, they are not commited
(and consequently, not sure the makefile rule is really needed...)

solution-base/mongodb/Makefile
```diff
@@ -19,5 +18,21 @@ fetch-mongodb-sharded:
 		--untardir ${CHART_DIR}
 
 patch:
-	@git apply --check ${PATCH_FILES}
-	@git apply ${PATCH_FILES}
+	@if [ -z "${PATCH_FILES}" ]; then \
+		echo "No patch files found in ${PATCH_DIR}"; \
+		exit 1; \
```
francoisferrand: not an error: if there is no patch to apply, all is well and job is done!

solution-base/mongodb/Makefile
```diff
@@ -19,5 +18,21 @@ fetch-mongodb-sharded:
...
+	fi
+	@echo "Checking patches..."
+	@for patch in ${PATCH_FILES}; do \
+		echo "Checking $$patch"; \
+		git apply --check "$$patch" || { echo "Patch check failed for $$patch"; exit 1; }; \
```
francoisferrand: what is the point to "check" first?
anyway it will fail when applying...

solution-base/mongodb/Makefile
```diff
@@ -19,5 +18,21 @@ fetch-mongodb-sharded:
...
+	done
+	@echo "Applying patches..."
+	@for patch in ${PATCH_FILES}; do \
+		echo "Applying $$patch"; \
+		git apply "$$patch" || { echo "Patch apply failed for $$patch"; exit 1; }; \
```
francoisferrand: not sure what is the benefit of "catching" the error: you already add a log indicating we are applying the patch, and `git apply` will fail with an error when the patch cannot be applied...
```suggestion
		git apply "$$patch" \
```

removing useless logs & duplication, the whole rule can be written simply:

```
patch:
    @for patch in ${PATCH_FILES} ; do \
		echo "Applying $$patch" ; \
		git apply "$$patch" ; \
    done
```

solution-base/mongodb/Makefile
```diff
@@ -5,11 +5,10 @@ CHART_REPO:="https://charts.bitnami.com/bitnami"
 CHART_MONGO_SHARDED_VERSION:="7.9.1"
 
 PATCH_DIR:="${ROOT_DIR}/patches"
-PATCH_FILES:="$(shell ls -d ${PATCH_DIR}/*)"
-
+PATCH_FILES:=$(shell ls ${PATCH_DIR}/* | grep -v mongodb-sharded-add-configsvr-service-file.patch)
```
francoisferrand: should not filter patches: if we need not to apply it, then we must actually remove it (`git rm`)...

```suggestion
PATCH_FILES:=$(shell ls ${PATCH_DIR}/*)
```

solution-base/mongodb/Makefile
benzekrimaha: Here if it fails we're exciting so it won't try to apply it , it's an added security only

solution-base/mongodb/Makefile
```diff
@@ -19,5 +18,18 @@ fetch-mongodb-sharded:
 		--untardir ${CHART_DIR}
 
 patch:
-	@git apply --check ${PATCH_FILES}
-	@git apply ${PATCH_FILES}
+	@if [ -z "${PATCH_FILES}" ]; then \
+		echo "No patch files found in ${PATCH_DIR}. Job is done"; \
+		exit 0; \
+	fi
+	@echo "Checking patches..."
```
francoisferrand: Not needed, it adds no value

solution-base/mongodb/Makefile
francoisferrand: Since we need all patches to apply anyway, there is no added security

solution-base/mongodb/Makefile
```diff
@@ -5,8 +5,7 @@ CHART_REPO:="https://charts.bitnami.com/bitnami"
 CHART_MONGO_SHARDED_VERSION:="7.9.1"
 
 PATCH_DIR:="${ROOT_DIR}/patches"
-PATCH_FILES:="$(shell ls -d ${PATCH_DIR}/*)"
-
+PATCH_FILES:=$(shell ls ${PATCH_DIR}/*)
```
francoisferrand: Nit: Why this change?

.github/workflows/end2end.yaml
```diff
@@ -159,6 +159,30 @@ jobs:
...
+  check-mongo-patches:
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: fetch mongo charts
+        run: make fetch-mongodb-sharded
+        working-directory: ./solution-base/mongodb
+      - name: remove configsrv file
+        run: rm -vf charts/mongodb-sharded/templates/config-server/config-server-service.yaml || echo "Failed to remove"
```
francoisferrand: why do we need to remove?
if this is needed, should be part of a patch...

(if helm does not does remove "external" files, then we may need remove the whole chart folder in `make fetch-mongodb-sharded`)

benzekrimaha: no the file isn't there when we fetch de mongo sharded , it has always been added manually , the chart creates it. That's why here I remove it otherwise i end up with an error saying that the file already exists ( I don't have the historic linked to this)

francoisferrand: that is an issue in the patch file,
```
--- /dev/null
+++ b/solution-base/mongodb/charts/mongodb-sharded/templates/config-server/config-server-service.yaml
```

and should be fixed by updating the patchfile, not the process.

francoisferrand: Just tested:
* "updating" the charts does not actually _remove_ files which were created... So upgrading "in-place" just overwrites the files, but leaves removed files in place...
* the patches work when we simply delete the charts directory before fetching: however, this causes the git diff to fail as we added some files:
   ```
   supprimé :        charts/mongodb-sharded/charts/common/values.schema.json
   supprimé :        charts/mongodb-sharded/files/docker-entrypoint-initdb.d/create-app-user-sharded.sh
   supprimé :        charts/mongodb-sharded/files/docker-entrypoint-initdb.d/set-default-write-concern-majority-sharded.sh
   supprimé :        charts/mongodb-sharded/templates/initialization-configmap.yaml
   ```
* Regarding `config-server-service.yaml`, this file is indeed not present in the chart: not related to patching, but I wonder why we added it (or kept it), and if we need it still...

- [x] clearing the folder before fetch is the way
- [x] need to update the patch files to "create" these removed files/links
- [x] removed `values.schema.json`, not present in chart & not referenced
- [ ] need to check if we really need `config--server-service.yaml`, of if that service is a leftover that should be removed

solution-base/mongodb/Makefile
francoisferrand: Applying patches in one lines works fine, as can be seen [here](https://github.com/scality/Zenko/actions/runs/14001929540/job/39209911779)

francoisferrand (approved):

welansari (approved):

.github/workflows/end2end.yaml
benzekrimaha: Update on this : when trying without `config-server-service.yaml`the zenko deployment timeouts ( keeps on hanging actually) waiting pod the pod to be up, gooing more on detail the file actually provides more granular control over how the config servers are exposed and accessed within the Kubernetes cluster, which wasn't present in the original template charts.

benzekrimaha: Following up on this based on yesterday's conversation and discoveries: The config-server-service creates a stable service endpoint that load-balances connections to all available config server replicas. Without it, mongos pods hardcode connections to specific config server pods (starting with configsvr-0), causing them to fail when that particular pod is unavailable, even if other healthy config server replicas exist.
**_Conclusion:_**  we need to keep the file for our specific use cases

ghost: Maybe a bit too late but worth documenting it somewhere, as it was painful to find (in the could would be great but in a confluence page is good as well).

ghost (approved):

PR #2203, opened 2025-03-18
Extend IP and SSL tests

ghost:
- Add an IP test that was missing, on the happy path with the correct IP range.
- Add SSL tests.

Issue: ZENKO-4929

### DIFF
--- a/tests/ctst/common/hooks.ts
+++ b/tests/ctst/common/hooks.ts
@@ -43,6 +43,9 @@ Before({ tags: '@Quotas', timeout: 1200000 }, async function (scenarioOptions) {
 });
 
 After(async function (this: Zenko, results) {
+    // Reset any configuration set on the endpoint (ssl, port)
+    CacheHelper.parameters.ssl = this.parameters.ssl;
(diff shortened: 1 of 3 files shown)


tests/ctst/common/hooks.ts
```diff
@@ -88,4 +88,9 @@ After({ tags: '@BP-ASSUME_ROLE_USER_CROSS_ACCOUNT'}, async function (this: Zenko
     await cleanupAccount(this, crossAccountName);
 });
 
+After({ tags: '@SSL'}, async function (this: Zenko) {
```
francoisferrand: should we not do this every time, to ensure it gets reset even if the `@SSL` is (by mistake) not set?

francoisferrand (approved):

tests/ctst/common/hooks.ts
ghost: Yes, that won't hurt... I'll update

benzekrimaha (approved):
LGTM

ghost:
I pushed new commits to rebase after the last zenko bump, to avoid one more merge build.

PR #2205, opened 2025-03-20
Improve MongoDB ReplicationLag alert

ghost:
ReplicationLagWarning alert to become dynamic:

- The threshold is not 30s anymore. This fixed value has two problems:
  - it is not consistent with the `flowControlTargetLagSeconds` config set as 10s.
  - it is creating a lot of alerts when a 3 node cluster is under load, when the lag may be in the 1min-30min range without impact on the replication.
- The new alert instead considers the largest oplog window of the PRIMARY mongod instance to dynamically create alerts. The "largest" is here to be future proof with multiple shards.
- Warning when the `optimeDate` (current lag) of any secondary exceeds 70% of the primary oplog window for 8min.
- Critical when it exceeds 95% for 2min. In this case there is a sizing issue and a high risk of requiring full init syncs.

Issue: [ZENKO-4986](https://scality.atlassian.net/browse/ZENKO-4986)

[ZENKO-4986]: https://scality.atlassian.net/browse/ZENKO-4986?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/monitoring/mongodb/alerts.yaml
+++ b/monitoring/mongodb/alerts.yaml
@@ -106,18 +106,18 @@ groups:
         max by(rs_nm, member_idx) (
           (
             max by(pod, rs_nm) (
-              mongodb_rs_members_optimeDate{namespace="${namespace}",pod=~"${service}.*",member_state="PRIMARY"}
+              mongodb_rs_members_optimeDate{namespace="${namespace}", pod=~"${service}.*", member_state="PRIMARY"}
             )
             - ignoring(member_idx) group_right min by(pod, rs_nm, member_idx) (
-              mongodb_rs_members_optimeDate{namespace="${namespace}",pod=~"${service}.*",member_state="SECONDARY"}
+              mongodb_rs_members_optimeDate{namespace="${namespace}", pod=~"${service}.*", member_state="SECONDARY"}
             )
           ) / scalar(
-            max (
-              (mongodb_mongod_replset_oplog_head_timestamp{namespace="${namespace}",pod=~"${service}.*"}
+            max by(rs_nm) (
+              (mongodb_mongod_replset_oplog_head_timestamp{namespace="${namespace}", pod=~"${service}.*"}
               - on(pod)
-              mongodb_mongod_replset_oplog_tail_timestamp{namespace="${namespace}",pod=~"${service}.*"})
+              mongodb_mongod_replset_oplog_tail_timestamp{namespace="${namespace}", pod=~"${service}.*"})
               * on(pod) group_left()
-              (mongodb_mongod_replset_my_state{namespace="${namespace}",pod=~"${service}.*"} == 1)
+              (mongodb_mongod_replset_my_state{namespace="${namespace}", pod=~"${service}.*"} == 1)
             )
           )
         ) * 100 > ${replicationLagWarningThreshold},
@@ -136,18 +136,18 @@ groups:
         max by(rs_nm, member_idx) (
           (
             max by(pod, rs_nm) (
-              mongodb_rs_members_optimeDate{namespace="${namespace}",pod=~"${service}.*",member_state="PRIMARY"}
+              mongodb_rs_members_optimeDate{namespace="${namespace}", pod=~"${service}.*", member_state="PRIMARY"}
             )
             - ignoring(member_idx) group_right min by(pod, rs_nm, member_idx) (
-              mongodb_rs_members_optimeDate{namespace="${namespace}",pod=~"${service}.*",member_state="SECONDARY"}
+              mongodb_rs_members_optimeDate{namespace="${namespace}", pod=~"${service}.*", member_state="SECONDARY"}
             )
           ) / scalar(
-            max (
-              (mongodb_mongod_replset_oplog_head_timestamp{namespace="${namespace}",pod=~"${service}.*"}
+            max by(rs_nm) (
+              (mongodb_mongod_replset_oplog_head_timestamp{namespace="${namespace}", pod=~"${service}.*"}
               - on(pod)
-              mongodb_mongod_replset_oplog_tail_timestamp{namespace="${namespace}",pod=~"${service}.*"})
+              mongodb_mongod_replset_oplog_tail_timestamp{namespace="${namespace}", pod=~"${service}.*"})
               * on(pod) group_left()
-              (mongodb_mongod_replset_my_state{namespace="${namespace}",pod=~"${service}.*"} == 1)
+              (mongodb_mongod_replset_my_state{namespace="${namespace}", pod=~"${service}.*"} == 1)
             )
           )
         ) * 100 > ${replicationLagCriticalThreshold},
--- a/monitoring/mongodb/alerts.yaml
+++ b/monitoring/mongodb/alerts.yaml
@@ -21,6 +21,12 @@ x-inputs:
   - name: replicas
     type: config
     value: 3
+  - name: replicationLagWarningThreshold
+    type: config
+    value: 70
+  - name: replicationLagCriticalThreshold
+    type: config
+    value: 90
 
 groups:
 - name: MongoDb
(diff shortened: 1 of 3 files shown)


monitoring/mongodb/alerts.yaml
```diff
@@ -97,20 +97,65 @@ groups:
...
-            min(mongodb_rs_members_optimeDate{namespace="${namespace}",pod=~"${service}.*",member_state="SECONDARY"})
-                by(pod, rs_nm, member_idx)
-        ) by(member_idx, rs_nm) / 1000 > 30
-      , "member_idx", "$1", "member_idx", "(${service}[^.]*)\\.${service}.*")
-    for: 10m
+        max by(rs_nm, member_idx) (
+          (
+            max by(pod, rs_nm) (
+              mongodb_rs_members_optimeDate{namespace="${namespace}",pod=~"${service}.*",member_state="PRIMARY"}
+            ) - ignoring(member_idx) group_right
```
francoisferrand: nit: the substraction is kind of hidden in the middle, may be better to wrap before?

francoisferrand, line 104: `ignoring(member_idx)` is useless, since `max by(pod, rs_nm)` will keep only the `pod` and `rs_nm` labels

francoisferrand: are there multiple `member_idx` by `pod` ?

if not, a more 'readable' way would be to just do `max by(pod, member_idx, rs_nm) - min by (pod, member_idx, rs_nm)` ?

monitoring/mongodb/alerts.yaml
```diff
@@ -97,20 +97,65 @@ groups:
...
+            )
+          ) / scalar(
+            max by(pod) (
+              (mongodb_mongod_replset_oplog_head_timestamp{namespace="${namespace}",pod=~"${service}.*"}
+              - on(pod)
+              mongodb_mongod_replset_oplog_tail_timestamp{namespace="${namespace}",pod=~"${service}.*"})
+              * on(pod) group_left()
+              (mongodb_mongod_replset_my_state{namespace="${namespace}",pod=~"${service}.*"} == 1)
+            )
+          ) > 0.7
```
francoisferrand: - should we make thresholds should be configurable, by templatizing them?
- as default values, 70% and 95 seems a little bit weird to me : either we should target something close to limit (95% and 85% maybe), or alert with more margin (70% and 90%) ?

monitoring/mongodb/alerts.yaml
```diff
@@ -97,20 +97,65 @@ groups:
...
+        max by(rs_nm, member_idx) (
+          (
+            max by(pod, rs_nm) (
+              mongodb_rs_members_optimeDate{namespace="${namespace}",pod=~"${service}.*",member_state="PRIMARY"}
+            ) - ignoring(member_idx) group_right
+            min by(pod, rs_nm, member_idx) (
+              mongodb_rs_members_optimeDate{namespace="${namespace}",pod=~"${service}.*",member_state="SECONDARY"}
+            )
+          ) / scalar(
+            max by(pod) (
```
francoisferrand: `scalar` is expected to work with a one-element vector : but here there is `max by(pod)`, which will return one element per pod...

```suggestion
            max (
```

monitoring/mongodb/alerts.yaml
ghost: To me with this suggestion, it tries to match the PRIMARY’s data to each SECONDARY’s `member_idx`, but the PRIMARY only has one index: it won't match with the secondaries and give empty data.
The original approach (that I am reusing in this PR) ignores `member_idx` to compare the PRIMARY to all SECONDARY members at once.

(in other words, the suggestion doesn't work)

monitoring/mongodb/alerts.yaml
ghost: Right, I updated the PR accordingly.

monitoring/mongodb/alerts.yaml
```diff
@@ -97,20 +103,65 @@ groups:
...
-      , "member_idx", "$1", "member_idx", "(${service}[^.]*)\\.${service}.*")
-    for: 10m
+        max by(rs_nm, member_idx) (
+          (
+            max by(pod, rs_nm) (
+              mongodb_rs_members_optimeDate{namespace="${namespace}",pod=~"${service}.*",member_state="PRIMARY"}
+            )
+            - ignoring(member_idx) group_right min by(pod, rs_nm, member_idx) (
+              mongodb_rs_members_optimeDate{namespace="${namespace}",pod=~"${service}.*",member_state="SECONDARY"}
+            )
```
francoisferrand: overall, this computation is done on each pod : we compare the primary's oplog date vs all the secondary oplog date, _as viewed by this pod_.
→ this gives the max optime delta _as viewed by this pod_.

should it not be other way, i.e. for each member_idx, diff the max(PRIMARY) with the max(SECONDARY) [ignoring pods], which gives a simpler expression (no ignoring, no need to perfrom 2 nested max) and more precisely mesures the lag of that member_idx vs the primary ?

monitoring/mongodb/alerts.yaml
```diff
@@ -97,20 +103,65 @@ groups:
...
+          ) / scalar(
+            max (
+              (mongodb_mongod_replset_oplog_head_timestamp{namespace="${namespace}",pod=~"${service}.*"}
+              - on(pod)
+              mongodb_mongod_replset_oplog_tail_timestamp{namespace="${namespace}",pod=~"${service}.*"})
+              * on(pod) group_left()
+              (mongodb_mongod_replset_my_state{namespace="${namespace}",pod=~"${service}.*"} == 1)
+            )
```
francoisferrand: * can we/should we only look at the oplog headroom from the PRIMARY ?
(i.e. if one SECONDARY has lots of headroom, it would affect the metrics even though it does not minimize the risk)
* should we not take the `min` headroom (instead of `max`)?

monitoring/mongodb/alerts.yaml
francoisferrand: right, that was the first part of the comment: i was not sure how the metrics where presented (i.e. each pod exposes his view of the state of each member)

monitoring/mongodb/alerts.yaml
ghost: I'll check that approach yes (I'm also trying that on multi shard clusters to ensure it's future proof)

(from what I tested, we hit issues because labels are not unique the other way around... giving errors)

monitoring/mongodb/alerts.yaml
ghost: This part is the journal window. The headroom would typically be something like

```
(avg(mongodb_mongod_replset_oplog_tail_timestamp - mongodb_mongod_replset_oplog_head_timestamp)
      - (avg(mongodb_mongod_replset_member_optime_date{state="PRIMARY"}) - avg(mongodb_mongod_replset_member_optime_date{state="SECONDARY"})))
```

Here, the whole alert is similar to the headroom, but computed as a percentage so we can define thresholds and not just alert when it's too late.

As the replication happens between the primary and secondaries only, not between secondaries, I don't think we need to consider secondary oplog window: its size has nothing to do with the secondary ability to converge its replication.

> should we not take the min headroom (instead of max)?

I asked myself this question and with only one replicaset we only have one value anyway. But I chose `max` to be on the optimistic side and avoid alerting too early. `min` makes sense if we want to detect the problem as early as possible, I'll go with it; I am aligned.

francoisferrand: > This part is the journal window. The headroom would typically be something like

ok, journal window :-)
But my point was really about looking only at the primary, since this is the "important" oplog in this case.

ghost: Yes, this is what we do already, by filtering on the "my state" being 1 (=== PRIMARY), did you have something else in mind?

monitoring/mongodb/alerts.yaml
ghost: Done in last commit.

monitoring/mongodb/alerts.yaml
ghost: So I kept this as-is because the operation already returns the PRIMARY only: the oplog metrics are process-level, so they don't have the primary/secondary information, but the "myState" can be used to map properly. Please let me know if you see another (better) approach.

monitoring/mongodb/alerts.yaml
```diff
@@ -97,20 +103,53 @@ groups:
...
+            (mongodb_mongod_replset_oplog_head_timestamp{namespace="${namespace}",pod=~"${service}.*"}
+            - on(pod)
+            mongodb_mongod_replset_oplog_tail_timestamp{namespace="${namespace}",pod=~"${service}.*"})
+            * on(pod) group_left()
+            (mongodb_mongod_replset_my_state{namespace="${namespace}",pod=~"${service}.*"} == 1)
+          )
+        ) * 100 > ${replicationLagWarningThreshold},
+        "member_idx", "$1", "member_idx", "(${service}[^.]*)\\.${service}.*"
+      )
+    for: 8m
```
welansari: Approach looks good to me, however these metrics return values for both of the `shardsrv` and `configsrv` instances. Is there a risk of mix between their values in the calculation that may trigger a false alert or the other way arround by not triggering the alert ?

I'm saying this because of the `max()` operators we use in multiple parts of the query and that remove the labels.

Note: `zenko-operator` replaces the `service` variable with `data-db-mongodb-sharded` which is the prefix of both the `shardsrv` and `configsrv` instances.

ghost: Practically this won't affect much but indeed it might create some <0 metrics sometimes. the solution to this is to rely on the `rs_nm`, as the config server is using a dedicated replicaset, ensuring metrics are independent. Last commit was updated.

welansari (approved):

benzekrimaha (approved):

PR #2208, opened 2025-04-04
Add miria location

francoisferrand:
- Bump cloudserver 9.0.7
- Bump pensieve-api 1.6.4
- Bump zenko-operator 1.7.3
- Create miria location
- gha: fix mongodb dump folder
- Use ubuntu-22.04 runner for doc

Issue: ZENKO-4987


.github/workflows/end2end.yaml
```diff
@@ -164,7 +165,7 @@ jobs:
         if: always()
 
   build-doc:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
```
ghost, line 168: Did you try with `24`?

ghost (approved):

.github/workflows/end2end.yaml
francoisferrand: yes, it's documented in the commit message: with 24 we get python 3.12, which breaks the current sphinx... So we need to upgrade all that stuff, too much for today :-)
(but i added a note in the zenko dependency bumps tickets)

welansari (approved):

PR #2214, opened 2025-04-22
Improvement/zenko 4838 sur api

ghost:
Adding SUR API backend (exposed) to Zenko. CTST tests are added.


ghost:
This looks like a good candidate for a new branch? We already use Zenko 2.11 for Artesca 3.1.

ghost:
I added a commit because now we are rejecting calls to GetAccountInfos without any filter provided. This affects one test. I might update this commit if the changes are not enough, but the SUR part won't be updated.

.github/scripts/end2end/configs/zenko.yaml
```diff
@@ -123,9 +127,6 @@ spec:
       azure:
         archiveTier: "hot"
         restoreTimeout: "15s"
-  scuba:
-    logging:
-      logLevel: debug
```
francoisferrand, line 128: should we not keep scuba log level to `debug` ? (but move it above, avoid having twice the same key)

.github/scripts/end2end/deploy-zenko.sh
```diff
@@ -17,6 +17,7 @@ export ZENKO_STS_INGRESS=${ZENKO_STS_INGRESS:-'sts.zenko.local'}
 export ZENKO_MANAGEMENT_INGRESS=${ZENKO_MANAGEMENT_INGRESS:-'management.zenko.local'}
 export ZENKO_S3_INGRESS=${ZENKO_S3_INGRESS:-'s3.zenko.local'}
 export ZENKO_UI_INGRESS=${ZENKO_UI_INGRESS:-'ui.zenko.local'}
+export ZENKO_SUR_API_INGRESS=${ZENKO_SUR_API_INGRESS:-'utilization.zenko.local'}
```
francoisferrand: nit: just `ZENKO_SUR_INGRESS`

```suggestion
export ZENKO_SUR_INGRESS=${ZENKO_SUR_API_INGRESS:-'utilization.zenko.local'}
```

(don't forget to update references below...)

solution/deps.yaml
```diff
@@ -106,7 +106,7 @@ scuba:
   sourceRegistry: ghcr.io/scality
   dashboard: scuba/scuba-dashboards
   image: scuba
-  tag: 1.0.11
+  tag: 1.1.0-preview.9
```
francoisferrand, line 109: we should not "release" (as in "Zenko release") a preview version of Scuba
- either this goes in next Zenko branch (2.12)
- or we need to at least wait for a Scuba 1.1.0 to merge

is there really an urge to merge this in Zenko 2.11, or can we just release this with Zenko 2.12.0 ?

tests/ctst/common/utils.ts
```diff
@@ -297,3 +303,102 @@ export async function cleanupAccount(world: Zenko, accountName: string) {
...
+    jobName?: string;
+    jobNamespace?: string;
+}
+
+/**
+ * Generic function to prepare scenarios that need cronjob to run (e.g., count-items)
+ * Can be used by both quota and utilization tests to avoid code duplication
+ * Creates accounts, buckets and runs cronjob once for all scenarios
+ */
+export async function prepareMetricsScenarios(
```
francoisferrand, line 318: "prepare scenario" does not seem very BDD/gherkin-like : the preparation should be the sets of "given", and the functions should reflect what they actually do...

esp. looking at the tests, it is not very clear/explicit what setup was done, and the scenario cannot be understood without looking at the code...

tests/ctst/features/cloudserverAuth.feature
```diff
@@ -5,9 +5,9 @@ Feature: AWS S3 Bucket operations
     @Cloudserver-Auth
     Scenario: Check Authentication on bucket object lock actions with Vault
         Given a IAM_USER type
-        And an IAM policy attached to the entity "user" with "Allow" effect to perform "CreateBucket" on "*"
-        And an IAM policy attached to the entity "user" with "<allow>" effect to perform "PutBucketObjectLockConfiguration" on "*"
-        And an IAM policy attached to the entity "user" with "<allow>" effect to perform "PutBucketVersioning" on "*"
+        And an IAM policy attached to the entity "user" with "Allow" effect to perform "s3" "CreateBucket" on "arn:aws:s3:::*"
+        And an IAM policy attached to the entity "user" with "<allow>" effect to perform "s3" "PutBucketObjectLockConfiguration" on "arn:aws:s3:::*"
+        And an IAM policy attached to the entity "user" with "<allow>" effect to perform "s3" "PutBucketVersioning" on "arn:aws:s3:::*"
         When the user tries to perform CreateBucket
```
francoisferrand, line 11: these tests updates go along with the Vault bump, which needs to land in 2.11 (patch version bump)
(i.e. depending on how we manage the scuba bump -cf earlier comment-, may need to split the PR)

tests/ctst/features/utilization/UtilizationAPIs.feature
```diff
@@ -0,0 +1,65 @@
+Feature: Scality Utilization Reporting (SUR) API
+  The utilization metrics API allows different authorized users to retrieve 
+  metrics for buckets, accounts, and locations.
+
+    @2.11.0
+    @PreMerge
+    @Utilization
+    @UtilizationAPI
+    Scenario Outline: Default roles can retrieve utilization metrics
+        Given a STORAGE_MANAGER type
```
francoisferrand: (not really related to this PR)

this "given" is not very clear : should it not read something like `Given a user with "STORAGE_MANAGER" role` ?
(probably worth a dedicated PR though, to avoid mixing concerns)

francoisferrand (commented):
Kind of feels we are fighting against cucumber here: to introduce some test fixture, we reduce the "descriptiveness" of the tests, and thus the benefit of cucumber...

LGTM otherwise, but we need to settle on the destination branch.

.github/scripts/end2end/configs/zenko.yaml
ghost: Not needed. It was done before for testing purposes...

solution/deps.yaml
ghost: We can wait for 2.12.0 to me. We can also wait for a proper 1.1.0 that should come before that hopefully, I am aligned. We use the same Zenko branch for Artesca 3.1 and 3.2 today, and will eventually want to use the 2.12 with Mongo 8...

tests/ctst/features/cloudserverAuth.feature
ghost: No these change are unrelated to Vault. The only part that is affected is https://github.com/scality/Zenko/pull/2214/commits/eb1e373b403206ec3477aaa2deaa0630e9bcaad3.

The other parts are only so that we can reuse the step with a different service name, so not s3, but sur.

ghost: We can cherry pick the above commit if we want to bump Vault in 2.11 meanwhile.

tests/ctst/common/utils.ts
ghost: This is compatible in the sense that Gherkins supports before hooks, and we just mutualize the logic we already implemented for count items with now the SUR tests. This allows to run only once the job and pre-create everything. The good part is: this will be removed with the real SUR backend... But we do not want to do these operations in sequence.

But Maybe we can rename the function, or add a step in the gherkin scenarios, just to say "given an environment setup [...]" or similar? That would give enough context to the test, if we don't look at the actual implementation...

tests/ctst/features/utilization/UtilizationAPIs.feature
ghost: Yes we can rename the step indeed... Not sure this is the highest priority though.
I created https://scality.atlassian.net/browse/ZENKO-4999 (next one will be 5000 :eyes: )

solution/deps.yaml
ghost: Or maybe we can already create the 2.12? Or just after the end of the current RC phase? We are mostly working on 1 or 2 branches maximum, this should remain feasible...

tests/ctst/common/utils.ts
ghost: I went with this improvement: https://github.com/scality/Zenko/pull/2214/commits/c1824efc6d7b2a69cba9a5df064443efc12e8f5c.

It has two benefits:
- We cannot unexpectedly remove the hook anymore and remain unotified
- We can freely describe what was done in the "prepare" function. Not programatically, but with english sentence(s).

tests/ctst/features/cloudserverAuth.feature
francoisferrand: based on https://github.com/scality/Zenko/pull/2214#discussion_r2066219998, indeed may be better to split the vault-only part (and possibly refactoring to support SUR, so we potentially avoid conflicts down the line), which can be merged "immediately" on 2.11 ; and deal with the rest of this PR confidently, when SUR is available or we have created 2.12

ghost: I created ZENKO-5000

PR #2216, opened 2025-04-23
Redirect mongodb 7 logs to stdout

ghost:
This PR adds a symlink so that logs are redirected to stdout but also accessible through the standard local path. This is adapted from https://github.com/bitnami/charts/pull/27510/files that fixed the issue for non-sharded mongodb.

This fixes two issues:
- Logs are not visible when doing `kubectl logs` anymore with mongodb v7.
- Logs are filling the empty dir, which can create sizing issues in the long run.


tcarmet (approved):

benzekrimaha (approved):
LGTM

PR #2219, opened 2025-05-06
Support N MongoDB shards

ghost:
Here, we only support deployment of multiple shards. Upgrades is managed by upper layers for now.

Issue: ZENKO-4641


.github/scripts/end2end/configs/zenko.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
   name: ${ZENKO_NAME}
   ${ZENKO_ANNOTATIONS}
     zenko.io/x-backbeat-oneshard-replicaset: data-db-mongodb-sharded-shard-0
-    zenko.io/x-backbeat-oneshard-replicaset-hosts: data-db-mongodb-sharded-shard0-data-0.data-db-mongodb-sharded-headless.default.svc.cluster.local:27017
+    zenko.io/x-backbeat-oneshard-replicaset-hosts: ${ZENKO_BACKBEAT_SHARD_HOSTS}
```
francoisferrand: why do we need to set these annotations?
it seems to me the whole config should be in `spec.mongodb` : and the `endpoints` array should provide enough info to setup mongo...

.github/scripts/end2end/generate-kustomization.sh
```diff
@@ -0,0 +1,80 @@
+#!/bin/bash
+
+set -euo pipefail
+
+# Constants for valid topologies for CI tests
+# We support multiple shards per host, or one shard per host, up to 9 nodes
```
francoisferrand: please document "what" is such topology: i.e. what are the first and last numbers...

.github/scripts/end2end/generate-kustomization.sh
```diff
@@ -0,0 +1,80 @@
...
+readonly VALID_TOPOLOGIES=(
+    "1:1" "1:2" "3:1" "3:3" "6:1" "6:3" "6:2" "6:6" "9:1" "9:3" "9:6" "9:9"
```
francoisferrand: - no "3:2" and "9:2" (seem to be missing from the logical series..)
- should be sorted : `"6:1" "6:3" "6:2"` --> `"6:1" "6:2" "6:3"`

francoisferrand: what is the logic for these specific topologies?
- i understand node counts (1, 3, 6 and 9)
- i understand shard count 1 and "N" (i.e. micro-sharding)
- i understand the 1:x config, may be useful for testing
- ...but the other shard count (2, 3, 6) seem arbitrary to me, we could do any kind of "striping" but probably cn't really support these cases at the moment?

.github/scripts/end2end/generate-kustomization.sh
```diff
@@ -0,0 +1,80 @@
...
+)
+
+generate_kustomization() {
+    local node_count=$1
+    local shard_count=$2
+    local kustomization_file="${DIR}/kustomization.yaml"
```
francoisferrand: - instead of creating the kustomization file, could we not apply it directly? (something like `echo ... | kustomize -f -`)
- better yet, could we not use the charts rendered into the ISO (which have the right kustomization and/or patches)? this would help test what we actually ship (and avoid pulling bitnami chart every time)...

.github/workflows/end2end.yaml
```diff
@@ -557,17 +500,12 @@ jobs:
       - name: Clean Up
         run: kind delete cluster
 
-  end2end-https:
+  end2end-2-shards-http:
```
francoisferrand, line 503: should we not run these tests on both 1 and 2-shards, to have a good comparison point?
i.e. keep a single `end2end-http(s)` job and add a matrix parameter for the number of shard/topology

.github/workflows/end2end.yaml
```diff
@@ -580,7 +518,7 @@ jobs:
         with:
           username: "${{ github.repository_owner }}"
           password: "${{ github.token }}"
-          registry: ghcr.io
+          registry: ghcr.io/zenko-dev/zenko-operator
```
francoisferrand: should not be merged?

monitoring/mongodb/alerts.yaml
```diff
@@ -234,7 +234,11 @@ groups:
 
   - alert: MongoDbRSNotSynced
     expr: |
-      group by(rs_nm) ( count by(rs_nm, pod) (mongodb_rs_members_state{namespace="${namespace}", pod=~"${service}.*", member_state="SECONDARY"}) != (${replicas} - 1) )
+      count by (rs_nm) (
+          group by (rs_nm, member_idx) (
```
francoisferrand, line 238: not sure I understand how the query is different: it does the same, just computed differently...

it used for group by pod (i.e. check that every instance has a consistent view of its' replicaset), while the new query goes the other way and check that the expected number of pods are reporting the correct state of each member...

monitoring/mongodb/alerts.yaml
```diff
@@ -234,7 +234,11 @@ groups:
...
+              mongodb_rs_members_state{member_state="SECONDARY", namespace="${namespace}", pod=~"${service}.*"}
+          )
+      ) != (${replicas} - 1)
     for: 10m
     labels:
       severity: warning
```
francoisferrand, line 245: we should add a `rs_nm` label

solution-base/mongodb/charts/mongodb-sharded/templates/shard/shard-data-statefulset.yaml
```diff
@@ -447,7 +447,11 @@ spec:
         {{- end }}
         {{- if $.Values.shardsvr.persistence.selector }}
         selector:
-{{ toYaml $.Values.shardsvr.persistence.selector | indent 10 }}
+          matchLabels:
+            {{- if gt $.Values.shards 1 }}
+            shard: {{ $i | quote }}
+            {{- end }}
+{{ toYaml $.Values.shardsvr.persistence.selector.matchLabels | indent 12 }}
```
francoisferrand: - why change this line, we do not use this variable?
- instead of patching the chart, could we not set this `shardsvr.persistence.selector` variable to the appropriate value to get the same result?

solution-base/mongodb/charts/mongodb-sharded/templates/shard/shard-data-podmonitor.yaml
```diff
@@ -4,11 +4,10 @@ SPDX-License-Identifier: APACHE-2.0
 */}}
 
 {{- if and .Values.shards .Values.metrics.enabled .Values.metrics.podMonitor.enabled }}
-{{- $i := 0 }}
 apiVersion: monitoring.coreos.com/v1
 kind: PodMonitor
 metadata:
-  name: {{ printf "%s-shard%d-data" (include "common.names.fullname" $ ) $i }}
+  name: {{ printf "%s-shards-data" (include "common.names.fullname" $ ) }}
```
francoisferrand: changing the `podMonitor` name will change the metrics I think (esp. the `job` and `service` label), and require change to ZKOP : do we really need to?

esp. since we use a single PodMonitor to match every mongo instance, so there is no confusion...

solution-base/build.sh
```diff
@@ -30,7 +30,22 @@ SKOPEO_OPTS="--override-os linux --insecure-policy"
...
+MONGODB_SHARDED_SIX_NODE_THREE_SHARDS_PATH=${ISO_ROOT}/deploy/mongodb-sharded-6-nodes-3-shards.yaml
+MONGODB_SHARDED_SIX_NODE_SIX_SHARDS_PATH=${ISO_ROOT}/deploy/mongodb-sharded-6-nodes-6-shards.yaml
+MONGODB_SHARDED_NINE_NODE_PATH=${ISO_ROOT}/deploy/mongodb-sharded-9-nodes.yaml
+MONGODB_SHARDED_NINE_NODE_THREE_SHARDS_PATH=${ISO_ROOT}/deploy/mongodb-sharded-9-nodes-3-shards.yaml
+MONGODB_SHARDED_NINE_NODE_SIX_SHARDS_PATH=${ISO_ROOT}/deploy/mongodb-sharded-9-nodes-6-shards.yaml
+MONGODB_SHARDED_NINE_NODE_NINE_SHARDS_PATH=${ISO_ROOT}/deploy/mongodb-sharded-9-nodes-9-shards.yaml
+MONGODB_SHARDED_TWELVE_NODE_PATH=${ISO_ROOT}/deploy/mongodb-sharded-12-nodes.yaml
+MONGODB_SHARDED_TWELVE_NODE_FOUR_SHARDS_PATH=${ISO_ROOT}/deploy/mongodb-sharded-12-nodes-4-shards.yaml
+MONGODB_SHARDED_TWELVE_NODE_EIGHT_SHARDS_PATH=${ISO_ROOT}/deploy/mongodb-sharded-12-nodes-8-shards.yaml
+MONGODB_SHARDED_TWELVE_NODE_TWELVE_SHARDS_PATH=${ISO_ROOT}/deploy/mongodb-sharded-12-nodes-12-shards.yaml
```
francoisferrand: does not match the supported topologies in CI...

solution-base/build.sh
```diff
@@ -74,7 +89,8 @@ function render_mongodb_sharded_yamls()
     local OUTPUT_PATH=${1:-${OPERATOR_PATH}}
     local SHARD_COUNT=${2:-1}
     local NODE_COUNT=${3:-1}
-    local ADD_OPTIONS=${4:-""}
+    local MAX_REPLICA_PER_SHARD=3
+    local REPLICA_COUNT=$(( $3 > $MAX_REPLICA_PER_SHARD ? $MAX_REPLICA_PER_SHARD : $3 ))
```
francoisferrand: * would be better to make REPLICA_COUNT as an extra parameter IMO
* rename to DATA_REPLICA_COUNT or SHARD_REPLICA_COUNT to avoid confusion

.github/scripts/end2end/configs/zenko.yaml
ghost: Yes, to me this is not used anywhere: can we drop it? Just wanted to have a confirmation from persons working on backbeat side here

.github/scripts/end2end/generate-kustomization.sh
ghost: > ...but the other shard count (2, 3, 6) seem arbitrary to me, we could do any kind of "striping" but probably cn't really support these cases at the moment?

Can you be more specific on the ones you don't understand?
As 6 nodes 2 shards is fine for example, as this is what we want to support.

.github/scripts/end2end/generate-kustomization.sh
ghost: I'll try the second approach.

.github/workflows/end2end.yaml
ghost: I'm not sure it's useful, the diff between http and https is the activation of the https, and otherwise, all these tests are covered by CTST today, so with https enabled and 1 shard. These tests bring no value as-is, and were kept only as sanity checks for the 2 shards deployment. Otherwise, if we really want to test things, we would only keep the CTST and use a matrix, but given the flakiness, I disagree with this approach...

monitoring/mongodb/alerts.yaml
ghost: The old query mistakenly counted how many exporter pods reported any secondaries (often resulting in 3), while the new query correctly counts the actual unique MongoDB members that are secondary (resulting in 2).
The issue was that we were comparing 2 with 3, triggering the alert...

solution-base/build.sh
ghost: Yes I need to update the CI part... Just a note on that: I am adding 12 nodes because of what was recently presented. It's really to be future proof and won't be deployed soon.

solution-base/mongodb/charts/mongodb-sharded/templates/shard/shard-data-statefulset.yaml
ghost: The variable is used, for instance we have this in the sts after deployment:
```
  selector:
    matchLabels:
      app.kubernetes.io/name: mongodb
      app.kubernetes.io/part-of: zenko
      shard: "1"
```

But it's actually set from the build.sh file:
```
        --set 'shardsvr.persistence.selector.matchLabels.app\.kubernetes\.io/part-of=zenko' \
```

This means, with the change, that we need to access the .matchLabels and not the parent variable. This is also something we use as a label selector in the artesca installer, so we cannot just remove it.

So in short:
- we pass this info when rendering the yamls with helm
- the patch is necessary because we need to add these labels for artesca
- now that we also have the shard, we cannot just pass a static value, we must compute it in the yaml dynamically
- this change forces us to access the .matchLabels

solution-base/mongodb/charts/mongodb-sharded/templates/shard/shard-data-podmonitor.yaml
ghost: This was done to actually use a single name for all mongo instances as we currently use the index 0. We could change that, as we shouldn't use index 0 for other shards, it's misleading...

If we keep the current behavior it will work indeed, but makes everything more complex to maintain (especially in dashboards?). From what I can see this would affect PRA & Backbeat dashboard only (where we use the job?).

I'm fine keeping it as-is for now and create a follow up ticket. Please let me know if you agree.

Edit: I kept the original name for now.

.github/scripts/end2end/generate-kustomization.sh
ghost: Note on the second point: it's already what we do. But we need to manage the volumes, as we don't have metal here. We use the yaml from the ISO already.

solution-base/mongodb/charts/mongodb-sharded/templates/shard/shard-data-statefulset.yaml
francoisferrand: > i'll check if we can add this to the values.yaml file

we don't use the `values.yaml` file (the one we have the default, part of the chart), but specify the values in argument to helm call

.github/workflows/end2end.yaml
francoisferrand: > kept only as sanity checks for the 2 shards deployment

I understand these tests are mostly redundant with ctst, and that we want to merge http and https, no problem there.

my point is really about sharding: why have a sanity check only on "2 shards" topology? should we not try to deploy the other topologies as well, as sanity checks?

.github/scripts/end2end/configs/zenko.yaml
francoisferrand: looking at ZKOP, this is only used when sharding is not supported:

```
func patchMongoConfigObjForSingleShard(cr *zenkov1alpha2.Zenko, currentState *zenkoReconcileState) (jsObj, error) {
	mongoConfigObj := commonMongoConfigObj(cr, currentState)

	if backbeatOplogPopulatorEnabled(cr, currentState) &&
		backbeatSupportsSingleExtensionQueuePopulators(currentState) {
		return mongoConfigObj, nil
	}

	replicaSet := cr.Annotations[backbeatReplicaSetAnnotation]
	replicaSetHosts := cr.Annotations[backbeatReplicaSetHostsAnnotation]

	if replicaSet == "" || replicaSetHosts == "" {
		return nil, fmt.Errorf("sharded mongodb temporarily requires annotations %s and %s", backbeatReplicaSetAnnotation, backbeatReplicaSetHostsAnnotation)
	}

	mongoConfigObj.Delete("shardCollections")
	mongoConfigObj["replicaSet"] = replicaSet
	mongoConfigObj["replicaSetHosts"] = replicaSetHosts

	return mongoConfigObj, nil
}
```

It most versions of Artesca (i.e. since ~1.3 or 1.4), we exit this function through the first `if` ; and the followup code should have been cleaned when we dropped non-shared mongo support a few months ago.

.github/scripts/end2end/generate-kustomization.sh
francoisferrand: > As 6 nodes 2 shards is fine for example, as this is what we want to support.

indeed

> Can you be more specific on the ones you don't understand?

* 6:3 : why have 2 instances/replicas per rs?
* 9:6 : how does it map 6 shards to 9 nodes?
* 3:2 and 9:2 I suggested earlier : same, how does that map 2 shards to 3 and 9 nodes, respectively?

ghost: Thanks, I removed them already, I just wanted to make sure I understood correctly. I am aligned.
What's left is:
- one shard per node
- one shard every 3 servers
- one shard in total

For 1, 3, 6, 9 & 12 nodes

solution-base/mongodb/charts/mongodb-sharded/templates/shard/shard-data-statefulset.yaml
ghost: My comment was updated in between, indeed, but we must still specify the shard number, so we need to edit this part, and also, we want to keep the selectors we had already. So unless you think this is wrong, to me the change is fine?

.github/workflows/end2end.yaml
ghost: > should we not try to deploy the other topologies as well, as sanity checks?

No, because we test one shard already with ctst, and ctst is doing much more than these sanity checks, while including the sanity checks. No reason to deploy twice the same topology: it will create more cost for no added value.

.github/scripts/end2end/generate-kustomization.sh
ghost: I tried various things, and finally I went with an edit of the base file, with minimum edits:
- just remove the match selectors, as not supported by the provisioner (and supporting them will create more complexity than needed? As we need the storage class to support them or pre-create the PVs with the right labels).
- we avoid patching the whole persistence.
- the base yaml is then used, so except for the images & match labels, we are using the base image without creating any temporary file anymore.

.github/scripts/end2end/configs/zenko.yaml
ghost: Thanks, so I will create a ZKOP cleanup ticket.

ghost: Created ZKOP-461.

.github/scripts/end2end/install-kind-dependencies.sh
```diff
@@ -137,24 +191,6 @@ build_solution_base_manifests() {
     sed -i "s/MONGODB_SHARDSERVER_RAM_LIMIT/${MONGODB_SHARDSERVER_RAM_LIMIT}/g" $DIR/_build/root/deploy/*
     sed -i "s/MONGODB_SHARDSERVER_RAM_REQUEST/${MONGODB_SHARDSERVER_RAM_REQUEST}/g" $DIR/_build/root/deploy/*
     sed -i "s/MONGODB_MONGOS_RAM_REQUEST/${MONGODB_MONGOS_RAM_REQUEST}/g" $DIR/_build/root/deploy/*
-
```
ghost, line 140: Note for reviewers: this was removed because we had this logic twice in the function, maybe a rebase issue at some point. See above.

.github/scripts/end2end/deploy-zenko.sh
```diff
@@ -18,6 +18,7 @@ export ZENKO_MANAGEMENT_INGRESS=${ZENKO_MANAGEMENT_INGRESS:-'management.zenko.lo
 export ZENKO_S3_INGRESS=${ZENKO_S3_INGRESS:-'s3.zenko.local'}
 export ZENKO_UI_INGRESS=${ZENKO_UI_INGRESS:-'ui.zenko.local'}
 export ZENKO_SUR_INGRESS=${ZENKO_SUR_INGRESS:-'utilization.zenko.local'}
+export MONGODB_SHARD_COUNT=${MONGODB_SHARD_COUNT:-1}
```
francoisferrand: not used in this file, seems useless?
(if MONGODB_SHARD_COUNT is set, it is already exported... and if not, the default has not much relevance here)

.github/scripts/end2end/install-kind-dependencies.sh
```diff
@@ -186,22 +222,31 @@ mongodb_wait_for_shards() {
...
-    kustomize edit set image \
-        $SOLUTION_REGISTRY/mongodb-sharded=$(get_image_from_deps mongodb-sharded) \
-        $SOLUTION_REGISTRY/os-shell=$(get_image_from_deps mongodb-shell) \
-        $SOLUTION_REGISTRY/mongodb-exporter=$(get_image_from_deps mongodb-sharded-exporter)
+    local base_yaml_path=$(get_mongodb_topology_file $node_count $shard_count)
 
-    kubectl apply -k .
+    sed -i "s|${SOLUTION_REGISTRY}/mongodb-sharded:.*|$(get_image_from_deps mongodb-sharded)|g" "$base_yaml_path"
+    sed -i "s|${SOLUTION_REGISTRY}/os-shell:.*|$(get_image_from_deps mongodb-shell)|g" "$base_yaml_path"
+    sed -i "s|${SOLUTION_REGISTRY}/mongodb-exporter:.*|$(get_image_from_deps mongodb-sharded-exporter)|g" "$base_yaml_path"
```
francoisferrand, line 238: why not keep using `kustomize edit set image`, which is more explicit and kind of safer than sed?
if we really need to change, best to use `yq` to edit the required fields...

.github/scripts/end2end/generate-kustomization.sh
francoisferrand: > as not supported by the provisioner

this is a good learning, to take into account later (ring10) : here we bake the selectors in the Zenko ISO, which actually creates some adherence to Metalk8s...and which we then need to remove in CI :/

.github/scripts/end2end/install-kind-dependencies.sh
```diff
@@ -114,6 +121,53 @@ stringData:
...
+        exit 1
+    fi
+
+    echo "$base_yaml_path"
+}
+
+# MongoDB selectors are not supported in the CI.
+# So we remove them and let the provisioner handle the
+# volume provisioning.
+patch_mongodb_selector() {
```
francoisferrand, line 155: this does the same as the original kustomize, with a few differences:
- handles multiple shards seamlessly (as was your goal i guess)
- does not overwrite/set the storageclass to `standard` (not sure if it is needed)
- does not set the requested storage to `8Gi` (how much gets requested? is it enough?)

francoisferrand (approved):

.github/scripts/end2end/install-kind-dependencies.sh
ghost: The base has 10Gi, which is enough, no need to change it.

> does not overwrite/set the storageclass to standard (not sure if it is needed)

It's done but elsewhere (it was the case already before, we were doing it twice): 

```
    sed -i 's/MONGODB_STORAGE_CLASS/standard/g' $DIR/_build/root/deploy/*
```

> handles multiple shards seamlessly (as was your goal i guess)

yes!

.github/scripts/end2end/install-kind-dependencies.sh
ghost: I now use yq.

Update: yq edits all the fields without distinction, even those where there is no image, meaning we must change every line manually, which lead to even more maintenance. Kept the sed.
The kustomize is not possible as the base yaml is not a kustomize file.

ghost:
Added a commit to remove the matchLabel for the shard0, so that we do not require any downtime during shard expansion.

VERSION
```diff
@@ -1,5 +1,5 @@
 VERSION="2.12.0"
 
-VERSION_SUFFIX=-preview.4
+VERSION_SUFFIX=-preview.5
```
welansari: need to rebase

welansari (approved):

PR #2220, opened 2025-05-07
Expose SUR APIs

ghost:
Adding SUR API backend (exposed) to Zenko. CTST tests are added.


VERSION
```diff
@@ -1,4 +1,4 @@
-VERSION="2.11.6"
+VERSION="2.11.7"
```
francoisferrand: * on 2.12 branch, `VERSION` needs to be `2.12.*`
* this is not a 2.12.0, but only a preview (if anything)

tests/ctst/Dockerfile
```diff
@@ -13,6 +13,7 @@ COPY ./steps /ctst/steps
 COPY ./world /ctst/world
 
 USER root
+RUN npm install typescript -g
```
francoisferrand: best to specify the version of typescript, for reproducible builds

```suggestion
RUN npm install typescript@5.8.3 -g
```

francoisferrand (approved):

.github/scripts/end2end/configs/zenko.yaml
```diff
@@ -85,7 +85,7 @@ spec:
     api:
       replicas: 1
       ingress:
-        hostname: ${ZENKO_SUR_API_INGRESS}
+        hostname: ${ZENKO_SUR_INGRESS}
```
benzekrimaha, line 88: can be squashed with https://github.com/scality/Zenko/pull/2220/commits/7606b5e39671c973ca4febe1c6c3f4db1c29e171 instead of having two separate commits as the referenced one has more context

benzekrimaha (approved):
LGTM

VERSION
ghost: I created the preview 1 of the 2.12.0. Preview 2 will (hopefully) be with multi shard support.

PR #2237, opened 2025-07-22
Create tests to verify that the we correctly handle version ID collisions during object creation

DarkIsDude:
Issue: ZENKO-5023

Define a new test to make sure we don't have collision of versionId.

Fixes https://scality.atlassian.net/browse/ZENKO-5023.

### DIFF
--- a/tests/ctst/features/put-object.feature
+++ b/tests/ctst/features/put-object.feature
@@ -1,0 +1,10 @@
+Feature: Put Object operations
+
+    @2.6.0
+    @PreMerge
+    @AfterAll
+    @ColdStorage
+    Scenario Outline: Handle versionId collision on PutObject
+    Given a "Versioned" bucket
+    And 1000 objects "obj" of size 100 bytes
+    Then no collision should occur on versionId
--- a/tests/ctst/features/put-object.feature
+++ b/tests/ctst/features/put-object.feature
@@ -1,6 +1,6 @@
 Feature: Put Object operations
 
-    @2.6.0
+    @2.12.0
     @PreMerge
     Scenario Outline: Handle versionId collision on PutObject
     Given a "Versioned" bucket
--- a/tests/ctst/features/put-object.feature
+++ b/tests/ctst/features/put-object.feature
@@ -2,8 +2,6 @@ Feature: Put Object operations
 
     @2.6.0
     @PreMerge
-    @AfterAll
-    @ColdStorage
     Scenario Outline: Handle versionId collision on PutObject
     Given a "Versioned" bucket
     And 1000 objects "obj" of size 100 bytes
--- a/tests/ctst/features/put-object.feature
+++ b/tests/ctst/features/put-object.feature
@@ -4,5 +4,5 @@ Feature: Put Object operations
     @PreMerge
     Scenario Outline: Handle versionId collision on PutObject
     Given a "Versioned" bucket
-    And 100 versions of objects "obj" of size 100 bytes with 10 threads
+    And 10 threads each uploading 100 versions of object "obj" of size 100 bytes
     Then 1000 versions of objects "obj" should exist 
--- a/tests/ctst/features/put-object.feature
+++ b/tests/ctst/features/put-object.feature
@@ -4,5 +4,5 @@ Feature: Put Object operations
     @PreMerge
     Scenario Outline: Handle versionId collision on PutObject
     Given a "Versioned" bucket
-    And 1000 objects "obj" of size 100 bytes
-    Then no collision should occur on versionId
+    And 100 versions of objects "obj" of size 100 bytes with 10 threads
+    Then 1000 versions of objects "obj" should exist 
--- a/tests/ctst/steps/put-object.ts
+++ b/tests/ctst/steps/put-object.ts
@@ -1,11 +1,10 @@
 import {Given, Then} from '@cucumber/cucumber';
 import Zenko from 'world/Zenko';
-import {S3, Utils} from 'cli-testing';
+import {S3} from 'cli-testing';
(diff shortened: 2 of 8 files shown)


tests/ctst/features/put-object.feature
```diff
@@ -0,0 +1,8 @@
+Feature: Put Object operations
+
+    @2.6.0
+    @PreMerge
+    Scenario Outline: Handle versionId collision on PutObject
+    Given a "Versioned" bucket
+    And 1000 objects "obj" of size 100 bytes
+    Then no collision should occur on versionId
```
ghost: Maybe we should add a comment : vID collision was quite hard to trigger, usually it's around several per 100,000 tries, because we need two S3 requests to reach the vID generation at the same millisecond. I believe this test can create that issue, because historically, we had many flakies in the "archive" tests, eventually due to that. However, not all runs will trigger the problem, so this makes the test not 100% reliable, but still nice to have...

> And 1000 objects "obj" of size 100 bytes

To be verified, if we do that in parallel, and if yes, how many in parallel.

I would just complete: you cannot detect vID conflict easily by looking for duplicate vID, because by definition, it means we replace an existing version. So the way of detecting would be to up with 99 versions (or less) at the end, meaning something was replaced.

DarkIsDude: @williamlardier thanks for your feedback. The PR was indeed under development and the way that I use to detect collision was not the right one. It's fixed 🙏.

For the other point, it's done ✅. I was using `development/2.11` where the fix was not there. With 1000 objects (100 objects for 10 threads), the difference was at least from 1000 / 995 (after several executions I had 988, 994, 991, 993, ...). So the scenario can reproduce, always, the issue and trigger collisions. Thanks to Walid, I understood that I was not targeting the right branch. The switch to development/2.12 fixed the pipeline, confirmed that the fix is the right one. So I guess we are good enough ?

tests/ctst/features/put-object.feature
```diff
@@ -0,0 +1,8 @@
+Feature: Put Object operations
+
+    @2.6.0
```
ghost:
```suggestion
    @2.12.0
```

The rationale is that we have a logic in ctst able to auto run or skip tests based on the version tag. The benefit is that we can decide to have the same test suite on all (maintained) branches to reduce the integration branches conflicts, by keeping the same implementation, and reuse it in other repos (like artesca) later if needed. So if we add a test starting 2.12 it's better to set 2.12 as the version tag

See https://github.com/scality/cli-testing/blob/main/hooks/versionTags.ts

(it also helps with upgrade tests, to run different versions of a same test if we were to automate the product upgrade in between, but not used in zenko so far).

tests/ctst/steps/put-object.ts
```diff
@@ -0,0 +1,86 @@
...
+    numberOfThread: number,
+    numberOfVerionPerThread: number,
+    objectName: string,
+    sizeBytes: number,
+) {
+    const bucketName = this.getSaved<string>('bucketName');
+    this.addToSaved('objectName', objectName);
+    let processedCounter = numberOfVerionPerThread * numberOfThread;
+
+    await Promise.all(Array.from({length: numberOfThread}, async () => {
```
ghost:
```suggestion
    await Promise.all(Array.from({ length: numberOfThread }, async () => {
```

tests/ctst/steps/put-object.ts
```diff
@@ -0,0 +1,86 @@
...
+        for (let i = 0; i < numberOfVerionPerThread; i++) {
+            const tempFileName = `${Utils.randomString()}_${objectName}`;
+            const objectBody = 'a'.repeat(sizeBytes);
+            await saveAsFile(tempFileName, objectBody);
+
+            const result = await S3.putObject({
```
ghost: To maximize our chance of hitting the race condition, maybe we should create the files first, then call the putObject in parallel, with a limit (async.parallelLimit?), as here we are doing the calls sequentially, so even with the 100 threads, we might be calling s3 almost never at the same times?

Another way to implement it would be to:
- First pre-create all files (or just use the same one for all as an optim?)
- Batch 100 calls to s3
- Wait for it to complete
- Do it 10 times

That way each batch will have much more chance to trigger the race condition, what do you think?

welansari: Just a note on this, the bug is purely on the metadata layer so we could simplify the tests and make them run faster by uploading empty objects with no data.

tests/ctst/steps/put-object.ts
DarkIsDude: Does it make sense to add this in the aslant config ? In the same PR ?

ghost: No, I think this is also not detected by it? It was just a small nit, usually we have spaces, definitely not a problem otherwise

tests/ctst/steps/put-object.ts
DarkIsDude: @williamlardier as checked together, I tried to rerun test suite to make sure we don't have more flaky tests. The CI seems to run with, from time to time, some flaky tests (never the new one). Seems that the current test suite is not degraded. So I guess we can move on.

@Kerkesni good point. I can update the code but I'm not sure it's a good move. We are in e2e test suite, not really a unit one or integration one. So we should try to be very close from production. That's why I would like to keep the file (even if it's not tested). Let me know what do you think, I don't have a strong opinion right now on it, as I'm still a new joiner.

ghost: I think we could at least reuse the same file, or create them outside of this for loop?

We also have the `uploadSetup` and `uploadTeardown` that coul dbe helpful here, as they handle removing the resources after the call, to avoid filling the test platform.

ghost (approved):

welansari (approved):

PR #2239, opened 2025-07-25
add tests for crr existing objects

SylvainSenechal:
Issue : ZENKO-5030

### DIFF
--- a/tests/ctst/features/crrReplicationS3utils.feature
+++ b/tests/ctst/features/crrReplicationS3utils.feature
@@ -1,0 +1,14 @@
+Feature: Replication
+    This feature tests replication using the script crrExistingObjects.js from S3utils,
+    that replicates objects created before the replication was enabled on source the bucket.
+
+    @2.12.0
+    @PreMerge
+    @ReplicationTest
+    Scenario Outline: Replicate objects created before creating the replication rule
+        Given an existing bucket "source-bucket" "with" versioning, "without" ObjectLock "without" retention mode
+        And an object "source-object-1" that "exists"
+        And a replication configuration to "awsbackendmismatch" location
+        When I run the job to replicate existing objects with status "NEW"
+        Then the object should eventually be replicated
+        And the replicated object should be the same as the source object
--- a/tests/ctst/steps/replication.ts
+++ b/tests/ctst/steps/replication.ts
@@ -1,0 +1,175 @@
+import { When, Then } from '@cucumber/cucumber';
+import Zenko from '../world/Zenko';
+import { createAndRunPod, getMongoDBConfig, getZenkoVersion } from 'steps/utils/kubernetes';
+import assert from 'assert';
+import { GetObjectCommand } from '@aws-sdk/client-s3';
+import { Utils } from 'cli-testing';
+import { getObject, headObject, getReplicationLocationConfig } from 'steps/utils/utils';
+import { safeJsonParse } from 'common/utils';
+
+When('I run the job to replicate existing objects with status {string}',
+    { timeout: 600000 },
+    async function (
+        this: Zenko,
+        sourceObjectStatus: string,
+    ) {
+        const sourceBucket = this.getSaved<string>('bucketName');
+        const replicationLocation = this.getSaved<string>('replicationLocation');
+        const { replicaSetHosts } = await getMongoDBConfig(this);
+        const { locationType } = await getReplicationLocationConfig(this, replicationLocation);
+        const zenkoVersion = await getZenkoVersion(this);
+        const s3utilsVersion = zenkoVersion.spec.versions.s3utils;
+        const podManifest = {
+            apiVersion: 'v1',
+            kind: 'Pod',
+            metadata: {
+                name: `s3utils-crr-existing-${Utils.randomString().toLowerCase()}`,
+                namespace: 'default',
+                labels: {
+                    app: 's3utils',
+                    script: 'crrExistingObjects.js'
+                }
+            },
+            spec: {
+                restartPolicy: 'Never',
+                containers: [
+                    {
+                        name: 's3utils',
+                        image: `${s3utilsVersion.image}:${s3utilsVersion.tag}`,
+                        command: ['node'],
+                        args: ['crrExistingObjects.js', sourceBucket],
+                        env: [
+                            {
+                                name: 'MONGODB_REPLICASET', 
+                                value: replicaSetHosts.join(',')
+                            },
+                            { 
+                                name: 'MONGODB_AUTH_USERNAME', 
+                                valueFrom: { 
+                                    secretKeyRef: { 
+                                        name: 'mongodb-db-creds', 
+                                        key: 'mongodb-username' 
+                                    } 
+                                } 
+                            },
+                            { 
+                                name: 'MONGODB_AUTH_PASSWORD', 
+                                valueFrom: { 
+                                    secretKeyRef: { 
+                                        name: 'mongodb-db-creds', 
+                                        key: 'mongodb-password' 
+                                    } 
+                                } 
+                            },
+                            { 
+                                name: 'MONGODB_DATABASE', 
+                                valueFrom: { 
+                                    secretKeyRef: { 
+                                        name: 'mongodb-db-creds', 
+                                        key: 'mongodb-database' 
+                                    } 
+                                } 
+                            },
+                            { name: 'MONGODB_SHARD_COLLECTIONS', value: 'true' },
+                            { name: 'STORAGE_TYPE', value: locationType },
+                            { name: 'TARGET_REPLICATION_STATUS', value: sourceObjectStatus },
+                            { name: 'SITE_NAME', value: replicationLocation },
+                        ]
+                    }
+                ]
+            }
+        };
+        
+        await createAndRunPod(this, podManifest);
+    });
+
+Then('the object should eventually be replicated',
+    async function (this: Zenko) {
+        const objectName = this.getSaved<string>('objectName');
+        const bucketSource = this.getSaved<string>('bucketName');
+        const startTime = Date.now();
+        const replicationTimeoutMs = 90_000;
+        while (Date.now() - startTime < replicationTimeoutMs) {
+            await new Promise(resolve => setTimeout(resolve, 3000));
+
+            const response = await headObject(this, objectName, bucketSource);
+            assert(response.stdout);
+            assert.strictEqual(response.statusCode, 200, `failed to headobject, ${response.statusCode}`);
+            const parsed = safeJsonParse<{
+                ReplicationStatus?: string;
+                LastModified?: string;
+                ETag?: string;
+                ContentLength?: number;
+                VersionId?: string;
+                Metadata?: Record<string, string>;
+            }>(response.stdout || '{}');
+            assert(parsed.ok);
+            const replicationStatus = parsed.result?.ReplicationStatus;
+            assert.notStrictEqual(replicationStatus, 'FAILED', `replication failed for object ${objectName}`);
+            if (replicationStatus === 'COMPLETED') {
+                return;
+            }
+            if (replicationStatus === 'PENDING' || replicationStatus === 'PROCESSING') {
+                continue;
+            }
+        }
+        assert.fail(`Timeout: Object '${objectName}' was not replicated successfully until timeout`);
+    });
+
+Then(
+    'the replicated object should be the same as the source object',
+    async function (
+        this: Zenko,
+    ) {
+        const objectName = this.getSaved<string>('objectName');
+        const bucketSource = this.getSaved<string>('bucketName');
+        const replicationLocation = this.getSaved<string>('replicationLocation');
+        const { destinationBucket, bucketMatch, awsS3Client } = 
+            await getReplicationLocationConfig(this, replicationLocation);
+        
+        // When bucketMatch is disabled on the destination bucket,
+        // replicated objects are named sourceBucket/objectName
+        let key = `${bucketSource}/${objectName}`;
+        if (bucketMatch) {
+            key = objectName;
+        }
+        
+        const command = new GetObjectCommand({
+            Bucket: destinationBucket,
+            Key: key, 
+        });
+        const replicaObj = await awsS3Client.send(command);
+        const sourceResponse = await getObject(this, objectName, bucketSource);
+        assert.strictEqual(sourceResponse.statusCode, 200, `failed to getObject, ${sourceResponse.statusCode}`);
+        const sourceObj = safeJsonParse<{
+            ReplicationStatus?: string;
+            LastModified?: string;
+            ETag?: string;
+            ContentLength?: number;
+            VersionId?: string;
+            Metadata?: Record<string, string>;
+        }>(sourceResponse.stdout || '{}');
+        assert(sourceObj.ok);
+
+        assert.strictEqual(sourceObj.result?.ReplicationStatus, 'COMPLETED');
+        assert.strictEqual(
+            sourceObj.result?.ContentLength,
+            replicaObj.ContentLength
+        );
+        assert.strictEqual(
+            sourceObj.result?.Metadata?.[`${replicationLocation}-version-id`],
+            replicaObj.VersionId
+        );
+        assert.strictEqual(
+            sourceObj.result?.Metadata?.[`${replicationLocation}-replication-status`],
+            'COMPLETED'
+        );
+        assert.strictEqual(
+            sourceObj.result?.VersionId,
+            replicaObj.Metadata?.['scal-version-id']
+        );
+        assert.strictEqual(
+            replicaObj.Metadata?.['scal-replication-status'],
+            'REPLICA'
+        );
+    });
--- a/tests/ctst/steps/utils/kubernetes.ts
+++ b/tests/ctst/steps/utils/kubernetes.ts
@@ -13,6 +13,7 @@ import {
     V1PersistentVolumeClaim,
     CoreV1Api,
     BatchV1Api,
+    V1Pod,
 } from '@kubernetes/client-node';
 
 type ZenkoStatusValue = {
@@ -24,6 +25,24 @@ type ZenkoStatusValue = {
 };
 
 type ZenkoStatus = ZenkoStatusValue[];
+
+type ZenkoVersion = {
+    apiVersion: string;
+    kind: string;
+    metadata: {
+        name: string;
+    };
+    spec: ZenkoVersionSpec;
+};
+
+type ZenkoVersionSpec = {
+    versions: {
+        s3utils: {
+            image: string;
+            tag: string;
+        };
+    };
+};
 
 export function createKubeBatchClient(world: Zenko): BatchV1Api {
     if (!KubernetesHelper.clientBatch) {
@@ -165,6 +184,74 @@ export async function createJobAndWaitForCompletion(
     }
 }
 
+export async function createAndRunPod(
+    world: Zenko,
+    podManifest: V1Pod,
+    waitForCompletion = true,
+    cleanup = false, // The pod will be visible in the artifacts is set to false
+    timeout = 300000,
+) {
+    const clientCore = createKubeCoreClient(world);
+    const watchClient = createKubeWatchClient(world);
+
+    try {
+        const response = await clientCore.createNamespacedPod('default', podManifest);
+        const podName = response.body.metadata?.name;
+        if (waitForCompletion && podName) {
+            world.logger.debug('Waiting for pod completion', { podName });
+
+            await new Promise<void>((resolve, reject) => {
+                const timeoutId = setTimeout(() => {
+                    reject(new Error(`Pod ${podName} did not complete within ${timeout}ms`));
+                }, timeout);
+
+                void watchClient.watch(
+                    '/api/v1/namespaces/default/pods',
+                    {},
+                    (type: string, apiObj, watchObj) => {
+                        if (watchObj.object?.metadata?.name === podName) {
+                            const phase = watchObj.object?.status?.phase;
+                            world.logger.debug('Pod status update', { podName, phase });
+                            
+                            if (phase === 'Succeeded') {
+                                clearTimeout(timeoutId);
+                                world.logger.debug('Pod completed successfully', { podName });
+                                resolve();
+                            } else if (phase === 'Failed') {
+                                clearTimeout(timeoutId);
+                                world.logger.error('Pod failed', { 
+                                    podName, 
+                                    status: watchObj.object?.status 
+                                });
+                                reject(new Error(`Pod ${podName} failed`));
+                            }
+                        }
+                    },
+                    err => {
+                        world.logger.debug('Watch error callback triggered', { podName, err });
+                        clearTimeout(timeoutId);
+                        reject(err);
+                    }
+                );
+            });
+        }
+
+        // Cleanup if requested
+        if (cleanup && podName) {
+            world.logger.debug('Cleaning up pod', { podName });
+            try {
+                await clientCore.deleteNamespacedPod(podName, 'default');
+            } catch (cleanupErr) {
+                world.logger.warn('Failed to cleanup pod', { podName, err: cleanupErr });
+            }
+        }
+
+        return response.body;
+    } catch (err: unknown) {
+        world.logger.error('Failed to create and run pod:', { err });
+        throw new Error(`Failed to create and run pod: ${err}`);
+    }
+}
 
 export async function waitForZenkoToStabilize(
     world: Zenko, needsReconciliation = false, timeout = 15 * 60 * 1000, namespace = 'default') {
@@ -445,3 +532,86 @@ export async function createSecret(
         throw err;
     }
 }
+
+export async function getMongoDBConfig(
+    world: Zenko,
+    namespace = 'default',
+) : Promise<{replicaSetHosts: string[]}> {
+    const customObjectClient = createKubeCustomObjectClient(world);
+    try {
+        // Get replicaSetHosts from Zenko CR
+        const zenkoCR = await customObjectClient.getNamespacedCustomObject(
+            'zenko.io',
+            'v1alpha2',
+            namespace,
+            'zenkos',
+            'end2end'
+        );
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        const mongodbSpec = (zenkoCR.body as any)?.spec?.mongodb;
+        const mongodbConfig = {
+            replicaSetHosts: mongodbSpec?.endpoints || [],
+        };
+
+        return mongodbConfig;
+    } catch (err) {
+        world.logger.debug('Error getting MongoDB config from secret and CR', {namespace, err});
+        throw err;
+    }
+}
+
+export async function getLocationConfigs(
+    world: Zenko,
+    namespace = 'default',
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+): Promise<Record<string, any>> {
+    const coreClient = createKubeCoreClient(world);
+    try {
+        // Get location configurations from connector-cloudserver-config secret
+        const secretList = await coreClient.listNamespacedSecret(
+            namespace,
+            undefined,
+            undefined,
+            undefined,
+            undefined,
+            'app.kubernetes.io/name=connector-cloudserver-config'
+        );
+
+        const secret = secretList.body.items[0];
+        const locationConfigData = secret.data?.['locationConfig.json'];
+        if (!locationConfigData) {
+            throw new Error('locationConfig.json not found in secret');
+        }
+
+        const locationConfigJson = Buffer.from(locationConfigData, 'base64').toString('utf-8');
+        return JSON.parse(locationConfigJson);
+    } catch (err) {
+        world.logger.debug('Error getting location configs from secret', { namespace, err });
+        throw err;
+    }
+}
+
+export async function getZenkoVersion(
+    world: Zenko,
+    namespace = 'default',
+): Promise<ZenkoVersion> {
+    const customObjectClient = createKubeCustomObjectClient(world);
+    try {
+        const zenkoVersionList = await customObjectClient.listNamespacedCustomObject(
+            'zenko.io',
+            'v1alpha1',
+            namespace,
+            'zenkoversions'
+        );
+        const zenkoVersionItems = (zenkoVersionList.body as { items: ZenkoVersion[] })?.items;
+        if (!zenkoVersionItems || zenkoVersionItems.length === 0) {
+            throw new Error('No ZenkoVersion resources found');
+        }
+
+        return zenkoVersionItems[0];
+    } catch (err) {
+        world.logger.debug('Error getting ZenkoVersion resource', { namespace, err });
+        throw err;
+    }
+}
+
--- a/tests/ctst/steps/utils/utils.ts
+++ b/tests/ctst/steps/utils/utils.ts
@@ -13,6 +13,8 @@ import { extractPropertyFromResults, s3FunctionExtraParams, safeJsonParse } from
 import Zenko from 'world/Zenko';
 import assert from 'assert';
 import constants from 'common/constants';
+import { getLocationConfigs } from './kubernetes';
+import { S3Client } from '@aws-sdk/client-s3';
 
 enum AuthorizationType {
     ALLOW = 'Allow',
@@ -300,6 +302,24 @@ async function putObject(world: Zenko, objectName?: string, content?: string) {
     return result;
 }
 
+async function getObject(world: Zenko, objectKey: string, bucketName: string): Promise<Utils.Command> {
+    const result = await S3.getObject({
+        key: objectKey,
+        bucket: bucketName,
+    });
+
+    return result;
+}
+
+async function headObject(world: Zenko, objectKey: string, bucketName: string): Promise<Utils.Command> {
+    const result = await S3.headObject({
+        key: objectKey,
+        bucket: bucketName,
+    });
+
+    return result;
+}
+
 function getAuthorizationConfiguration(world: Zenko): AuthorizationConfiguration {
     return {
         Identity: world.getSaved<AuthorizationConfiguration>('authzConfiguration')?.Identity
@@ -371,6 +391,71 @@ async function addTransitionWorkflow(this: Zenko, location: string, enabled = tr
         conditionOk = res.err === null;
         // Wait for the transition to be accepted because the deployment of the location's pods can take some time
         await Utils.sleep(5000); 
+    }
+}
+
+async function getReplicationLocationConfig(world: Zenko, location: string): Promise<{
+        destinationBucket: string;
+        locationType: string;
+        bucketMatch: boolean;
+        awsS3Client: S3Client;
+    }> {
+    const locationsConfigs = await getLocationConfigs(world);
+    if (!locationsConfigs[location]) {
+        throw new Error(`Unsupported replication location: '${location}'`);
+    }
+    return {
+        destinationBucket: locationsConfigs[location].details.bucketName,
+        locationType: locationsConfigs[location].type,
+        bucketMatch: locationsConfigs[location].details.bucketMatch,
+        awsS3Client: new S3Client({
+            region: locationsConfigs[location].details.region,
+            endpoint: `https://${locationsConfigs[location].details.awsEndpoint}`,
+            credentials: {
+                accessKeyId: locationsConfigs[location].details.credentials.accessKey,
+                secretAccessKey: locationsConfigs[location].details.credentials.secretKey,
+            },
+            tls: false,
+            maxAttempts: 1,
+            forcePathStyle: true,
+        }),
+    };
+}
+
+async function putBucketReplication(
+    this: Zenko,
+    srcBucket: string,
+    replicationLocation: string
+) {
+    this.resetCommand();
+    this.addCommandParameter({ bucket: srcBucket });
+    this.addCommandParameter({
+        replicationConfiguration: JSON.stringify({
+            Role: 'arn:aws:iam::root:role/s3-replication-role',
+            Rules: [
+                {
+                    Prefix: '',
+                    Destination: {
+                        // eslint-disable-next-line max-len
+                        // https://documentation.scality.com/Artesca/4.0.1/data_management/bucket_operations/replication_workflow/create_a_replication_workflow.html
+                        Bucket: `arn:aws:s3:::${srcBucket}`,
+                        StorageClass: replicationLocation,
+                    },
+                    Status: 'Enabled',
+                },
+            ],
+        }),
+    });
+    
+    const commandParameters = this.getCommandParameters();
+    const res = await S3.putBucketReplication(commandParameters);
+    if (res.err) {
+        this.logger.error('Failed to put bucket replication', {
+            srcBucket,
+            replicationLocation,
+            error: res.err,
+        });
+        throw new Error(`Failed to put bucket replication, err : ${res.err}`);
     }
 }
 
@@ -475,10 +560,14 @@ export {
     putMpuObject,
     copyObject,
     putObject,
+    getObject,
+    headObject,
     emptyNonVersionedBucket,
     emptyVersionedBucket,
     verifyObjectLocation,
     getObjectNameWithBackendFlakiness,
     restoreObject,
     addTransitionWorkflow,
+    getReplicationLocationConfig,
+    putBucketReplication,
 };
--- a/tests/ctst/common/common.ts
+++ b/tests/ctst/common/common.ts
@@ -15,6 +15,7 @@ import {
     verifyObjectLocation,
     restoreObject,
     addTransitionWorkflow,
+    putBucketReplication,
 } from 'steps/utils/utils';
 import { ActionPermissionsType } from 'steps/bucket-policies/utils';
 import constants from './constants';
@@ -258,6 +259,12 @@ Then('object {string} should have the user metadata with key {string} and value 
 Given('a transition workflow to {string} location', async function (this: Zenko, location: string) {
     await addTransitionWorkflow.call(this, location);
 });
+
+Given('a replication configuration to {string} location',
+    async function (this: Zenko, replicationLocation: string) {
+        this.addToSaved('replicationLocation', replicationLocation);
+        await putBucketReplication.call(this, this.getSaved<string>('bucketName'), replicationLocation);
+    });
 
 When('i restore object {string} for {int} days', async function (this: Zenko, objectName: string, days: number) {
     await restoreObject.call(this, objectName, days);


tests/ctst/features/replicatateExistingObjects.feature
```diff
@@ -0,0 +1,19 @@
+Feature: Replicate existing objects created before enabling replication
+    This feature tests the script crrExistingObjects.js from S3utils that replicates
+    objects created before the replication was enabled on source the bucket.
+
+    @2.7.0
```
SylvainSenechal: Let me know if some tags should be different

.github/scripts/end2end/run-e2e-ctst.sh
```diff
@@ -207,4 +221,4 @@ kubectl run $POD_NAME \
       }
     ]
   }
-}' -- ./run "$COMMAND" $WORLD_PARAMETERS --parallel $PARALLEL_RUNS --retry $RETRIES --retry-tag-filter @Flaky --format junit:$JUNIT_REPORT_PATH "$@"
+}' -- ./run "$COMMAND" $WORLD_PARAMETERS --parallel $PARALLEL_RUNS --retry $RETRIES --retry-tag-filter @Flaky --format junit:$JUNIT_REPORT_PATH "$@" --tags '@ReplicationTest'
```
SylvainSenechal: todo : remove

tests/ctst/world/Zenko.ts
```diff
@@ -89,6 +90,15 @@ export interface ZenkoWorldParameters extends ClientOptions {
     SorbetdRestoreTimeout: string;
     UtilizationServiceHost: string;
     UtilizationServicePort: string;
+    AwsBackendDestinationReplicationLocation: string;
+    AwsReplicationBucketName: string;
+    AwsEndpoint: string;
```
welansari: Consider making the variable names more explicit to indicate they are specific to our AWS location

tests/ctst/world/Zenko.ts
```diff
@@ -194,6 +206,20 @@ export default class Zenko extends World<ZenkoWorldParameters> {
...
+                endpoint: this.parameters.AwsEndpoint,
+                region: 'us-east-1',
+                forcePathStyle: true,
+                credentials: {
+                    accessKeyId: this.parameters.AwsAccessKey,
+                    secretAccessKey: this.parameters.AwsSecretKey,
+                },
+                maxAttempts: 1,
+            });
+        }
```
welansari: Should we have this in a utility file somewhere ? Not sure this is the best place to have it.

tests/ctst/features/replicatateExistingObjects.feature
```diff
@@ -0,0 +1,19 @@
...
+    @PreMerge
+    @ReplicationTest
+    Scenario Outline: Object created before adding the replication setting is replicated with the script
+        Given an existing bucket "<sourceBucket>" "with" versioning, "without" ObjectLock "without" retention mode
+        And an object "<sourceObject>" that "exists"
+        And a replication configuration from "<sourceBucket>" bucket to "<replicationLocation>" location
+        Then the object "<sourceObject>" should "not exists" in AwsS3 bucket "<replicationBucket>"
```
welansari:
```suggestion
        Then the object "<sourceObject>" should "not exists" in the destination's "<replicationBucket>" bucket
```

tests/ctst/features/replicatateExistingObjects.feature
```diff
@@ -0,0 +1,19 @@
...
+    @2.7.0
+    @PreMerge
+    @ReplicationTest
+    Scenario Outline: Object created before adding the replication setting is replicated with the script
+        Given an existing bucket "<sourceBucket>" "with" versioning, "without" ObjectLock "without" retention mode
+        And an object "<sourceObject>" that "exists"
+        And a replication configuration from "<sourceBucket>" bucket to "<replicationLocation>" location
+        Then the object "<sourceObject>" should "not exists" in AwsS3 bucket "<replicationBucket>" 
+        When I run the job "replication" to replicate existing objects for bucket "<sourceBucket>" to location "<replicationLocation>"
+        And I wait 60 seconds
```
welansari: Should wait in the next step while continuously checking the replication status. This systimatically wait for 60 seconds every time even if the replication itself takes only a few seconds.

We should use the replication status to check if the object got replicated before verifying the destination's bucket, the replication status is available in the headObject response.

tests/ctst/features/replicatateExistingObjects.feature
```diff
@@ -0,0 +1,19 @@
...
+
+    @2.7.0
+    @PreMerge
+    @ReplicationTest
+    Scenario Outline: Object created before adding the replication setting is replicated with the script
+        Given an existing bucket "<sourceBucket>" "with" versioning, "without" ObjectLock "without" retention mode
+        And an object "<sourceObject>" that "exists"
+        And a replication configuration from "<sourceBucket>" bucket to "<replicationLocation>" location
+        Then the object "<sourceObject>" should "not exists" in AwsS3 bucket "<replicationBucket>" 
+        When I run the job "replication" to replicate existing objects for bucket "<sourceBucket>" to location "<replicationLocation>"
```
welansari: should make it clear that we are running crrExistingObjects

tests/ctst/steps/k8s-jobs/replication.ts
```diff
@@ -0,0 +1,82 @@
...
+            Key: objectName,
+        });
+
+        if (objectExists === 'exists') {
+            try {
+                await awsS3Client.send(command);
+                this.logger.info(`Success: Object '${objectName}' was found as expected.`);
+            } catch (error: unknown) {
+                assert.fail(`Expected object '${objectName}' to exist, but it was not found. Error: ${error}`);
+            }
```
welansari: Should check the integrity of the object just like we do in the other non ctst tests.

tests/ctst/steps/k8s-jobs/replication.ts
```diff
@@ -0,0 +1,82 @@
...
+
+When('I run the job {string} to replicate existing objects for bucket {string} to location {string}',
+    { timeout: 600000 },
+    async function (this: Zenko, jobName: string, sourceBucket: string, location: string) {
+        const timestamp = Date.now();
+        const podManifest = {
+            apiVersion: 'v1',
+            kind: 'Pod',
+            metadata: {
+                name: `s3utils-crr-existing-${jobName}-${timestamp}`,
```
welansari: I don't see the use of `jobName` here. A simple uuid is enough, the `timestamp` might not be enough to guarantee we don't have collisions when running multiple tests in parallel

tests/ctst/steps/k8s-jobs/replication.ts
```diff
@@ -0,0 +1,82 @@
+import { When, Then } from '@cucumber/cucumber';
```
welansari: I don't understand why we store this file inside `k8s-jobs` folder instead of keeping it with other files ?

tests/zenko_tests/e2e-config.yaml.template
```diff
@@ -30,6 +30,15 @@ locations:
...
       bucketMatch: yes
       repoId: []
+  - name: "${AWS_BACKEND_DESTINATION_REPLICATION_CTST_LOCATION}"
+    locationType: "location-aws-s3-v1"
+    details:
+      bucketName: "${AWS_REPLICATION_CTST_BUCKET_NAME}"
+      endpoint: "${AWS_ENDPOINT}"
+      accessKey: "${AWS_ACCESS_KEY}"
+      secretKey: "${AWS_SECRET_KEY}"
+      bucketMatch: yes
```
welansari: `bucketMatch` should be disabled to allow usage of this destination accross multiple tests in parallel.
Context for other reviewers: Disabling `bucketMatch` makes object keys get prefixed with the source bucket's name. (ex. `source-bucket/key0`)

tests/ctst/steps/utils/utils.ts
```diff
@@ -374,6 +374,54 @@ async function addTransitionWorkflow(this: Zenko, location: string, enabled = tr
...
+
+    switch (location) {
+    case 'aws-replication-location':
+        storageClass = this.parameters.AwsBackendDestinationReplicationLocation;
+        destinationBucket = this.parameters.AwsReplicationBucketName;
+        this.logger.debug('Using AWS replication location', { storageClass, destinationBucket });
+        break;
+    default:
+        throw new Error(`Unsupported replication location: '${location}'`);
+    }
```
welansari: should have a separate helper function for this ?

tests/ctst/steps/k8s-jobs/replication.ts
```diff
@@ -0,0 +1,82 @@
...
+                    app: 's3utils',
+                    script: 'crrExistingObjects.js'
+                }
+            },
+            spec: {
+                restartPolicy: 'Never',
+                containers: [
+                    {
+                        name: 's3utils',
+                        image: 'ghcr.io/scality/s3utils:1.14.18',
```
welansari: should not hardcode the version of s3utils. The goal of these tests is testing the currently deployed version

tests/ctst/steps/k8s-jobs/replication.ts
```diff
@@ -0,0 +1,82 @@
...
+                        command: ['node'],
+                        args: ['crrExistingObjects.js', sourceBucket],
+                        env: [
+                            { name: 'MONGODB_REPLICASET', value: this.parameters.MongodbReplicaSet },
+                            { name: 'MONGODB_AUTH_USERNAME', value: this.parameters.MongodbAuthUsername },
+                            { name: 'MONGODB_AUTH_PASSWORD', value: this.parameters.MongodbAuthPassword },
+                            { name: 'MONGODB_SHARD_COLLECTIONS', value: 'true' },
+                            { name: 'MONGODB_DATABASE', value: this.parameters.MongodbDatabase },
+                            { name: 'STORAGE_TYPE', value: 'aws_s3' },
+                            { name: 'TARGET_REPLICATION_STATUS', value: 'NEW' },
```
welansari: Should not hard-code these two values.

`STORAGE_TYPE` => Should check if this is an appropriate solution, but i think an easy way to handle the locations is to pass the full list of location details to ctst. This is available in either cloudserver or backbeat config secrets in the file `locationConfig.json` => you can get it using `kubectl get secrets -l app.kubernetes.io/name=connector-cloudserver-config -o jsonpath='{.items[0].data.locationConfig\.json}'`

`TARGET_REPLICATION_STATUS` => we will need to make it configurable when wanting to test retrying objects that failed replication

tests/ctst/world/Zenko.ts
SylvainSenechal: yes I'm gonna move it somewhere in utils

SylvainSenechal: Actually I'm gonna keep it on the class as it's using the env variable, but not store it in the class member. Just call getAwsReplicationS3Client when needed

tests/ctst/steps/k8s-jobs/replication.ts
SylvainSenechal: I moved this file to the parent folder, it doesn't make much sense to create this k8s-job folder. But I think it deserves its own replication.ts file (looking at the rest of the of the folder we have notifications.ts for example)

tests/ctst/features/replicatateExistingObjects.feature
SylvainSenechal: yes I added a proper waiting loop

tests/ctst/features/replicatateExistingObjects.feature
francoisferrand: should work for "every" version of Zenko, so no tag should be needed?
@williamlardier can you please confirm?

.github/scripts/end2end/patch-coredns.sh
```diff
@@ -14,6 +14,7 @@ corefile="
     rewrite name exact ci-zenko-aws-crr-target-bucket.aws-mock.zenko.local ingress-nginx-controller.ingress-nginx.svc.cluster.local
     rewrite name exact ci-zenko-aws-fail-target-bucket.aws-mock.zenko.local ingress-nginx-controller.ingress-nginx.svc.cluster.local
     rewrite name exact ci-zenko-aws-target-bucket.aws-mock.zenko.local ingress-nginx-controller.ingress-nginx.svc.cluster.local
+    rewrite name exact ci-zenko-aws-replication-ctst-bucket.aws-mock.zenko.local ingress-nginx-controller.ingress-nginx.svc.cluster.local
```
francoisferrand: why do we need a new endpoint?
can't we use the existing locations (and/or mocks) for the tests?

(this would simplify setup quite a bit, by reducing setup and avoiding the introduction of yet more variables ; plus avoid creating an extra location and associated delay...)

Note that it may be better to use a local/mock location instead of AWS, to avoid "integration" issues that may happen if AWS is down and ensure we fully control (in the tests) all buckets. I think we already have such locations: $RING_S3C_BACKEND_SOURCE_LOCATION and $RING_S3C_BACKEND_SOURCE_NON_VERSIONED_LOCATION

tests/ctst/features/replicatateExistingObjects.feature
```diff
@@ -0,0 +1,18 @@
+Feature: Replicate existing objects created before enabling replication
```
francoisferrand: maybe the "Feature" should be `replication`; and we will later expend it to test live replication as well.

tests/ctst/features/replicatateExistingObjects.feature
```diff
@@ -0,0 +1,18 @@
...
+    This feature tests the script crrExistingObjects.js from S3utils that replicates
+    objects created before the replication was enabled on source the bucket.
+
+    @2.7.0
+    @PreMerge
+    @ReplicationTest
+    Scenario Outline: Object created before adding the replication setting is replicated with the script
```
francoisferrand:
```suggestion
    Scenario Outline: Replicate objects created before creating replication rule
```

tests/ctst/features/replicatateExistingObjects.feature
```diff
@@ -0,0 +1,18 @@
...
+        Given an existing bucket "<sourceBucket>" "with" versioning, "without" ObjectLock "without" retention mode
+        And an object "<sourceObject>" that "exists"
+        And a replication configuration from "<sourceBucket>" bucket to "<replicationLocation>" location
+        When I run the job to replicate existing objects with status "<sourceObjStatus>" from bucket "<sourceBucket>" to location "<replicationLocation>"
+        Then the object "<sourceObject>" should be replicated within 60 seconds from "<sourceBucket>" bucket
+        And the replicated object "<sourceObject>" from "<sourceBucket>" bucket to "<replicationBucket>" bucket with location "<replicationLocation>" should be the same as its source object
+
+        Examples:
+            | sourceBucket  | sourceObjStatus |                    replicationBucket |    sourceObject |       replicationLocation |
+            | source-bucket |             NEW | ci-zenko-aws-replication-ctst-bucket | source-object-1 | awsbackendreplicationctst |
```
francoisferrand: Replicating "existing" objects is just one use for crrExistingObjects.

There is however another use case we will want to test: it can also be used to trigger replication of objects which failed to replicate.
It may be done in a separate PR and/or ticket though, as making the replication fail (in the test setup) may be complicated...

(one way to do that may be by setting bucket policy that prevents writing in the bucket, then setting up live replication and waiting
until the object has indeed failed to replicate; then we can remove bucket policy and run crrExistingObjects, and finally watch as it works)

.github/scripts/end2end/patch-coredns.sh
SylvainSenechal: I wanted to have a separate for these replications test, it didn't seem to expensive to do it for me.
I can use an existing one, not sure if it's problematic that existing ones are already used for other tests

tests/ctst/features/replicatateExistingObjects.feature
SylvainSenechal: Created Zenko 5050 : https://scality.atlassian.net/jira/software/c/projects/OS/boards/268?selectedIssue=ZENKO-5050

tests/ctst/steps/utils/utils.ts
```diff
@@ -374,6 +393,72 @@ async function addTransitionWorkflow(this: Zenko, location: string, enabled = tr
...
+        locationType: string;
+        awsS3Client: S3Client;  
+    } {
+    switch (location) {
+    case 'awsbackendreplicationctst':
+        return {
+            storageClass: world.parameters.AwsBackendDestinationReplicationLocation,
+            destinationBucket: world.parameters.AwsReplicationBucketName,
+            locationType: world.parameters.AwsReplicationLocationType,
+            awsS3Client: new S3Client({
```
francoisferrand: `awsS3Client` is not part of "replication location _config_"...

.github/scripts/end2end/patch-coredns.sh
francoisferrand: - it is kind of expensive in terms of setup (hoping that some day we want to run the tests locally and/or against any artesca, the more location there are the more it may be complicated), and passing all the parameters to the tests
- also zenko _today_ does not scale well with the number of locations. There is no (enforced) hard limit, but we document in artesca that we should not have more than 10 locations. (There are multiple reasons to this: the list is added to some k8s CR; in places the code traverses the whole list of locations; many locations create many kafka consumers in some backbeat processes, which can overload the node workers; ....)

.github/scripts/end2end/run-e2e-ctst.sh
```diff
@@ -78,6 +78,20 @@ SORBETD_RESTORE_TIMEOUT=$(kubectl get zenko ${ZENKO_NAME} -o jsonpath='{.spec.so
...
+
+S3_UTILS_TAG=$(yq eval ".s3utils.tag" ../../../solution/deps.yaml)
+
+LOCATION_CONFIGS=$(kubectl get secrets -l app.kubernetes.io/name=connector-cloudserver-config -o jsonpath='{.items[0].data.locationConfig\.json}' | base64 -di | jq -r .)
+AWS_REPLICATION_ENDPOINT="http://$(echo "$LOCATION_CONFIGS" | jq -r --arg loc "$AWS_BACKEND_DESTINATION_REPLICATION_CTST_LOCATION" '.[$loc].details.awsEndpoint')"
+AWS_REPLICATION_ACCESS_KEY=$(echo "$LOCATION_CONFIGS" | jq -r --arg loc "$AWS_BACKEND_DESTINATION_REPLICATION_CTST_LOCATION" '.[$loc].details.credentials.accessKey')
+AWS_REPLICATION_SECRET_KEY=$(echo "$LOCATION_CONFIGS" | jq -r --arg loc "$AWS_BACKEND_DESTINATION_REPLICATION_CTST_LOCATION" '.[$loc].details.credentials.secretKey')
+AWS_REPLICATION_LOCATION_TYPE=$(echo "$LOCATION_CONFIGS" | jq -r --arg loc "$AWS_BACKEND_DESTINATION_REPLICATION_CTST_LOCATION" '.[$loc].type')
+AWS_REPLICATION_REGION=$(echo "$LOCATION_CONFIGS" | jq -r --arg loc "$AWS_BACKEND_DESTINATION_REPLICATION_CTST_LOCATION" '.[$loc].details.region')
+
```
francoisferrand: in configure-e2e.sh, we pass AWS params from env directly:
```
  --env=AWS_ACCESS_KEY=${AWS_ACCESS_KEY} \
  --env=AWS_SECRET_KEY=${AWS_SECRET_KEY} \
  --env=AWS_ENDPOINT=${AWS_ENDPOINT} \
```

so we should probably do the same?

.github/scripts/end2end/run-e2e-ctst.sh
```diff
@@ -78,6 +78,20 @@ SORBETD_RESTORE_TIMEOUT=$(kubectl get zenko ${ZENKO_NAME} -o jsonpath='{.spec.so
 UTILIZATION_SERVICE_HOST=$(kubectl get zenko ${ZENKO_NAME} -o jsonpath='{.spec.scuba.api.ingress.hostname}')
 UTILIZATION_SERVICE_PORT="80"
 
+MONGODB_REPLICASET=$(kubectl get secrets -l app.kubernetes.io/name=connector-cloudserver-config -o jsonpath='{.items[0].data.config\.json}' | base64 -di | jq -r .mongodb.replicaSetHosts)
+MONGODB_AUTH_USERNAME=$(kubectl get secrets -l app.kubernetes.io/name=connector-cloudserver-config -o jsonpath='{.items[0].data.config\.json}' | base64 -di | jq -r .mongodb.authCredentials.username)
+MONGODB_AUTH_PASSWORD=$(kubectl get secrets -l app.kubernetes.io/name=connector-cloudserver-config -o jsonpath='{.items[0].data.config\.json}' | base64 -di | jq -r .mongodb.authCredentials.password)
+MONGODB_DATABASE=$(kubectl get secrets -l app.kubernetes.io/name=connector-cloudserver-config -o jsonpath='{.items[0].data.config\.json}' | base64 -di | jq -r .mongodb.database)
+
+S3_UTILS_TAG=$(yq eval ".s3utils.tag" ../../../solution/deps.yaml)
+
```
francoisferrand: should be done within the tests, similar to how a user (or CS procedure) uses crrExistingObjects

.github/workflows/end2end.yaml
```diff
@@ -66,6 +67,7 @@ env:
   AWS_BUCKET_NAME: ci-zenko-aws-target-bucket
   AWS_CRR_BUCKET_NAME: ci-zenko-aws-crr-target-bucket
   AWS_FAIL_BUCKET_NAME: ci-zenko-aws-fail-target-bucket
+  AWS_REPLICATION_CTST_BUCKET_NAME: ci-zenko-aws-replication-ctst-bucket
```
francoisferrand: would be great if we can just use AWS_BACKEND_DESTINATION_LOCATION and AWS_CRR_BUCKET_NAME

tests/ctst/steps/replication.ts
```diff
@@ -0,0 +1,148 @@
...
+                        args: ['crrExistingObjects.js', sourceBucket],
+                        env: [
+                            { 
+                                name: 'MONGODB_REPLICASET', 
+                                value: Array.isArray(replicaSetHosts) ? replicaSetHosts.join(',') : replicaSetHosts
+                            },
+                            { name: 'MONGODB_AUTH_USERNAME', value: username },
+                            { name: 'MONGODB_AUTH_PASSWORD', value: password },
+                            { name: 'MONGODB_DATABASE', value: database },
+                            { name: 'MONGODB_SHARD_COLLECTIONS', value: 'true' },
```
SylvainSenechal: These are required for s3utils 1.14
In s3utils 1.16, it will be replaced with aws ak/sk. We bumped s3utils in Zenko 2.13, so I'll handle it in the waterflow

tests/ctst/features/replicatateExistingObjects.feature
ghost, line 1: The file has a strange name :smile:

tests/ctst/features/replicatateExistingObjects.feature
ghost: If the test should work in all versions then we should target development/2.6 and add the @2.6.0 tag.

The tag is used to dynamically skip a test if the current Zenko version does not match. If we set 2.7.0 then we expect the test to work starting 2.7.0. If we want to start at the current base branch, then we want to add @2.12.0

tests/ctst/features/replicatateExistingObjects.feature
```diff
@@ -0,0 +1,18 @@
...
+        Given an existing bucket "<sourceBucket>" "with" versioning, "without" ObjectLock "without" retention mode
+        And an object "<sourceObject>" that "exists"
+        And a replication configuration from "<sourceBucket>" bucket to "<replicationLocation>" location
+        When I run the job to replicate existing objects with status "<sourceObjStatus>" from bucket "<sourceBucket>" to location "<replicationLocation>"
+        Then the object "<sourceObject>" should be replicated within 90 seconds from "<sourceBucket>" bucket
+        And the replicated object "<sourceObject>" from "<sourceBucket>" bucket to "<replicationBucket>" bucket with location "<replicationLocation>" should be the same as its source object
+
+        Examples:
+            | sourceBucket  | sourceObjStatus |              replicationBucket |    sourceObject | replicationLocation |
+            | source-bucket |             NEW | ci-zenko-aws-crr-target-bucket | source-object-1 |  awsbackendmismatch |
```
ghost: If we only have one variant then the examples is likely not needed?

tests/ctst/steps/replication.ts
```diff
@@ -0,0 +1,148 @@
...
+                    app: 's3utils',
+                    script: 'crrExistingObjects.js'
+                }
+            },
+            spec: {
+                restartPolicy: 'Never',
+                containers: [
+                    {
+                        name: 's3utils',
+                        image: `ghcr.io/scality/s3utils:${this.parameters.S3UtilsTag}`,
```
ghost: (minor comment) I believe we have the image locally and do not need to pull it from the ghcr repository? That shouldn't change anything as we never recreate the same tag though...

tests/ctst/steps/utils/kubernetes.ts
```diff
@@ -165,6 +166,20 @@ export async function createJobAndWaitForCompletion(
     }
 }
 
+export async function createAndRunS3UtilsPod(
```
ghost: This function can be more generic, it runs a pod given a manifest, not just s3utils

tests/ctst/steps/utils/kubernetes.ts
```diff
@@ -445,3 +460,59 @@ export async function createSecret(
...
+        if (!configJsonBase64) {
+            throw new Error('config.json not found in secret data');
+        }
+
+        // Decode base64 and parse JSON
+        const configJson = Buffer.from(configJsonBase64, 'base64').toString('utf-8');
+        const config = JSON.parse(configJson);
+
+        // Extract MongoDB configuration values
+        const mongodbConfig = {
```
ghost, line 552: Isn't this already in a dedicated secret? We do create a mongodb secret with all the info. The only one missing is the replicaSetHosts and it's in zenko CR. I would think it's better to rely on the input secrets than the output of the operator, as the config structure might change without notice?

tests/ctst/steps/utils/kubernetes.ts
```diff
@@ -445,3 +460,59 @@ export async function createSecret(
...
+            replicaSetHosts: config.mongodb?.replicaSetHosts,
+            username: config.mongodb?.authCredentials?.username,
+            password: config.mongodb?.authCredentials?.password,
+            database: config.mongodb?.database,
+        };
+
+        return mongodbConfig;
+    } catch (err) {
+        world.logger.error('Error getting MongoDB config from secret', {namespace, err});
```
ghost: can be a debug log: if we throw an error, the test will auto dump all logs in this case, if it failed.

tests/ctst/features/replicatateExistingObjects.feature
```diff
@@ -0,0 +1,18 @@
...
+    @2.7.0
+    @PreMerge
+    @ReplicationTest
+    Scenario Outline: Replicate objects created before creating the replication rule
+        Given an existing bucket "<sourceBucket>" "with" versioning, "without" ObjectLock "without" retention mode
+        And an object "<sourceObject>" that "exists"
+        And a replication configuration from "<sourceBucket>" bucket to "<replicationLocation>" location
+        When I run the job to replicate existing objects with status "<sourceObjStatus>" from bucket "<sourceBucket>" to location "<replicationLocation>"
+        Then the object "<sourceObject>" should be replicated within 90 seconds from "<sourceBucket>" bucket
+        And the replicated object "<sourceObject>" from "<sourceBucket>" bucket to "<replicationBucket>" bucket with location "<replicationLocation>" should be the same as its source object
```
ghost:
```suggestion
        And the replicated object "<sourceObject>" from "<sourceBucket>" bucket replicated to "<replicationBucket>" bucket with location "<replicationLocation>" should be the same as the source object
```

tests/ctst/features/replicatateExistingObjects.feature
SylvainSenechal: Then i guess it should be 2.12, as this test will not exist in previous branches anyways 🤔

tests/ctst/features/replicatateExistingObjects.feature
SylvainSenechal: yeah I'll rename it

tests/ctst/features/replicatateExistingObjects.feature
SylvainSenechal: I'll remove the variables used only once. But some of them are used 2/3 times, I preferred to have the value defined in one place

.github/scripts/end2end/run-e2e-ctst.sh
```diff
@@ -78,6 +78,13 @@ SORBETD_RESTORE_TIMEOUT=$(kubectl get zenko ${ZENKO_NAME} -o jsonpath='{.spec.so
 UTILIZATION_SERVICE_HOST=$(kubectl get zenko ${ZENKO_NAME} -o jsonpath='{.spec.scuba.api.ingress.hostname}')
 UTILIZATION_SERVICE_PORT="80"
 
+S3_UTILS_TAG=$(yq eval ".s3utils.tag" ../../../solution/deps.yaml)
+
+LOCATION_CONFIGS=$(kubectl get secrets -l app.kubernetes.io/name=connector-cloudserver-config -o jsonpath='{.items[0].data.locationConfig\.json}' | base64 -di | jq -r .)
+echo "Location Configs: $LOCATION_CONFIGS"
```
francoisferrand: debug, should be removed

.github/scripts/end2end/run-e2e-ctst.sh
```diff
@@ -128,7 +135,15 @@ WORLD_PARAMETERS="$(jq -c <<EOF
   "DRAdminAccessKey":"${DR_ADMIN_ACCESS_KEY_ID}",
   "DRAdminSecretKey":"${DR_ADMIN_SECRET_ACCESS_KEY}",
   "UtilizationServiceHost":"${UTILIZATION_SERVICE_HOST}",
-  "UtilizationServicePort":"${UTILIZATION_SERVICE_PORT}"
+  "UtilizationServicePort":"${UTILIZATION_SERVICE_PORT}",
+  "AwsBackendDestinationReplicationLocation":"${AWS_BACKEND_DESTINATION_LOCATION}",
+  "AwsReplicationBucketName":"${AWS_CRR_BUCKET_NAME}",
+  "AwsReplicationEndpoint":"${AWS_ENDPOINT}",
+  "AwsReplicationAccessKey":"${AWS_ACCESS_KEY}",
+  "AwsReplicationSecretKey":"${AWS_SECRET_KEY}",
```
francoisferrand: does not exist anymore?

.github/workflows/end2end.yaml
```diff
@@ -396,9 +396,9 @@ jobs:
       - name: Install ctst test dependencies
         working-directory: tests/ctst
         run: yarn install
-      - name: Lint ctst tests
-        working-directory: tests/ctst
-        run: yarn lint
+      # - name: Lint ctst tests
+      #   working-directory: tests/ctst
+      #   run: yarn lint
```
francoisferrand: fixed as soon as you rebase

tests/ctst/steps/utils/kubernetes.ts
francoisferrand: indeed, it should be in some secret already: just waiting to be mounted in the crrExistingObject pod

tests/ctst/steps/utils/utils.ts
```diff
@@ -374,6 +392,60 @@ async function addTransitionWorkflow(this: Zenko, location: string, enabled = tr
     }
 }
 
+function getReplicationLocationConfig(world: Zenko, location: string): {
+        storageClass: string;
+        destinationBucket: string;
+        locationType: string;
+    } {
+    switch (location) {
+    case 'awsbackendmismatch':
```
francoisferrand: if we hardcode this "locationName", we may just as well hard code the other parameters...

tests/ctst/steps/utils/utils.ts
```diff
@@ -374,6 +392,60 @@ async function addTransitionWorkflow(this: Zenko, location: string, enabled = tr
...
+        throw new Error(`Unsupported replication location: '${location}'`);
+    }
+}
+
+async function putBucketReplication(
+    this: Zenko,
+    srcBucket: string,
+    replicationLocation: string
+) {
+    const { storageClass, destinationBucket } = getReplicationLocationConfig(this, replicationLocation);
```
francoisferrand: do we need to make things more generic? versus something like:
```suggestion
    const storageClass = world.parameters.AwsBackendDestinationReplicationLocation; 	
    const destinationBucket = world.parameters.AwsReplicationBucketName;
```

Or, if we really want to make things generic : give CTST access to the locationConfig secret (either retrieved and parsed whenever needed ; or mounted in the pod), and retrieve all needed informations by only specifying the name of the location in the tests...

.github/workflows/end2end.yaml
SylvainSenechal: yeah I was trying something different

.github/scripts/end2end/run-e2e-ctst.sh
SylvainSenechal: what do you mean ?

tests/ctst/steps/utils/utils.ts
```diff
@@ -374,6 +392,60 @@ async function addTransitionWorkflow(this: Zenko, location: string, enabled = tr
...
+    this.resetCommand();
+    this.addCommandParameter({ bucket: srcBucket });
+    this.addCommandParameter({
+        replicationConfiguration: JSON.stringify({
+            Role: 'arn:aws:iam::root:role/s3-replication-role',
+            Rules: [
+                {
+                    Prefix: '',
+                    Destination: {
+                        Bucket: `arn:aws:s3:::${destinationBucket}`,
```
francoisferrand: Not sure this is really used in Zenko's multi-backend replication (since the target bucket is actually part of the storage location definition), but [Artesca documentation](https://documentation.scality.com/Artesca/4.0.1/data_management/bucket_operations/replication_workflow/create_a_replication_workflow.html) mentions to set `srcBucket` here:

```suggestion
                        Bucket: `arn:aws:s3:::${srcBucket}`,
```

(It will however be relevant for CRR which allows replicating to "any" bucket for the CRR location/cluster)

tests/ctst/steps/replication.ts
francoisferrand: `imagePullPolicy` is not specified, so we will use the default (`IfNotPresent`), and thus pull from ghcr only if not already present locally

.github/scripts/end2end/run-e2e-ctst.sh
francoisferrand: I don't see the definitions of these variables (AWS_ENDPOINT, AWS_ACCESS_KEY, AWS_SECRET_KEY) in the file

tests/ctst/steps/replication.ts
```diff
@@ -0,0 +1,148 @@
...
+                        ]
+                    }
+                ]
+            }
+        };
+        
+        await createAndRunPod(this, podManifest);
+    });
+
+Then('the object {string} should be replicated within {int} seconds from {string} bucket',
```
francoisferrand: we should be able to store the source bucket (`from {string} bucket` part) and location (`to location {string}`) during the "When" step, so it is available for reuse in followup steps (which avoids more complex & redundant setup, making the test scenario easier to read; and ensure consistency of the steps)

francoisferrand: this approach may also be used to automatically reuse the name of the object in followup steps

tests/ctst/steps/replication.ts
```diff
@@ -0,0 +1,148 @@
...
+    });
+
+Then(
+    'the replicated object {string} from {string} bucket to {string} bucket ' +
+    'with location {string} should be the same as the source object',
+    async function (
+        this: Zenko,
+        objectName: string,
+        bucketSource: string,
+        bucketDestination: string,
```
francoisferrand: bucketDestination is redundant: it is actually part of the location, not a parameter of the test/step

tests/ctst/steps/utils/kubernetes.ts
SylvainSenechal: I changed it, im pulling it from the secret now

tests/ctst/features/crrReplicationS3utils.feature
```diff
@@ -0,0 +1,18 @@
...
+    @2.12.0
+    @PreMerge
+    @ReplicationTest
+    Scenario Outline: Replicate objects created before creating the replication rule
+        Given an existing bucket "<sourceBucket>" "with" versioning, "without" ObjectLock "without" retention mode
+        And an object "<sourceObject>" that "exists"
+        And a replication configuration from "<sourceBucket>" bucket to "<replicationLocation>" location
+        When I run the job to replicate existing objects with status "NEW" from bucket "<sourceBucket>" to location "<replicationLocation>"
+        Then the object "<sourceObject>" should be replicated within 90 seconds from "<sourceBucket>" bucket
+        And the replicated object "<sourceObject>" from "<sourceBucket>" bucket to "ci-zenko-aws-crr-target-bucket" bucket with location "<replicationLocation>" should be the same as the source object
```
francoisferrand: hard to understand, should it not be the other way around?
```suggestion
        And the replicated object "<sourceObject>" from bucket "ci-zenko-aws-crr-target-bucket" in location <replicationLocation>" should be the same as the source object "<sourceObject>" from bucket "<sourceBucket>"
```

.github/scripts/end2end/run-e2e-ctst.sh
SylvainSenechal: They are defined in end2end.yaml

tests/ctst/features/crrReplicationS3utils.feature
SylvainSenechal: ~I find this part weird because it's mixing the source object with the destination bucket :
`the replicated object "<sourceObject>" from bucket "ci-zenko-aws-crr-target-bucket"`
The replicated object is from the sourceBucket, not destinationBucket~

I completely reformatted the text after starting using save/get parameters from the world, to make it clearer

tests/ctst/steps/replication.ts
SylvainSenechal: I agree. I was using it because I need it line 109 for the getObject command, but turns out it's also available in the getReplicationLocationConfig() function that I created so I'll use it instead

tests/ctst/steps/utils/utils.ts
SylvainSenechal: I want to keep the getReplicationLocationConfig() because with the other PR where I add another test for the replication failure case, I will use a different location, so I think this function will make more sense and become more useful with more tests

tests/ctst/steps/utils/utils.ts
SylvainSenechal: yeah I'm gonna use a variable from the locations

tests/ctst/steps/utils/utils.ts
SylvainSenechal: I'm trying to get the locations list from inside the test now though, it will reduce the number of env variable we have to pass to the test

tests/ctst/steps/replication.ts
SylvainSenechal: yeah i did this, the text description of the test should be clearer now

tests/ctst/steps/utils/utils.ts
SylvainSenechal: done

.github/scripts/end2end/run-e2e-ctst.sh
```diff
@@ -78,6 +78,8 @@ SORBETD_RESTORE_TIMEOUT=$(kubectl get zenko ${ZENKO_NAME} -o jsonpath='{.spec.so
 UTILIZATION_SERVICE_HOST=$(kubectl get zenko ${ZENKO_NAME} -o jsonpath='{.spec.scuba.api.ingress.hostname}')
 UTILIZATION_SERVICE_PORT="80"
 
+S3_UTILS_TAG=$(yq eval ".s3utils.tag" ../../../solution/deps.yaml)
```
francoisferrand: retrieving the s3utils image/tag should be done from within the tests as well : it will remove hard-coded image name, and make the test easier to run against "custom" Zenko.

this can be done quite easily by simply retrieving the ZenkoVersion CR (or portion of it), something like:
```
kubectl get zenkoversion -o=jsonpath='{.items[0].spec.versions.s3utils.image}:{.items[0].spec.versions.s3utils.tag}'
```

(note that this would however break if we have multiple versions [i.e. on an "old" artesca instance which has been upgraded a few times]. To be precise we'd need to find out "which" zenkoversion by looking at the `version` field of the Zenko instance.....and so we either need to hope there is a single zenko instance, pass the name of the target zenko instance, or discover zenko by browsing through ingresses looking for the subdomain [overkill])

tests/ctst/steps/utils/kubernetes.ts
francoisferrand: do we need to actually _pull it_ ? Can't we just "mount" that secret into the pod and bind the fields to env variables ?

.github/scripts/end2end/run-e2e-ctst.sh
SylvainSenechal: ok I'm gonna do it

tests/ctst/features/crrReplicationS3utils.feature
```diff
@@ -0,0 +1,14 @@
+Feature: Replication
+    This feature tests replication using the script crrExistingObjects.js from S3utils,
+    that replicates objects created before the replication was enabled on source the bucket.
+
+    @2.12.0
+    @PreMerge
+    @ReplicationTest
+    Scenario Outline: Replicate objects created before creating the replication rule
+        Given an existing bucket "source-bucket" "with" versioning, "without" ObjectLock "without" retention mode
+        And an object "source-object-1" that "exists"
```
francoisferrand, line 10: no code should be needed if the object does not exist?
```suggestion
        And an object "source-object-1"
```

tests/ctst/features/crrReplicationS3utils.feature
```diff
@@ -0,0 +1,14 @@
...
+        And a replication configuration to "awsbackendmismatch" location
+        When I run the job to replicate existing objects with status "NEW"
+        Then the object should be replicated within 90 seconds
```
francoisferrand: is `within 90 seconds` really part of the test?
(or should it be the timeout, in the implementation of this step)

tests/ctst/features/crrReplicationS3utils.feature
```diff
@@ -0,0 +1,14 @@
...
+    This feature tests replication using the script crrExistingObjects.js from S3utils,
+    that replicates objects created before the replication was enabled on source the bucket.
+
+    @2.12.0
+    @PreMerge
+    @ReplicationTest
+    Scenario Outline: Replicate objects created before creating the replication rule
+        Given an existing bucket "source-bucket" "with" versioning, "without" ObjectLock "without" retention mode
+        And an object "source-object-1" that "exists"
+        And a replication configuration to "awsbackendmismatch" location
```
francoisferrand, line 11: here we may insert another check:

```suggestion
        And a replication configuration to "awsbackendmismatch" location
        Then the object should not be replicated
```

francoisferrand: (pre-existing object should not automatically be replicated after setting up replication)

It may be better to do it in separate test though, as this can only be done by waiting a bit : and we may not want to create "long" tests...

tests/ctst/steps/utils/kubernetes.ts
```diff
@@ -25,6 +26,25 @@ type ZenkoStatusValue = {
...
+type ZenkoVersionSpec = {
+    versions: {
+        s3utils: {
+            tag: string;
+        };
+    };
+};
+
+
+
```
francoisferrand:
```suggestion
```

tests/ctst/steps/utils/kubernetes.ts
```diff
@@ -25,6 +26,25 @@ type ZenkoStatusValue = {
...
+    kind: string;
+    metadata: {
+        name: string;
+    };
+    spec: ZenkoVersionSpec;
+};
+
+type ZenkoVersionSpec = {
+    versions: {
+        s3utils: {
```
francoisferrand, line 40: also pick up the "image" to use

```suggestion
        s3utils: {
            image: string;
```

tests/ctst/steps/utils/kubernetes.ts
```diff
@@ -445,3 +479,105 @@ export async function createSecret(
...
+    }
+}
+
+export async function getZenkoVersion(
+    world: Zenko,
+    namespace = 'zenko',
+): Promise<ZenkoVersion> {
+    const customObjectClient = createKubeCustomObjectClient(world);
+    try {
+        console.log('ZenkoVersionList 0', namespace);
```
francoisferrand: remember to remove the logs eventually

tests/ctst/features/crrReplicationS3utils.feature
SylvainSenechal: Created https://scality.atlassian.net/jira/software/c/projects/OS/boards/268?selectedIssue=ZENKO-5057

tests/ctst/features/crrReplicationS3utils.feature
SylvainSenechal: It's a little weird but it's the way the object creation was implemented in common/common.ts line 374.
It's because some tests are parametrized to sometimes not create an object

tests/ctst/steps/utils/kubernetes.ts
SylvainSenechal: Gonna try something like this, if it works I'll keep it : 
```
{
    name: 'MONGODB_AUTH_USERNAME',
    valueFrom: {
        secretKeyRef: {
            name: 'mongodb-db-creds',
            key: 'mongodb-username'
        }
    }
},
```

SylvainSenechal: Yes it works

francoisferrand (approved):

tests/ctst/steps/utils/kubernetes.ts
```diff
@@ -165,6 +184,20 @@ export async function createJobAndWaitForCompletion(
...
+    const clientCore = createKubeCoreClient(world);
+
+    try {
+        const response = await clientCore.createNamespacedPod('default', podManifest);
+        return response.body;
+    } catch (err: unknown) {
+        world.logger.error('Failed to create pod, API server response:', { err });
+        throw new Error(`Failed to create pod: ${err}`);
+    }
+}
```
ghost: Shouldn't we wait for the pod to enter a specific state? could be running, but here, as this is more like a job, could we wait for the Completed state? Otherwise the next step will have to wait "for nothing" that the pod completed? Maybe using createJobAndWaitForCompletion would do the trick

ghost: We should also add a cleanup step in that test to cleanup the created pod, maybe?

tests/ctst/package.json
```diff
@@ -18,7 +18,8 @@
     "prometheus-query": "^3.4.0",
     "proper-lockfile": "^4.1.2",
     "qs": "^6.13.0",
-    "scubaclient": "git+https://github.com/scality/scubaclient#^1.1.2"
+    "scubaclient": "git+https://github.com/scality/scubaclient#^1.1.2",
+    "uuid": "^11.1.0"
```
ghost: From what i can see we add this dependency just to create a unique pod name, is that really needed? With ctst we have some utility function, used for instance to generate object names, that could be used instead?

tests/ctst/steps/utils/utils.ts
```diff
@@ -374,6 +394,72 @@ async function addTransitionWorkflow(this: Zenko, location: string, enabled = tr
...
+    this.resetCommand();
+    this.addCommandParameter({ bucket: srcBucket });
+    this.addCommandParameter({
+        replicationConfiguration: JSON.stringify({
+            Role: 'arn:aws:iam::root:role/s3-replication-role',
+            Rules: [
+                {
+                    Prefix: '',
+                    Destination: {
+                        // Source bucket is not a typo :
```
ghost:
```suggestion
```

tests/ctst/package.json
SylvainSenechal: True, I found Utils.randomString()

tests/ctst/steps/utils/kubernetes.ts
SylvainSenechal: Yes I'm gonna add a wait & cleanup. Previously it was working because the next step is waiting anyways, but it's better to have a wait here.
I don't wanna reuse the function createJobAndWaitForCompletion though, I will reuse some of its logic, but I noticed it has some specific logic (related to cronjob) that makes it not straightforward to use it for this case

tests/ctst/steps/utils/kubernetes.ts
```diff
@@ -165,6 +184,74 @@ export async function createJobAndWaitForCompletion(
...
+                                clearTimeout(timeoutId);
+                                world.logger.error('Pod failed', { 
+                                    podName, 
+                                    status: watchObj.object?.status 
+                                });
+                                reject(new Error(`Pod ${podName} failed`));
+                            }
+                        }
+                    },
+                    (err) => {
```
ghost:
```suggestion
                    err => {
```

tests/ctst/steps/utils/kubernetes.ts
```diff
@@ -165,6 +184,74 @@ export async function createJobAndWaitForCompletion(
...
+                        world.logger.debug("Watch error callback triggered", { podName, err });
```
ghost:
```suggestion
                        world.logger.debug('Watch error callback triggered', { podName, err });
```

ghost (approved):

PR #2246, opened 2025-07-31
Rework ReplicationLagWarning alert

ghost:
The previous version was creating alerts after a fresh install:
- The oplog would not be full in this case
- No operation might cause the lag to be wrongly high, or
- ...the platform is in a transient / startup phase

We would create this alert even if the oplog was not full, because the delta was too small. We here avoid that by ensuring this alert is only created if the oplog circular buffer is fully utilized. This makes more sense because in other cases, it means we have margin.

The strategy here is to use the oplog size as reported by mongodb PRIMARY and detect if the usage is reaching the max. We start the alert at 95% to keep a small margin in case of heavy workloads.

Issue: ZENKO-5047

### DIFF
--- a/monitoring/mongodb/alerts.yaml
+++ b/monitoring/mongodb/alerts.yaml
@@ -103,24 +103,38 @@ groups:
   - alert: ReplicationLagWarning
     expr: |
       label_replace(
-        max by(rs_nm, member_idx) (
-          (
-            max by(pod, rs_nm) (
-              mongodb_rs_members_optimeDate{namespace="${namespace}", pod=~"${service}.*", member_state="PRIMARY"}
+        (
+          max by(rs_nm, member_idx) (
+            (
+              max by(pod, rs_nm) (
+                mongodb_rs_members_optimeDate{namespace="${namespace}", pod=~"${service}.*", member_state="PRIMARY"}
+              )
+              - ignoring(member_idx) group_right min by(pod, rs_nm, member_idx) (
+                mongodb_rs_members_optimeDate{namespace="${namespace}", pod=~"${service}.*", member_state="SECONDARY"}
+              )
+            ) / scalar(
+              max by(rs_nm) (
+                (mongodb_mongod_replset_oplog_head_timestamp{namespace="${namespace}", pod=~"${service}.*"}
+                - on(pod)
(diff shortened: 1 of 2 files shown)


monitoring/mongodb/alerts.yaml
```diff
@@ -103,24 +103,38 @@ groups:
...
+              mongodb_oplog_stats_storageStats_size{namespace="${namespace}", pod=~"${service}.*"}
+              / 
+              mongodb_oplog_stats_storageStats_maxSize{namespace="${namespace}", pod=~"${service}.*"}
             )
-          )
-        ) * 100 > ${replicationLagWarningThreshold},
+            * on(pod) group_left()
+            (mongodb_mongod_replset_my_state{namespace="${namespace}", pod=~"${service}.*"} == 1)
+          ) >= 0.95
+        ),
```
welansari: We don't necessarily have to have a fully utilized oplog buffer to encounter replication issues right ? In that case, this fixes the issue with false positives at init but might hide lag issues happening early after the creation of the platform. The oplog size is configurable (not sure if we have customers with a larger size than the default) which means that it can be fairly large and can take a while before it's full, which will make the alert trigger much later than it's supposed to.

ghost: > We don't necessarily have to have a fully utilized oplog buffer to encounter replication issues right ?

No indeed, but product wise, we do not want to alert if there is no risk. The replication lag can be higher during this period with no issue for customers. When it reaches 95% (not the max), then we can start having problems. If a platform was already lagging before the oplog reaches the max storage, then it will need a resync anyway, it's not a replication lag issue anymore.

The oplog size is indeed configurable although we don't change it, so it's 5% (minimum 990MB) of the available storage.

ghost: A trade-off could be to choose a lower threshold, like 50%, leaving enough room for the startup phases, and some operations. What do you think?

welansari: I think that works yes, LGTM otherwise

welansari: Maybe we could just make the threshold configurable ?

welansari (approved):

monitoring/mongodb/alerts.yaml
ghost: Done (and set as 0.5)

benzekrimaha (approved):

PR #2249, opened 2025-08-04
zenko 5050 : Test replication of objects with replicationStatus=failed 

SylvainSenechal:
ZENKO-5050


tests/ctst/features/crrReplicationS3utils.feature
```diff
@@ -9,6 +9,20 @@ Feature: Replication
         Given an existing bucket "source-bucket" "with" versioning, "without" ObjectLock "without" retention mode
         And an object "source-object-1" that "exists"
         And a replication configuration to "awsbackendmismatch" location
-        When I run the job to replicate existing objects with status "NEW"
-        Then the object should eventually be replicated
+        When we run the job to replicate existing objects with status "NEW"
```
francoisferrand: nit: rephrase to avoid having to identify the subject :)
```
When the job to replicate existing objects with status "NEW" is executed
```

tests/ctst/steps/replication.ts
```diff
@@ -83,12 +88,12 @@ When('I run the job to replicate existing objects with status {string}',
         await createAndRunPod(this, podManifest);
     });
 
-Then('the object should eventually be replicated',
-    async function (this: Zenko) {
+Then('the object should eventually {string} replicated', { timeout: 360_000 },
```
francoisferrand, line 92: nit: alternatively, we could make this a bit more generic by specifying the replication status in the test?

```suggestion
Then('the object replication state should eventually be {string}', { timeout: 360_000 },
```
where string can be COMPLETED or FAILED (or PENDING, but this could create flakoness)

It avoids the "mapping", not sure if it is more useful or it makes the tests more easier to read...

tests/ctst/steps/replication.ts
```diff
@@ -173,3 +190,28 @@ Then(
             'REPLICA'
         );
     });
+
+Given('a deleted destination bucket on that location', async function (this: Zenko) {
```
francoisferrand, line 195: this approach will not "scale" well if multiple tests are using that location : since we will not be able to run multiple tests targeting it in parallel... (and which is why you created a new, dedicated location)

I don't really see an easy alternative option (except introducing "faults" in the system, either via code changes or network tinkering), and probably acceptable at the moment; but I think we should explicitly "document" this limitation and try to prepare for the case where this is reused to avoid introducing flakiness then

* Maybe adding a comment in e2e-config.yaml-template, so that future readers who may be interested can understand how this location is used;
* Creating already a tag for tests using this location, and preventing tests which have this tag from running in parallel;
* Could we add an assertion in the test (maybe in `a replication configuration to "{string}" location` step) to verify that tag is set as expected?

What do you think?

francoisferrand (approved):

tests/ctst/steps/replication.ts
SylvainSenechal: Yes. investigating all 3 remarks

tests/ctst/steps/replication.ts
SylvainSenechal: I will reconsider this one with the next pr which adds another test, because this new test will introduce a third new status variable

tests/ctst/steps/replication.ts
SylvainSenechal: Yeah, I added a tag which enforce non parallel run of tests using this tag. This tag is then read in the "delete bucket" step. If the tag is absent, the test returns an error.

tests/ctst/common/hooks.ts
```diff
@@ -19,13 +20,25 @@ import {
...
 
 setParallelCanAssign(noParallelRun);
 
-Before(async function (this: Zenko) {
+Before(async function (this: Zenko, scenario: ITestCaseHookParameter) {
     this.resetSaved();
     Identity.resetIdentity();
+    // Store scenario tags for access in step definitions
+    const scenarioTags = scenario.pickle.tags?.map(tag => tag.name) || [];
+    this.addToSaved('scenarioTags', scenarioTags);
```
ghost, line 41: Can we add this in CTST directly? Looks like an info we would like to have in other places too.

tests/ctst/features/crrReplicationS3utils.feature
```diff
@@ -9,6 +9,21 @@ Feature: Replication
...
-        When I run the job to replicate existing objects with status "NEW"
-        Then the object should eventually be replicated
+        When the job to replicate existing objects with status "NEW" is executed
+        Then the object should eventually "be" replicated
+        And the replicated object should be the same as the source object
+
+    @2.12.0
+    @PreMerge
+    @ReplicationTest
+    @Lockawsbackendreplicationctstfail
```
ghost, line 19: If we only have one scenario needed a lock, do we actually need a lock at all? Or we are planning to add more?

If we add more, not here, but later, maybe we should consider creating a dynamic list of locations before the tests start (& move the setup logic out of zenko but in ctst, so tests are easier to run if the tool is able to auto configure the cluster), and let CTST handle it in an official way, like real clients. Could be a test, but better to have it as a global "before" that CTST can eventually start supporting if cucumberjs does not allow it.

We did a similar approach for quota where we were pre-creating multiple buckets, populate them and run once the count items jobs instead of many times, affecting all other tests stability and actual metrics.

ghost (approved):

tests/ctst/common/hooks.ts
SylvainSenechal: I'm not too sure I understand, because with this, the tags will already be accessible in all functions (which have a Zenko parameters) 🤔

tests/ctst/features/crrReplicationS3utils.feature
SylvainSenechal: yeah tbh I think the locking is a bit overkill, I have already documented in template.yaml that it should avoid being used for something else. Your other solutions also doesn't look very straightforward, so I think either we remove the lock or keep this one for now

tests/ctst/common/hooks.ts
ghost: Yes, but it could be added as a global hook in CTST directly, so tags are always available in that saved variable, for all repositories. It looks like a good candidate because this is unrelated with Zenko.

PR #2253, opened 2025-08-11
release: automatically pickup artifacts

francoisferrand:
- release: automatically discover artifact
- gha: fix check-workflow artifacts archival

Issue: ZENKO-5046

### DIFF
--- a/.github/scripts/get-build-artifact.js
+++ b/.github/scripts/get-build-artifact.js
@@ -1,0 +1,52 @@
+/**
+ * Auto-derive artifacts name from successful staging build
+ * 
+ * This script is used in GitHub Actions to automatically find and construct
+ * the artifacts name from a successful staging workflow run that contains
+ * the build-iso-and-end2end-test job.
+ * 
+ * @param {Object} github - GitHub Actions toolkit github object
+ * @param {Object} context - GitHub Actions context object
+ * @param {Object} core - GitHub Actions core object for logging/errors
+ * @returns {Promise<string>} The constructed artifacts name
+ */
+async function getBuildArtifact(github, context, core) {
+  const workflow_id = 'build-iso-and-end2end-test'; // The workflow ID for the build job
+
+  // Get the commit SHA for the tag
+  const tagCommit = context.sha;
+  core.info(`Looking for successful builds for commit: ${tagCommit}`);
+  
+  // Get workflow runs for this repository``
+  const { data: workflowRuns } = await github.rest.actions.listWorkflowRuns({
+    owner: context.repo.owner,
+    repo: context.repo.repo,
+    workflow_id,
+    head_sha: tagCommit,
+    status: 'completed',
+    conclusion: 'success'
+  });
+  
+  // Find the first successful staging workflow run
+  const run = workflowRuns.workflow_runs[0];
+  if (!run) {
+    throw new Error(`No successful end2end workflow run found for commit ${tagCommit}`);
+  }
+  
+  core.info(`Found staging run: ${run.id} with conclusion: ${run.conclusion}`);
+  
+  // Construct artifacts name, like scality/action-artifacts
+  const commitHash = tagCommit.substring(0, 10); // use first 10 chars of commit hash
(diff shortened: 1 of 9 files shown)


.github/scripts/get-build-artifact.js
```diff
@@ -0,0 +1,52 @@
...
+  // Find the first successful staging workflow run
+  const run = workflowRuns.workflow_runs[0];
+  if (!run) {
+    throw new Error(`No successful end2end workflow run found for commit ${tagCommit}`);
+  }
+  
+  core.info(`Found staging run: ${run.id} with conclusion: ${run.conclusion}`);
+  
+  // Construct artifacts name
+  const commitHash = tagCommit.substring(0, 7); // Use first 7 chars of commit hash
```
welansari: Shouldn't this be 10 characters ? Previous builds have a 10 character hash.

Examples:
- `github:scality:Zenko:staging-2e37bb5673.build-iso-and-end2end-test.7286`
- `github:scality:Zenko:staging-18ddd8823d.build-iso-and-end2end-test.7297`
- `github:scality:Zenko:staging-24eb369cde.build-iso-and-end2end-test.7295`
- `github:scality:Zenko:staging-6994ac4ee7.build-iso-and-end2end-test.7292`

francoisferrand: This is based on the check which is implemented in the ˋartifacts` action, when promoting an artifact: so it works for a check, but indeed 10 chars are actually used (in the action) when creating the artifact name...

.github/scripts/get-build-artifact.js
```diff
@@ -0,0 +1,52 @@
...
+  const buildNumber = run.run_number;
+  const artifactsName = `github:${context.repo.owner}:${context.repo.repo}:staging-${commitHash}.${workflow_id}.${buildNumber}`;
```
benzekrimaha, line 41: shouldn't we use here the `tagCommit` instead of the `commitHash` ? ( to have the 10 chars instead of 7 ?

francoisferrand: `tagCommit` has the full commit hash, so much larger

.github/workflows/release.yaml
```diff
@@ -9,13 +9,15 @@ on:
         description: Tag
         required: true
       artifacts-name:
-        description: Artifacts name to promote
-        required: true
+        description: Artifacts name to promote (optional - will be auto-detected if not provided)
```
ghost, line 12: is there any risk of not passing it, or any reason to keep this option if we always want to release on top of a branch (which is the behavior of the script)?

.github/scripts/get-build-artifact.js
```diff
@@ -0,0 +1,52 @@
...
+ * the artifacts name from a successful staging workflow run that contains
+ * the build-iso-and-end2end-test job.
+ * 
+ * @param {Object} github - GitHub Actions toolkit github object
+ * @param {Object} context - GitHub Actions context object
+ * @param {Object} core - GitHub Actions core object for logging/errors
+ * @returns {Promise<string>} The constructed artifacts name
+ */
+async function getBuildArtifact(github, context, core) {
+  const workflow_id = 'build-iso-and-end2end-test'; // The workflow ID for the build job
```
ghost, line 14: Could that be auto-retrieved from the end2end.yaml workflow?

ghost (approved):

.github/workflows/release.yaml
francoisferrand: eventually, I think not ; but I'd rather keep the option for now, to avoid finding a bug when making an (urgent, obviously...) release

.github/scripts/get-build-artifact.js
francoisferrand: we could parse the workflow, but we would need anyway to _hard-code_ the name of that workflow file (`end2end.yaml`) : so benefit seems minimal, especially vs the extra amount of code...

benzekrimaha (approved):

PR #2258, opened 2025-08-21
zenko 5057 : Replication should not happen if object created before the replication config

SylvainSenechal:
Issue: ZENKO-5057


tests/ctst/steps/replication.ts
```diff
@@ -112,12 +115,22 @@ Then('the object should eventually {string} replicated', { timeout: 360_000 },
             assert(parsed.ok);
             const replicationStatus = parsed.result?.ReplicationStatus;
             
-            if (replicate === 'be') {
+            if ((expectedOutcome === 'succeed' || expectedOutcome === 'fail') && 
+                (replicationStatus === 'PENDING' || 
+                    replicationStatus === 'PROCESSING' ||
+                    replicationStatus === undefined // If replication hasn't started, status is still undefined
```
francoisferrand: > // If replication hasn't started, status is still undefined

this is not correct: if replicationStatus is undefined, then replication will not be started at all.
the way replication works is that `cloudserver` sets this field to PENDING when modifying the object (and replication is configured); then backbeat listens to the epilog and can filter objects which may require replication.

→ undefined does not mean replication is pending

tests/ctst/steps/replication.ts
```diff
@@ -126,12 +139,20 @@ Then('the object should eventually {string} replicated', { timeout: 360_000 },
...
+                    'expected object to have no ReplicationStatus (should remain undefined)'
+                );
+                break;
             }
-            if (replicationStatus === 'PENDING' || replicationStatus === 'PROCESSING') {
-                continue;
-            }
         }
-        assert.fail(`Timeout: Object '${objectName}' is still pending/processing after timeout`);
+        if (expectedOutcome === 'never happen') {
```
francoisferrand: 5min is a long time to checking it never happened...
And worth double checking if expecting "undefined" status may be enough?

tests/ctst/steps/replication.ts
```diff
@@ -112,12 +115,22 @@ Then('the object should eventually {string} replicated', { timeout: 360_000 },
             assert(parsed.ok);
             const replicationStatus = parsed.result?.ReplicationStatus;
             
-            if (replicate === 'be') {
+            if ((expectedOutcome === 'succeed' || expectedOutcome === 'fail') && 
+                (replicationStatus === 'PENDING' || 
+                    replicationStatus === 'PROCESSING' ||
```
francoisferrand: this looks kind of weird next the switch on status: for better consistency & readability, either
* we should move the condition in pending in the `expectedOutcome` switch;
* or we should do a switch on the status, which should have less cases I think

```suggestion
switch(replicationStatus) {
case 'PENDING', 'PROCESSING':
   assert(expectedOutcome != 'never happen');
   break;

case 'COMPLETED':
   assert(expectedOutcome === 'succeed');
   return;

case 'FAILED':
   assert(expectedOutcome === 'fail');
   return;

case undefined:
   assert(expectedOutcome == 'never happen');
   break; // or return if we don't need to wait the timeout
}
```

tests/ctst/steps/replication.ts
```diff
@@ -89,8 +89,11 @@ When('the job to replicate existing objects with status {string} is executed',
         await createAndRunPod(this, podManifest);
     });
 
-Then('the object should eventually {string} replicated', { timeout: 360_000 },
-    async function (this: Zenko, replicate: 'be' | 'fail to be') {
+Then('the object replication should eventually {string}', { timeout: 360_000 },
```
francoisferrand: nit: removing the `eventually` makes it easier to read, esp. in the "never happen" case
```suggestion
Then('the object replication should {string}', { timeout: 360_000 },
```

tests/ctst/steps/replication.ts
SylvainSenechal: I switched to configurable timeouts because failure takes longer than success

tests/ctst/steps/replication.ts
SylvainSenechal: I've had a situation where at the very start of the replication status check, the object replication status hasn't had the time to be set to PENDING, so it's still undefined

tests/ctst/steps/replication.ts
SylvainSenechal: Yes, I changed the switch to something similar to this

tests/ctst/features/crrReplicationS3utils.feature
```diff
@@ -2,15 +2,24 @@ Feature: Replication
...
     @PreMerge
     @ReplicationTest
     Scenario Outline: Replicate objects created before creating the replication rule
-        Given an existing bucket "source-bucket" "with" versioning, "without" ObjectLock "without" retention mode
+        Given an existing bucket "source-bucket-1" "with" versioning, "without" ObjectLock "without" retention mode
         And an object "source-object-1" that "exists"
         And a replication configuration to "awsbackendmismatch" location
         When the job to replicate existing objects with status "NEW" is executed
-        Then the object should eventually "be" replicated
+        Then the object replication should "succeed" within 300 seconds
```
ghost, line 22: For consistency if we expect a replication to succeed up to 300s, we should do the same for the "never happen" case, no?
But as pointed out this would artificially make tests very long. If we thus think 30s is enough, we should probably use it here as well.

Another approach for such "test to see if this never happens" would be to look more in depth at the system to see if the old object was scanned after enabling the replication configuration. This allows to have a condition to stop on. But this is probably not realistic at the end-to-end level. Do we think the scenario `Objects created before setting up replication should not be replicated automatically` can only be tested at the Zenko level, and not Backbeat as a functional test?

SylvainSenechal: I think if we wanted to be perfect we should use the same timeout everywhere. In practice I've found the failure case to take some more time to be marked as a failure (versus the success case), and the 'never happen' case doesn't really require a lot of waiting. I think the timings are good this way.

About the other point, I think technically by looking inside the backbeat codebase, we should be able to assert that replication doesn't happen to objects created before the replication configuration is set. But I guess it's quite hard to determine that this does not happen, because it doesnt happen so there is nothing see. But the test is still nice to enforce that the behavior remains the same over time, as any modification to any service could break this later

tests/ctst/steps/replication.ts
francoisferrand: that is weird : the status is set to PENDING in cloudserver, directly when writing the metadata of the new object/version c.f. [1]

[1] https://github.com/scality/cloudserver/blob/d0ae11ef265a765de2fb6f775daf66d0e971de11/lib/api/apiUtils/object/createAndStoreObject.js#L139

If this is not the case, it means we are starting this check without waiting for the "return" of the s3:putObject - or maybe multiple tests are hitting the same object?

tests/ctst/features/crrReplicationS3utils.feature
francoisferrand: technically the "decision" to replicate is made by cloudserver, and reflected by the status being set to PENDING only when replication is needed : so just checking this is be enough, with the current design.

tests/ctst/steps/replication.ts
SylvainSenechal: You can see on this run https://github.com/scality/Zenko/actions/runs/17275504190/job/49031281060
If you grep "replicationStatus", the first few ones are "undefined", then "pending", and then "completed"

The crrExistingObject script is setting the PENDING status of the object with a PutMetadata method : 
PUT (/_/backbeat/metadata/{Bucket}/{Key+})
Maybe there is some delay, or something asynchronous in the way the meta data is updated.

SylvainSenechal: Hum, actually 👀 , the logic that I added to wait for the pod to complete was added after I encountered this issue, and the run I just linked above was not waiting for the pod to complete ([waitForCompletion = false here](https://github.com/scality/Zenko/commit/95a9fbf35a5b6a311adef50bbd7725aa7c31e80f#diff-0f654639a87d87f1f058feb4b7e271e7668a7644c47f824931ef00f0760f038d)), so that's the reason why it was undefined. So I'll change the logic, as you said, the replicationStatus should directly be PENDING

tests/ctst/steps/replication.ts
```diff
@@ -89,13 +89,16 @@ When('the job to replicate existing objects with status {string} is executed',
         await createAndRunPod(this, podManifest);
     });
 
-Then('the object should eventually {string} replicated', { timeout: 360_000 },
-    async function (this: Zenko, replicate: 'be' | 'fail to be') {
+Then('the object replication should {string} within {int} seconds', { timeout: 600_000 },
```
francoisferrand, line 92: do we actually need to increase timeout from 6 to 10 minutes?
did you experience cases of timeout?

francoisferrand (approved):

tests/ctst/steps/replication.ts
SylvainSenechal: I found sometimes I couldn't make it work within 200/300 seconds so I increased it, hopefully no problems with larger timeout

ghost (approved):

PR #2259, opened 2025-08-22
Fix codespace

francoisferrand:
- Move ZKOP deployment to dedicated script
- codespace: handle `GITHUB_ENV`
- codespace: eval whole block in a single call
- codespace: need persmissions for metadata & mdop
- codespace: pass variables from `env` field
- codespace: json lint
- codespace: fix kind feature
- codespace: fix setup.sh shbang
- Fix warnings in ctst image

Issue: ZENKO-5060

### DIFF
--- a/.github/actions/deploy/action.yaml
+++ b/.github/actions/deploy/action.yaml
@@ -55,20 +55,7 @@ runs:
       working-directory: ./.github/scripts/end2end
     - name: Install operator
       shell: bash
-      run: |-
-        set -exu;
-        export OPERATOR_IMAGE_NAME=$(yq eval '."zenko-operator" | .sourceRegistry + "/" + .image' solution/deps.yaml)
-        if ! [ -n "${OPERATOR_IMAGE_TAG}" ]; then
-          export OPERATOR_IMAGE_TAG=$(yq eval '."zenko-operator".tag' solution/deps.yaml)
-        fi
-        docker pull ${OPERATOR_IMAGE_NAME}:${OPERATOR_IMAGE_TAG}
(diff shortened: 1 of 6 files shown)


.github/actions/deploy/action.yaml
```diff
@@ -55,20 +55,7 @@ runs:
...
-        fi
-        docker pull ${OPERATOR_IMAGE_NAME}:${OPERATOR_IMAGE_TAG}
-        kind load docker-image ${OPERATOR_IMAGE_NAME}:${OPERATOR_IMAGE_TAG}
-        cd ./.github/scripts/end2end
-        git init operator
-        cd operator
-        git fetch --depth 1 --no-tags https://git:${GIT_ACCESS_TOKEN}@github.com/scality/zenko-operator.git ${OPERATOR_IMAGE_TAG}
-        git checkout FETCH_HEAD
-        tilt ci
+      run: ./.github/scripts/end2end/deploy-zkop.sh
```
welansari, line 58:
```suggestion
      run: bash deploy-zkop.sh
      working-directory: ./.github/scripts/end2end
```
Consistent with the other steps and cleaner IMO ?

welansari (approved):

.github/actions/deploy/action.yaml
francoisferrand: * `bash` should not be specified: if needed, it must be set in the script's shbang. I did not fix the other instances though, to avoid overloading this PR
* this specific script must be run from root, as it accesses various other resources (deps.yaml...)

benzekrimaha (approved):

PR #2261, opened 2025-08-26
Pull mongo images from bitnamilegacy

francoisferrand:
- Pull mongo images from bitnamilegacy
- Remove unused mongodb_options.yaml

Issue: ZENKO-5062


DarkIsDude (approved):

DarkIsDude:
Should we create an other JIRA task to finish the migration ? `All existing container images, including older or versioned tags (e.g., 2.50.0, 10.6), will be moved from the public catalog (docker.io/bitnami) to the [Bitnami Legacy repository](https://hub.docker.com/u/bitnamilegacy) (docker.io/bitnamilegacy). This legacy catalog will receive no further updates or support and should only be used for temporary migration purposes.`

welansari (approved):

francoisferrand:
> Should we create an other JIRA task to finish the migration ? `All existing container images, including older or versioned tags (e.g., 2.50.0, 10.6), will be moved from the public catalog (docker.io/bitnami) to the [Bitnami Legacy repository](https://hub.docker.com/u/bitnamilegacy) (docker.io/bitnamilegacy). This legacy catalog will receive no further updates or support and should only be used for temporary migration purposes.`

Did I miss some package, or do you mean older zenko branches?
I don't think we will make any more release on branch earlier than 2.12, so I voluntarily avoided the extra builds and flakiness...

DarkIsDude:
> > Should we create an other JIRA task to finish the migration ? `All existing container images, including older or versioned tags (e.g., 2.50.0, 10.6), will be moved from the public catalog (docker.io/bitnami) to the [Bitnami Legacy repository](https://hub.docker.com/u/bitnamilegacy) (docker.io/bitnamilegacy). This legacy catalog will receive no further updates or support and should only be used for temporary migration purposes.`
> 
> Did I miss some package, or do you mean older zenko branches? I don't think we will make any more release on branch earlier than 2.12, so I voluntarily avoided the extra builds and flakiness...

No my point is that we should get rid of this legacy catalog in a future and don't use them at all regarding the doc. How we'll make sure that we'll address that in the future ?

PR #2274, opened 2025-09-19
increase timeout for replication success check

SylvainSenechal:
Issue: ZENKO-5073


francoisferrand (approved):

welansari (approved):

SylvainSenechal:
Closing this pr : These tests are never flaky in Codespace, ticket closed as won't do
This doesn't mean the test isn't flaky, but we suspect other issue around cluster stability, and backbeat pods going down at the wrong time due to other tests. Increasing timeout isn't the solution

PR #2279, opened 2025-10-23
Fix MongoDB dashboard

ghost:
This is a set of fixes to ensure the MongoDB dashboard is properly handling multi shards. From this:

image

to this:

image

Note for reviewers: the json file is not to be reviewed, the python one is used with `scalgrafanalib` to generate the dashboard.

### DIFF
--- a/.github/scripts/end2end/fix-zookeeper.sh
+++ b/.github/scripts/end2end/fix-zookeeper.sh
@@ -1,0 +1,97 @@
+# This file is a temporary fix. Pravega zookeeper is not
+# maintained anymore and is as such not compatible with
+# recent Linux kernel versions. Disabling container support
+# is not acceptable for production, and only done here
+# for the CI.
+set -eu
+
+ZENKO_NAME="${1:?Missing ZENKO_NAME argument}"
+NAMESPACE="${2:?Missing NAMESPACE argument}"
+
+ZK_STS_NAME="${ZENKO_NAME}-base-quorum"
+ZK_CONTAINER_NAME="zookeeper"
+ZK_POD_NAME="${ZK_STS_NAME}-0"
+
+OPERATOR_WAIT_TIMEOUT=120
+STATEFULSET_WAIT_TIMEOUT=180
+
+# Helper function to get elapsed time
+get_elapsed() {
+    local start_time=$1
+    echo $(($(date +%s) - start_time))
+}
+
+# Wait for the Zenko operator to process the CR
+OPERATOR_WAIT_START=$(date +%s)
+
+while true; do
+    ELAPSED=$(get_elapsed ${OPERATOR_WAIT_START})
+    
+    if [ ${ELAPSED} -ge ${OPERATOR_WAIT_TIMEOUT} ]; then
+        echo "ERROR: Timed out after ${OPERATOR_WAIT_TIMEOUT}s waiting for Zenko operator to process CR."
+        kubectl -n "${NAMESPACE}" get zenko/"${ZENKO_NAME}" -o yaml 2>&1 || echo "Could not get Zenko CR."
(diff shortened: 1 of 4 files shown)


.github/scripts/end2end/fix-zookeeper.sh
```diff
@@ -0,0 +1,93 @@
+# This file is a temporary fix.
```
DarkIsDude: Maybe you can describe why is it and how to fix it ? And when ?

DarkIsDude (approved):

.github/scripts/end2end/fix-zookeeper.sh
ghost: No idea yet, we are using a zookeeper image [not maintained since 2y](https://github.com/pravega/zookeeper-operator/issues/628), we need to discuss what to use instead, as the new kernel versions are incompatible...

ghost: I completed the comment

monitoring/mongodb/dashboard.py
```diff
@@ -518,7 +518,7 @@ def mongodb_state_timeline(title, expr, description="", mappings=None, **kwargs)
     orientation="horizontal",
     targets=[
         Target(
-            expr="sum(" + Metrics.UP() + ") by (pod)",
+            expr="sum(" + Metrics.UP() + ") by (job)",
```
francoisferrand: should not be changed : the legend is `{{pod}}`, so the sum must be by `pod` as well (since the result series will not that label)

i.e. either:
```suggestion
            expr="sum(" + Metrics.UP() + ") by (pod)",
            templateLabel = "{{pod}}"
```
or
```suggestion
            expr="sum(" + Metrics.UP() + ") by (job)",
            templateLabel = "{{job}}"
```

monitoring/mongodb/dashboard.py
```diff
@@ -665,8 +667,8 @@ def mongodb_state_timeline(title, expr, description="", mappings=None, **kwargs)
     unit="short",
     targets=[
         Target(
-            expr="sum by (job) (" + Metrics.DBSTATS_OBJECTS() + ")",
-            legendFormat="{{ job }}",
+            expr="sum by (pod) (" + Metrics.DBSTATS_OBJECTS() + ")",
```
francoisferrand, line 679: nit: since you now use the `pod` label, for consistency it should be added to the metrics definitions (in `Metrics`). Not a real issue at the moment (since the expression is not "validated"), but would help to document how we can use the metrics...

monitoring/mongodb/dashboard.py
```diff
@@ -665,8 +667,8 @@ def mongodb_state_timeline(title, expr, description="", mappings=None, **kwargs)
...
+            legendFormat="{{ pod }}",
```
francoisferrand, line 680: Why do we now operate on `pod` instead of `job` ? Does this not create some inconsistency with the "filters" on top of the dashboard?

(don't we have one `job` per `pod` ?)

francoisferrand:
Also, in the "distribution" charts, the value displayed on chart should probably be modified to include only the percentage (and not the job/pod label)

francoisferrand (approved):

monitoring/mongodb/dashboard.py
ghost: I actually prefer using `cluster_role`: so we merge. Job is fine but it creates many entries for multiple shards, not readable at all... I'll update the legend as well
image

monitoring/mongodb/dashboard.py
ghost: > Why do we now operate on pod instead of job ?

Because both shards actually share the same `job`, as we chose to keep a single pod monitor for all shards (it's name should probably be changed...)

PR #2287, opened 2025-11-18
Fix codespace

francoisferrand:
- Fix codespace
- codespace: use default input values
- Do not recreate kind cluster if aready present
- codespace: disable gcp tests
- codespace: add minimum doc on running tests

Issue: ZENKO-5096


.github/scripts/end2end/bootstrap-kind.sh
```diff
@@ -78,6 +78,11 @@ EOF
 }
 
 create_cluster() {
+  if kind get clusters | grep -q "^${CLUSTER_NAME}$"; then
+    echo "Kind cluster ${CLUSTER_NAME} already exists. Skipping creation."
```
SylvainSenechal, line 82: in what situation do we end up calling create_cluster when it already exists ?

SylvainSenechal (approved):
Later I wanna discuss, being able to setup a codespace and run tests inside is nice, but hopefully there is a way to make code modifications in that codespace, otherwise the benefits arent that great since when a test fails and we want to make changes to fix it, we would have to recreate a codespace which should take about the same amont of time as running the pipeline 🤔 

Although if the problem is not coming from the components but from the zenko test itself I guess editing the test is possible which is already a nice feature

.github/scripts/end2end/bootstrap-kind.sh
francoisferrand: When setup.sh fails for some reason, and you want to re-run it

Does not (and should not!) happen in CI, but needed when a human is in command, for "local" (or codespace) development

francoisferrand:
> Later I wanna discuss, being able to setup a codespace and run tests inside is nice, but hopefully there is a way to make code modifications in that codespace, otherwise the benefits arent that great since when a test fails and we want to make changes to fix it, we would have to recreate a codespace which should take about the same amont of time as running the pipeline 🤔

Yes, that is the whole point of codespace: it is basically a VM "ready to use". You can (in theory) do your whole dev in codespace, edit / run tests / debug / commit and push!

> Although if the problem is not coming from the components but from the zenko test itself I guess editing the test is possible which is already a nice feature  

You can edit the tests (though I tested running the tests only through docker, which may add extra step) and thus ultimately edit/run/fix them.

But you can also more easily inspect the state of the system before/after the test or "tail" logs if that helps.

And you can -just like on any lab- change the image to use in zenko, directly in cluster to avoid full setup.

(And then we could probably do more advanced stuff: checkout another repo in a subdirectory, to be able to edit it directory ; and from there possibly "remote" debug a container running in k8s)

----
To summarize, codespace is "like" a local environment, with slight twist / different compromise:
- it is a controlled env vs depending on packages and setup on your Pc: less "customized" but more reproducible
- it is a remote VM automatically started/stopped: you can work from anywhere, but need network connectivity and wait for restarts/...
- due to the approach, it needs to handle security more strictly: can't just put your ssh key. GitHub has made it quite seamless, but still there are some impacts (declaring the repos we depend on in particular)
- also it runs "in vscode": though you still can access shell, it may not fit everybody's preferences

Note: a large part of codespace is actually just the "devcontainer", which in theory could be run locally (even on mac/arm, if we can specify an x86 container). I quickly tried once, long time ago, and it did not work; but that could be another compromise.

In my experience today the main issues are:
- it often breaks - partly because we don't prebuilt the image and don't use it often
- codespace creation (and Zenko deployment) is very slow - possibly due to use of `kind`
- it allows deploying zenko, but that is about where we stopped efforts : more work is needed to make it trivial to run/debug tests

.devcontainer/README.md
```diff
@@ -10,6 +10,20 @@ To run the CTST tests in the codespace, simply head to `.github/script/end2end/`
...
 
+## Running e2e tests in the codespace
+
+To run the CTST tests in the codespace, head to `.github/script/end2end/` and run `run-e2e-ctst.sh`
+script. Some variables need to be exported before running the tests.
+
+```bash
+    env_variables=$(yq eval '.env | to_entries | .[] | .key + "=" + .value' .github/workflows/end2end.yaml | sed 's/\${{[^}]*}}//g') && export $env_variables
+    export E2E_IMAGE_TAG=latest
+    export GCP_BACKEND_DESTINATION_LOCATION=
```
tcarmet, line 21: You can set those two vars directly in the `devcontainer.json` if you'd like:

```json
// .devcontainer/devcontainer.json
...
  "containerEnv": {
    "E2E_IMAGE_TAG": "latest",
    "GCP_BACKEND_DESTINATION_LOCATION": ""
  },
...
```

tcarmet (approved):

.devcontainer/README.md
francoisferrand: Not sure, they need to be set after "loading" the other ones from the workflow (yes, we need to clean this up, and do it in a way that avoids duplication, weird intrication and dependency on gha...)

PR #2290, opened 2025-11-19
Fix mongo dashboard for multiple shards

francoisferrand:
- Fix MongoDB service state overview
- Show sharding distribution by replicaset name (`rs_nm`) instead of pod

image

("mongodb balancing process" is empty because this panel works only with mongodb 8, and I tested in mongodb 7 ; but nothing changed there)

Issue: ZENKO-5098

### DIFF
--- a/monitoring/mongodb/dashboard.json
+++ b/monitoring/mongodb/dashboard.json
@@ -1334,14 +1334,14 @@
       "maxDataPoints": 100,
       "targets": [
         {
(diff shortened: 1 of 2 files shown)


SylvainSenechal (approved):
its not clear what rs_nm is
Copilot said "replicaset_name" but wouldnt have guessed by myself

francoisferrand:
> its not clear what rs_nm is Copilot said "replicaset_name" but wouldnt have guessed by myself

not our choice, this comes from the metrics in mongo.
But indeed, this is the name of the (mongo) "replicaset" (not to be confused with k8s replicaset), i.e. _group of mongod instances that maintain the same data set_ : practically, this means there is one RS for each shard and one for configsvr...

DarkIsDude (approved):

benzekrimaha (approved):

PR #2305, opened 2026-01-09
Dedup zenkoversion.yaml file

francoisferrand:
Use the same (official!) zenkoversion.yaml file in CI and in solution,
to ensure we don't introduce discrepency and run the tests in a

Issue: ZENKO-4817


solution/zenkoversion.yaml
```diff
@@ -3,129 +3,129 @@
...
+      image: ${BACKBEAT_IMAGE}
+      tag: ${BACKBEAT_TAG}
     drctl:
-      image: zenko-drctl
+      image: ${DRCTL_IMAGE}
       tag: ${DRCTL_TAG}
-    utapi:
-      image: utapi
-      tag: '${UTAPI_TAG}'
+    # utapi:
```
SylvainSenechal: why did you comment utapi and mongodb ?

SylvainSenechal (approved):

solution/zenkoversion.yaml
francoisferrand: both are actually not used, and lead to an invalid YAML if we remove the quotes which are otherwise useless...

benzekrimaha: as I read they are not used, can't we just remove them ?

benzekrimaha (approved):

PR #2316, opened 2026-02-10
Fix putObject warning

benzekrimaha:
Issue: ZENKO-5128

### DIFF
--- a/tests/ctst/steps/utils/utils.ts
+++ b/tests/ctst/steps/utils/utils.ts
@@ -1,3 +1,4 @@
+import { Buffer } from 'buffer';
 import { promises as fsp } from 'fs';
 import { join } from 'path';
 import {
@@ -47,6 +48,11 @@ export async function uploadSetup(world: Zenko, action: string, body?: string) {
         const objectBody = body || 'a'.repeat(objectSize);
         await saveAsFile(tempFileName, objectBody);
         world.addCommandParameter({ body: world.getSaved<string>('tempFileName') });
+        const contentLength = body ? Buffer.byteLength(objectBody) : objectSize;
+        world.addCommandParameter({ contentLength: `${contentLength}` });
+    } else if (action === 'PutObject') {
+        world.addCommandParameter({ body: '' });
+        world.addCommandParameter({ contentLength: '0' });
     }
 }
 
@@ -58,6 +64,10 @@ export async function uploadTeardown(world: Zenko, action: string) {
     if (objectSize > 0) {
         await deleteFile(world.getSaved<string>('tempFileName'));
         world.deleteKeyFromCommand('body');
+        world.deleteKeyFromCommand('contentLength');
+    } else if (action === 'PutObject') {
+        world.deleteKeyFromCommand('body');
+        world.deleteKeyFromCommand('contentLength');
     }
 }
 
--- a/tests/zenko_tests/node_tests/cloudserver/keyFormatVersion/tests/versionedBucket.js
+++ b/tests/zenko_tests/node_tests/cloudserver/keyFormatVersion/tests/versionedBucket.js
@@ -38,9 +38,12 @@ const ownerInfo = {};
 async function putObjectVersions(objName, vFormat) {
     for (let i = 0; i < 3; i++) {
         // eslint-disable-next-line no-await-in-loop
-        const result = await s3.send(new PutObjectCommand(
-            { Bucket: BUCKET_NAME[vFormat], Key: objName },
-        ));
+        const result = await s3.send(new PutObjectCommand({
+            Bucket: BUCKET_NAME[vFormat],
+            Key: objName,
+            Body: '',
+            ContentLength: 0,
+        }));
         versionIds[vFormat][objName].push(result.VersionId);
     }
 }
@@ -204,7 +207,12 @@ describe('Cloudserver : keyFormatVersion : versioned bucket', () => {
         });
 
         it(`Should delete specified version ${vFormat}`, async () => {
-            const res = await s3.send(new PutObjectCommand({ Bucket: BUCKET_NAME[vFormat], Key: 'first-key' }));
+            const res = await s3.send(new PutObjectCommand({
+                Bucket: BUCKET_NAME[vFormat],
+                Key: 'first-key',
+                Body: '',
+                ContentLength: 0,
+            }));
             const tmpVersionId = res.VersionId;
 
             await s3.send(new DeleteObjectCommand({
--- a/tests/zenko_tests/node_tests/s3SDK.js
+++ b/tests/zenko_tests/node_tests/s3SDK.js
@@ -7,7 +7,11 @@ const sharedHttpHandler = new NodeHttpHandler({
     connectionTimeout: 0,
 });
 
-const scalityS3Client = new S3Client({
+function createS3Client(config) {
+    return new S3Client(config);
+}
+
+const scalityS3Client = createS3Client({
     credentials: {
         accessKeyId: process.env.ZENKO_ACCESS_KEY,
         secretAccessKey: process.env.ZENKO_SECRET_KEY,
@@ -47,7 +51,7 @@ const scalityIAMClient = new IAMClient({
 const verifyCerts = process.env.VERIFY_CERTIFICATES
     ? process.env.VERIFY_CERTIFICATES : true;
 
-const awsS3Client = new S3Client({
+const awsS3Client = createS3Client({
     credentials: {
         accessKeyId: process.env.AWS_ACCESS_KEY,
         secretAccessKey: process.env.AWS_SECRET_KEY,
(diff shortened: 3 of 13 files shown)


tests/ctst/steps/utils/utils.ts
```diff
@@ -47,6 +48,11 @@ export async function uploadSetup(world: Zenko, action: string, body?: string) {
         const objectBody = body || 'a'.repeat(objectSize);
         await saveAsFile(tempFileName, objectBody);
         world.addCommandParameter({ body: world.getSaved<string>('tempFileName') });
+        const contentLength = body ? Buffer.byteLength(objectBody) : objectSize;
+        world.addCommandParameter({ contentLength: `${contentLength}` });
+    } else if (action === 'PutObject') {
```
francoisferrand, line 53: why not do this on `UploadPart` as well?


```suggestion
    } else {
```

tests/ctst/steps/utils/utils.ts
```diff
@@ -58,6 +64,10 @@ export async function uploadTeardown(world: Zenko, action: string) {
     if (objectSize > 0) {
         await deleteFile(world.getSaved<string>('tempFileName'));
         world.deleteKeyFromCommand('body');
+        world.deleteKeyFromCommand('contentLength');
+    } else if (action === 'PutObject') {
```
francoisferrand, line 68: same


```suggestion
    } else {
```

tests/zenko_tests/node_tests/s3SDK.js
```diff
@@ -7,7 +7,42 @@ const sharedHttpHandler = new NodeHttpHandler({
     connectionTimeout: 0,
 });
 
-const scalityS3Client = new S3Client({
+function attachContentLengthMiddleware(client) {
```
francoisferrand: not clear why we need this middleware : it would be required to automatically compute the ContentLength from Body.... but in `uploadSetup()` you systematically set the value already (and conversely: if we already add this middleware, no point doing it as well in uploadSetup)

tests/zenko_tests/node_tests/s3SDK.js
```diff
@@ -7,7 +7,42 @@ const sharedHttpHandler = new NodeHttpHandler({
...
+    client.middlewareStack.add(
+        (next, context) => async args => {
+            const input = args.input || {};
+            if (context?.commandName === 'PutObjectCommand'
```
francoisferrand: - why only `PutObjectCommand` ? why not as well put part ?
- is it not the same logic as below, and just another kind of `Body` ?

```suggestion
if (input.ContentLength !== undefined || context?.commandName !== 'PutObjectCommand') {
    return next(args);
}

if (!input.Body) {
   input.Body = '';
   input.ContentLength = 0;
} else if (Buffer.isBuffer(input.Body)) {
   input.ContentLength = input.Body.length;
} else [...]
```

francoisferrand (changes requested):

tests/zenko_tests/node_tests/s3SDK.js
benzekrimaha: In uploadSetup it's setup when the body is empty only , this one is to avoid having to put the contentLength each time we call the method , the middleWare is called upon client creation

tests/ctst/steps/utils/utils.ts
benzekrimaha: We're expection contentLength for putOnly , upload handles this internally already and doesn't fire any error or warning

tests/zenko_tests/node_tests/s3SDK.js
```diff
@@ -7,7 +7,42 @@ const sharedHttpHandler = new NodeHttpHandler({
...
+                && (input.Body === undefined || input.Body === null)) {
+                input.Body = '';
```
SylvainSenechal: are you sure you need to modify body, I think we just need to have content length set to a value

SylvainSenechal (approved):
Works for me but might be other less intrusive method than a global middleware, I think passing the value ContentLength: xx to all putObjectCommand could've fixed it. We have ~25 putObjectCommand in zenko test, but I think everytime, the content length is very well known in advance as an hardcoded value (or 0)

tests/zenko_tests/node_tests/s3SDK.js
benzekrimaha: we only do so when it's undefined or null as an additionnal safety gate

tests/ctst/steps/utils/utils.ts
benzekrimaha: For multipart uploads (UploadPart) we always enter the first branch of uploadTeardown(): if a temp file was created we delete it and remove both body and contentLength keys there. That covers every multipart case because the harness never issues UploadPart without staging a file first. Because UploadPart can’t reach the zero size case, broadening it to a generic else would do nothing for multiparts.

tests/zenko_tests/node_tests/s3SDK.js
benzekrimaha: after checking => UploadSetup runs only in the CLI cucumber steps, so it has to inject contentLength (including the zero-byte PutObject fallback). The middleware sits on the Node SDK clients used by the node test suite, which never call uploadSetup, so it fills ContentLength there. Two separate harnesses, two helpers; both are needed

tests/zenko_tests/node_tests/s3SDK.js
benzekrimaha: The PutObject only guard is just for the case where a test creates an empty object without providing a body. In that situation the SDK can’t compute contentLength, so we synthesise both values before sending the command. Every other flow (including all UploadPart calls) already pass a real body, for which we calculate the payload's contentLength automatically.

francoisferrand: - ok-ish for first question (maybe we don't go into that path, but it would not harm to handle it as well?)
- suggestion for simplification still stands, code is nested and harder to read than it should...

tests/zenko_tests/node_tests/s3SDK.js
francoisferrand: on second read : I understand we can't use the same "fix" as CTST, but I wonder if we have that many cases where adding the length is needed?

If it is only in a few places, would be much simpler to just update these few calls to make "correct" calls to the SDK v3. This may increase the length of the change (PutObjectCommand is used 36 times, though we may not need to fix them all), but it will have much lower cognitive complexity, and will avoid leaving incorrect code here which may be copy-pasted (to some other repo) and fail there. Especially since this is not seamless anyway, as the client must be created in some custom way.

francoisferrand (commented):
still a few open comments (including Sylvain's suggestion)

benzekrimaha:
@francoisferrand For Sylvain's comment The middleware fixes it once for everybody. Most of those putObjectCommand calls build their bodies dynamically (buffers, strings, streams), so “just hardcode ContentLength” means re-calculating it at ~25 sites today plus any new ones later, and risking it drifting whenever a test changes its payload. By keeping the logic in one middleware we consistently set ContentLength no matter how the body is built, and we never have to revisit every call when tests evolve.

francoisferrand:
> Most of those putObjectCommand calls build their bodies dynamically (buffers, strings, streams), so “just hardcode ContentLength” means re-calculating it at ~25 sites today plus any new ones later, and risking it drifting whenever a test changes its payload.

"New tests" is not an issue: people will notice when they introduce it.
And in most (if not all) cases this can be trivially done : each tests "knows" what it is supposed to do, e.g. if it passed an empty body, a string (can trivially compute the string length)...

> By keeping the logic in one middleware we consistently set ContentLength no matter how the body is built, and we never have to revisit every call when tests evolve.

Another way of seeing this is that we make it seem like the AWS SDK is behaving in a way that it is not : and so people may get used to the idea the field is not needed, while it actually MUST be specified with the new SDK...

maeldonn (approved):

tests/zenko_tests/node_tests/cloudserver/keyFormatVersion/tests/versionedBucket.js
```diff
@@ -39,7 +39,12 @@ async function putObjectVersions(objName, vFormat) {
...
         // eslint-disable-next-line no-await-in-loop
         const result = await s3.send(new PutObjectCommand(
-            { Bucket: BUCKET_NAME[vFormat], Key: objName },
+            {
+                Bucket: BUCKET_NAME[vFormat],
+                Key: objName,
+                Body: '',
+                ContentLength: 0,
+            },
         ));
```
francoisferrand: nit: consistent style

```suggestion
       const result = await s3.send(new PutObjectCommand({
            Bucket: BUCKET_NAME[vFormat],
            Key: objName,
            Body: '',
            ContentLength: 0,
        }));
```

francoisferrand: (a few others below)

francoisferrand (approved):

PR #2319, opened 2026-02-11
Run ctst locally

SylvainSenechal:
ISSUE: ZENKO-5180

Run ctst inside Codespace :
 - Update ./run.sh file
 - Use cli-testing as a library 
Some other refactoring


.devcontainer/setup.sh
```diff
@@ -74,8 +79,8 @@ done
 
 docker image prune -af
 
-CTST_TAG=$(sed 's/.*"cli-testing": ".*#\(.*\)".*/\1/;t;d' ./tests/ctst/package.json)
+GH_TOKEN=$GITHUB_TOKEN # Needed to install cli-testing (private repo)
```
SylvainSenechal: Cli-testing is a private repo. Because we now use cli-testing as library (you will see later in the pr, before it was used in a weird way), we now need to pull it, so I added here and there some way to authenticate.

Tbh I would prefer if cli-testing was public, maybe we can discuss it, because i don't really like the way I added all these token passing around

SylvainSenechal:
I will clean my commits after reviews

.github/scripts/end2end/run-e2e-ctst.sh
SylvainSenechal, line 1: Trying to minimize the utilisation of global env variable that are used only once, or sometimes even useless, there is more work to do on this topic but this will be for other pr

.github/scripts/end2end/run-e2e-ctst.sh
```diff
@@ -175,6 +167,7 @@ kubectl run $POD_NAME \
         --env=AZURE_BLOB_URL=$AZURE_BACKEND_ENDPOINT  \
         --env=AZURE_QUEUE_URL=$AZURE_BACKEND_QUEUE_ENDPOINT \
         --env=VERBOSE=1 \
+        --env=SDK=true \
```
SylvainSenechal, line 179: Cli testing can be run in sdk mode, or CLI mode. 

This shouldn't have impact, because actually this env used to be set in the Dockerfile of the E2E_IMAGE. 
I put it here instead, because imo putting it in the Dockerfile was kinda hidding this parameter, and we are already defining all the other envs variable here

.github/scripts/end2end/run-e2e-ctst.sh
```diff
@@ -212,4 +205,11 @@ kubectl run $POD_NAME \
       }
     ]
   }
-}' -- ./run "$COMMAND" $WORLD_PARAMETERS --parallel $PARALLEL_RUNS --retry $RETRIES --retry-tag-filter @Flaky --format junit:$JUNIT_REPORT_PATH "$@"
+}' -- yarn cucumber-js \
```
SylvainSenechal, line 217: Imo much better, before it was hard to see, but a lot of things were happening behind the ./run commands, which wasn't even defined in Zenko, but in Cli-testing.

Here, we are now just naturally using cucumber syntax to run cucumber tests

.github/workflows/end2end.yaml
```diff
@@ -392,9 +388,12 @@ jobs:
             cli-testing
       - name: Configure GIT
         run: |
-          git config --global url.https://x-access-token:${{ env.GIT_ACCESS_TOKEN }}@github.com/.insteadOf https://github.com/
-          git config --global url.https://x-access-token:${{ env.GIT_ACCESS_TOKEN }}@github.com/.insteadOf github.com:
-          git config --global url.https://x-access-token:${{ env.GIT_ACCESS_TOKEN }}@github.com/.insteadOf ssh://git@github.com/
+          git config --global url."https://x-access-token:${{ env.GIT_ACCESS_TOKEN }}@github.com/".insteadOf "https://github.com/"
+          git config --global url."https://x-access-token:${{ env.GIT_ACCESS_TOKEN }}@github.com/".insteadOf "github.com:"
+          git config --global url."https://x-access-token:${{ env.GIT_ACCESS_TOKEN }}@github.com/".insteadOf "ssh://git@github.com/"
+          git config --global url."https://x-access-token:${{ env.GIT_ACCESS_TOKEN }}@github.com/".insteadOf "git+ssh://git@github.com/"
```
SylvainSenechal: this is what I was talking about when I said I struggled to deal with cli-testing auth. It only worked when I added the ~/.netrc thing, which kinda looks smelly

.github/workflows/end2end.yaml
```diff
@@ -405,9 +404,9 @@ jobs:
       - name: Install ctst test dependencies
         working-directory: tests/ctst
         run: yarn install
-      - name: Lint ctst tests
-        working-directory: tests/ctst
-        run: yarn lint
+      # - name: Lint ctst tests
```
SylvainSenechal: remind me

.github/workflows/end2end.yaml
SylvainSenechal, line 666: Also nicer, bye bye """"""""""""", and now we can clearly see which tags are triggered (yeah cucumber allows to use multiple tags with and/or syntax)

tests/ctst/cucumber.config.cjs
SylvainSenechal, line 1: The cucumber configuration. 
Should be about the same as what we used to use, that was defined in cli-testing here : 
https://github.com/scality/cli-testing/blob/85204dd35d1c9a76c1eb9aa8452a83fac13708ca/package.json#L39

We should definitly have another Jira ticket to investigate what changes can we make to the cucumber config.
The ctst tests in the ci are outputting +50K lines of unreadable stuff, and it evens ends up being truncated, and hard to download because +100mb, so hopefully later we find a way to make the output look better.

Also some testing to do around parallel stuff : I suspect 98% of the tests can be run in parallel, and very fast, and only a few tests are slow/problematic => would be nice to separate them

tests/ctst/Dockerfile
SylvainSenechal, line 1: This is the equivalent of what we had in Cli-testing too : https://github.com/scality/cli-testing/blob/development/1.3/Dockerfile

I kept it mostly the same, but defined it here, so that cli-testing is used as a library, and not something where we store dockerfiles, commands, configs for Zenko.

tests/ctst/package.json
SylvainSenechal, line 1: Many packages here can be updated to a higher version, but let's keep it for another pr

tests/ctst/run-ctst-locally.sh
```diff
@@ -0,0 +1,68 @@
...
+#   ./run-ctst-locally2.sh getObject ctst-image:local      # Uses local image
+
+CUCUMBER_TAG="${1#@}"
+IMAGE_NAME="${2:-ghcr.io/scality/zenko/zenko-e2e-ctst:latest}"
+
+# Version is used with a Before Hook (in ctst/common/hooks.ts) to skip 
+# certain tests based on their @version tag.
+VERSION=$(cat ../../VERSION | grep -Po 'VERSION="\K[^"]*')
+POD_NAME="ctst-end2end"
+WORLD_PARAMETERS="$(jq -c <<EOF
```
SylvainSenechal, line 17: This is where we still have frictions : The current setup is basically enough to run tests locally, but these tests needs a lot of env variables that are passed as "world parameter". Here, I used a few that I defined manually to make one test work.

Later gonna work on using kube client to get secrets directly at the beginning of the tests, this will give us 2 things : 
- Reduce setup needed to run test locally
- Reduce and refactor the number of things happening in .sh setup files. Final objective is also to have variables defined only once at the same place

.github/scripts/end2end/deploy-zenko.sh
```diff
@@ -137,7 +137,10 @@ env $(dependencies_env) envsubst < ${ZENKO_CR_PATH} | kubectl -n ${NAMESPACE} ap
 
 # Fix Zookeeper memory issues on newer Ubuntu GHA runners
 # this is a temporary fix.
-bash ${DIR}/fix-zookeeper.sh "${ZENKO_NAME}" "${NAMESPACE}"
+
+# Update February 2026 : Fix useless in github runners (and flaky)
```
SylvainSenechal: Just got this in the CI : 

```
+ bash ./fix-zookeeper.sh end2end default
Waiting for Zookeeper StatefulSet (end2end-base-quorum) to be created by operator...
statefulset.apps/end2end-base-quorum patched
pod "end2end-base-quorum-0" deleted
Error from server (NotFound): pods "end2end-base-quorum-0" not found
ERROR: Zookeeper pod end2end-base-quorum-0 failed to become Ready after patching.
+ rm -f zenko-key.pub.FBuDLQ zenko-key.yxMuZ4
```

.devcontainer/setup.sh
```diff
@@ -1,5 +1,10 @@
 #!/usr/bin/env bash
 
+# Note : You can inspect the logs of the Codespace creation this way :
+# Ctrl+Shift+P -> "Codespaces: Export Logs" 
+# You will download a zip file on your local machine,
+# In the zip, you want to look at the creation.log file
+
```
francoisferrand: should this be in the README instead ?
or is really just a Codespace usage, nothing specific to this repo → remove from here, but may need a doc or presentation "introduction to codespace", and/or a "Codespace tips&tricks" page ?

.devcontainer/setup.sh
francoisferrand: since we do a checkout, do we still need to build a docker image?
(or maybe this is followup work and one of the next steps)

.devcontainer/setup.sh
```diff
@@ -74,8 +79,8 @@ done
 
 docker image prune -af
 
-CTST_TAG=$(sed 's/.*"cli-testing": ".*#\(.*\)".*/\1/;t;d' ./tests/ctst/package.json)
+GH_TOKEN=$GITHUB_TOKEN # Needed to install cli-testing (private repo)
 SORBET_TAG=$(yq eval '.sorbet.tag' solution/deps.yaml)
 DRCTL_TAG=$(yq eval '.drctl.tag' solution/deps.yaml)
-docker build --build-arg CTST_TAG=$CTST_TAG --build-arg SORBET_TAG=$SORBET_TAG --build-arg DRCTL_TAG=$DRCTL_TAG -t $E2E_CTST_IMAGE_NAME:$E2E_IMAGE_TAG ./tests/ctst
+docker build --build-arg GH_TOKEN=$GH_TOKEN --build-arg SORBET_TAG=$SORBET_TAG --build-arg DRCTL_TAG=$DRCTL_TAG -t $E2E_CTST_IMAGE_NAME:$E2E_IMAGE_TAG ./tests/ctst
```
francoisferrand: passing the token as build arg must never be done, as the token ends up being stored in the image.

the way to do a checkout in docker build is to use [build secrets](https://docs.docker.com/build/building/secrets/#git-authentication-for-remote-contexts) :

```suggestion
GIT_AUTH_TOKEN=$GH_TOKEN docker build --secret id=GIT_AUTH_TOKEN --build-arg SORBET_TAG=$SORBET_TAG --build-arg DRCTL_TAG=$DRCTL_TAG -t $E2E_CTST_IMAGE_NAME:$E2E_IMAGE_TAG ./tests/ctst
```

(note: there is also a way for [ssh credentials](https://docs.docker.com/build/building/secrets/#ssh-mounts))

.github/scripts/end2end/deploy-zenko.sh
francoisferrand: if useless, just remove it: it is still in the git history anyway

.github/workflows/end2end.yaml
francoisferrand: which command failed in which context (codespace, ci, inside docker image...) ?
- we need to use the GIT_ACCESS_TOKEN to have access to other repos (instead of GitHub usual reds, which are restricted the current repo + public repos)
- the git config above is already a trick to inject the GIT_ACCESS_TOKEN in git urls (so we don't need to modify the url in every place it is used: package.json, ...)
- netrc is an alternative to configure git auth, which should not be needed if the git config are made (but maybe we miss some, etc...)

tests/ctst/Dockerfile
```diff
@@ -1,28 +1,62 @@
...
-RUN chmod 0777 -R /ctst/
-
-ENV SDK=true
+RUN mkdir -p /ctst
 
 WORKDIR /ctst
 
+LABEL maintainer="Object Squad <object-squad@scality.com>"
+LABEL name="zenko-e2e-ctst"
+LABEL vendor="scality"
```
francoisferrand: - the name of these labels are non standard
- does not seem really useful to put them in Dockerfile : they should be derived from the image tag used when building, so can best be managed at build time IMHO

tests/ctst/Dockerfile
```diff
@@ -1,28 +1,62 @@
...
+LABEL version=$TAG
```
francoisferrand: what is the point/usage of this `version` label (and accordingly the TAG arg) ?
nothing in the test use it, so why bother?

tests/ctst/README.md
```diff
@@ -1,63 +1,66 @@
...
-The script runs the end to end tests in a pod withing the kubernetes cluster,
-this was done to give the tests access to internal resources such as the Kafka cluster.
+```bash
+cd tests/ctst
+./run-ctst-locally.sh @getObject
+```
 
-The `kubectl run` command uses a custom image of CTST containing the required test
-folders. The image can be built and pushed with the following steps:
+This uses the pre-built image `ghcr.io/scality/zenko/zenko-e2e-ctst:latest` which is loaded into kind during Codespace setup.
```
francoisferrand: `latest` does not work since we have multiple branches : so should always be rebuilt (maybe use an image tag which is unique to the branch, but not sure how to build images & tag them so it is practical...)

tests/ctst/README.md
```diff
@@ -1,63 +1,66 @@
...
+### Using a Custom Image
 
-# Building the image
-docker build --build-arg CTST_TAG=0.2.0 . -t ghcr.io/scality/playground/<username>/custom-ctst:0.2.0
+If you need to test changes to the Dockerfile or dependencies:
 
-# Pushing the custom image into a repository
-docker push ghcr.io/scality/playground/<username>/custom-ctst:0.2.0
+```bash
+# 1. Login to GHCR (required for private base images)
```
francoisferrand: not needed in codespace, this should be done automatically...

tests/ctst/run-ctst-locally.sh
```diff
@@ -0,0 +1,67 @@
...
+
+# Run tests using standard cucumber-js syntax
+kubectl exec "$POD_NAME" -- env \
+    TARGET_VERSION="$VERSION" \
+    SDK=true \
+    yarn cucumber-js \
+        --config cucumber.config.cjs \
+        --world-parameters "$WORLD_PARAMETERS" \
+        --tags "@$CUCUMBER_TAG" \
+        --exit
```
francoisferrand, line 67: * is this faster than just `docker build` and `kubectl run` ?
(docker (re)build should now be very fast, since we only copy files to the container...)
* instead of copying, we could use volume mount ; this would however be less generic, as it requires the cluster to run locally : so that approach would not work against a remote artesca lab (only for codespace or running locally)

tests/ctst/run-ctst-locally.sh
francoisferrand: seems now we are running as a regular node program: so it should work as well without docker/pod?
→ maybe some tests (and world) need to be adapted, but we should try to investigate what changes it would require (which commands to tweak, ...). Esp. we already need special tricks to run kubectl commands, so it may actually simplify stuff...

tests/ctst/run-ctst-locally.sh
```diff
@@ -0,0 +1,67 @@
+#!/bin/bash
+set -exu
+
+# Usage: ./run-ctst-locally.sh <tag> [image]
+# Examples:
+#   ./run-ctst-locally.sh getObject                  # Uses default GHCR image
+#   ./run-ctst-locally.sh getObject ctst-image:local # Uses local image
+
+CUCUMBER_TAG="${1#@}"
```
francoisferrand: instead of passing specifically "tags", we could just forward"all" arguments to yarn, with "$@" : so the script is a "transparent" wrapper, and more flexible (if we need to pass extra cucumber params, for exemple to run a single test or multiple tags...)

may be a bit more tricky with the optional image though, may need to add more (bash) argument parsing, either to support running like this `./run-ctst-locally.sh [image] [-- <extra params>]` or to use a flag `./run-ctst-locally.sh [-i <image>] <extra params>`...

tests/ctst/run-ctst-locally.sh
```diff
@@ -0,0 +1,67 @@
...
+kubectl cp ./steps "$POD_NAME":/ctst/steps
+kubectl cp ./world "$POD_NAME":/ctst/world
+
+kubectl cp ./tsconfig.json "$POD_NAME":/ctst/tsconfig.json
+kubectl cp ./cucumber.config.cjs "$POD_NAME":/ctst/cucumber.config.cjs
+
+# Run tests using standard cucumber-js syntax
+kubectl exec "$POD_NAME" -- env \
+    TARGET_VERSION="$VERSION" \
+    SDK=true \
```
francoisferrand: SDK is repeated, would be better to have a default in the image (or even better in CTST)...
(not sure we really want to use the CLI mode, it is very much slower and a lot more verbose)

tests/ctst/run-ctst-locally.sh
francoisferrand, line 1: just `run.sh`?
(since it may be used not just locally -we could/should use it in CI, to avoid having multiple code-path- and already in ctst folder)

.devcontainer/setup.sh
SylvainSenechal: moved to the readme

.devcontainer/setup.sh
SylvainSenechal: We could pull latest and then load it in kind directly. But we will have an image that's from whichever CI was run last. Should be good enough 99.9% of the time as we don't expect the image to change often, and we can still rebuild it locally if really needed

.devcontainer/setup.sh
```diff
@@ -74,8 +74,9 @@ done
...
 
-CTST_TAG=$(sed 's/.*"cli-testing": ".*#\(.*\)".*/\1/;t;d' ./tests/ctst/package.json)
+# Build CTST image from current branch 
 SORBET_TAG=$(yq eval '.sorbet.tag' solution/deps.yaml)
 DRCTL_TAG=$(yq eval '.drctl.tag' solution/deps.yaml)
-docker build --build-arg CTST_TAG=$CTST_TAG --build-arg SORBET_TAG=$SORBET_TAG --build-arg DRCTL_TAG=$DRCTL_TAG -t $E2E_CTST_IMAGE_NAME:$E2E_IMAGE_TAG ./tests/ctst
-kind load docker-image  ${E2E_CTST_IMAGE_NAME}:${E2E_IMAGE_TAG}
+TAG_NAME=ctst_codespace_setup
+GIT_AUTH_TOKEN=$GITHUB_TOKEN docker build --secret id=GIT_AUTH_TOKEN --build-arg SORBET_TAG=$SORBET_TAG --build-arg DRCTL_TAG=$DRCTL_TAG -t $E2E_CTST_IMAGE_NAME:$TAG_NAME ./tests/ctst
+kind load docker-image ${E2E_CTST_IMAGE_NAME}:$TAG_NAME
```
DarkIsDude: Force your IDE to always add a new line at the end of each file and avoid this forbidden icon

tests/ctst/cucumber.config.cjs
DarkIsDude: I remember a comment from William explaining that tests are already in parallel ?

DarkIsDude: And ca you create the JIRA ticket in the EPIC that Francois presented us ?

tests/ctst/README.md
```diff
@@ -1,63 +1,71 @@
-# CTST tests for Zenko
```
DarkIsDude, line 1: Does it make sense to mention somehow the .devcontainer ?

tests/ctst/run-ctst-locally.sh
DarkIsDude: I strongly agree to avoid several .sh script. Also to make sure everything will continue to work, I think we should find a way to test that in the CI ?

tests/ctst/run-ctst-locally.sh
DarkIsDude: I'm not sure to understand what you are trying to do here @francoisferrand

.devcontainer/setup.sh
SylvainSenechal: added   "files.insertFinalNewline": true 👍

tests/ctst/cucumber.config.cjs
SylvainSenechal: Yeah, in the run-ctst.sh file, we run cucumber with --parallel $PARALLEL_RUNS, so they are run in parallel.
There is some more work to do to about the config, because for example, what happens when we have parallel run defined in the config file, but also provide as an argument to the run command ? 

Will be sorted in
https://scality.atlassian.net/browse/ZENKO-5192

tests/ctst/Dockerfile
SylvainSenechal: I reused the ones we used to define in cli-testing. But I agree, can be removed completely 
https://github.com/scality/cli-testing/blob/0ef7b6b0bc9e0c24dce6c91427dc345b1087ec4d/Dockerfile#L17

tests/ctst/Dockerfile
SylvainSenechal: Removed because useless (see previous comment)

tests/ctst/README.md
SylvainSenechal: why not, it's free, i added a small line

tests/ctst/run-ctst-locally.sh
SylvainSenechal: I'm not a big fan of run.sh, it's too short and grepping "run" returns thousands of results. Actually this is the reason i originally got confused, because in run-e2e-ctst, the pods was executing run, and it took me some time to realize it was the cli-testing run, and not this file. I'm not fixed on this specific name, but I think it's nice to have a more identifiable name, the current name, or "run-cucumber-tests.sh" for example leaves no room for ambiguity. 

Also, more importantly, there is indeed an objective which would be that local tests and github ci tests are run using the same same file, but at the moment there isn't done, as there are still some difference between the 2 files. I think eventually we will do it, and when it happens we can think again about the naming

tests/ctst/run-ctst-locally.sh
SylvainSenechal: I tried something but it got a little bit dirty with the parsing of the image indeed. 
I did make a change so that we can now use natural Cucumber tags expression though

tests/ctst/run-ctst-locally.sh
SylvainSenechal: Yes, we may get there eventually, this should be doable, but probably lots of variables related to endpoint needs to change with port forward

tests/ctst/run-ctst-locally.sh
SylvainSenechal: I removed it from here, and put it in the dockerfile (like it was before)

tests/ctst/run-ctst-locally.sh
SylvainSenechal: But if we do a docker build, we have to reload the image into kind, and this step is quite long

tests/ctst/cucumber.config.cjs
francoisferrand: > Also some testing to do around parallel stuff : I suspect 98% of the tests can be run in parallel, and very fast, and only a few tests are slow/problematic => would be nice to separate them

the analysis has already been done https://scality.atlassian.net/browse/ZENKO-5116?focusedCommentId=403493, and followup tickets created in https://scality.atlassian.net/browse/ZENKO-5116

→ as for the test & analysis, we should be sorted out (all that's left is to just do it)
→ there may be something you want to achieve with ZENKO-5192, but not clear what (the description seems to be about "what to check", not "what we want to achieve"...) : stability? speed? parallelism? [we have tickets on that in ZENKO-5116] "reducing log verbosity" ? [indeed we have no ticket on this I think] ...

tests/ctst/run-ctst-locally.sh
francoisferrand: > I'm not sure to understand what you are trying to do here @francoisferrand

avoiding to duplicate the logic from the Dockerfile in a separate script: rebuilding docker image or copying and restarting node process is typically about the same duration, so we could consider to just replace this script with just `docker build && kubect run`.

but the issue is really the time it takes to load an image in `kind`, thanks @SylvainSenechal

tests/ctst/run-ctst-locally.sh
francoisferrand: agreed this is not for this PR

> port forward

in most case we use (or should use) the ingress : so no port forward...
we may however need to tweak etc/host or some other trick for DNS resolution, though

tests/ctst/README.md
```diff
@@ -1,63 +1,76 @@
...
+GIT_AUTH_TOKEN=$GITHUB_TOKEN docker build \
+  --secret id=GIT_AUTH_TOKEN \
+  --build-arg SORBET_TAG=$SORBET_TAG \
+  --build-arg DRCTL_TAG=$DRCTL_TAG \
+  -t ctst-local:dev \
+  .
+
+# 2. Load into kind and reset the pod
+kind load docker-image ctst-local:dev
+kubectl delete pod ctst-end2end
```
francoisferrand, line 39: why would it be created already?
is it created automatically on devcontainer startup?

tests/ctst/Dockerfile
```diff
@@ -1,28 +1,59 @@
 ARG SORBET_TAG=latest
-ARG CTST_TAG=latest
 ARG DRCTL_TAG=latest
 
 FROM ghcr.io/scality/sorbet:$SORBET_TAG AS sorbet
 FROM ghcr.io/scality/zenko-drctl:$DRCTL_TAG AS drctl
```
francoisferrand, line 5: note for later (more when we want to run CTST tests outside of k8s) : maybe we do not actually need these binaries, we could run them directly in pods...

(typically, esp. for Sorbet, this is often how it is done: kubectl exec into `sorbetctl` container...)

tests/ctst/cucumber.config.cjs
```diff
@@ -0,0 +1,15 @@
+
+module.exports = {
+    default: {
+        requireModule: ['ts-node/register', 'tsconfig-paths/register'],
+        require: ['steps/**/*.ts', 'common/**/*.ts', 'world/**/*.ts'],
+        paths: ['features/**/*.feature'],
+        format: [
+            'progress-bar',
+            '@cucumber/pretty-formatter',
+            'json:reports/cucumber-report.json',
```
francoisferrand, line 10: does this generate a JUnit report? (needed for the overall Zenko workflow, to report failed tests summary)

it is the "stock" reporter, or the one implemented in cli-testing ?

francoisferrand (commented):
LGTM except maybe for the https://github.com/scality/Zenko/pull/2319/changes#r2813543342 question : need to make sure we don't "break" failed tests reporting via JUnit test report

tests/ctst/Dockerfile
```diff
@@ -1,28 +1,59 @@
...
+# Install AWS CLI
+RUN curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64-${AWSCLI_VERSION}.zip" -o "awscliv2.zip" -k \
+    && unzip awscliv2.zip \
+    && ./aws/install \
+    && rm -rf awscliv2.zip aws
+
+COPY package.json ./
+COPY yarn.lock ./
+
+RUN --mount=type=secret,id=GIT_AUTH_TOKEN \
```
francoisferrand, line 46: nit: this will work in CI and codespace, but fail when running locally if SSH is used to auth with GitHub.
So (eventually, if we still use this image) we will need to further improve this to support both GIT_AUTH_TOKEN and SSH... e.g. something like adding the 2 mounts (if possible...), and replacing the URL depending on what we get...

tests/ctst/README.md
SylvainSenechal: I updated the comment. Basically, only needed if the pod was already created before with a different image

tests/ctst/cucumber.config.cjs
SylvainSenechal: okay I reviewed this last comment : my code changes do not modify the existing logic, in run-e2e-ctst, cucumber is ran with the same previous parameter (-- format junitxxxx) to generate the report. 

What I checked though, is if this report was exported to the artifact, and I found it was not.
So I updated one of the action file to export the report to the artifact, and also found that the xml report is not really human readable, so I also exported an html version (while keeping the first format). Also bumped the action action-junit-report from v4 to v6

tests/ctst/README.md
```diff
@@ -1,63 +1,76 @@
...
-docker build --build-arg CTST_TAG=0.2.0 . -t ghcr.io/scality/playground/<username>/custom-ctst:0.2.0
+### Rebuilding the CTST Image
 
-# Pushing the custom image into a repository
-docker push ghcr.io/scality/playground/<username>/custom-ctst:0.2.0
+If you need to test changes to the Dockerfile or dependencies:
+
+```bash
+# 1. Build the CTST image
+
```
DarkIsDude, line 27:
```suggestion
```

tests/ctst/README.md
```diff
@@ -1,63 +1,76 @@
...
-The `kubectl run` command uses a custom image of CTST containing the required test
-folders. The image can be built and pushed with the following steps:
+Run tests with a tag:
 
-``` bash
-cd ./tests/ctst/
+```bash
+cd tests/ctst
+./run-ctst-locally.sh @getObject
+./run-ctst-locally.sh "@PreMerge and not @Flaky"  # Use quotes for complex expressions
```
DarkIsDude, line 16:
```suggestion
./run-ctst-locally.sh "@PreMerge and not @Flaky"
```

You have the same letter. I think it's pretty a standard for bash command and it's not needed to explain it ?

tests/ctst/run-ctst-locally.sh
DarkIsDude: My goal was mainly to avoid duplicate logic. I'm fine to iterate later, we should track that in a JIRA and in the epic and add some comments and the beginning of each 🙏

DarkIsDude (approved):
For me we are good. Some small comments. If you have a re-review, just retrigger the flow. Thanks for your work !

.github/actions/archive-artifacts/action.yaml
```diff
@@ -51,6 +51,16 @@ runs:
         token: ${{ inputs.trunk_token }}
       continue-on-error: true
 
+    - name: Archive Cucumber test reports to artifacts
```
francoisferrand, line 54: we already generated (and archived) JUnit report : why add another report?

.github/scripts/end2end/run-e2e-ctst.sh
```diff
@@ -1,14 +1,15 @@
...
-ZENKO_NAME=${1:-end2end}
-COMMAND=${2:-"premerge"}
+# Usage: run-e2e-ctst.sh <TAGS>
+# Examples:
+#   run-e2e-ctst.sh "@PreMerge"
+#   run-e2e-ctst.sh "@PreMerge and not @PRA"
+#   run-e2e-ctst.sh "@PRA"
+
+TAGS=${1:?'Error: TAGS argument is required (e.g., "@PreMerge", "@PRA")'}
+ZENKO_NAME="end2end"
```
francoisferrand, line 11: good to allow overriding the name:
```suggestion
ZENKO_NAME="${2:-end2end}"
```

.github/workflows/end2end.yaml
```diff
@@ -392,9 +391,11 @@ jobs:
             cli-testing
       - name: Configure GIT
         run: |
-          git config --global url.https://x-access-token:${{ env.GIT_ACCESS_TOKEN }}@github.com/.insteadOf https://github.com/
-          git config --global url.https://x-access-token:${{ env.GIT_ACCESS_TOKEN }}@github.com/.insteadOf github.com:
-          git config --global url.https://x-access-token:${{ env.GIT_ACCESS_TOKEN }}@github.com/.insteadOf ssh://git@github.com/
+          echo "machine github.com login x-access-token password ${GIT_ACCESS_TOKEN}" > ~/.netrc
+          chmod 600 ~/.netrc
```
francoisferrand, line 395: nit: can be improved later, but there is definitely something wrong here... The match below should work : if it does not, it means we have another kind of remote (maybe `https://git@github.com`)

tests/ctst/cucumber.config.cjs
francoisferrand: the Junit report is not human readable, but already parsed by the CI to display the failed builds (and kind of lingua-franca of test reports, can be used in other tools if needed)

is that (html) report useful? What information do you actually need to consume (and in which context / for what purpose) from this artifact ?

(not necessarily to change it now, not an issue to store it as well ; but not sure just storing the artifact is useful, so I just want to understand what we are really looking for.... so we may eventually tackle that need)

francoisferrand (approved):

SylvainSenechal:
/approve
will address nits comments in next pr as the ci here is passing

PR #2324, opened 2026-02-16
Run zenko tests in codespace + some cleanup of unused files

SylvainSenechal:
Issue: ZENKO-5181


.devcontainer/devcontainer.json
```diff
@@ -13,6 +13,7 @@
         "ghcr.io/devcontainers/features/docker-in-docker": {},
         "ghcr.io/devcontainers/features/github-cli:1": {},
         "ghcr.io/devcontainers/features/node:1": {},
+        "ghcr.io/devcontainers/features/python:1": {},
```
SylvainSenechal, line 16: Python is already available in the codespace, but some yarn install command because something "gyp" related to python failed to install. This should fix it

tests/zenko_tests/node_tests/npm_chain.sh
SylvainSenechal, line 1: Unrelated to running zenko tests in Codespace, but I did some decluttering in this pr and removed 4 unused files (removed docker entry point, and the associated files that werent used by anything else)

tests/zenko_tests/node_tests/run-node-tests-locally.sh
SylvainSenechal, line 1: This structure and usage is very similar to what I did for the Cucumber tests.
Tbh, I was also able to run the test without a pod, but it required port forward, and using different values for some variables, which later will probably become annoying because we don't want to have to deal with 2 values for the same variables depending on whether the test is running in github ci or codespace.
So maybe later we'll come back to this if we decide to run all tests outside of cluster, but don't really wanna deal with it for now

tests/zenko_tests/node_tests/run-node-tests-locally.sh
```diff
@@ -0,0 +1,75 @@
...
+set -e
+
+ZENKO_NAME=${ZENKO_NAME:-end2end}
+IMAGE_NAME="${3:-ghcr.io/scality/zenko/zenko-e2e:latest}"
+POD_NAME="node-tests-local"
+
+TEST_GREP="${1:-should list objects in V2 format}"
+TEST_PATH="${2:-cloudserver}"
+
+ZENKO_ACCESS_KEY=$(kubectl get secret end2end-management-account-creds.v1 -o jsonpath='{.data.accessKey}' | base64 -d)
```
SylvainSenechal, line 32: These env variables will eventually be removed (calculated automatically from within the test)

tests/zenko_tests/node_tests/run-node-tests-locally.sh
```diff
@@ -0,0 +1,75 @@
...
+#   kubectl delete pod node-tests-local
+
+set -e
+
+ZENKO_NAME=${ZENKO_NAME:-end2end}
+IMAGE_NAME="${3:-ghcr.io/scality/zenko/zenko-e2e:latest}"
+POD_NAME="node-tests-local"
+
+TEST_GREP="${1:-should list objects in V2 format}"
+TEST_PATH="${2:-cloudserver}"
```
francoisferrand: should it not run "all" tests by default?

tests/zenko_tests/node_tests/run-node-tests-locally.sh
francoisferrand: in any case, we should not have separate scripts for local vs CI setups, as they will diverge over time and/or add duplicate work : so a good first step, but it will need to be reworked indeed
- ideally, running "locally" (even in codespace) will be easier (frictionless), faster (no copy/reload), and allow debugging more easily
- we may run in CI "locally" as well (which avoids a docker rebuild, and can leverage gha cache to speedup yarn install)
- or at least somehow "merge" the scripts used in CI and locally

francoisferrand (approved):

delthas (approved):
Will be really helpful. Thanks!

tests/zenko_tests/node_tests/run-node-tests-locally.sh
```diff
@@ -0,0 +1,75 @@
...
+# Usage: ./run-node-tests-locally.sh <test-grep> <test-path> [image]
+#
+# Examples:
+#   ./run-node-tests-locally.sh "should list objects" "cloudserver/bucketGetV2"
+#   ./run-node-tests-locally.sh "should list objects" "cloudserver" my-e2e:local
+#   ./run-node-tests-locally.sh "" "smoke_tests"                    # Run all smoke tests
+#   ./run-node-tests-locally.sh "" "cloudserver"                    # Run all cloudserver tests
+#
+# If you need to use a different image, delete pod first:
+#   kubectl delete pod node-tests-local
```
maeldonn, line 15: A `--cleanup` flag would be great

tests/zenko_tests/node_tests/run-node-tests-locally.sh
maeldonn, line 1: Should we enforce the launch directory or cd within the script? Or is it useless?

maeldonn (approved):
LGTM

PR #2325, opened 2026-02-17
Base setup so that tests automatically load variables they need

SylvainSenechal:
Issue: ZENKO-5182

WIP


SylvainSenechal:
used for exploration, some of the work re-used in other prs already, wont work or merge this one

PR #2326, opened 2026-02-19
Fix CI cgroupv2 failures by disabling JMX

delthas:
Issue: ZENKO-5194
See: https://github.com/actions/runner-images/issues/13684


delthas:
Hm likely not the right option since we probably need JMX for Zenko.

PR #2327, opened 2026-02-19
Fix CI cgroupv2 failures by enabling cgroupv2 controller subtree control

delthas:
Issue: ZENKO-5194
See: https://github.com/actions/runner-images/issues/13684


delthas:
Will not be done at this time. See #2328

PR #2328, opened 2026-02-19
Fix CI cgroupsv2 failures by upgrading Zookeeper & Kafka

delthas:
Issue: ZENKO-5194

### DIFF
--- a/solution/deps.yaml
+++ b/solution/deps.yaml
@@ -133,7 +133,7 @@ zenko-operator:
   tag: v1.8.3
   envsubst: ZENKO_OPERATOR_TAG
 zookeeper:
-  sourceRegistry: pravega
+  sourceRegistry: ghcr.io/adobe/zookeeper-operator
   image: zookeeper
-  tag: 0.2.15
+  tag: 3.8.4-0.2.15-adobe-20250923
   envsubst: ZOOKEEPER_TAG
--- a/solution/kafka/Dockerfile
+++ b/solution/kafka/Dockerfile
@@ -22,7 +22,7 @@ RUN tar -xzf $kafka_distro
 RUN rm -r kafka_$scala_version-$kafka_version/bin/windows
 
 ####################################################################################################
-FROM eclipse-temurin:17.0.3_7-jre
+FROM eclipse-temurin:17.0.18_8-jre
 
 ARG scala_version=2.13
 ARG kafka_version=3.1.0


delthas:
Needs a proper tagging of the Kafka image (to avoid putting a hash and commenting out the image tag comparison), but otherwise works.

.github/workflows/end2end.yaml
```diff
@@ -318,7 +318,7 @@ jobs:
 
       - name: Check kafka & kafka-connect versions match
         run: |-
-          [ "${{ env.KAFKA_TAG }}-${{ env.MONGODB_CONNECTOR_TAG }}" = "${{ env.KAFKA_CONNECT_TAG }}" ]
+          [ "${{ env.KAFKA_TAG }}-${{ env.MONGODB_CONNECTOR_TAG }}" = "${{ env.KAFKA_CONNECT_TAG }}" ] || true
```
francoisferrand: this check is mandatory, don't forget to re-enable it

solution/deps.yaml
```diff
@@ -51,7 +51,7 @@ kaf: # CLI Kafka client all-in-one
 kafka:
   sourceRegistry: ghcr.io/scality/zenko
   image: kafka
-  tag: 2.13-3.1.2
+  tag: 2.13-3.1.2-fc929acc6e4eb763f1bdf972cbf8d0ecc48c7897
```
francoisferrand: why change this?
we are rebuilding anyway (and adding a source-tree hash I think), so we should be fine...

delthas: We were using `2.13-3.1.2`, not the one appended with the hash (so not the one we built in the previous step) in the ZenkoVersion (causing the change not to be used and the CI to fail)

francoisferrand: I beg to differ:

```
      - name: Build and push kafka
        uses: docker/build-push-action@v5
        with:
          push: true
          context: ./solution/kafka
          build-args: |-
            scala_version=${{ env.scala_version }}
            kafka_version=${{ env.kafka_version }}
          tags: "${{ env.KAFKA_IMAGE }}:${{ env.KAFKA_TAG }}-${{ env.BUILD_TREE_HASH }}"
          cache-from: type=gha,scope=kafka-${{ env.KAFKA_TAG }}
          cache-to: type=gha,mode=max,scope=kafka-${{ env.KAFKA_TAG }}
```

so if something is missing, it may be in the "deployment" itself (.github/scripts/end2end/deploy-zenko.sh), not the "spec" (i.e. deps.yaml)

francoisferrand (approved):

PR #2330, opened 2026-02-19
Use generated Kafka image in CI tests

delthas:
Issue: ZENKO-5196

### DIFF
--- a/solution/zenkoversion.yaml
+++ b/solution/zenkoversion.yaml
@@ -96,10 +96,10 @@ spec:
         tag: ${KAFKA_CLEANER_TAG}
       cluster:
         image: ${KAFKA_IMAGE}
-        tag: ${KAFKA_TAG}
+        tag: ${KAFKA_TAG}-${BUILD_TREE_HASH}
       connect:
         image: ${KAFKA_CONNECT_IMAGE}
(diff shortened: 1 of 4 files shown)


solution/zenkoversion.yaml
```diff
@@ -96,10 +96,10 @@ spec:
         tag: ${KAFKA_CLEANER_TAG}
       cluster:
         image: ${KAFKA_IMAGE}
-        tag: ${KAFKA_TAG}
+        tag: ${KAFKA_TAG}-${BUILD_TREE_HASH}
```
francoisferrand, line 99: Since you update zenkoversion, you must also update build.sh : to ensure the ISO does indeed contain this image...

(Note: unfortunately, the CI does not currently use the generated ISO itself, for various reasons. This is one of the known issues of zenko ci, we have a ticket for this)

francoisferrand (changes requested):
~~Does not embed the correct Kafka, ISO would not work (in artesca).
Fixed in https://github.com/scality/Zenko/pull/2331~~

francoisferrand (dismissed):

delthas:
Will use #2331 instead.

PR #2332, opened 2026-02-20
Allow building ISO on macos & verify it contains all required images

francoisferrand:
Add sanity check to verify ISO is correct: it verifies that every image referenced in the zenkoversion CR (which is bundled in the ISO) is indeed in the ISO.
This is not an actual "functional" test, but simply a sanity check to mitigate risk if the zenkoversion is updated in a way that is inconsistent with the registry images.

Also adds fixes to allow building ISO on macos, so that it possible to iterate locally.

Issue: ZENKO-5197

### DIFF
--- a/solution/build.sh
+++ b/solution/build.sh
@@ -267,6 +267,28 @@ function dedupe()
     ${HARDLINK} -c ${IMAGES_ROOT}
 }
 
+function validate_registry()
+{
+    echo "Validating ISO registry completeness against zenkoversion.yaml..."
+    local missing=0
+
+    while IFS= read -r image_ref ; do
+        local image="${image_ref%:*}"
+        local tag="${image_ref##*:}"
+        local manifest="${IMAGES_ROOT}/${image}/${tag}/manifest.json"
+        if [ ! -f "${manifest}" ] ; then
+            echo "::error::Missing image in ISO registry: ${image}:${tag}"
+            missing=$((missing + 1))
+        fi
+    done < <(yq eval '.. | select(.image and .tag) | .image + ":" + .tag' ${ISO_ROOT}/zenkoversion.yaml | sort -u)
+
+    if [ ${missing} -gt 0 ] ; then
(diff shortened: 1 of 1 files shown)


solution/build.sh
```diff
@@ -267,6 +267,28 @@ function dedupe()
...
+
+    while IFS= read -r image_ref ; do
+        local image="${image_ref%:*}"
+        local tag="${image_ref##*:}"
+        local manifest="${IMAGES_ROOT}/${image}/${tag}/manifest.json"
+        if [ ! -f "${manifest}" ] ; then
+            echo "::error::Missing image in ISO registry: ${image}:${tag}"
+            missing=$((missing + 1))
+        fi
+    done < <(yq eval '.. | select(has("image") and has("tag")) | .image + ":" + .tag' ${ISO_ROOT}/zenkoversion.yaml | sort -u)
```
delthas: Maybe `select(.image)`? Corresponding to your changes in the previous commit

delthas (approved):

benzekrimaha (approved):

solution/build.sh
```diff
@@ -316,7 +353,7 @@ function download_tools()
     done
 
     # Extract tools from images
-    yq eval '.[] | select(has("toolName") and (has("toolUrl")|not)) | .sourceRegistry + "/" + .image + ":" + .tag + " " + .toolName' deps.yaml |\
+    yq eval '.[] | select(.toolName and (has("toolUrl")|not)) | .sourceRegistry + "/" + .image + ":" + .tag + " " + .toolName' deps.yaml |\
```
benzekrimaha, line 356:
```suggestion
    yq eval '.[] | select(.toolName and (.toolUrl |not)) | .sourceRegistry + "/" + .image + ":" + .tag + " " + .toolName' deps.yaml |\
```

francoisferrand: tried, but does not actually work (syntax error)

PR #2333, opened 2026-02-23
Support new API GetObjectAttributes

maeldonn:
Issue: ZENKO-5117

### DIFF
--- a/solution/deps.yaml
+++ b/solution/deps.yaml
@@ -125,7 +125,7 @@ vault:
   dashboard: vault2/vault-dashboards
   policy: vault2/vault-policies
   image: vault2
-  tag: 8.11.1
+  tag: 8.11.2
   envsubst: VAULT_TAG
 zenko-operator:
   sourceRegistry: ghcr.io/scality
--- a/solution/deps.yaml
+++ b/solution/deps.yaml
@@ -16,7 +16,7 @@ cloudserver:
   sourceRegistry: ghcr.io/scality
   dashboard: cloudserver/cloudserver-dashboards
   image: cloudserver
-  tag: 9.3.2
+  tag: 9.3.4
   envsubst: CLOUDSERVER_TAG
 drctl:
   sourceRegistry: ghcr.io/scality
--- a/tests/ctst/features/user-defined-metadata/get-object-attributes.feature
+++ b/tests/ctst/features/user-defined-metadata/get-object-attributes.feature
@@ -37,13 +37,14 @@ Feature: GetObjectAttributes
         And an IAM policy attached to the entity "user" with "Allow" effect to perform "s3" "GetObjectAttributes" on "arn:aws:s3:::*"
         When the user calls GetObjectAttributes for "object-no-metadata" requesting "<attributes>"
         Then the operation finished without error
+        And the GetObjectAttributes response should contain "<attributes>"
 
         Examples:
             | attributes      |
             | ObjectSize      |
             | ETag            |
+            | StorageClass    |
             | ObjectSize,ETag |
-
 
     @2.14.0
     @PreMerge
@@ -104,6 +105,7 @@ Feature: GetObjectAttributes
         And an IAM policy attached to the entity "user" with "Allow" effect to perform "s3" "GetObjectVersionAttributes" on "arn:aws:s3:::*"
         When the user calls GetObjectAttributes for "object-no-metadata-versioned" requesting "<attributes>" with the latest version
         Then the operation finished without error
+        And the GetObjectAttributes response should contain "<attributes>"
 
         Examples:
             | attributes      |
--- a/tests/ctst/features/user-defined-metadata/get-object-attributes.feature
+++ b/tests/ctst/features/user-defined-metadata/get-object-attributes.feature
@@ -1,6 +1,6 @@
 Feature: GetObjectAttributes
 
-    @2.13.5
+    @2.14.0
     @PreMerge
     @GetObjectAttributes
     Scenario: Access Denied for GetObjectAttributes without user metadata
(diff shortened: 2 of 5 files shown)


solution/deps.yaml
```diff
@@ -16,7 +16,7 @@ cloudserver:
   sourceRegistry: ghcr.io/scality
   dashboard: cloudserver/cloudserver-dashboards
   image: cloudserver
-  tag: 9.3.2
+  tag: 9.3.3
```
maeldonn:
```suggestion
  tag: 9.3.4
```
Update after [CLDSRV-856](https://scality.atlassian.net/browse/CLDSRV-856) merge and release to fix the tests.

tests/ctst/steps/user-defined-metadata/get-object-attributes.ts
```diff
@@ -0,0 +1,48 @@
...
+    const attributesList = JSON.stringify(attributes.split(',').map(attr => attr.trim()));
+
+    const result = await S3.getObjectAttributes({
+        bucket: bucketName,
+        key: objectName,
+        objectAttributes: attributesList,
+        ...(versionId && { versionId }),
+    });
+
+    // The AWS SDK may fail to deserialize non-standard responses
```
DarkIsDude: We should find a way to check the value of some metadata ? Maybe with a new `Then the metadata <name> should be equal to <value>` and create a new `Given the user metadata <name> on the object with <value>` ?

tests/ctst/features/user-defined-metadata/get-object-attributes.feature
SylvainSenechal, line 1: Here we are adding a test for a new api, it seems we don't really have other ctst tests specifically focused on uniquely doing api tests, but for organisation purpose, it might be nice to have a folder features/api/get-object-attributes. If we were to have many apis tested, the other feature folder would be "drowned" in the api tests

tests/ctst/steps/user-defined-metadata/get-object-attributes.ts
maeldonn: These CTST tests are focused on validating the authorization model, verifying that the correct IAM policy combinations are properly enforced through all components. The return values themselves are already covered by cloudserver's own tests.

Adding value assertions at the E2E level would increase maintenance cost without meaningfully increasing confidence, since there's no inter-service transformation between cloudserver reading metadata and returning the response that could silently corrupt a value while still returning HTTP 200.

francoisferrand: * "may" is weird : when does it fail, when does it pass?
* it obviously cannot return "custom" attributes, but it should be possible to get access to "all" headers (maybe through some possibly custom middleware), so you can get the requested information. @DarkIsDude added some exemple code (for listobjectv2) in the design or TAD I think (or somewhere else :-P ), it should probably look similar : and we should definitely do it here, both to verify it works (the actual expected value is returned) and have some exemple code

tests/ctst/steps/user-defined-metadata/list-objects-v2.ts
```diff
@@ -0,0 +1,30 @@
...
+    const result = await S3.listObjectsV2({
+        bucket: bucketName,
+        ...(optionalAttributes && {
+            optionalObjectAttributes: JSON.stringify(
+                optionalAttributes.split(',').map(attr => attr.trim()),
+            ),
+        }),
+    });
+
+    // The AWS SDK may fail to deserialize non-standard responses
```
francoisferrand: same here

tests/ctst/steps/user-defined-metadata/get-object-attributes.ts
maeldonn: I’ve updated the comment. Since CTST tests rely on the cli-testing library rather than the SDK, we can't override the behavior as we did before. Instead, I've added assertions for the standard attributes.

francoisferrand: you should still be able to call the SDK directly, even from ctst tests, right?
and/or make a PR in cli-testing repo to add the missing code on top of SDK...

francoisferrand:
> # Incorrect fix version
> The `Fix Version/s` in issue [ZENKO-5117](https://scality.atlassian.net/browse/ZENKO-5117) contains:
> 
> * _None_
> 
> Considering where you are trying to merge, I ignored possible hotfix versions and I expected to find:
> 
> * `2.14.0`
> 
> Please check the `Fix Version/s` of [ZENKO-5117](https://scality.atlassian.net/browse/ZENKO-5117), or the target branch of this pull request.

@maeldonn This is not the right ticket: should reference ZENKO-5177 .... and thus you will need to rename your branch, and reopen the PR...

[ZENKO-5117]: https://scality.atlassian.net/browse/ZENKO-5117?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ZENKO-5117]: https://scality.atlassian.net/browse/ZENKO-5117?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

tests/ctst/features/user-defined-metadata/get-object-attributes.feature
```diff
@@ -0,0 +1,133 @@
+Feature: GetObjectAttributes
+
+    @2.14.0
+    @PreMerge
+    @GetObjectAttributes
+    Scenario: Access Denied for GetObjectAttributes without user metadata
```
SylvainSenechal, line 6: Nit: I think should rework the scenario title a bit, it's kinda misleading : makes it looks like the access denied is because there are no userMetadata, should be more like 

"Access Denied for GetObjectAttributes when missing IAM permissions" + "no metadata case" 

Also, more importantly, it's probably not gonna be (easily) doable given the rigidity of cucumber steps, but  I see you have scenarios "without user metadata", "with user metadata" : This is the kind of situation where we usually want to keep a single scenario, but have the with/without inside a parameter to be put in the Examples array. edit : yeah not possible unless we wanna refactor 5000 tests :)

 There might be the same issue though with the versioned/non versioned bucket, and this one should be fixable

tests/ctst/features/user-defined-metadata/get-object-attributes.feature
```diff
@@ -0,0 +1,133 @@
...
+            | ETag                |
+            | x-amz-meta-foo      |
+            | x-amz-meta-bar      |
+            | ETag,x-amz-meta-foo |
+
+    @2.14.0
+    @PreMerge
+    @GetObjectAttributes
+    Scenario Outline: Access Denied for GetObjectAttributes with versioned bucket
+        Given a "Non versioned" bucket
```
SylvainSenechal, line 85: title is versioned 
given is non versioned
this should be merged with the very first scenario, and versioned or not added to Example table parameters

tests/ctst/features/user-defined-metadata/get-object-attributes.feature
```diff
@@ -0,0 +1,133 @@
...
+    @GetObjectAttributes
+    Scenario Outline: GetObjectAttributes returns expected attributes for an object with user metadata
+        Given a "Non versioned" bucket
+        And 1 objects "object-with-metadata" of size 100 bytes with user metadata "x-amz-meta-foo=bar"
+        And a IAM_USER type
+        And an IAM policy attached to the entity "user" with "Allow" effect to perform "s3" "GetObject" on "arn:aws:s3:::*"
+        And an IAM policy attached to the entity "user" with "Allow" effect to perform "s3" "GetObjectAttributes" on "arn:aws:s3:::*"
+        And an IAM policy attached to the entity "user" with "Allow" effect to perform "scality" "GetObjectAttributesCustom" on "arn:aws:s3:::*"
+        When the user calls GetObjectAttributes for "object-with-metadata-1" requesting "<attributes>"
+        Then the operation finished without error
```
SylvainSenechal, line 72: Are you not checking the returned attribute values ?

maeldonn:
Moved to #2337

PR #2335, opened 2026-02-24
Use Zookeeper image from Adobe fork

delthas:
Issue: ZENKO-5198

### DIFF
--- a/solution/deps.yaml
+++ b/solution/deps.yaml
@@ -133,7 +133,7 @@ zenko-operator:
   tag: v1.8.3
   envsubst: ZENKO_OPERATOR_TAG
 zookeeper:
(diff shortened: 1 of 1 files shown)


solution/deps.yaml
```diff
@@ -130,10 +130,10 @@ vault:
 zenko-operator:
   sourceRegistry: ghcr.io/scality
   image: zenko-operator
-  tag: v1.8.3
+  tag: 9452d6fd6a0a9fabd6e7cbb4a6b4f227e6e8396f
```
francoisferrand: to be updated with proper tag

francoisferrand (approved):

SylvainSenechal (approved):

PR #2336, opened 2026-02-26
Run zenko tests out of cluster

SylvainSenechal:
Issue: ZENKO-5201


.devcontainer/setup.sh
```diff
@@ -47,23 +54,6 @@ for i in $(seq 0 $array_length); do
     )
 done
 
-(
```
SylvainSenechal, line 50: No need for this anymore as we don't need the container, and config is handled differently

.github/actions/deploy/action.yaml
```diff
@@ -87,9 +87,5 @@ runs:
       if: ${{ inputs.deploy_metadata == 'true' }}
     - name: End-to-end configuration
       shell: bash
-      run: bash configure-e2e.sh "end2end" ${E2E_IMAGE_NAME}:${E2E_IMAGE_TAG} "default"
-      working-directory: ./.github/scripts/end2end
-    - name: Linting
-      shell: bash
-      run: bash run-e2e-test.sh "end2end" ${E2E_IMAGE_NAME}:${E2E_IMAGE_TAG} "lint" "default"
```
SylvainSenechal, line 94: No need to do the linting in the deploy action..
+ this deploy action is run multiple times in the ci
+ it was building an image, taking time, all that just for linting
Now you will see later in end2end.yaml there is a simple yarn run lint

.github/scripts/end2end/configure-e2e-ctst.sh
```diff
@@ -3,6 +3,9 @@ set -exu
 
 DIR=$(dirname "$0")
 
+# Set up ingress endpoints and /etc/hosts for out-of-cluster access
+source "$DIR/configure-e2e-endpoints.sh"
```
SylvainSenechal, line 7: This setup will be common to ctst and end2end tests. 
Hopefully we will soon be able to have a common setup for both type of tests, to avoid having to go from configure-ctst to configure e2e, with many things that are common

.github/scripts/end2end/configure-e2e-endpoints.sh
```diff
@@ -0,0 +1,139 @@
+#!/bin/bash
```
SylvainSenechal, line 1: This should be idempotent. I used it multiple times in Codespace without issues

.github/scripts/end2end/configure-e2e.sh
```diff
@@ -6,49 +6,26 @@ DIR=$(dirname "${0}")
 . "$DIR"/common.sh
 
 ZENKO_NAME=${1:-end2end}
-E2E_IMAGE=${2:-ghcr.io/scality/zenko/zenko-e2e:latest}
-NAMESPACE=${3:-default}
-
-SERVICE_ACCOUNT="${ZENKO_NAME}-config"
-POD_NAME="${ZENKO_NAME}-config"
-MANAGEMENT_ENDPOINT="http://${ZENKO_NAME}-management-orbit-api:5001"
```
SylvainSenechal, line 14: This is the kind of things that we have to change now that the test doesn't run in a pod within the Zenko cluster

.github/scripts/end2end/configure-e2e.sh
```diff
@@ -72,52 +49,56 @@ kubectl run kafka-topics \
...
-  --env=CRR_ROLE_NAME=${CRR_ROLE_NAME} \
-  --env=DEPLOY_CRR_LOCATIONS=${DEPLOY_CRR_LOCATIONS} \
-  --command -- python3 configuration.py
+# Run configuration.py directly
+ZENKO_TESTS_DIR="$DIR/../../../tests/zenko_tests"
+pip3 install --break-system-packages -r "$ZENKO_TESTS_DIR/requirements.txt"
+
+cd "$ZENKO_TESTS_DIR"
+
+envsubst < e2e-config.yaml.template > e2e-config.yaml
```
SylvainSenechal, line 58: This used to be setup in the Dockerfile that I deleted

.github/scripts/end2end/setup-e2e-env.sh
```diff
@@ -0,0 +1,130 @@
+#!/usr/bin/env bash
```
SylvainSenechal, line 1: I came up with this to deal with the env variables.
It's a bit tricky as I wanted it to work for both github ci and codespace

Variables setup here are pretty much the same as the ones that were setup in the (now deleted) run-e2e-test.sh file

.github/workflows/end2end.yaml
```diff
@@ -483,8 +471,6 @@ jobs:
         working-directory: ./.github/scripts/end2end
       - name: Deploy second Zenko for PRA
         run: bash deploy-zenko.sh end2end-pra default './configs/zenko.yaml'
-        env:
-          ZENKO_MONGODB_DATABASE: pradb
```
SylvainSenechal, line 487: setup in prepare-pra.sh directly

.github/workflows/end2end.yaml
```diff
@@ -350,33 +349,22 @@ jobs:
           cache-from: type=gha,scope=kafka-connect-${{ env.KAFKA_CONNECT_TAG }}
           cache-to: type=gha,mode=max,scope=kafka-connect-${{ env.KAFKA_CONNECT_TAG }}
 
-  build-test-image:
```
SylvainSenechal, line 353: No more image, just a new lint step as I removed it from the deploy action

.github/workflows/end2end.yaml
```diff
@@ -550,27 +536,32 @@ jobs:
         env:
           GIT_ACCESS_TOKEN: ${{ steps.app-token.outputs.token }}
       - name: Run init CI test
-        run: bash run-e2e-test.sh "end2end" ${E2E_IMAGE_NAME}:${E2E_IMAGE_TAG} "end2end" "default"
-        working-directory: ./.github/scripts/end2end
+        run: |
+          source .github/scripts/end2end/setup-e2e-env.sh
+          yarn run test_operator
```
SylvainSenechal: I prefer this, simple, direct use of yarn commands instead of going through scripts with FOUR arguments 🧐 

I just don't really like having to source the env var for each run so if you have suggestion I'm open

tests/zenko_tests/e2e_config/accounts.py
```diff
@@ -25,7 +25,7 @@ def get_credentials(token, account_id):
         RoleArn="arn:aws:iam::%s:role/scality-internal/storage-manager-role" % (account_id),
         RoleSessionName='end2end',
         WebIdentityToken=token,
-        DurationSeconds=60 * 60 * 12, # 12 hrs
+        DurationSeconds=60 * 60 * 12, # 12 hrs (max allowed by STS for assume role)
```
SylvainSenechal, line 28: I wanna keep this comment, I tried to use 30 days because in my codespace it was expiring after 12h, but when I used 30 days it broke because the api only accepts 12h

tests/zenko_tests/node_tests/package.json
```diff
@@ -28,33 +28,32 @@
     "mocha": "^10.0.0",
     "mocha-junit-reporter": "^2.2.1",
     "mocha-multi-reporters": "^1.1.7",
-    "mocha-tags": "^1.0.1",
```
SylvainSenechal, line 31: - Never used,
- Could be useful in theory,
- Will never be used in practice, so just add again if needed (will never be needed 🌚 )

tests/zenko_tests/node_tests/package.json
```diff
@@ -28,33 +28,32 @@
...
+    "test_bucket_policy": "mocha --exit -t 10000 --reporter mocha-multi-reporters --reporter-options configFile=config.json,cmrOutput=mocha-junit-reporter+testsuitesTitle+test_bucket_policy --recursive cloudserver/bucketPolicy/tests",
+    "test_operator": "mocha --exit -t 10000 --reporter mocha-multi-reporters --reporter-options configFile=config.json,cmrOutput=mocha-junit-reporter+testsuitesTitle+test_operator ./init_test.js",
+    "test_smoke": "mocha --exit -t 10000 --reporter mocha-multi-reporters --reporter-options configFile=config.json,cmrOutput=mocha-junit-reporter+testsuitesTitle+test_smoke --recursive smoke_tests",
+    "test_iam_policies": "mocha --exit -t 15000 --reporter mocha-multi-reporters --reporter-options configFile=config.json,cmrOutput=mocha-junit-reporter+testsuitesTitle+test_iam_policies --recursive iam_policies",
     "test_all_extensions": "run-p --aggregate-output test_crr test_aws_crr test_expiration test_transition test_ingestion_oob_s3c",
```
SylvainSenechal, line 55: Btw, I just had a look at end2end-sharded. The step "Run backbeat end to end tests", which is running test_all_extensions, now runs within ~12 min, against ~28min before.

I had to double check, as I thought some tests were simply not run, but we run the same tests. I think previously the parallel run was not working, maybe because of some weird stuff happening when this command is run in a pod, the pod can't run stuff in parallel or something.

francoisferrand: it seems this step perform a yarn install: by having it running directly in the runner, it will now be cached (automatically!) by setup-node, and would thus be much faster?

.github/workflows/end2end.yaml
francoisferrand: you can instead add a first step which updates the env for followup steps

c.f. https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#setting-an-environment-variable

.github/scripts/end2end/setup-e2e-env.sh
```diff
@@ -0,0 +1,130 @@
...
+    if [[ "${BASH_SOURCE[0]:-}" != "${0}" ]]; then
+        # Sourced — register cleanup only if not already registered
+        if [ -z "${_SETUP_E2E_CLEANUP_SET:-}" ]; then
+            trap "kill ${_MONGO_PF_PID} 2>/dev/null || true" EXIT
+            export _SETUP_E2E_CLEANUP_SET=1
+        fi
+    else
+        trap "kill ${_MONGO_PF_PID} 2>/dev/null || true" EXIT
+    fi
+    sleep 2
```
francoisferrand: why sleep?

.github/scripts/end2end/setup-e2e-env.sh
```diff
@@ -0,0 +1,130 @@
...
+    _MONGO_PF_PID=$!
+    # Clean up on exit only if we started it (not when sourced for interactive use)
+    if [[ "${BASH_SOURCE[0]:-}" != "${0}" ]]; then
+        # Sourced — register cleanup only if not already registered
+        if [ -z "${_SETUP_E2E_CLEANUP_SET:-}" ]; then
+            trap "kill ${_MONGO_PF_PID} 2>/dev/null || true" EXIT
+            export _SETUP_E2E_CLEANUP_SET=1
+        fi
+    else
+        trap "kill ${_MONGO_PF_PID} 2>/dev/null || true" EXIT
```
francoisferrand:
```suggestion
    if [[ "${BASH_SOURCE[0]:-}" == "${0}" ]]; then
        trap "kill ${_MONGO_PF_PID} 2>/dev/null || true" EXIT
    elif [ -z "${_SETUP_E2E_CLEANUP_SET:-}" ]; then
        # Sourced — register cleanup only if not already registered
        trap "kill ${_MONGO_PF_PID} 2>/dev/null || true" EXIT
        export _SETUP_E2E_CLEANUP_SET=1
    fi
```

.github/scripts/end2end/setup-e2e-env.sh
```diff
@@ -0,0 +1,130 @@
...
+MONGO_PORT=$(echo "${MONGO_REPLICA_SET_HOSTS_ORIG}" | cut -d: -f2)
+MONGO_SVC=$(echo "${MONGO_FQDN}" | cut -d. -f1)
+MONGO_NS=$(echo "${MONGO_FQDN}" | cut -d. -f2)
+
+# Only start port-forward if not already listening
+if ! ss -tlnp 2>/dev/null | grep -q ":${MONGO_PORT}" && \
+   ! lsof -i ":${MONGO_PORT}" &>/dev/null; then
+    kubectl port-forward -n "${MONGO_NS}" "svc/${MONGO_SVC}" "${MONGO_PORT}:${MONGO_PORT}" &
+    _MONGO_PF_PID=$!
+    # Clean up on exit only if we started it (not when sourced for interactive use)
```
francoisferrand: comment is not correct anymore, seems we now always kill mongo? (i.e. except I already registered)

so ust handling the "double registration" should be enough:
```
if [ -z "${_SETUP_E2E_CLEANUP_SET:-}" ]; then
    trap "kill ${_MONGO_PF_PID} 2>/dev/null || true" EXIT
    export _SETUP_E2E_CLEANUP_SET=1
fi
```

.devcontainer/setup.sh
```diff
@@ -2,12 +2,19 @@
...
+echo 'export GIT_ACCESS_TOKEN="${GITHUB_TOKEN}"' >> "$ZENKO_ENV_FILE"
 # Disable GCP tests as we don't have credentials setup in devcontainer
-export GCP_BACKEND_DESTINATION_LOCATION=
+echo 'export GCP_BACKEND_DESTINATION_LOCATION=' >> "$ZENKO_ENV_FILE"
+
+# Source now for this session
+source "$ZENKO_ENV_FILE"
+
+# Auto-source in future Codespace terminals
+echo '[ -f "$HOME/.zenko.env" ] && source "$HOME/.zenko.env"' >> "$HOME/.bashrc"
```
francoisferrand: should probably be added to the devcontariner's Dockerfile instead

francoisferrand (approved):

tests/zenko_tests/node_tests/package.json
SylvainSenechal: Yeah also possible

.github/scripts/end2end/setup-e2e-env.sh
SylvainSenechal: the port forward isn't instant, but yeah this is a flaky risk, i changed it so we wait till its ready instead

.github/scripts/end2end/configure-e2e.sh
```diff
@@ -72,52 +49,56 @@ kubectl run kafka-topics \
...
-  --env=CRR_SOURCE_LOCATION_NAME=${CRR_SOURCE_LOCATION_NAME} \
-  --env=CRR_DESTINATION_LOCATION_NAME=${CRR_DESTINATION_LOCATION_NAME} \
-  --env=CRR_SOURCE_ACCOUNT_NAME=${CRR_SOURCE_ACCOUNT_NAME} \
-  --env=CRR_DESTINATION_ACCOUNT_NAME=${CRR_DESTINATION_ACCOUNT_NAME} \
-  --env=CRR_ROLE_NAME=${CRR_ROLE_NAME} \
-  --env=DEPLOY_CRR_LOCATIONS=${DEPLOY_CRR_LOCATIONS} \
-  --command -- python3 configuration.py
+# Run configuration.py directly
+ZENKO_TESTS_DIR="$DIR/../../../tests/zenko_tests"
+pip3 install --break-system-packages -r "$ZENKO_TESTS_DIR/requirements.txt"
```
delthas, line 54: I wouldn't want to break my system packages 😁 If we need to install dependencies, could we perhaps use a venv (maybe use uv?)

SylvainSenechal: I was thinking about using uv, maybe in another pr, but i could add it now 🤔

delthas (approved):

PR #2337, opened 2026-02-27
Support new API GetObjectAttributes

maeldonn:
Issue: [ZENKO-5177](https://scality.atlassian.net/browse/ZENKO-5177)

This is the same PR as https://github.com/scality/Zenko/pull/2333 but with a different branch name.

[ZENKO-5177]: https://scality.atlassian.net/browse/ZENKO-5177?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/ctst/steps/user-defined-metadata/get-object-attributes.ts
+++ b/tests/ctst/steps/user-defined-metadata/get-object-attributes.ts
@@ -1,0 +1,121 @@
+import assert from 'assert';
+import { Then, When } from '@cucumber/cucumber';
+import { parseStringPromise } from 'xml2js';
+import { Identity } from 'cli-testing';
+import Zenko from '../../world/Zenko';
+import { safeJsonParse } from '../../common/utils';
+
+async function getObjectAttributes(
+    world: Zenko,
+    objectName: string,
+    attributes: string,
+    versionId?: string,
+) {
+    world.resetCommand();
+
+    const bucketName = world.getSaved<string>('bucketName');
+    const attributesList = attributes.split(',').map(attr => attr.trim());
+    const credentials = Identity.getCurrentCredentials();
+
+    let path = `/${bucketName}/${objectName}?attributes`;
+    if (versionId) {
+        path += `&versionId=${versionId}`;
+    }
+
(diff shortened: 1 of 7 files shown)


tests/ctst/types/xml2js.d.ts
```diff
@@ -0,0 +1,3 @@
+declare module 'xml2js' {
```
DarkIsDude, line 1: https://www.npmjs.com/package/@types/xml2js this one is not working ?

DarkIsDude (approved):
Congrats, fine for me 👏. Maybe we can try to merge some logic in `ts` file but I'm not sure it'll help a lot for reading. So I'm fine to keep them separated

tests/ctst/steps/user-defined-metadata/get-object-attributes.ts
```diff
@@ -0,0 +1,121 @@
...
+    const bucketName = world.getSaved<string>('bucketName');
+    const attributesList = attributes.split(',').map(attr => attr.trim());
+    const credentials = Identity.getCurrentCredentials();
+
+    let path = `/${bucketName}/${objectName}?attributes`;
+    if (versionId) {
+        path += `&versionId=${versionId}`;
+    }
+
+    const result = await world.awsS3Request(
```
francoisferrand, line 25: nit: this uses "raw" curl requests, not the SDK. Ok functionally, but it would have been nice to "document" (through code) how to do this with the SDK.... or possibly using `cloudserverclient`
→ please check if we already integrated our "extensions" of `ListObjectv2` and `GetObjectAttributes` in there, and create missing tickets accordingly (in cloudserverclient and zenko for integrating it)

francoisferrand (approved):

tests/ctst/types/xml2js.d.ts
maeldonn: It works but I didn't think about it. Will be updated in a new ticket.

PR #2339, opened 2026-03-02
Add CTST tests for live replication with oplog location stripping

delthas:
Add replication.feature with a Scenario Outline testing live object
replication across 4 sizes (0B, 1MiB, 2MiB, 4MiB) against a 2MiB
stripping threshold. Configure objectRefreshSizeThresholdMB in the
E2E Zenko CR so tests exercise both stripped and preserved paths.

Issue: ZENKO-5204

### DIFF
--- a/tests/ctst/features/replication.feature
+++ b/tests/ctst/features/replication.feature
@@ -1,0 +1,26 @@
+Feature: Live Object Replication
+    This feature tests live replication of objects across different sizes,
+    verifying that oplog location stripping does not affect replication
+    correctness.
+
+    The objectRefreshSizeThresholdMB is set to 2 in the Zenko CR, meaning
+    objects >= 2 MiB have their location stripped from the MongoDB change
+    stream event. The replication consumer detects this and re-fetches
+    metadata from the source before replicating.
+
+    @2.14.0
+    @PreMerge
+    @ReplicationTest
+    Scenario Outline: Object of <objectDescription> replicates with location stripping threshold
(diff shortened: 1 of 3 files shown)


tests/ctst/features/replication.feature
```diff
@@ -0,0 +1,26 @@
...
+    The objectRefreshSizeThresholdMB is set to 2 in the Zenko CR, meaning
+    objects >= 2 MiB have their location stripped from the MongoDB change
+    stream event. The replication consumer detects this and re-fetches
+    metadata from the source before replicating.
+
+    @2.14.0
+    @PreMerge
+    @ReplicationTest
+    Scenario Outline: Object of <objectDescription> replicates with location stripping threshold
+        Given an existing bucket "" "with" versioning, "without" ObjectLock "without" retention mode
```
SylvainSenechal: no bucket name, is it because no bucket name => uses default bucket name ? 

For this, and for the "1 objects "repl-obj"", if you're brave enough to repush some code and rerun the ci, you can use names that are even more custom than this, so that in the event of a bug, we can grep in the artifact and not end up with 20000+ matches

SylvainSenechal (approved):

tests/ctst/features/replication.feature
delthas: no bucket name -> uses autogenerated name with random suffix (to avoid collisions)

delthas: Updated to use custom, non-colliding names.

maeldonn (approved):

Issue #18, opened 2017-07-25
Tutorial instances don't match doc


baslking:
The Nginx instance is not named as the tutorial documentation indicates, or maybe that's a Mac version thing?  In any case the docker-stack.yml file uses this docker not one named Nginx

ID                  NAME                MODE                REPLICAS            IMAGE                       PORTS
awuju96gnhho        zenko-testing_lb    replicated          1/1                 zenko/loadbalancer:latest   *:80->80/tcp
p3us1marp8hq        zenko-testing_s3    replicated          1/1                 scality/s3server:latest     *:0->8000/tcp

LaureVergeron:
Fixed, thanks :)

PR #2340, opened 2026-03-02
Add CTST tests for storage usage reporting API

delthas:
## Summary
- Add end-to-end CTST tests for the new `GET /instance/{uuid}/reporting/usage` Pensieve route
- Tests cover: StorageManager happy path (200 + structure validation), unauthorized Keycloak user (403), multi-account presence, and metric accuracy (objectsTotal/bytesTotal strict equality)
- Extend `managementAPIRequest` with optional `username`/`password` params for testing different Keycloak users
- Extend `prepareMetricsScenarios` with `objectSize`/`objectCount` options for uploading non-empty test objects

Issue: ZENKO-5202


delthas:
Depends on https://github.com/scality/pensieve-api/pull/351 and https://github.com/scality/vault2/pull/189

solution/deps.yaml
```diff
@@ -78,9 +78,9 @@ mongodb-connector:
   tag: 1.13.0
   envsubst: MONGODB_CONNECTOR_TAG
 pensieve-api:
-  sourceRegistry: ghcr.io/scality
```
DarkIsDude, line 81: Be careful with that, you need to revert some changes 🙏

tests/ctst/steps/reporting/storageUsageReporting.ts
```diff
@@ -0,0 +1,122 @@
...
+    await prepareMetricsScenarios(world, scenarioConfiguration, {
+        versioning: '',
+        jobNamespace: 'storage-usage-reporting-setup',
+        jobName: 'end2end-ops-count-items',
+        objectSize: 200,
+        objectCount: 3,
+    });
+}
+
+Given('two additional accounts', async function (this: Zenko) {
```
DarkIsDude: To have something generic, we can put `two` as a parameter ? Also I'm not sure this should be inside the reporting, it's very generic ?

tests/ctst/steps/reporting/storageUsageReporting.ts
```diff
@@ -0,0 +1,122 @@
...
+    const accountNames: string[] = [];
+    const accountIds: string[] = [];
+    for (let i = 0; i < 2; i++) {
+        const name = `reporting-test-${Date.now()}-${i}`;
+        await this.createAccount(name, true);
+        const account = await SuperAdmin.getAccount({ accountName: name });
+        accountNames.push(name);
+        accountIds.push(account.id);
+    }
+    this.addToSaved('additionalAccountNames', accountNames);
```
DarkIsDude: IMO we should also refactor the step `Given an account` to reuse this one (and then refactor our code), even if it's more work just adding seems adding a step because we don't want to touch the other one

tests/ctst/steps/reporting/storageUsageReporting.ts
```diff
@@ -0,0 +1,122 @@
...
+    this.addToSaved('additionalAccountIds', accountIds);
+});
+
+When('the user retrieves the storage usage report',
```
DarkIsDude: This one should go inside a steps `pensieve`. That make more sense to regroup step not by 'how we use them' but more by 'where they are'.

tests/ctst/common/hooks.ts
```diff
@@ -97,6 +104,16 @@ After({ tags: '@AzureArchive' }, async function (this: Zenko) {
     );
 });
 
+After({ tags: '@StorageUsageReporting' }, async function (this: Zenko, results) {
```
DarkIsDude: Why adding a tag ? We can clean it always ? Because here if we forgot to add the tags, the test will pass and we don't notice the cleanup ?

DarkIsDude: And if we do that, we can https://github.com/scality/Zenko/pull/2340/changes#diff-92390ef75e6d193a08d69991e395f4ddb3d1155908a53ed4c7caed1f9830c025R66 remove this one to make a unique new tag `@PrepareStorageUsageReportingScenarios` for example ?

tests/ctst/steps/reporting/storageUsageReporting.ts
```diff
@@ -0,0 +1,122 @@
...
+            'GET',
+            `/instance/${this.parameters.InstanceID}/reporting/usage`,
+        );
+        this.addToSaved('reportingResponse', result);
+    });
+
+When('the user retrieves the storage usage report as a no-rights user',
+    async function (this: Zenko) {
+        const noRightsUsername =
+            `${this.parameters.KeycloakUsername || 'storage_manager'}-norights`;
```
DarkIsDude: Why do we use `-norights` ?

tests/ctst/steps/reporting/storageUsageReporting.ts
```diff
@@ -0,0 +1,122 @@
...
+            `/instance/${this.parameters.InstanceID}/reporting/usage`,
+            {},
+            {},
+            noRightsUsername,
+            this.parameters.KeycloakTestPassword || '123',
+        );
+        this.addToSaved('reportingResponse', result);
+    });
+
+Then('the storage usage report response status is {int}',
```
DarkIsDude:
```suggestion
Then('the storage usage report http response code is {int}',
```

tests/ctst/features/reporting/StorageUsageReporting.feature
```diff
@@ -0,0 +1,45 @@
...
+    @2.14.0
+    @PreMerge
+    @StorageUsageReporting
+    @CronJob
+    Scenario Outline: Storage usage report returns accurate metrics
+        Given the environment is set up with bucket created, test data uploaded, and count-items ran
+        And a STORAGE_MANAGER type
+        When the user retrieves the storage usage report
+        Then the storage usage report response status is 200
+        And the storage usage report contains the test account with location "<locationName>"
```
DarkIsDude: Why `<>` ?

DarkIsDude (changes requested):

solution/deps.yaml
delthas: Right, it is in the [Do not merge] commit that I will edit before merging.

tests/ctst/steps/reporting/storageUsageReporting.ts
delthas: Done — parameterized to `{int} additional accounts` and moved to `common/common.ts` so it's reusable.

tests/ctst/steps/reporting/storageUsageReporting.ts
delthas: Agreed this would be nice, but refactoring `Given an account` / `setupEntity` is high-risk — it's deeply integrated with the Identity system, command parameter preservation, and cross-account flows. I've moved and parameterized our step to `common/common.ts` for now. Full refactor of the existing account step could be a follow-up.

tests/ctst/steps/reporting/storageUsageReporting.ts
delthas: The existing codebase convention is feature-based organization (`steps/quotas/`, `steps/utilization/`, `steps/bucket-policies/`, etc.), not service-based. There's no `steps/pensieve/` directory today. Keeping it in `steps/reporting/` is consistent with the current convention.

tests/ctst/common/hooks.ts
delthas: Good point — done. Removed the tag from the After hook (the `!additionalAccountNames` guard already makes it a no-op). Also replaced `@StorageUsageReporting and @CronJob` with a single dedicated `@PrepareStorageUsageReportingScenarios` tag for the Before hook.

tests/ctst/steps/reporting/storageUsageReporting.ts
delthas: `storage_manager-norights` is a pre-existing Keycloak user defined in `.github/scripts/end2end/configs/keycloak_config.json`. It has no functional realm roles (only default `uma_authorization` and `offline_access`), so it's suitable for testing that users without the `StorageManager` role get rejected (403) by the endpoint.

tests/ctst/steps/reporting/storageUsageReporting.ts
delthas: Done 👍

tests/ctst/features/reporting/StorageUsageReporting.feature
delthas: The `prepareMetricsScenarios` function extracts scenario IDs from the Examples table via `pickle.astNodeIds[1]`, which only exists for Scenario Outline rows. A plain Scenario only has `astNodeIds[0]` (the scenario ID), so accessing `[1]` would be `undefined` and break the shared metrics environment setup. The single-row Outline is required by this constraint, and also makes it easy to add more locations later.

DarkIsDude: oO yes ok, it's related to the examples below. Thanks 🙏

tests/ctst/steps/reporting/storageUsageReporting.ts
DarkIsDude: 🤝. I have a weird feeling but can't pinpoint it. So let's move forward (maybe it's just the naming 🤷)

tests/ctst/steps/reporting/storageUsageReporting.ts
DarkIsDude: Can we find a way to avoid duplicating that and make the link obvious ? For example adding a suffix ? Because here we 1. duplicate logic (the password is never used? and default password duplicated) 2. Introduce a non obvious link between the setup and this logic 3. Introduce more complexity that seems not needed

tests/ctst/common/common.ts
```diff
@@ -110,6 +110,20 @@ Given('an account', async function (this: Zenko) {
     await this.createAccount();
 });
 
+Given('{int} additional accounts', async function (this: Zenko, count: number) {
+    const accountNames: string[] = [];
+    const accountIds: string[] = [];
+    for (let i = 0; i < count; i++) {
+        const name = `additional-account-${Date.now()}-${i}`;
+        await this.createAccount(name, true);
```
DarkIsDude: Why do we use the force parameter here ?

tests/ctst/common/common.ts
```diff
@@ -110,6 +110,20 @@ Given('an account', async function (this: Zenko) {
     await this.createAccount();
 });
 
+Given('{int} additional accounts', async function (this: Zenko, count: number) {
+    const accountNames: string[] = [];
+    const accountIds: string[] = [];
+    for (let i = 0; i < count; i++) {
+        const name = `additional-account-${Date.now()}-${i}`;
```
DarkIsDude: We have something already used in the createAccount to manage randomness. 

```
        const accountName = name || this.getSaved<string>('accountName') ||
            `${Constants.ACCOUNT_NAME}${Utils.randomString()}`;
```

We should reuse that no ?

tests/ctst/common/common.ts
```diff
@@ -110,6 +110,20 @@ Given('an account', async function (this: Zenko) {
...
+        await this.createAccount(name, true);
+        const account = await SuperAdmin.getAccount({ accountName: name });
+        accountNames.push(name);
+        accountIds.push(account.id);
+    }
+    this.addToSaved('additionalAccountNames', accountNames);
```
DarkIsDude: `addToSaved` can't be used with object ?

tests/ctst/steps/reporting/storageUsageReporting.ts
DarkIsDude: @delthas I checked the code, didn't find where we use the `Given('an account', async function (this: Zenko) {
` ? Did I miss something ? We can then maybe just drop it ?

tests/ctst/common/common.ts
DarkIsDude: My last point on it is that createAccount can return the data. It's also weird that `createAccount` save data and then you save it again `saveIdentityInformation` but I'm not sure we should tackle it. Let's see what the 2nd reviewer will say.

```            return callback(null, {
                account: result.account.data,
            });
```

tests/ctst/steps/reporting/storageUsageReporting.ts
delthas: Good catch! The reason refactoring it was tricky is that `Given an account` sets up the account as the active identity (for subsequent S3/IAM calls), while our `{int} additional accounts` step just creates accounts as test fixtures for assertions. Different purposes, so merging them wasn't straightforward. But since it's indeed unused in any feature file, we can just drop it. Done.

tests/ctst/steps/reporting/storageUsageReporting.ts
delthas: Good points. Fixed:

1. **Dropped the `password` parameter** from `managementAPIRequest` entirely — nobody was using it. The password now always comes from `this.parameters.KeycloakPassword || '123'`, which is the same for all Keycloak test users including the norights one.

2. **Added a `KeycloakUsernameNoRights` world parameter** instead of building the username dynamically with string concatenation. The flow is now explicit end-to-end:

   - `end2end.yaml` defines `OIDC_USERNAME: 'storage_manager'`
   - `keycloak_config.json` creates the user as `${OIDC_USERNAME}-norights`
   - `run-e2e-ctst.sh` derives `"KeycloakUsernameNoRights":"${KEYCLOAK_TEST_USER}-norights"` in WORLD_PARAMETERS
   - `ZenkoWorldParameters` interface declares `KeycloakUsernameNoRights: string`
   - The step uses `this.parameters.KeycloakUsernameNoRights || 'storage_manager-norights'`

   The `|| 'storage_manager-norights'` fallback is for local development where WORLD_PARAMETERS may not be set — same pattern as `managementAPIRequest` which falls back to `this.parameters.KeycloakUsername || 'storage_manager'`.

   So the link between Keycloak config and test code is now through the same env var chain as `KeycloakUsername`, with a consistent fallback convention.

tests/ctst/common/common.ts
delthas: You're right, it's not needed — the names are unique (generated with `Date.now()` + index) so they'll never collide with the identity cache. Dropped.

tests/ctst/common/common.ts
delthas: Good call, much cleaner. I made `createAccount` return the generated account name (`Promise<string>`) so the step can just do `const name = await this.createAccount()` without needing to know the naming convention. Also dropped the unused `additionalAccountIds` and `SuperAdmin` import in the process.

tests/ctst/common/common.ts
delthas: Moot now — I simplified the step to a single `addToSaved('additionalAccountNames', accountNames)` call. The `additionalAccountIds` was unused so I dropped it entirely.

.github/scripts/end2end/run-e2e-ctst.sh
```diff
@@ -112,6 +112,7 @@ WORLD_PARAMETERS="$(jq -c <<EOF
   "KafkaHosts":"${KAFKA_HOST_PORT}",
   "KafkaAuthHosts":"${KAFKA_AUTH_HOST_PORT}",
   "KeycloakUsername":"${KEYCLOAK_TEST_USER}",
+  "KeycloakUsernameNoRights":"${KEYCLOAK_TEST_USER}-norights",
```
francoisferrand: instead of "no rights", could we map this on "personas" ?
e.g. KeycloakStorageAdmin, KeycloakAccountOwner, KeycloakStorageConsumer...

tests/ctst/common/common.ts
```diff
@@ -106,8 +106,13 @@ async function getTopicsOffsets(topics: string[], kafkaAdmin: Admin) {
     return offsets;
 }
 
-Given('an account', async function (this: Zenko) {
```
francoisferrand, line 109: this one wasn' not used ? 🤦‍♂️

food for thought (not for this PR): is there some kind of "cucumber linter" to detect unused step definitions?

tests/ctst/common/common.ts
francoisferrand: since createAccount already saves the account, can't we just make it so it always appends to a list?

i.e. `saveIdentityInformation` would not simply "overwrite" identityNameForScenario/identityTypeForScenario/accountNameForScenario, but create a list (either in these variables, and update `useSavedIdentity` to return the "latest" -or first?- by default ; or keep the existing variables as is and also create a list)

tests/ctst/steps/reporting/storageUsageReporting.ts
```diff
@@ -0,0 +1,104 @@
...
+    objectsTotal: number;
+}
+
+interface ReportingUsageResponse {
+    isTruncated: boolean;
+    marker: string | null;
+    accounts: Record<string, Record<string, LocationUsage>>;
+}
+
+export async function prepareStorageUsageReportingScenarios(
```
francoisferrand: what is the point of this function? Does it really make sense to keep the extra indirection/abstraction level, or should it be done directly in `Before({ tags: '@PrepareStorageUsageReportingScenarios', timeout: 1200000 }` ?

tests/ctst/common/utils.ts
```diff
@@ -330,7 +332,9 @@ export async function prepareMetricsScenarios(
     const {
         versioning = '',
         jobName = 'end2end-ops-count-items',
-        jobNamespace = `${featureName}-setup`
+        jobNamespace = `${featureName}-setup`,
+        objectSize = 0,
+        objectCount = 1,
     } = options;
 
     if (!fs.existsSync(filePath)) {
```
francoisferrand, line 340: nit: I think there is now (or at least in upcoming CTST release) a helper function/class to handle this atomic/shared init.... out of scope of this ticket/PR, but please create a ticket so we remember to migrate (once 1.3 is integrated)

tests/ctst/common/utils.ts
```diff
@@ -365,7 +369,12 @@ export async function prepareMetricsScenarios(
         for (const scenarioId of scenarioIds) {
             await world.createAccount(scenarioId, true);
             await createBucketWithConfiguration(world, scenarioId, versioning);
-            await putObject(world);
+            for (let i = 0; i < objectCount; i++) {
+                if (objectSize > 0) {
+                    world.addToSaved('objectSize', objectSize);
+                }
```
francoisferrand: ~independant of both loops : should be moved outside ?~

this is really hard to read/follow : saving to world is useful (or even needed) to pass information between steps ; but here we are purely in javascript : so putObject should be refactored to allow (optionally) passing the length (and if neither content nor length is passed, then it should -as before- picked the objectSize from world...)

tests/ctst/common/hooks.ts
```diff
@@ -97,6 +104,16 @@ After({ tags: '@AzureArchive' }, async function (this: Zenko) {
     );
 });
 
+After(async function (this: Zenko, results) {
```
francoisferrand: - why do we want/need to delete "additional accounts"? Are there that many?
- cleanup is almost a philosophical question, and for the most part is done currently (not saying it's good or bad, simply this is the current status in these tests). Is there something specific in these new tests that require the cleanup? Or should we more generally settle the debate, and update the code to _always_ cleanup (but in that case, it should be done in a separate PR) ?

tests/ctst/features/reporting/StorageUsageReporting.feature
```diff
@@ -0,0 +1,45 @@
+Feature: Storage Usage Reporting API
+  The storage usage reporting API allows authorized Keycloak users to retrieve
+  aggregated storage usage metrics across all accounts and locations.
+
+    @2.14.0
+    @PreMerge
+    @StorageUsageReporting
+    Scenario: StorageManager can retrieve the storage usage report
+        Given a STORAGE_MANAGER type
```
francoisferrand: what is a "type" ?

should it be something like this instead ? (I think we have such a step)
```suggestion
        Given a user with "StorageManager" role
```

or maybe introduce identity/persona concepts/names :
```suggestion
        Given an identity with "StorageManager" persona
```

tests/ctst/features/reporting/StorageUsageReporting.feature
```diff
@@ -0,0 +1,45 @@
...
+        When the user retrieves the storage usage report
+        Then the storage usage report http response code is 200
+        And the storage usage report response has a valid structure
+
+    @2.14.0
+    @PreMerge
+    @StorageUsageReporting
+    Scenario: No-rights Keycloak user cannot retrieve the storage usage report
+        When the user retrieves the storage usage report as a no-rights user
```
francoisferrand: "no-rights" : same as above :)
and "user" is ambiguous, I think there is a concensus (at least on Florent side) that we should use `identity` for keycloak users

tests/ctst/features/reporting/StorageUsageReporting.feature
```diff
@@ -0,0 +1,45 @@
...
+
+    @2.14.0
+    @PreMerge
+    @StorageUsageReporting
+    Scenario: Storage usage report contains multiple accounts
+        Given 2 additional accounts
+        And a STORAGE_MANAGER type
+        When the user retrieves the storage usage report
+        Then the storage usage report http response code is 200
+        And the storage usage report contains the additional accounts
```
francoisferrand: this test is actually the "same" as `StorageManager can retrieve the storage usage report`, just with different "content" for the report.

it seems to me we are kind of mixing things here, and there is a better way:
- there is a series of tests to verify "permissions", e.g. allowing (or not) access to the report for storageManager, storageAccountowner, ...
- validating the content: locations, accounts...

→ to simplify tests, reduce redundancy and more easily write (and read/review!) combinations, may be better to use parametrized tests? (not sure how practical, but still)

tests/ctst/features/reporting/StorageUsageReporting.feature
```diff
@@ -0,0 +1,45 @@
...
+        Given the environment is set up with bucket created, test data uploaded, and count-items ran
+        And a STORAGE_MANAGER type
+        When the user retrieves the storage usage report
+        Then the storage usage report http response code is 200
+        And the storage usage report contains the test account with location "<locationName>"
+        And the location metrics show 3 objects and 600 bytes
+
+        Examples:
+          | locationName |
+          | us-east-1    |
```
francoisferrand, line 48: - missing tests for "AccountOwner" persona (and multiple accounts): to verify he can see only his own account only
- missing tests for "StorageReporter" persona

tests/ctst/steps/reporting/storageUsageReporting.ts
```diff
@@ -0,0 +1,104 @@
...
+    });
+
+When('the user retrieves the storage usage report as a no-rights user',
+    async function (this: Zenko) {
+        const result = await this.managementAPIRequest(
+            'GET',
+            `/instance/${this.parameters.InstanceID}/reporting/usage`,
+            {},
+            {},
+            this.parameters.KeycloakUsernameNoRights || 'storage_manager-norights',
```
francoisferrand: what is this "storage_manager-norights" default ?
should we not just ensure we have a value for this parameter (on startup, or even just assert here) ?

tests/ctst/steps/reporting/storageUsageReporting.ts
```diff
@@ -0,0 +1,104 @@
...
+        versioning: '',
+        jobNamespace: 'storage-usage-reporting-setup',
+        jobName: 'end2end-ops-count-items',
+        objectSize: 200,
+        objectCount: 3,
+    });
+}
+
+When('the user retrieves the storage usage report',
+    async function (this: Zenko) {
```
francoisferrand: nit: I think in most case this is at the end of the `When` line; should we do the same?
This makes the whole block look like a regular function...

```suggestion
    When('the user retrieves the storage usage report', async function (this: Zenko) {
```

solution/deps.yaml
delthas: pensieve-api was merged. Fixed.

tests/ctst/steps/reporting/storageUsageReporting.ts
delthas: `this.managementAPIRequest` does `this.parameters.KeycloakUsername || 'storage_manager'`. I supposed this was intended for when you ran the tests locally without using the entire setup script thats sets all the environment variables... I replicated the behavior here.

delthas: Here: https://github.com/scality/Zenko/pull/2340/changes#diff-371f6719b5883f728d9d3218b4ed2eaf1b7ada8faf798acf95b686a7c347c7ddR945

tests/ctst/common/common.ts
delthas: Done — refactored `saveIdentityInformation` to append to a typed `SavedIdentity[]` array instead of overwriting three scalar keys.

Changes:
- Added `SavedIdentity` interface (`identityName`, `identityType`, `accountName`)
- `saveIdentityInformation` now pushes to the array
- Added `getSavedIdentities()` (returns the full array) and `getSavedIdentity(index = -1)` (defaults to last entry, supports negative indexing)
- `useSavedIdentity()` now goes through `getSavedIdentity()` — all 15 existing callers are unchanged since "last entry" matches the previous overwrite semantics
- Migrated the direct `getSaved('identityNameForScenario')` reads in `bucket-policies/common.ts` and `iam-policies/common.ts` to use `getSavedIdentity()`
- Simplified the `{int} additional accounts` step — no more manual `additionalAccountNames` tracking, since `createAccount()` already appends to the list
- The reporting assertion now queries `getSavedIdentities().filter(id => id.identityType === IdentityEnum.ACCOUNT)`
- Also dropped the After cleanup hook for additional accounts (see thread #18)

tests/ctst/common/hooks.ts
delthas: Agreed — dropped the cleanup hook. There's no practical issue from not cleaning up: account names are randomized (no collisions), the CI environment is ephemeral (spun up per run), and the test assertions check for presence, not exclusivity. Cleanup philosophy can be settled codebase-wide in a separate effort.

.github/scripts/end2end/run-e2e-ctst.sh
delthas: Addressed: replaced the `norights` user with the existing `data_consumer` Keycloak user, which already has the `DataConsumer` role (not `StorageManager`), so it correctly gets 403. Removed `KeycloakUsernameNoRights` from world parameters entirely — the step now uses the pre-existing `DataConsumerUsername` parameter.

Regarding the broader "personas" naming (`KeycloakStorageAdmin`, `KeycloakAccountOwner`, etc.): the current world parameters already follow a `{Role}Username` convention (`StorageManagerUsername`, `StorageAccountOwnerUsername`, `DataConsumerUsername`, `DataAccessorUsername`). Adopting a unified persona naming scheme would touch `run-e2e-ctst.sh`, `Zenko.ts` (interface), and every step definition that references these parameters — probably best as a separate PR.

tests/ctst/features/reporting/StorageUsageReporting.feature
delthas: `Given a STORAGE_MANAGER type` is an existing step defined in `common/identity.ts` — it uses a custom `{type}` Cucumber parameter type that maps to the `EntityType` enum and calls `setupEntity()`. It's used across 14 feature files with 47 total occurrences (`IAM_USER`, `STORAGE_MANAGER`, `DATA_CONSUMER`, `ASSUME_ROLE_USER`, etc.).

Renaming to something like `Given a user with "StorageManager" role` would require updating all 47 usages across those 14 feature files, plus the step definition and parameter type — a codebase-wide refactor that should be done in a separate PR.

tests/ctst/features/reporting/StorageUsageReporting.feature
delthas: The "no-rights" part is addressed — replaced with the `data_consumer` user (see thread above).

Regarding "user" vs "identity": "the user" is the standard subject used in Gherkin steps throughout the entire CTST suite — 85+ occurrences across feature files and 18+ step definitions in 9 step files. Switching to "identity" terminology would be a large cross-cutting refactor. It's beyond the scope of this PR.

tests/ctst/steps/reporting/storageUsageReporting.ts
delthas: Done — removed the `prepareStorageUsageReportingScenarios` wrapper. The Before hook now calls `prepareMetricsScenarios` directly with the options inline.

tests/ctst/common/utils.ts
delthas: Created [ZENKO-5214](https://scality.atlassian.net/browse/ZENKO-5214) to track this migration once CTST 1.3 is integrated.

tests/ctst/common/utils.ts
delthas: Done — added an optional `objectSize` parameter to `putObject` (and `uploadSetup`), so callers can pass the size directly. Falls back to `world.getSaved('objectSize')` for backward compatibility with step definitions. `prepareMetricsScenarios` now calls `putObject(world, undefined, undefined, objectSize)` instead of saving to World first.

tests/ctst/steps/reporting/storageUsageReporting.ts
delthas: Done — moved `async function` onto the same line as `When(`/`Then(` for all step definitions in this file.

tests/ctst/common/common.ts
delthas: Done — `cucumber-js --dry-run --format usage` does exactly this out of the box. Added as a CI check and cleaned up the existing unused steps.

Ticket: [ZENKO-5215](https://scality.atlassian.net/browse/ZENKO-5215)
PR: #2349

tests/ctst/features/reporting/StorageUsageReporting.feature
delthas: Done — restructured the feature file into two groups:

**Permission tests** — a single parameterized Scenario Outline with `role`/`expectedStatus` examples (storage_manager → 200, data_consumer → 403). Easy to extend with new roles.

**Content tests** — separate scenarios for response structure, multi-account presence, and metric accuracy.

Also replaced the two hardcoded `When` steps with a single `When('the user retrieves the storage usage report as {string}')` step backed by a role-to-username lookup map.

tests/ctst/features/reporting/StorageUsageReporting.feature
delthas: - **AccountOwner**: The `/reporting/usage` endpoint doesn't support per-account permissions — authorization is all-or-nothing (403 fails fast before any account filtering). An AccountOwner test would behave identically to StorageManager (200 with all accounts) or get 403, neither of which tests account-scoped visibility. We'd need per-account filtering in pensieve-api first.
- **StorageReporter**: No `StorageReporter` Keycloak role exists in the test environment yet. On it

delthas: Actually, for StorageReporter this needs a new user, which is a bit cumbersome in the current version of Zenko. This will change in https://github.com/scality/cli-testing/pull/96 / https://github.com/scality/Zenko/pull/2341 

I suggest keeping this aside for now, and adding the user + tests after that Zenko PR is merged. Will be easier. I tested it end to end on my end.

DarkIsDude (changes requested):
Can you check the CI ?

tests/ctst/features/reporting/StorageUsageReporting.feature
francoisferrand: 👍 (though we can work incrementally, and start introducing "identity" step by step, in the new ones at least...)

tests/ctst/features/reporting/StorageUsageReporting.feature
francoisferrand: we can work incrementally, but:
- we need these extra tests to complete the work; so if we don't do this yet, the topic (and EPIC) stays open for longer.... esp. since most of the work was done so that we could have non-admin users with this capability...
- what is the ETA for that PR/change ? If this is a couple of hours or maybe days, why not ; but if there is no ETA we should really consider an alternative, less optimal, approach (i.e what is the actual cost of this "cumbersome-ness") and/or put more focus on that prerequisite change
- in that case need to "split" the current ticket, so we have a followup (and put it in sprint)

tests/ctst/features/reporting/StorageUsageReporting.feature
```diff
@@ -0,0 +1,49 @@
+Feature: Storage Usage Reporting API
+  The storage usage reporting API allows authorized Keycloak users to retrieve
+  aggregated storage usage metrics across all accounts and locations.
+
+    # Permission tests
+    @2.14.0
+    @PreMerge
+    @StorageUsageReporting
+    Scenario Outline: Storage usage report access control per role
+        When the user retrieves the storage usage report as "<role>"
```
francoisferrand: here we should really mention "persona" (or "keycloak user" at least), as "role" looks like the _IAM role_ and make the scenario very confusing

```suggestion
    Scenario Outline: Storage usage report access control per persona
        When the user retrieves the storage usage report as "<persona>"
```

tests/ctst/features/reporting/StorageUsageReporting.feature
```diff
@@ -0,0 +1,49 @@
+Feature: Storage Usage Reporting API
+  The storage usage reporting API allows authorized Keycloak users to retrieve
+  aggregated storage usage metrics across all accounts and locations.
+
+    # Permission tests
```
francoisferrand: nit: are these comments useful?

```suggestion
```

(same below for `# Content tests`)

tests/ctst/features/reporting/StorageUsageReporting.feature
```diff
@@ -0,0 +1,49 @@
...
+        Then the storage usage report http response code is 200
+        And the storage usage report contains the additional accounts
+
+    @2.14.0
+    @PreMerge
+    @StorageUsageReporting
+    @PrepareStorageUsageReportingScenarios
+    Scenario Outline: Storage usage report returns accurate metrics
+        Given the environment is set up with bucket created, test data uploaded, and count-items ran
+        When the user retrieves the storage usage report as "storage_manager"
```
francoisferrand: nit: maybe not important (or relevant) at the moment, but I have the _feeling_ that it may be more future-proof/reusage to use 2 steps here, to decouple the identity/user setup (will be needed for all routes, basically) from the specific handling of storage report. I.e. something like

```
Given an identity with the "storage_manager" keyclick role
   And the user retrieves the storage usage report
   ...
```

(the first step may be different, like `the user has the "storage_manager" keycloak role` or `an identity with the "storage_manager" persona`... not sure what is best there)

tests/ctst/features/reporting/StorageUsageReporting.feature
```diff
@@ -0,0 +1,49 @@
...
+        Then the storage usage report http response code is 200
```
francoisferrand: nit: `the user retrieves the storage usage report as "storage_manager"` hints at success, which makes this next step kind of redundant (and makes scenario harder to read)

maybe what we need is 
* `the user tries to retrieve the storage usage report` : just issue the command, and require a followup `the http response is 200` (which would be used mostly in the permission test I guess)
* `the user retrieves the storage usage report` : automatically checks for success in the same step

(note also that for better reusability we may want to have a generic `the http response is <result>` instead of making it specific to this API. This may be done simply saving/using a generic "lastHttpResponse" param, so mainly need to introduce a const variable and/or helper function somewhere)

tests/ctst/features/reporting/StorageUsageReporting.feature
```diff
@@ -0,0 +1,49 @@
...
+    @PreMerge
+    @StorageUsageReporting
+    Scenario Outline: Storage usage report access control per role
+        When the user retrieves the storage usage report as "<role>"
+        Then the storage usage report http response code is <expectedStatus>
+
+        Examples:
+          | role             | expectedStatus |
+          | storage_manager  | 200            |
+          | data_consumer    | 403            |
```
francoisferrand: what are really these "values" : the names of keycloak users? the keycloak role? the IAM role?

overal, I think this may actually be the name of the keycloak user, as defined in keycloak_config.json file ?
- we should probably rename these to make it clear what this is this → `storage_manager_identity`, etc...
- the is a fundamental inconsistency here, as we have hard-coded references here while we pass a KeycloakUsername in the Zenko World... To make/keep the tests portable, we will need to either handle some mapping in the world, or just create "our own" identities in the test (and just require a keycloak user to do this setup). Not sure how far we can go in this PR (certainly not the whole cleanup), but worth reviewing and checking we don't introduce more implicit coupling than there already is...

tests/ctst/features/reporting/StorageUsageReporting.feature
delthas: Done — removed both section comments.

tests/ctst/features/reporting/StorageUsageReporting.feature
delthas: Done — renamed to "persona" throughout. Also replaced hardcoded Keycloak usernames with World parameter keys (`StorageManagerUsername`, `DataConsumerUsername`) so the feature file no longer has implicit coupling to `keycloak_config.json`.

tests/ctst/features/reporting/StorageUsageReporting.feature
delthas: Done — added `Given an identity with the "{string}" keycloak persona` step that saves the resolved username from World parameters. All scenarios now have persona setup as a Given precondition, and the When step reads it from World.

tests/ctst/features/reporting/StorageUsageReporting.feature
delthas: Done — split into `tries to retrieve` (permission tests) and `retrieves` (content tests, auto-asserts 200). Added generic `Then the http response code is {int}` in `common/common.ts`, using a `lastHttpResponse` saved key.

tests/ctst/features/reporting/StorageUsageReporting.feature
delthas: Addressed: the Examples values are now World parameter keys (`StorageManagerUsername`, `DataConsumerUsername`) rather than raw Keycloak usernames, and the column is labeled `persona`. The step resolves them via `this.parameters[persona]` with a fallback for local dev.

tests/ctst/features/reporting/StorageUsageReporting.feature
delthas: Addressed incrementally — new step uses "identity" language: `Given an identity with the "{string}" keycloak persona`.

tests/ctst/features/reporting/StorageUsageReporting.feature
delthas: The PR https://github.com/scality/Zenko/pull/2341 already has 2 approves, just a few remaining comments. Planned to be merged by early next week. 

Ticket created at https://scality.atlassian.net/browse/ZENKO-5219 , assigned, added to sprint.

tests/ctst/features/reporting/StorageUsageReporting.feature
delthas: Reverted to raw Keycloak usernames (`storage_manager`, `data_consumer`). Here's why using World parameters doesn't work:

**Two separate user populations exist in Keycloak:**

1. **Base users** (`storage_manager`, `data_consumer`, `storage_account_owner`) — created by importing `keycloak_config.json` during cluster setup. These have the `instanceIds` attribute (set inline in the JSON), which the management API (pensieve) requires for token validation.

2. **CTST users** (`ctst_storage_manager`, `ctst_data_consumer`, etc.) — created by `seedKeycloak.sh` during test setup. These do **not** have `instanceIds` set. They're used for ARWWI (AssumeRoleWithWebIdentity) flows in S3/IAM tests, where `instanceIds` isn't needed.

**World parameters point to the wrong population:**
- `StorageManagerUsername` → `ctst_storage_manager` (no `instanceIds` → 401 from management API)
- `DataConsumerUsername` → `ctst_data_consumer` (same issue)
- `KeycloakUsername` → `storage_manager` (works, but only exists for one user)

There's no World parameter for the base `data_consumer` username.

**The base usernames have no env var source either:**
- `storage_manager` is hardcoded in `keycloak_config.json` line 81. `OIDC_USERNAME` in `end2end.yaml` is a separate duplicate of the same value — it's used for `keycloak_user.json` (admin user template) and the `-norights` user, not for `keycloak_config.json`.
- `data_consumer` and `storage_account_owner` are hardcoded in `keycloak_config.json` with no env var at all.

So there's already duplication between `end2end.yaml` (`OIDC_USERNAME: 'storage_manager'`) and `keycloak_config.json` (`"username": "storage_manager"`). Adding World parameters would just be a third copy of the same hardcoded strings with no actual decoupling.

Properly fixing this would mean either: (a) templating all usernames in `keycloak_config.json` through env vars, or (b) setting `instanceIds` on the `ctst_` users too. Both are out of scope for this PR.

delthas: I think for users we need a proper refactoring, the env vars + json config + ctst thing + cli testing is a bit all over the place. Maybe in a separate PR later?

delthas:
> Can you check the CI ?

CI now passes

DarkIsDude (approved):
Congrats. Thanks for all the refactoring done. LGTM 🙏

tests/ctst/features/reporting/StorageUsageReporting.feature
francoisferrand: > Two separate user populations exist in Keycloak:

not really a reason why we can't use other names : we could change the names in keycloak already.

> So there's already duplication between end2end.yaml (OIDC_USERNAME: 'storage_manager') and keycloak_config.json ("username": "storage_manager"). Adding World parameters would just be a third copy of the same hardcoded strings with no actual decoupling.

that is the broken window effect. Since it's not clean already, there is no point trying to improve (i.e. make the test understandable).

The World parameters is useful for adaptability (and probably not done as well as it should in every case) : the goal is really that the test can be run independently from the setup, i.e. either on a plain/official Artesca or a CI zenko...

- some parameters should be derived/retrieved (e.g. from k8s cluster, based only on the name of the Zenko cr? ; possibly from other parameters)
- some parameters will be needed either way (e.g. credentials...)
- some parameters should be replaced by dynamically created entities (like users 'seeded' by CTST, which should be part of the tests instead of the setup)

True we are not there yet, but this is still the direction we want to go. So we must keep it in mind, and try to move in the right direction - so we don't face a bigger mountain once we have made the first few steps and are ready to proceed.

> Maybe in a separate PR later?

👍
esp. since it is already much cleaner when the text of the scenario/column mention identities and personas.

solution/deps.yaml
```diff
@@ -80,7 +80,7 @@ mongodb-connector:
 pensieve-api:
   sourceRegistry: ghcr.io/scality
   image: pensieve-api
-  tag: 1.9.0
+  tag: 1.9.1
```
francoisferrand, line 83: depending on timing, we may want to bump 1.9.2 and update VERSION (when rebasing before the merge) so we can make the release (c.f. https://scality.atlassian.net/browse/ZENKO-5210)
→ ping me if in doubt how to "optimize" rebuilds

francoisferrand (approved):

PR #2341, opened 2026-03-02
Bump Cli-testing to 1.3.0, update kafka libraries

SylvainSenechal:
Issue: ZENKO-5203

- Bump cli testing to 1.3.0 => Breaking changes around the keycloak seeder, so updated a few things related to that
- Updated node runner to 24
- Updated cucumber version from 10.x to 12.x
- Updated kubernetes client from 0.2 to 1.4...
- Kafka libraries : align all kafka interactions on the new JS native kafka library : platformatic/kafka => drop kafkajs, drop node-rdkafka (and drop the annoying node-gyp that was needed for node-rdkafka, which was always causing weird python warnings when installing). => This new kafka library is compatible with Bunm unlike the previous ones, so later we can investigate if using Bun is worth it (I believe it is 🌚 ), before that we could'nt even try because of this lib

Ended up adding some extra stuff as I was testing my changes in Codespace and found a few stuff bothering me. I think the pr is still not too hard to review.

Can review commit by commit too

### DIFF
--- a/.github/scripts/end2end/run-e2e-ctst.sh
+++ b/.github/scripts/end2end/run-e2e-ctst.sh
@@ -156,13 +156,6 @@ E2E_IMAGE=$E2E_CTST_IMAGE_NAME:$E2E_IMAGE_TAG
 POD_NAME="${ZENKO_NAME}-ctst-tests"
 CTST_VERSION=$(sed 's/.*"cli-testing": ".*#\(.*\)".*/\1/;t;d' ../../../tests/ctst/package.json)
 
-# Configure keycloak
-docker run \
-    --rm \
-    --network=host \
-    "${E2E_IMAGE}" /bin/bash \
-    -c "SUBDOMAIN=${SUBDOMAIN} CONTROL_PLANE_INGRESS_ENDPOINT=${OIDC_ENDPOINT} ACCOUNT=${ZENKO_ACCOUNT_NAME} KEYCLOAK_REALM=${KEYCLOAK_TEST_REALM_NAME} STORAGE_MANAGER=${STORAGE_MANAGER_USER_NAME} STORAGE_ACCOUNT_OWNER=${STORAGE_ACCOUNT_OWNER_USER_NAME} DATA_CONSUMER=${DATA_CONSUMER_USER_NAME} DATA_ACCESSOR=${DATA_ACCESSOR_USER_NAME} /ctst/node_modules/cli-testing/bin/seedKeycloak.sh"; [[ $? -eq 1 ]] && exit 1 || echo 'Keycloak Configured!'
-
 # Grant access to Kube API (insecure, only for testing)
 kubectl create clusterrolebinding serviceaccounts-cluster-admin \
   --clusterrole=cluster-admin \
@@ -178,6 +171,14 @@ kubectl run $POD_NAME \
         --attach=True \
         --image-pull-policy=IfNotPresent \
         --env=TARGET_VERSION=$VERSION  \
+        --env=ACCOUNT=${ZENKO_ACCOUNT_NAME} \
+        --env=STORAGE_MANAGER=${STORAGE_MANAGER_USER_NAME} \
+        --env=STORAGE_ACCOUNT_OWNER=${STORAGE_ACCOUNT_OWNER_USER_NAME} \
+        --env=DATA_CONSUMER=${DATA_CONSUMER_USER_NAME} \
+        --env=DATA_ACCESSOR=${DATA_ACCESSOR_USER_NAME} \
+        --env=SEED_KEYCLOAK_DEFAULT_ROLES=true \
+        --env=KEYCLOAK_HOST=${KEYCLOAK_TEST_HOST} \
+        --env=KEYCLOAK_REALM=${KEYCLOAK_TEST_REALM_NAME} \
         --env=AZURE_BLOB_URL=$AZURE_BACKEND_ENDPOINT  \
         --env=AZURE_QUEUE_URL=$AZURE_BACKEND_QUEUE_ENDPOINT \
         --env=VERBOSE=1 \
--- a/.github/scripts/end2end/run-e2e-ctst.sh
+++ b/.github/scripts/end2end/run-e2e-ctst.sh
@@ -227,5 +227,6 @@ kubectl run $POD_NAME \
     --parallel $PARALLEL_RUNS \
     --retry 3 \
     --retry-tag-filter @Flaky \
+    --format pretty \
     --format junit:/reports/ctst-junit.xml \
     --format html:/reports/report.html
--- a/tests/ctst/Dockerfile
+++ b/tests/ctst/Dockerfile
@@ -20,19 +20,12 @@ WORKDIR /ctst
 RUN apt-get update \
     && apt-get install -y --no-install-recommends \
     apt-utils \
-    python3 \
-    build-essential \
     ssh \
     git \
     curl \
     unzip \
     jq \
     ca-certificates \
-    librdkafka-dev \
-    zlib1g-dev \
-    libssl-dev \
-    libffi-dev \
-    libzstd-dev \
     && rm -rf /var/lib/apt/lists/* \
     && apt-get clean
 
--- a/tests/ctst/Dockerfile
+++ b/tests/ctst/Dockerfile
@@ -6,7 +6,7 @@ ARG DRCTL_TAG=latest
 FROM $SORBET_IMAGE:$SORBET_TAG AS sorbet
 FROM $DRCTL_IMAGE:$DRCTL_TAG AS drctl
 
-FROM node:22.19.0-bookworm-slim
+FROM node:24.14.0-bookworm-slim
 
 ARG AWSCLI_VERSION=2.17.39
 
--- a/tests/ctst/common/common.ts
+++ b/tests/ctst/common/common.ts
@@ -4,7 +4,7 @@ import { CacheHelper, Constants, Identity, IdentityEnum, S3, Utils } from 'cli-t
 import Zenko from 'world/Zenko';
 import { parseGoDuration, safeJsonParse } from './utils';
 import assert from 'assert';
-import { Admin, Kafka } from 'kafkajs';
+import { Admin } from '@platformatic/kafka';
 import {
     createBucketWithConfiguration,
     putMpuObject,
@@ -99,8 +99,24 @@ async function addUserMetadataToObject(this: Zenko, objectName: string | undefin
 async function getTopicsOffsets(topics: string[], kafkaAdmin: Admin) {
     const offsets = [];
     for (const topic of topics) {
-        const partitions: ({ high: string; low: string; })[] =
-            await kafkaAdmin.fetchTopicOffsets(topic);
+        const metadata = await kafkaAdmin.metadata({ topics: [topic] });
+        const partitionCount = metadata.topics.get(topic)?.partitionsCount ?? 0;
+        const partitionIndexes = Array.from({ length: partitionCount }, (_, i) => ({
+            partitionIndex: i,
+            timestamp: BigInt(-2),
+        }));
+        const earliestResult = await kafkaAdmin.listOffsets({
+            topics: [{ name: topic, partitions: partitionIndexes }],
+        });
+        const latestResult = await kafkaAdmin.listOffsets({
+            topics: [{ name: topic, partitions: partitionIndexes.map(p => ({ ...p, timestamp: BigInt(-1) })) }],
+        });
+        const partitions = [];
+        for (let i = 0; i < partitionCount; i++) {
+            const low = earliestResult[0]?.partitions.find(p => p.partitionIndex === i)?.offset ?? BigInt(0);
+            const high = latestResult[0]?.partitions.find(p => p.partitionIndex === i)?.offset ?? BigInt(0);
+            partitions.push({ low: String(low), high: String(high) });
+        }
         offsets.push({ topic, partitions });
     }
     return offsets;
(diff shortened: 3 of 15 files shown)


tests/ctst/package.json
```diff
@@ -37,5 +36,8 @@
   "scripts": {
     "build": "tsc --build tsconfig.json",
     "lint": "eslint ."
+  },
+  "resolutions": {
```
SylvainSenechal, line 41: related to issues newer azure libraries : https://scality.atlassian.net/browse/ZENKO-5213

tests/ctst/steps/utils/utils.ts
SylvainSenechal, line 1: This was an infinite loop, although catched by global test timeout, still nicer to properly control while loops

tests/ctst/steps/utils/kubernetes.ts
SylvainSenechal, line 1: These are all small syntax changes because I did a huge bump of the kubernetes client from version 0.x to 1.4.0

tests/ctst/steps/utils/kafka.ts
```diff
@@ -0,0 +1,70 @@
...
+
+
+/**
+ * Polls the Kafka Connect REST API until the given bucket appears in
+ * at least one connector's MongoDB change-stream pipeline (`$match` →
+ * `ns.coll.$in`).  This ensures the oplog-populator has propagated a
+ * new `putBucketNotificationConfiguration` to the connector before the
+ * test proceeds to trigger events.
+ */ 
+export async function waitForBucketInConnectorPipeline(
```
SylvainSenechal, line 27: Used in bucket notification : Before, we were doing 
- putBucketNotificationConfiguration
- sleep for 10sec and pray the oplog has picked up the change and updated kafka connector.

It was risky, because triggering the notification event before kafka connector was setup would imply the notification would never be sent

Now we detect the kafka connector is really setup.
 
I'm kinda sad though : I thought this would fix bucket notif flaky, but it doesn't, so there are other things to check

tests/ctst/steps/bucket-policies/common.ts
```diff
@@ -499,10 +499,9 @@ Then('the authorization result is correct', function (this: Zenko) {
             assert.strictEqual(this.getResult().stdout?.includes('AccessDenied') ||
                 this.getResult().err?.includes('AccessDenied'), true);
         } else if (action.action === 'HeadObject' || action.action === 'HeadBucket') {
-            // SDK return Unknown errors for HeadObject, but error code from
-            // S3 is correct.
             assert.strictEqual(this.getResult().err?.includes('AccessDenied') ||
-                this.getResult().err?.includes('403'), true);
+                this.getResult().err?.includes('403')||
+                this.getResult().statusCode === 403, true);
```
SylvainSenechal, line 504: For some reason, I got issues with this part not catching the 403 anymore.
Not fully sure why, but with the bump of the cli testing, the version of the sdk has increased quite a lot, from ~3.900 to ~3.100, maybe something changed I don't know

tests/ctst/features/azureArchive.feature
```diff
@@ -260,39 +260,47 @@ Feature: Azure Archive
...
-    Given a "<versioningConfiguration>" bucket
-    And <objectCount> objects "obj" of size <objectSize> bytes
-    And a transition workflow to "e2e-azure-archive" location
-    And that lifecycle is "paused" for the "e2e-azure-archive" location
-    Then the storage class of object "obj-1" must stay "" for <timeout> seconds
-    And the storage class of object "obj-2" must stay "" for <timeout> seconds
-    Given that lifecycle is "resumed" for the "e2e-azure-archive" location
-    Then object "obj-1" should be "transitioning" and have the storage class "e2e-azure-archive"
-    And object "obj-2" should be "transitioning" and have the storage class "e2e-azure-archive"
+# This test is flaky, and doesn't make much sense as it is :
```
SylvainSenechal, line 277: @francoisferrand Let's discuss this. I only commented for now, maybe we wanna keep it ? 

But looking at the test it doesnt make much sense (cf the comment I added in the code). 
But at the same time I feel like I'm missing something, there must be a reason why this test was created

tests/ctst/common/hooks.ts
```diff
@@ -16,6 +16,10 @@ import {
     cleanupAccount,
 } from './utils';
 
+import 'cli-testing/hooks/KeycloakSetup';
```
SylvainSenechal, line 20: We have BeforeAll in Cli-testing, no import => they are not run
Imo should be in this repo but its a story for another day

.github/scripts/end2end/run-e2e-ctst.sh
```diff
@@ -156,13 +156,6 @@ E2E_IMAGE=$E2E_CTST_IMAGE_NAME:$E2E_IMAGE_TAG
 POD_NAME="${ZENKO_NAME}-ctst-tests"
 CTST_VERSION=$(sed 's/.*"cli-testing": ".*#\(.*\)".*/\1/;t;d' ../../../tests/ctst/package.json)
 
-# Configure keycloak
```
SylvainSenechal, line 159: Seed keycloak is now done in a before all from Cli-testing

tests/ctst/common/common.ts
```diff
@@ -4,7 +4,7 @@ import { CacheHelper, Constants, Identity, IdentityEnum, S3, Utils } from 'cli-t
 import Zenko from 'world/Zenko';
 import { safeJsonParse } from './utils';
 import assert from 'assert';
-import { Admin, Kafka } from 'kafkajs';
+import { Admin } from '@platformatic/kafka';
```
SylvainSenechal, line 7: Switched the kafka library
Before we were using kafkajs, and node-rdkafka. 

These internally relied on librdkafka, a C library. Because of that, we needed to setup quite a few extra stuff to make them work : node-gyp (which was throwing weird python errors when doing yarn install in the Codespace), and also need a bunch of things to be installed in the dockerfile (you'll see later in the pr, librd-XXX). 
Now this library is pure js, it removes some dependency, and also it will be compatible with Bun if we wanna use it later

tests/ctst/steps/azureArchive.ts
```diff
@@ -302,12 +302,13 @@ Then('blob for object {string} must be rehydrated',
     async function (this: Zenko, objectName: string) {
         const tarName = await isObjectRehydrated(this, objectName);
         assert(tarName);
-        await AzureHelper.sendBlobCreatedEventToQueue(
+        const sent = await AzureHelper.sendBlobCreatedEventToQueue(
```
SylvainSenechal, line 305: Lost crazy amount of time on this : 
sendBlobCreatedEventToQueue was erroring, the error wasn't thrown but instead the function return a boolean that was never checked -_-

tests/ctst/common/common.ts
francoisferrand: notes:
- such comments would be better in the PR comment, so reviewers know before even reading the code
- would be good to keep such change in a separate PR (even if it means stacking multiple PRs) - not saying to change it _now_ (too late), but as a goal for next changes

tests/ctst/common/common.ts
```diff
@@ -99,8 +99,24 @@ async function addUserMetadataToObject(this: Zenko, objectName: string | undefin
...
+        const earliestResult = await kafkaAdmin.listOffsets({
+            topics: [{ name: topic, partitions: partitionIndexes }],
+        });
+        const latestResult = await kafkaAdmin.listOffsets({
+            topics: [{ name: topic, partitions: partitionIndexes.map(p => ({ ...p, timestamp: BigInt(-1) })) }],
+        });
+        const partitions = [];
+        for (let i = 0; i < partitionCount; i++) {
+            const low = earliestResult[0]?.partitions.find(p => p.partitionIndex === i)?.offset ?? BigInt(0);
+            const high = latestResult[0]?.partitions.find(p => p.partitionIndex === i)?.offset ?? BigInt(0);
```
francoisferrand, line 117: partitions are lists, right? So searching is not efficient...
- are partitions "ordered" predictably?
- do earliest and latest have the same order?
- ...could we not just do a "zip" merge, something like:

```
for (let i = 0; i < partitionCount; i++) {
   partitions.push({ low: earliestResult[0]?.partitions[I], high: latestResult[0]?.partitions[I] });
```

tests/ctst/steps/notifications.ts
```diff
@@ -327,30 +328,70 @@ Then('notifications should be enabled for {string} event in destination {int}',
 
 Then('i should {string} a notification for {string} event in destination {int}',
     async function (this: Zenko, receive: string, notificationType: string, destination: number) {
+        const topic = this.getSaved<NotificationDestination[]>('notificationDestinations')[destination].topic;
+        const hosts = this.getSaved<NotificationDestination[]>('notificationDestinations')[destination].hosts;
```
francoisferrand:
```suggestion
        const { topic, hosts } = this.getSaved<NotificationDestination[]>('notificationDestinations')[destination];
```

tests/ctst/steps/notifications.ts
```diff
@@ -327,30 +328,70 @@ Then('notifications should be enabled for {string} event in destination {int}',
...
+                    if (Date.now() - startTime >= KAFKA_TESTS_TIMEOUT) {
+                        break;
+                    }
+                    this.logger.debug('Kafka message received', {
+                        topic: msg.topic,
+                        partition: msg.partition,
+                        offset: msg.offset?.toString(),
+                        value: msg.value,
+                    });
+                    try {
```
francoisferrand, line 369: nested try blocks (3 here!) are kind of hard to read, should we not introduce helper functions?

tests/ctst/steps/notifications.ts
```diff
@@ -327,30 +328,70 @@ Then('notifications should be enabled for {string} event in destination {int}',
...
-            (msg: Message) => {
-                try {
-                    const notification = (JSON.parse(msg.value?.toString() as string
-                        || '{Records:[]}') as { Records: Notification[] }).Records[0] ;
-                    const bucketNameMatches = this.getSaved<string>('bucketName') === notification?.s3.bucket.name;
-                    const objectNameMatches = this.getSaved<string>('objectName') === notification?.s3.object.key;
-                    const eventTypeMatches = notificationType === notification?.eventName;
-                    if (bucketNameMatches && objectNameMatches && eventTypeMatches) {
-                        return true;
+            try {
```
francoisferrand, line 358: instead of integrating the lib here, should we modify KafkaHelper directly in cli-testing to change the kafka lib?
or at least we need to remove that code (and dependency) from cli-testing?

francoisferrand (approved):

tests/ctst/steps/notifications.ts
SylvainSenechal: I removed the kafka helper from cli-testing, it was only used once in Zenko, no need for such a helper, and this way, no more node-rdkafka in cli testing

tests/ctst/cucumber.config.cjs
```diff
@@ -5,8 +5,7 @@ module.exports = {
         require: ['steps/**/*.ts', 'common/**/*.ts', 'world/**/*.ts'],
         paths: ['features/**/*.feature'],
         format: [
-            'progress-bar',
-            '@cucumber/pretty-formatter',
+            'pretty',
```
francoisferrand, line 8: why are we removing `@cucumber/pretty-formatter` ?
seems a good option to auto-format .freature files?

SylvainSenechal: Pretty formatter is archived (https://github.com/cucumber/cucumber-js-pretty-formatter?tab=readme-ov-file), It's integrated directly in cucumber now (thats why i removed pretty formatter in cli-testing), and pretty formatter is now just "pretty" : https://github.com/cucumber/cucumber-js/blob/main/docs/formatters.md

tests/ctst/common/common.ts
SylvainSenechal: I think we don't have guarantee about the order, but the list should be small enough that in practice this find is not too impactfull

tests/ctst/Dockerfile
```diff
@@ -4,7 +4,7 @@ ARG DRCTL_TAG=latest
 FROM ghcr.io/scality/sorbet:$SORBET_TAG AS sorbet
 FROM ghcr.io/scality/zenko-drctl:$DRCTL_TAG AS drctl
 
-FROM node:22.19.0-bookworm-slim
+FROM node:24-bookworm-slim
```
maeldonn: Shouldn't we pin the version like before ?

tests/ctst/features/azureArchive.feature
maeldonn: Don't forget to resolve this before merging 😉

tests/ctst/steps/bucket-policies/common.ts
```diff
@@ -499,10 +499,9 @@ Then('the authorization result is correct', function (this: Zenko) {
             assert.strictEqual(this.getResult().stdout?.includes('AccessDenied') ||
                 this.getResult().err?.includes('AccessDenied'), true);
         } else if (action.action === 'HeadObject' || action.action === 'HeadBucket') {
-            // SDK return Unknown errors for HeadObject, but error code from
-            // S3 is correct.
             assert.strictEqual(this.getResult().err?.includes('AccessDenied') ||
-                this.getResult().err?.includes('403'), true);
+                this.getResult().err?.includes('403')||
```
maeldonn, line 503: Do we really need to keep this condition ?

maeldonn (approved):
LGTM

tests/ctst/steps/bucket-policies/common.ts
SylvainSenechal: I am not 100% sure anymore tbh, might have been something that failed while i was working, but probably only need one of the two now.
But a/b testing by launching the ci again with and without i too long to just fix this

tests/ctst/common/common.ts
delthas: Thanks for the kafka change!

tests/ctst/Dockerfile
SylvainSenechal: yeah gonna use 24.14

PR #2348, opened 2026-03-12
Update GetObjectAttributes CTST tests to use aws SDK client 

maeldonn:
Issue: ZENKO-5208

### DIFF
--- a/tests/ctst/package.json
+++ b/tests/ctst/package.json
@@ -27,7 +27,7 @@
   },
   "devDependencies": {
     "@aws-sdk/client-iam": "^3.901.0",
-    "@aws-sdk/client-s3": "^3.901.0",
+    "@aws-sdk/client-s3": "^3.931.0",
     "@aws-sdk/client-sts": "^3.901.0",
     "@eslint/compat": "^1.1.1",
     "@scality/cloudserverclient": "1.0.7",
--- a/tests/ctst/package.json
+++ b/tests/ctst/package.json
@@ -30,6 +30,7 @@
     "@aws-sdk/client-s3": "^3.901.0",
     "@aws-sdk/client-sts": "^3.901.0",
     "@eslint/compat": "^1.1.1",
+    "@scality/cloudserverclient": "1.0.7",
     "eslint": "^9.9.1",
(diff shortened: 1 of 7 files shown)


tests/ctst/package.json
```diff
@@ -30,6 +30,7 @@
     "@aws-sdk/client-s3": "^3.901.0",
     "@aws-sdk/client-sts": "^3.901.0",
     "@eslint/compat": "^1.1.1",
+    "@scality/cloudserverclient": "github:scality/cloudserverclient#improvement/CLDSRVCLT-10/user-metadata",
```
maeldonn:
```suggestion
    "@scality/cloudserverclient": "^1.0.5",
```

tests/ctst/world/Zenko.ts
```diff
@@ -924,8 +925,40 @@ export default class Zenko extends World<ZenkoWorldParameters> {
         }
     }
 
+    createS3Client(): S3Client {
```
francoisferrand, line 928: don't we already have this in CTST ? can't we reuse CTST's s3client?

tests/ctst/world/Zenko.ts
```diff
@@ -924,8 +925,40 @@ export default class Zenko extends World<ZenkoWorldParameters> {
...
+            const result = await client.send(command);
+            this.setResult({ stdout: JSON.stringify(result), err: null, statusCode: 200 });
+        } catch (err: unknown) {
+            if (err instanceof S3ServiceException) {
+                this.setResult({ stdout: '', err: err.name, statusCode: err.$metadata.httpStatusCode || 403 });
+            } else {
+                throw err;
+            }
+        } finally {
+            client.destroy();
```
francoisferrand: nit: creating and destroying the client each time may be slow, should we not keep it throughout?
(may not be so important since it is used only ni a few tests, but if/when generalized may prevent reusing resources?)

francoisferrand (approved):

tests/ctst/world/Zenko.ts
maeldonn: No we can't for now, it would require to update cli-testing.

benzekrimaha (approved):
LGTM

PR #2349, opened 2026-03-12
Detect unused Cucumber step definitions in CI

delthas:
## Summary

- Add a CI check that runs `cucumber-js --dry-run --format usage` to detect unused step definitions, and fails the `lint-and-build-ctst` job if any are found
- Remove 4 unused step definitions flagged by this check

## How it works

`cucumber-js --dry-run` loads all step definitions and cross-references them against every feature file without actually executing any steps. The `--format usage` output lists each step definition alongside its usage count, marking unreferenced ones as `UNUSED`. The new `check-unused-steps.sh` script greps for `UNUSED` in this output and exits non-zero if any are found.

## Why this branch

This approach requires `cucumber-js --dry-run` to load all step definition files. On the main Zenko branches, this fails because `node-rdkafka` (a native C++ addon) is imported transitively by some step files, and it requires system libraries (`librdkafka`) that aren't available in a plain `yarn install` environment. On this branch (`improvement/ZENKO-5203`), `node-rdkafka` has been replaced with `@platformatic/kafka` (a pure JS implementation), so the dry-run loads cleanly without native dependencies.

## Alternative approach

An alternative would be to detect unused steps at test runtime rather than as a static CI check — for example by hooking into Cucumber's `AfterAll` to inspect which step definitions were matched. However, this would only catch unused steps when the full test suite runs, not at lint time, and would be harder to enforce as a gate. The dry-run approach is simpler and catches issues earlier in the pipeline.


tests/ctst/package.json
```diff
@@ -35,7 +35,8 @@
   },
   "scripts": {
     "build": "tsc --build tsconfig.json",
-    "lint": "eslint ."
+    "lint": "eslint .",
+    "unused-steps": "bash check-unused-steps.sh"
```
francoisferrand: script not needed, this should do exactly the same:

```suggestion
    "unused-steps": "! cucumber-js --config cucumber.config.cjs --dry-run --format usage --parallel 1 2>&1 | grep UNUSED"
```

tests/ctst/check-unused-steps.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+set -eu
+
+# Detect unused Cucumber step definitions by running a dry-run
+# and checking for UNUSED entries in the usage report.
+# Exit code: 0 if none found, 1 if unused steps exist.
+
+unused=$(cucumber-js --config cucumber.config.cjs --dry-run --format usage --parallel 1 2>&1 | grep UNUSED || true)
```
francoisferrand: can't we enable more parallelism (to make it faster, always! 😀) ?
or at least just keep the default, for conciseness....

francoisferrand (approved):

maeldonn (approved):
Could you squash or fixup these changes into the relevant commits? Keeping the history clean makes it much easier to follow the logic during future audits.

francoisferrand:
must be rebased, this is not up-to-date with improvement/ZENKO-5203

delthas:
> must be rebased, this is not up-to-date with improvement/[ZENKO-5203](https://scality.atlassian.net/browse/ZENKO-5203)

Yes, waiting for [that PR](https://github.com/scality/Zenko/pull/2341) to be merged/stabilized

[ZENKO-5203]: https://scality.atlassian.net/browse/ZENKO-5203?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2350, opened 2026-03-13
Add Fluent Bit DaemonSet to CI for persistent log collection

delthas:
## Summary

- Deploy a Fluent Bit DaemonSet on all kind nodes to capture container logs in real-time
- Logs are written to the shared `/data` hostPath volume and included in the `logs-volumes.tgz` artifact
- Lightweight: ~50m CPU / 64Mi RAM per node, estimated ~100-200 MB compressed output

## Why this is needed

When the Zenko operator reconciles, it triggers rolling restarts that replace pods. Kubelet deletes container logs from `/var/log/pods/` on pod deletion, and `kind export logs` only captures logs from pods alive at export time. **Every log from before the restart is permanently lost.**

### Concrete example

While investigating [ctst-end2end-sharded #66912102578](https://github.com/scality/Zenko/actions/runs/23023382868/job/66912102578), all 11 Azure archive tests failed because objects were stuck at `transition-in-progress: true`. The transition pipeline (lifecycle → cloudserver → Kafka → sorbet-fwd → sorbet-azure) was broken, but we could not determine where because:

- The operator's reconciliation loop triggered a mass rolling restart at 07:20, replacing ALL data service pods
- `kind export logs` ran at 08:22, capturing only the post-restart pod logs
- All container logs from the 06:36–06:38 failure window were gone
- We could not determine whether sorbet-fwd crashed (it has a known pattern of committing Kafka offsets during shutdown, permanently losing in-flight messages), whether sorbet-azure was unreachable, or whether internal-cloudserver had the wrong config

With Fluent Bit tailing logs continuously, all of these would have been preserved regardless of pod lifecycle.

## Changes

- **`.github/scripts/end2end/configs/fluentbit.yaml`** — ConfigMap + DaemonSet manifest. Tails `/var/log/containers/*.log`, writes raw CRI log lines (with timestamps) to `/data/fluentbit-logs/`, one file per container instance
- **`.github/scripts/end2end/install-kind-dependencies.sh`** — Deploys Fluent Bit early (before nginx-controller) so it starts capturing logs before any other workloads
- **`.github/actions/archive-artifacts/action.yaml`** — Copies Fluent Bit output into `kind-logs/fluentbit-logs/` before tarring

Issue: [ZENKO-5216](https://scality.atlassian.net/browse/ZENKO-5216)

[ZENKO-5216]: https://scality.atlassian.net/browse/ZENKO-5216?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/actions/archive-artifacts/action.yaml
+++ b/.github/actions/archive-artifacts/action.yaml
@@ -74,6 +74,12 @@ runs:
         kubectl get zenkodrsink -A -o yaml > /tmp/artifacts/data/${STAGE}/kind-logs/all-zenkodrsinks.log || true
         kind export logs /tmp/artifacts/data/${STAGE}/kind-logs/kind-export || true
         cp -r /artifacts/data/fluentbit-logs /tmp/artifacts/data/${STAGE}/kind-logs/fluentbit-logs 2>/dev/null || true
+        for container_log in /tmp/artifacts/data/${STAGE}/kind-logs/kind-export/*/containers/*.log; do
+          [ -f "$container_log" ] || continue
+          fb_file="/tmp/artifacts/data/${STAGE}/kind-logs/fluentbit-logs/$(basename "$container_log")"
+          [ -f "$fb_file" ] || continue
+          ln -sf "../../../fluentbit-logs/$(basename "$container_log")" "$container_log"
+        done
         tar zcvf /tmp/artifacts/${{ github.sha }}-${STAGE}-logs-volumes.tgz /tmp/artifacts/data/${STAGE}/kind-logs
       env:
         STAGE: ${{ inputs.stage }}
--- a/.github/actions/archive-artifacts/action.yaml
+++ b/.github/actions/archive-artifacts/action.yaml
@@ -73,6 +73,7 @@ runs:
         kubectl get zenkodrsource -A -o yaml > /tmp/artifacts/data/${STAGE}/kind-logs/all-zenkodrsources.log || true
         kubectl get zenkodrsink -A -o yaml > /tmp/artifacts/data/${STAGE}/kind-logs/all-zenkodrsinks.log || true
         kind export logs /tmp/artifacts/data/${STAGE}/kind-logs/kind-export || true
+        cp -r /artifacts/data/fluentbit-logs /tmp/artifacts/data/${STAGE}/kind-logs/fluentbit-logs 2>/dev/null || true
         tar zcvf /tmp/artifacts/${{ github.sha }}-${STAGE}-logs-volumes.tgz /tmp/artifacts/data/${STAGE}/kind-logs
       env:
         STAGE: ${{ inputs.stage }}
--- a/.github/scripts/end2end/configs/fluentbit.yaml
+++ b/.github/scripts/end2end/configs/fluentbit.yaml
@@ -1,0 +1,77 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: fluent-bit-config
+data:
+  fluent-bit.conf: |
+    [SERVICE]
+        Flush        5
+        Log_Level    warn
+
+    [INPUT]
+        Name         tail
+        Tag          kube.*
+        Path         /var/log/containers/*.log
+        DB           /var/lib/fluent-bit/flb_kube.db
+        Mem_Buf_Limit 16MB
+        Skip_Long_Lines On
+
+    [OUTPUT]
+        Name         file
+        Match        *
+        Path         /data/fluentbit-logs/
+        Mkdir        true
+        Format       template
+        Template     {log}
+---
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: fluent-bit
+  labels:
+    app: fluent-bit
+spec:
+  selector:
+    matchLabels:
+      app: fluent-bit
+  template:
+    metadata:
+      labels:
+        app: fluent-bit
+    spec:
+      tolerations:
+        - key: node-role.kubernetes.io/control-plane
+          effect: NoSchedule
+      containers:
+        - name: fluent-bit
+          image: cr.fluentbit.io/fluent/fluent-bit:4.0
+          resources:
+            requests:
+              cpu: 50m
+              memory: 64Mi
+            limits:
+              cpu: 200m
+              memory: 128Mi
+          volumeMounts:
+            - name: varlog
+              mountPath: /var/log
+              readOnly: true
+            - name: fluentbit-output
+              mountPath: /data/fluentbit-logs
+            - name: fluentbit-db
+              mountPath: /var/lib/fluent-bit
+            - name: config
+              mountPath: /fluent-bit/etc
+      volumes:
+        - name: varlog
+          hostPath:
+            path: /var/log
+        - name: fluentbit-output
+          hostPath:
+            path: /data/fluentbit-logs
+            type: DirectoryOrCreate
+        - name: fluentbit-db
+          emptyDir: {}
+        - name: config
+          configMap:
+            name: fluent-bit-config
--- a/.github/scripts/end2end/configs/fluentbit.yaml
+++ b/.github/scripts/end2end/configs/fluentbit.yaml
@@ -10,7 +10,8 @@ data:
 
     [INPUT]
         Name         tail
-        Tag          kube.*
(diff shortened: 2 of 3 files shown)


.github/actions/archive-artifacts/action.yaml
```diff
@@ -73,6 +73,7 @@ runs:
         kubectl get zenkodrsource -A -o yaml > /tmp/artifacts/data/${STAGE}/kind-logs/all-zenkodrsources.log || true
         kubectl get zenkodrsink -A -o yaml > /tmp/artifacts/data/${STAGE}/kind-logs/all-zenkodrsinks.log || true
         kind export logs /tmp/artifacts/data/${STAGE}/kind-logs/kind-export || true
+        cp -r /artifacts/data/fluentbit-logs /tmp/artifacts/data/${STAGE}/kind-logs/fluentbit-logs 2>/dev/null || true
```
SylvainSenechal, line 76: Are we gonna have 2 folders with similar data now ?

we already have thorough logs in the artifacts, aren't the new ones in /data/fluentbit-logs basically the same (with some extra data regarding the pod that were deleted during the tests)

delthas: The two folders serve different purposes and have minimal overlap:

**`kind-export/`** — node-level diagnostics from `kind export logs`, organized per node:
```
kind-export/
├── docker-info.txt
├── kind-version.txt
├── kind-worker/
│   ├── journal.log              ← systemd journal
│   ├── kubelet.log              ← kubelet logs
│   ├── containerd.log           ← container runtime logs
│   ├── images.log               ← pulled images list
│   ├── inspect.json             ← docker inspect of the node
│   ├── kubernetes-version.txt
│   ├── serial.log
│   ├── containers/              ← container logs (flat, one per container)
│   │   ├── cert-manager-..._cert-manager-controller-<id>.log
│   │   ├── data-db-mongodb-sharded-configsvr-0_default_mongodb-<id>.log
│   │   └── ...  (118 files — only pods alive at export time)
│   └── pods/                    ← structured per pod, includes rotated logs
│       └── default_end2end-backbeat-lifecycle-producer-..._<uid>/
│           └── backbeat-lifecycle-producer/
│               ├── 0.log                       ← current log
│               ├── 0.log.20260313-130753       ← rotated
│               ├── 0.log.20260313-130253.gz    ← rotated + compressed
│               └── 0.log.20260313-125753.gz
├── kind-worker2/
│   └── (same structure)
└── kind-control-plane/
    └── (same structure)
```

**`fluentbit-logs/`** — complete container log history, one file per container instance:
```
fluentbit-logs/
├── kube.var.log.containers.end2end-cold-sorbet-azure-...-gml8f_..._cold-sorbet-azure-<id>.log
├── kube.var.log.containers.end2end-cold-sorbet-azure-...-642bc_..._cold-sorbet-azure-<id>.log
├── ...  (797 files — every container instance that ever ran)
```

The container logs in `kind-export/*/containers/` and `kind-export/*/pods/` are a **point-in-time snapshot**: only pods alive at export time. Fluentbit tails logs continuously from the moment the cluster starts, so it also captures pods that were deleted during the run (rolling restarts, operator reconciliation, etc.).

Concrete example from this CI run: `cold-sorbet-azure` (the azure archive backend) was replaced 23 times during operator reconciliation. kind-export captured **1** instance, fluentbit captured all **23**. Same for `ops-scuba-external-api`: 16 pod replacements in fluentbit, 1 in kind-export.

The overlap is only the ~118 containers alive at the end. The other ~679 files in fluentbit are genuinely new data that we couldn't access before. The node-level files (journal, kubelet, containerd, images, inspect) are only in kind-export and not captured by fluentbit at all.

delthas:
Pushing a tentative fix to make sure the log files have the same name. Let's see how it goes.

delthas:
Working fine. Ready.

.github/actions/archive-artifacts/action.yaml
francoisferrand: > The overlap is only the ~118 containers alive at the end. The other ~679 files in fluentbit are genuinely new data that we couldn't access before

the overlap is "only" these 118 containers, but still there is no point extracting and storing these extra 118 container logs: since we would now gather them already through fluentbit

francoisferrand:
Seems fine for the CI use case: but I wonder if this is appropriate for the local (or codespace) cases, where platform may run for much longer and where space may be kind of a premium.

Should this use of fluentbit be configurable/optional? Or do we estimate impact will still be acceptable in these context?

.github/actions/archive-artifacts/action.yaml
delthas: I'd like to keep the 118 contianer logs in their place for ksnap to parse them. But I guess I can try replacing them with symlinks to the fluentd logs.

francoisferrand: so we should drop `kind export logs` ?

francoisferrand:
> Seems fine for the CI use case: but I wonder if this is appropriate for the local (or codespace) cases, where platform may run for much longer and where space may be kind of a premium.
> 
> Should this use of fluentbit be configurable/optional? Or do we estimate impact will still be acceptable in these context?

@delthas did you see this comment?

.github/actions/archive-artifacts/action.yaml
delthas: No — `kind export logs` still provides unique per-node diagnostics that fluentbit doesn't capture:

- `journal.log` — systemd journal
- `kubelet.log` — kubelet logs
- `containerd.log` — container runtime logs
- `images.log` — list of pulled images
- `inspect.json` — docker inspect of the node
- `serial.log`
- `kubernetes-version.txt`
- `pods/` directory — some containers are not included in the fluentbit output

The latest commit replaces the ~104 overlapping container logs (in `kind-export/*/containers/`) with symlinks to the fluentbit files, so we avoid the duplication while keeping the paths in place for ksnap.

delthas:
> Seems fine for the CI use case: but I wonder if this is appropriate for the local (or codespace) cases, where platform may run for much longer and where space may be kind of a premium.
> 
> Should this use of fluentbit be configurable/optional? Or do we estimate impact will still be acceptable in these context?

Addressed. fluentbit will only be deployed in CI (which makes sense anyway because we are not running the archive artifacts step in local/devcontainer).

delthas:
Ready for review. Symlinks resolve correctly.

.github/actions/archive-artifacts/action.yaml
SylvainSenechal: I got used to using the data from kind export logs, and maybe it's nice to keep it for debugging sometimes because I'm worried the new folder with fluent bit is gonna be extra large with 1gb tar file and too many files to be efficient, I would probably used it to debug when the existing tar is not sufficient

.github/actions/archive-artifacts/action.yaml
```diff
@@ -74,6 +74,12 @@ runs:
         kubectl get zenkodrsink -A -o yaml > /tmp/artifacts/data/${STAGE}/kind-logs/all-zenkodrsinks.log || true
         kind export logs /tmp/artifacts/data/${STAGE}/kind-logs/kind-export || true
         cp -r /artifacts/data/fluentbit-logs /tmp/artifacts/data/${STAGE}/kind-logs/fluentbit-logs 2>/dev/null || true
+        for container_log in /tmp/artifacts/data/${STAGE}/kind-logs/kind-export/*/containers/*.log; do
```
SylvainSenechal, line 77: Wait so, with this, can I still download individual parts of the archive, o rdo i need to download everything
Usually I only download this : 

Image

delthas: This will still work. It's just that some files inside that archive will point to other files inside that archive.

.github/actions/archive-artifacts/action.yaml
delthas: Yes, the kind folder will still contain the same logs (though some will be symlinked), and fluentbit contains the entire export which is more a machine-readable thing.

francoisferrand (approved):

SylvainSenechal (approved):

PR #2352, opened 2026-03-13
Fix Go duration parsing in kafkaCleaner test

delthas:
## Summary

The `kafkaCleaner` end-to-end test (`features/zzz.kafkaCleaner.feature`) reads
`KafkaCleanerInterval` from the Zenko CR's `spec.kafkaCleaner.interval` field,
which is a Go duration string (e.g. `"1m"`). The test used `parseInt("1m")` to
convert this to seconds, but `parseInt` stops at the first non-numeric character
and silently returned `1` instead of `60`.

This caused the test to compute a timeout of ~60s (`1 × 6000ms × 10`) instead
of the intended ~600s (`60 × 6000ms × 10`), giving the kafkacleaner only **one**
cleaning cycle to process all topics rather than ten. Under normal conditions one
cycle was often enough, but when operator reconciliation triggered topic
recreation mid-run, the kafkacleaner needed several cycles to catch up — and the
test failed with:

```
AssertionError: Kafka cleaner did not clean the topics within the expected time
```

### Fix

Replace `parseInt` with a `parseDurationToSeconds` utility that properly handles
Go-style duration strings — including compound durations (`"2h45m"`), fractional
values (`"1.5s"`), and all standard time units (`ns`, `us`/`µs`, `ms`, `s`, `m`, `h`).

For the current `"1m"` value, this correctly returns `60` seconds, restoring the
intended 10-cycle timeout window.

Issue: ZENKO-5218

### DIFF
--- a/tests/ctst/common/common.ts
+++ b/tests/ctst/common/common.ts
@@ -2,7 +2,7 @@ import { ListObjectVersionsOutput } from '@aws-sdk/client-s3';
 import { Given, setDefaultTimeout, Then, When } from '@cucumber/cucumber';
 import { CacheHelper, Constants, Identity, IdentityEnum, S3, Utils } from 'cli-testing';
 import Zenko from 'world/Zenko';
-import { safeJsonParse } from './utils';
+import { parseGoDuration, safeJsonParse } from './utils';
 import assert from 'assert';
 import { Admin, Kafka } from 'kafkajs';
 import {
@@ -300,7 +300,7 @@ Then('i {string} be able to add user metadata to object {string}',
 
 Then('kafka consumed messages should not take too much place on disk', { timeout: -1 },
     async function (this: Zenko) {
-        const kfkcIntervalSeconds = parseInt(this.parameters.KafkaCleanerInterval);
+        const kfkcIntervalSeconds = parseGoDuration(this.parameters.KafkaCleanerInterval);
         const checkInterval = kfkcIntervalSeconds * (1000 + 5000);
 
         const timeoutID = setTimeout(() => {
--- a/tests/ctst/common/utils.ts
+++ b/tests/ctst/common/utils.ts
@@ -100,6 +100,32 @@ export const s3FunctionExtraParams: { [key: string]: Record<string, unknown>[] }
     }],
 };
 
+/**
+ * Parses a duration string in Go's time.ParseDuration format
+ * (https://pkg.go.dev/time#ParseDuration) and returns the equivalent in seconds.
+ * @param {string} duration - the duration string to parse (e.g. "1h30m", "500ms")
+ * @return {number} - the duration in seconds
+ */
+export function parseGoDuration(duration: string): number {
+    const units: Record<string, number> = {
+        ns: 1e-9, us: 1e-6, µs: 1e-6, ms: 1e-3, s: 1, m: 60, h: 3600,
+    };
+    let remaining = duration;
+    if (remaining.length === 0) {
+        throw new Error(`Invalid duration: "${duration}"`);
+    }
+    let totalSeconds = 0;
+    while (remaining.length > 0) {
+        const match = remaining.match(/^(\d+(?:\.\d*)?)(ns|us|µs|ms|s|m|h)/);
+        if (!match) {
+            throw new Error(`Invalid duration: "${duration}" (unparsed: "${remaining}")`);
+        }
+        totalSeconds += parseFloat(match[1]) * units[match[2]];
+        remaining = remaining.slice(match[0].length);
+    }
+    return totalSeconds;
+}
+
 export function safeJsonParse<T>(jsonString: string): { ok: boolean, result: T | null, error?: Error | null } {
     let result: T;
     try {
--- a/tests/ctst/common/parseGoDuration.test.ts
+++ b/tests/ctst/common/parseGoDuration.test.ts
@@ -1,0 +1,34 @@
+import assert from 'assert';
+import { parseGoDuration } from './utils';
+
+const cases: [string, number][] = [
+    ['1m', 60],
+    ['2h', 7200],
+    ['30s', 30],
+    ['2h45m', 9900],
+    ['500ms', 0.5],
+    ['1.5s', 1.5],
+    ['1h30m10s', 5410],
+    ['100ns', 1e-7],
+    ['10us', 1e-5],
+    ['10µs', 1e-5],
+    ['0s', 0],
+];
+
+for (const [input, expected] of cases) {
+    const result = parseGoDuration(input);
+    assert.strictEqual(
+        Math.abs(result - expected) < 1e-12, true,
+        `parseGoDuration("${input}") = ${result}, expected ${expected}`,
+    );
+}
+
+const invalid = ['', 'abc', '1x', '5h 3m', '1', 'm', ' 1m'];
+for (const input of invalid) {
+    assert.throws(
+        () => parseGoDuration(input),
+        { message: /Invalid duration/ },
+        `parseGoDuration("${input}") should throw`,
+    );
+}
+


SylvainSenechal (approved):
Hey I'm approving just some thoughts : 
This new helper function is kinda ugly and will probably only be used once, 
- I was thinking we could almost just drop the "KafkaCleanerInterval" parameters and just give that function 300sec to timeout.
- If you do keep the parseDurationSeconds, consider a few documentation example of what "duration: string" => result this function would process



More importantly : In that codebaase, we have Zenko.yaml, and it's got the kafkaCleaner.interval set to 1m, considering that all the tests we've got probably don't produce that many messages (maybe a few hundreds), and that the issue was probably situation where that kafka cleaner would start a few seconds before the wrong 60sec timeout instead of 600sec that we had, I would suggest lowering the value, somewhere between 15s and 30

delthas:
> Hey I'm approving just some thoughts : This new helper function is kinda ugly and will probably only be used once,
> 
>     * I was thinking we could almost just drop the "KafkaCleanerInterval" parameters and just give that function 300sec to timeout.
> 
>     * If you do keep the parseDurationSeconds, consider a few documentation example of what "duration: string" => result this function would process
> 
> 
> More importantly : In that codebaase, we have Zenko.yaml, and it's got the kafkaCleaner.interval set to 1m, considering that all the tests we've got probably don't produce that many messages (maybe a few hundreds), and that the issue was probably situation where that kafka cleaner would start a few seconds before the wrong 60sec timeout instead of 600sec that we had, I would suggest lowering the value, somewhere between 15s and 30

Right. Testing with 15s, timeout of 5 minutes. Let's see how it goes.

.github/scripts/end2end/configs/zenko.yaml
```diff
@@ -54,7 +54,7 @@ spec:
         size: 20Gi
         storageClassName: "standard"
   kafkaCleaner:
-    interval: 1m
+    interval: 15s
```
francoisferrand: Why do we want to reduce from 1m to 15s ?

tests/ctst/common/common.ts
```diff
@@ -298,12 +298,12 @@ Then('i {string} be able to add user metadata to object {string}',
 
 Then('kafka consumed messages should not take too much place on disk', { timeout: -1 },
     async function (this: Zenko) {
-        const kfkcIntervalSeconds = parseInt(this.parameters.KafkaCleanerInterval);
-        const checkInterval = kfkcIntervalSeconds * (1000 + 5000);
+        const checkIntervalMs = 30_000;
```
francoisferrand: Why hardcode the value here?
This creates coupling with the CR, which
- makes it harder to maintain
- will further prevent running the tests against any zenko or artesca

hence the solution that was put in place to "derive" the timeout from the behavior configured.

Why not go with the solution you describe in the PR title, and simply parse the "unit"?

francoisferrand (changes requested):

.github/scripts/end2end/configs/zenko.yaml
delthas: See discussion above: https://github.com/scality/Zenko/pull/2352#pullrequestreview-3944759653

tests/ctst/common/common.ts
delthas: See discussion above: https://github.com/scality/Zenko/pull/2352#pullrequestreview-3944759653

I don't have a strong opinion on this, either [my first commit](https://github.com/scality/Zenko/pull/2352/changes/7e50c31541c328905b15679406abf07d416311c4) or @SylvainSenechal solution work for me.

.github/scripts/end2end/configs/zenko.yaml
francoisferrand: 1min was done to be conservative, and avoid too much load : is there a significant need or benefit to lower the value?

> situation where that kafka cleaner would start a few seconds before the wrong 60sec timeout instead of 600sec that we had, I would suggest lowering the value, somewhere between 15s and 30

the test used to work (before we changed to a duration I guess), and need to have the retry to ensure we eventually see the cleanup -whatever the value. It now fails because there is no retry : changing the period will not change this....

tests/ctst/common/common.ts
francoisferrand: I'd really rather keep coupling low : esp. since we are on a path (c.f. all the tickets) to allow running Zenko tests "everywhere" (locally, against artesca...)

delthas: Reverted.

.github/scripts/end2end/configs/zenko.yaml
delthas: Reverted.

delthas:
>     * If you do keep the parseDurationSeconds, consider a few documentation example of what "duration: string" => result this function would process

Documented.

tests/ctst/common/utils.ts
```diff
@@ -100,6 +100,32 @@ export const s3FunctionExtraParams: { [key: string]: Record<string, unknown>[] }
     }],
 };
 
+/**
+ * Parses a duration string in Go's time.ParseDuration format
+ * (https://pkg.go.dev/time#ParseDuration) and returns the equivalent in seconds.
+ * @param {string} duration - the duration string to parse (e.g. "1h30m", "500ms")
+ * @return {number} - the duration in seconds
+ */
+export function parseDurationToSeconds(duration: string): number {
```
francoisferrand: nit: `parseGoDuration`

tests/ctst/common/utils.ts
```diff
@@ -100,6 +100,32 @@ export const s3FunctionExtraParams: { [key: string]: Record<string, unknown>[] }
...
+    const units: Record<string, number> = {
+        ns: 1e-9, us: 1e-6, µs: 1e-6, ms: 1e-3, s: 1, m: 60, h: 3600,
+    };
+    let remaining = duration;
+    if (remaining.length === 0) {
+        throw new Error(`Invalid duration: "${duration}"`);
+    }
+    let totalSeconds = 0;
+    while (remaining.length > 0) {
+        const match = remaining.match(/^(\d+(?:\.\d*)?)(ns|us|µs|ms|s|m|h)/);
```
francoisferrand, line 119: does (go) ParseDuration not support "spaces" in between fragments (e.g. "5h 3m") ?
it does not seem to be supported here...

tests/ctst/common/utils.ts
```diff
@@ -100,6 +100,32 @@ export const s3FunctionExtraParams: { [key: string]: Record<string, unknown>[] }
...
+        if (!match) {
+            throw new Error(`Invalid duration: "${duration}" (unparsed: "${remaining}")`);
+        }
+        totalSeconds += parseFloat(match[1]) * units[match[2]];
+        remaining = remaining.slice(match[0].length);
+    }
+    return totalSeconds;
+}
+
```
francoisferrand, line 128: should we not have a few test for this function, whose logic is not trivial - to ensure we don't unexpectedly fail?

francoisferrand (approved):

tests/ctst/common/utils.ts
delthas: Done, renamed to `parseGoDuration`.

tests/ctst/common/utils.ts
delthas: Go's `time.ParseDuration` does not support spaces — `"5h 3m"` is invalid in Go and returns an error. Our parser matches Go's behavior here and correctly rejects it (covered in the test file).

tests/ctst/common/utils.ts
delthas: Added `parseGoDuration.test.ts` with 11 valid cases (simple units, compound durations, fractional values, sub-second units) and 7 invalid cases (empty string, missing unit, spaces between fragments, etc.).

PR #2354, opened 2026-03-16
Merge tests reports from "re-runs" to automatically ignore flaky test results

francoisferrand:
In order to workaround flaky tests, the test command result will now be ignored. Instead, extra processing is done at the end of each job, in the archive-artifacts action:
- retrieve the junit report from every earlier attempt (note: this only considers the current _run_, there is no general notion that a build is flaky)
- merge these reports, by simply adding every test results in a single report. This is done using the python script, as it does not require installing any extra package for XML processing.
- rely on `mikepenz/action-junit-report` action to compute the job status. We were already using it, and it has this functionality already AND knows how to handle such flaky tests (where there are multiple results for the same test).

In addition, to give some visibility on these flaky test, the summary (which was already present) will now show the results form every attempt. This required running the `mikepenz/action-junit-report` action twice though: once with the individual reports for building an easy to read report, and another time with the "merged" report to compute the job status. It looks like this:

image

This approach does not solve/reduce flakiness, nor allow easily blacklisting some flaky test results; however it helps mitigate the worse case scenario, by statistically ensuring much faster convergence: as every attempt can only decrease the number of failed tests.
Instead of requiring all tests to pass in a single run (where probabilty can quickly degrade when multiple tests are flaky), we only require that each test passes once overall.

Issue: ZENKO-5142


.github/actions/archive-artifacts/action.yaml
```diff
@@ -32,38 +32,115 @@ inputs:
...
+          report_paths=("/tmp/junit-previous/${name}/*.xml" "${report_paths[@]}")
+        done
+
+        set_output() { echo "$1<<EOF"; printf '%s\n' "${@:2}"; echo "EOF"; } >> "$GITHUB_OUTPUT"
+
+        set_output "check-names" "${check_names[@]}"
+        set_output "report-paths" "${report_paths[@]}"
+      id: download-previous-reports
+
+    # TODO: if there is no report generated in the current run, we should create one to show that
```
SylvainSenechal: checking: Is it something that you wanna do now and forgot to do, or for later ?

.github/workflows/end2end.yaml
```diff
@@ -549,21 +548,26 @@ jobs:
       - name: Run init CI test
         run: bash run-e2e-test.sh "end2end" ${E2E_IMAGE_NAME}:${E2E_IMAGE_TAG} "end2end" "default"
         working-directory: ./.github/scripts/end2end
+        continue-on-error: true
```
SylvainSenechal, line 551: Can you explain the continue on error ?
Is it because it's required for the merging part to continue ?

SylvainSenechal:
> In order to workaround flaky tests, the test command result will now be ignored. Instead, extra processing is done at the end of each job, in the archive-artifacts action:
> 
> * retrieve the junit report from every earlier attempt (note: this only considers the current _run_, there is no general notion that a build is flaky)
> * merge these reports, by simply adding every test results in a single report. This is done using the python script, as it does not require installing any extra package for XML processing.
> * rely on `mikepenz/action-junit-report` action to compute the job status. We were already using it, and it has this functionality already AND knows how to handle such flaky tests (where there are multiple results for the same test).
> 
> In addition, to give some visibility on these flaky test, the summary (which was already present) will now show the results form every attempt. This required running the `mikepenz/action-junit-report` action twice though: once with the individual reports for building an easy to read report, and another time with the "merged" report to compute the job status. It looks like this:
> 
> image
> This approach does not solve/reduce flakiness, nor allow easily blacklisting some flaky test results; however it helps mitigate the worse case scenario, by statistically ensuring much faster convergence: as every attempt can only decrease the number of failed tests. Instead of requiring all tests to pass in a single run (where probabilty can quickly degrade when multiple tests are flaky), we only require that each test passes once overall.
> 
> Issue: [ZENKO-5142](https://scality.atlassian.net/browse/ZENKO-5142)

So on this run : 
https://github.com/scality/Zenko/actions/runs/23207141364/job/67446761946
We have this result : 
image

I see the Run e2e ctst step is marked as if it passed, but with this pr, it will always be marked as passed even when there are failures, and instead we will have to look at the Archive step or the Summary to see what happened, is that correct ? 
- Is it possible to still show the "run e2e ctst" step as failed, or is it something controlled by github that we can't do much about ?  
- Is it possible to have a dedicated in between step, maybe just before the archive, that would be called something like "merge ctst tests reports" : Because here I find it a bit weird that we see the failure on the archive step

Only partially related to this pr, but might be an opportunity for that : The Archive artifact step is super long, usually around ~7000 lines, and it's super annoying to have to scroll down to find the artifact, could we find a way to either, significantly reduce the amount of printed lines on this step (i dont think whats printed is super useful). Or better, does github offers some kind of feature so that we can display the artifact url maybe in the step title with a variable or something ? Also, I just checked rapidly, we should be able to have this link be visible in the github summary, and I think we already kinda do it, but not sure this is correct because the 4 artifacts I see here are dockerbuilds, not exactly the kind of url that I usually use to download artifacts (https://artifacts.scality.net/builds/github:scality:Zenko:staging-f6f5ab8b81.build-iso-and-end2end-test.8791) 🤔 
image
  

[ZENKO-5142]: https://scality.atlassian.net/browse/ZENKO-5142?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

francoisferrand:
> So on this run : [scality/Zenko/actions/runs/23207141364/job/67446761946](https://github.com/scality/Zenko/actions/runs/23207141364/job/67446761946) I see the Run e2e ctst step is marked as if it passed, but with this pr, it will always be marked as passed even when there are failures, and instead we will have to look at the Archive step or the Summary to see what happened, is that correct ?

is this just about changing habits, or is there a real problem?
- The status of the _job_ is still displayed front and foremost (in the sidebar typically) - which IMO is the most important information
- The summary (also on the front page) shows the details of the failure, without having to dig through the logs or steps. Though unfortunately it is now displayed below the "Docker build summary", maybe we can improve that later.
- Indeed, the step is not marked as failed : so we cannot as "immediately" answer the question _did test fail_ ?
- However, we can still easily answer to the reverse question: _what failed" ? either it is explicitely one of the deployment steps, or it is indeed the "archive artifacts"....which is a proxy for test failure, since most -if not all- steps there are continue-on-failure
 
> * Is it possible to still show the "run e2e ctst" step as failed, or is it something controlled by github that we can't do much about ?

Not possible unfortunately. In order for the build to be able to success, we must ignore the the error (`continue-on-error`), and GitHub marks the build as success in that case.

> * Is it possible to have a dedicated in between step, maybe just before the archive, that would be called something like "merge ctst tests reports" : Because here I find it a bit weird that we see the failure on the archive step

Technically yes, but it has other drawbacks
- requires even more modifications to each job
- there is a lot of coupling between this merging/analysis and the upload of artifacts, so I'd really keep them together

> Only partially related to this pr, but might be an opportunity for that : The Archive artifact step is super long, usually around ~7000 lines, and it's super annoying to have to scroll down to find the artifact, could we find a way to either, significantly reduce the amount of printed lines on this step (i dont think whats printed is super useful). Or better, does github offers some kind of feature so that we can display the artifact url maybe in the step title with a variable or something ? Also, I just checked rapidly, we should be able to have this link be visible in the github summary, and I think we already kinda do it, but not sure this is correct because the 4 artifacts I see here are dockerbuilds, not exactly the kind of url that I usually use to download artifacts 

- reducing the length of archive artifacts logs is out of scope of this PR, but feel free to improve :)
- you don't need to scroll, there is another summary with the link to artifacts, directly on the front page of the build:

image

.github/workflows/end2end.yaml
francoisferrand: if we don't continue-on-error, the build will fail immediately on error: whatever we do after, we cannot "recover" the build and make it pass.

.github/actions/archive-artifacts/action.yaml
francoisferrand: forgot to remove the comment :)

SylvainSenechal:
Ok for the response, yeah it will shift a bit the way we investigate errors, as I'm used to looking at the "run ctst" job directly, and also usually don't look at the summary.
But this is just habits

SylvainSenechal (approved):

maeldonn (approved):

delthas (approved):

.github/actions/archive-artifacts/action.yaml
```diff
@@ -150,3 +229,4 @@ runs:
         user: ${{ inputs.user }}
         password: ${{ inputs.password }}
         source: /tmp/artifacts
+      if: always()
```
DarkIsDude, line 232: don't forget to add new lines 🙏

.github/scripts/merge-junit-reports.py
```diff
@@ -0,0 +1,132 @@
+#!/usr/bin/env python3
```
DarkIsDude, line 1: Should we have test somehow ?

PR #2356, opened 2026-03-17
rework website ctst test to preconfigure the overlay before running t…

SylvainSenechal:
Issue: ZENKO-5221

Rule 3 from "how to write tests" : 

## 3. Do not reconfigure the environment if it affects other tests' resources.

If a reconfiguration triggers a reconciliation of the resources, running tests
might be affected. Also, it restarts the services, which leads to missing logs,
harder debugging, and longer test execution.

For examples, location creation or bucket Website endpoints, that triggers a
restart of multiple services, must be created before the test starts, or in a
dedicated set of tests, that is, a set of test executed before all tests having
a specific tag used for identifying the feature(s).

Note: Testing the reconfiguration of the environment is recommended, but it
should be done carefully to not affect other tests.

### DIFF
--- a/.github/scripts/end2end/configure-e2e-ctst.sh
+++ b/.github/scripts/end2end/configure-e2e-ctst.sh
@@ -2,6 +2,7 @@
 set -exu
 
 DIR=$(dirname "$0")
+. "$DIR"/common.sh
 
 # Get kafka image name and tag
 kafka_image() {
@@ -78,6 +79,21 @@ UUID=${UUID:1}
 
 echo "127.0.0.1 iam.zenko.local s3-local-file.zenko.local keycloak.zenko.local \
     sts.zenko.local management.zenko.local s3.zenko.local website.mywebsite.com utilization.zenko.local" | sudo tee -a /etc/hosts
+
+# Add website endpoint to the overlay so it doesn't trigger reconciliation during tests
+INSTANCE_ID=$(kubectl get zenko ${ZENKO_NAME} -o jsonpath='{.status.instanceID}')
+MGMT_TOKEN=$(get_token)
+WEBSITE_HTTP_CODE=$(curl -s -o /dev/null -w "%{http_code}" -X POST \
+    -H "Content-Type: application/json" \
+    -H "X-Authentication-Token: ${MGMT_TOKEN}" \
+    "http://management.zenko.local/api/v1/config/${INSTANCE_ID}/website/endpoint" \
+    -d '"mywebsite.com"')
+if [ "$WEBSITE_HTTP_CODE" = "200" ] || [ "$WEBSITE_HTTP_CODE" = "409" ]; then
+    echo "Website endpoint registered successfully (HTTP $WEBSITE_HTTP_CODE)"
+else
+    echo "Failed to register website endpoint (HTTP $WEBSITE_HTTP_CODE)" >&2
+    exit 1
+fi
 
 # Add bucket notification target
 envsubst < ./configs/notification_destinations.yaml | kubectl apply -f -
--- a/tests/ctst/steps/website/website.ts
+++ b/tests/ctst/steps/website/website.ts
@@ -29,8 +29,10 @@ When('the user puts the bucket website configuration', async function (this: Zen
     });
 });
 
-When('the {string} endpoint is added to the overlay', async function (this: Zenko, endpoint: string) {
-    await this.addWebsiteEndpoint(endpoint);
+When('the {string} endpoint is present in the overlay', async function (this: Zenko, endpoint: string) {
+    const found = await this.verifyWebsiteEndpoint(endpoint);
+    assert.strictEqual(found, true,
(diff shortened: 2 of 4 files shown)


.github/scripts/end2end/configure-e2e-ctst.sh
SylvainSenechal, line 1: @francoisferrand 
Let me know if you find this PR irrelevant. 
I understand there is a conflict between what we ultimately want to achieve in the tests :
- On one side, I guess this PR should probably not exists, and instead we should aim to have a software thats stable and functional even we update the overlay
- On the other side, if we really just wanna check that the feature bucket website works, and avoid affecting other tests, then I guess this pr is ok, doing the overlay configuration before the tests still allows us to test that the management api is working

.github/scripts/end2end/configure-e2e-ctst.sh
```diff
@@ -79,6 +80,15 @@ UUID=${UUID:1}
 echo "127.0.0.1 iam.zenko.local s3-local-file.zenko.local keycloak.zenko.local \
     sts.zenko.local management.zenko.local s3.zenko.local website.mywebsite.com utilization.zenko.local" | sudo tee -a /etc/hosts
 
+# Add website endpoint to the overlay so it doesn't trigger reconciliation during tests
+INSTANCE_ID=$(kubectl get zenko ${ZENKO_NAME} -o jsonpath='{.status.instanceID}')
+MGMT_TOKEN=$(get_token)
+curl -sf -X POST \
```
francoisferrand: this is not idem-potent, so it would prevent "re-running" the tests (no problem in CI, but locally or in codespace) : is this not an issue?

.github/scripts/end2end/configure-e2e-ctst.sh
francoisferrand: > On one side, I guess this PR should probably not exists, and instead we should aim to have a software thats stable and functional even we update the overlay

I agree it should not indeed:
- because it should work in any case. Even if they involve modifying the overlay, this is a normal operation which should not disrupt the cluster....and thus not disrupt the other tests either
- this "fix" does not solve the actual problem causing the flakiness, it just avoids it....and we will still have the underlying problem

> On the other side, if we really just wanna check that the feature bucket website works, and avoid affecting other tests, then I guess this pr is ok, doing the overlay configuration before the tests still allows us to test that the management api is working

- "avoid affecting other tests" is not a goal in itself ; at most a mean to avoid flakiness
- the approach works with the current test, but only because the "website CRUD test" actually only test the creation.... when we add update/delete, it will not work anymore
- moving the creation out of the test makes more coupling between setup & tests:
   - which is the opposite direction than where we want to go ("portable" tests that can be played against any cluster, idempotentcy, minimize `World` and configuration variables...)
   - which makes it harder to understand the test (issues may come from more places, or all the machinery in between) and troubleshoot it (less temporary locality, the test actually spans from setup to the moment it is actually executed)

Overall,
→ we are missing a few tests here, should create a followup ticket
→ this confirms some flakiness related to overlay update : not a solution, but knowing we should search there is progress, and should us to think both of ways this could make test fail & ways to better observe such failure path [*]
→ the overlay is a big blob, and actually quite a lot of things may be updated when it changes. We may want to optimize this eventually, and start by identifying exactly what depends on it / reviewing if it is relevant / maybe finding ways to be more granular

[*] the other day, you had a graph/timeline of tests. If we also display k8s "events" (pod restarts...) and reconciliations on top of it, this may help identify patterns of failures or correlations...

.github/scripts/end2end/configure-e2e-ctst.sh
SylvainSenechal: I added a 409 error check that I tested in codespace, should be idempotent now

.github/scripts/end2end/configure-e2e-ctst.sh
SylvainSenechal: Ok I think we will rediscuss this

tests/ctst/steps/website/website.ts
```diff
@@ -29,8 +29,10 @@ When('the user puts the bucket website configuration', async function (this: Zen
     });
 });
 
-When('the {string} endpoint is added to the overlay', async function (this: Zenko, endpoint: string) {
-    await this.addWebsiteEndpoint(endpoint);
+When('the {string} endpoint is present in the overlay', async function (this: Zenko, endpoint: string) {
+    const found = await this.verifyWebsiteEndpoint(endpoint);
+    assert.strictEqual(found, true,
+        `Website endpoint '${endpoint}' not found in overlay after 60 retries`);
```
maeldonn, line 35: Where are the retries ?

maeldonn (approved):
Just a question about the retry, otherwise LGTM for the PR content. For the goal, I read the thread with @francoisferrand, but I don't really have an opinion for now...

SylvainSenechal:
Delivered by https://scality.atlassian.net/browse/ZENKO-5245

PR #2362, opened 2026-03-25
e2e scripts: add rootless podman and local environment support

delthas:
## Summary
- Auto-detect rootless podman and configure kind with delegated systemd scope
- Fix DNS resolution issues caused by podman injecting `dns.podman` search domains
- Support HOST_DNS override for CoreDNS patching in podman environments
- Use `127.0.0.1` instead of `localhost` and add scheme variable for get_token
- Override VOLUME_ROOT for local development, remove unused local registry setup
- Allow using a local zenko-operator checkout if available
- Add /etc/hosts pre-setup documentation and idempotent hosts file checks

### DIFF
--- a/.devcontainer/README.md
+++ b/.devcontainer/README.md
@@ -3,6 +3,16 @@
 
 A [VS Code extension](https://marketplace.visualstudio.com/items?itemName=GitHub.codespaces) is available for Codespaces.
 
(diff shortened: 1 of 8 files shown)


delthas:
Oops this was auto-created. Sorry. :sweat_smile:

PR #2363, opened 2026-03-25
e2e scripts: make setup idempotent and fix test runner credentials

delthas:
## Summary
- Make kubectl create calls idempotent with --dry-run=client | kubectl apply
- Fix re-run failures for helm repos, root-ca, prometheus and kafka CRDs
- Reuse existing RSA keypair on re-runs to avoid OAEP decryption errors
- Persist auth ZooKeeper storage so SCRAM credentials survive broker restarts
- Use stable MongoDB database name for local and codespace runs
- Refactor setup step loop to early exits, skip metadata deploy when disabled
- Fix CTST test runner to read vault admin credentials from cluster secrets

### DIFF
--- a/.devcontainer/setup.sh
+++ b/.devcontainer/setup.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-set -e
+set -ex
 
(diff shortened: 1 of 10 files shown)


delthas:
Oops this was auto-created. Sorry. :sweat_smile:

PR #2364, opened 2026-03-25
Add local environment support

delthas:
## Summary

The end-to-end scripts were designed to run in GitHub Codespaces (Docker-in-Docker).
Running them locally with rootless podman on Linux exposed several issues. This PR
fixes them so the full Zenko stack can be deployed and tested on a local machine.

### DNS timeouts from podman search domains

Podman injects `search dns.podman` into kind node resolv.conf. CoreDNS can't resolve
this domain, adding an 8-second timeout to every DNS lookup. This killed pods that
rely on fast DNS during init (e.g. ZooKeeper exceeded its liveness probe). Fixed by
adding a `config.yaml` for kind with `networking: { dnsSearch: [] }`, which strips
all search clauses.

### curl hangs to localhost

`get_token()` in `common.sh` hardcoded `https://localhost`. Two issues: curl resolves
`localhost` to `::1` (IPv6) first but kind only binds on IPv4, and the scheme was
hardcoded to `https` but keycloak ingress uses `http` when `ENABLE_KEYCLOAK_HTTPS=false`.
Fixed with a scheme variable and `127.0.0.1`.

### CoreDNS can't resolve host DNS

In podman environments, the host DNS server address differs from Docker's default
(`172.17.0.1`). Added `HOST_DNS` override support in `patch-coredns.sh` so it
reads the nameserver from `/etc/resolv.conf` when running outside Docker.

### Other changes

- Auto-detect rootless podman and configure kind with a delegated systemd scope
- Override `VOLUME_ROOT` for local development (no `/mnt` in local environments)
- Remove unused local registry setup from `bootstrap-kind.sh`
- Allow using a local `zenko-operator` checkout instead of always fetching from GitHub
- Make `/etc/hosts` entries idempotent (check before appending)
- Add documentation for required `/etc/hosts` pre-setup

Issue: ZENKO-5243

### DIFF
--- a/.devcontainer/README.md
+++ b/.devcontainer/README.md
@@ -3,6 +3,16 @@
 
 A [VS Code extension](https://marketplace.visualstudio.com/items?itemName=GitHub.codespaces) is available for Codespaces.
 
+## Pre-setup: /etc/hosts (local environments only)
+
+When running locally (not in a Codespace/devcontainer where you are root),
+the CTST configuration script needs Zenko hostnames to resolve to localhost.
+Add them before running setup to avoid a `sudo` prompt mid-run:
+
+```bash
+echo "127.0.0.1 iam.zenko.local s3-local-file.zenko.local keycloak.zenko.local sts.zenko.local management.zenko.local s3.zenko.local website.mywebsite.com utilization.zenko.local aws-mock.zenko.local azure-mock.zenko.local blob.azure-mock.zenko.local queue.azure-mock.zenko.local devstoreaccount1.blob.azure-mock.zenko.local devstoreaccount1.queue.azure-mock.zenko.local dr.zenko.local" | sudo tee -a /etc/hosts
+```
+
 ## Running CTST tests in the codespace
 
 ```bash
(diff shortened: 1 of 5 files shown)


.github/scripts/end2end/configure-e2e-ctst.sh
SylvainSenechal, line 1: Okay yeah, on the latest merged pr, I created this file that handles hosts (amongst other things) : https://github.com/scality/Zenko/blob/development/2.14/.github/scripts/end2end/configure-e2e-endpoints.sh

Might need a rebase and some sync between the 2

.devcontainer/README.md
```diff
@@ -3,6 +3,16 @@
 
 A [VS Code extension](https://marketplace.visualstudio.com/items?itemName=GitHub.codespaces) is available for Codespaces.
 
+## Pre-setup: /etc/hosts (local environments only)
```
SylvainSenechal, line 6: Not sure if you can already run on your pc with this pr, or if another one is coming, but maybe there needs to be a more detailed readme for local run on linux if you can already do it

SylvainSenechal (approved):

.devcontainer/README.md
delthas: Yeah there will probably be more PRs coming, this was a first pass.

benzekrimaha (approved):
LGTM

PR #2365, opened 2026-03-25
Make setup idempotent for re-runs

delthas:
## Summary

The end-to-end setup scripts used imperative `kubectl create` commands and other
one-shot operations that fail with `AlreadyExists` on re-runs. This forces a full
cluster teardown between attempts, which wastes time. This PR makes all scripts
safe to re-run on an existing cluster.

### kubectl create not idempotent

Many scripts used `kubectl create` which fails with `AlreadyExists` on re-run.
Fixed with `--dry-run=client -o yaml | kubectl apply -f -` for imperative creates,
and `kubectl apply --server-side` for large CRDs (prometheus, kafka) that exceed
the annotation size limit.

### Keypair overwrite breaks encryption

The deploy-zenko idempotency fix (`--dry-run=client | kubectl apply`) regenerated
the RSA keypair on every run, but location configs were encrypted with the old key,
causing OAEP decryption errors. Fixed by extracting the existing public key from
the secret when it already exists instead of generating a new one.

### SCRAM credentials lost on broker restart

The auth ZooKeeper cluster was cloned from the main one with persistence explicitly
stripped (`storageType: ephemeral`). SCRAM-SHA-512 user registrations are stored in
ZooKeeper, so any ZK pod restart wiped all credentials. This caused
`notification-processor-destination4` to fail with SASL authentication errors.
Fixed by keeping the original persistence config on the cloned ZK cluster.

### Helm repo and kafka clone failures on re-run

`helm repo add` fails if the repo already exists. The kafka-operator git clone
failed if the directory was already present. Fixed with existence checks and
`--force-update` for helm repos.

### CTST test runner used wrong credentials

`run-ctst-locally.sh` had hardcoded access keys that didn't survive vault restarts.
Also used account name `zenko` instead of `zenko-ctst`. Fixed to read vault admin
credentials from the `management-vault-admin-creds` secret, matching the CI script.

### Other changes

- Replace `docker image prune -af` with targeted `rmi` (less destructive)
- Remove unnecessary operator image pull in `deploy-zkop.sh`
- Use stable MongoDB database name via `ZENKO_MONGODB_DATABASE` env var
- Use `PVC_NAME` env var instead of sed hack for miria deploy
- Refactor setup step loop to early exits, skip metadata deploy when disabled

Issue: ZENKO-5244

### DIFF
--- a/tests/ctst/run-ctst-locally.sh
+++ b/tests/ctst/run-ctst-locally.sh
@@ -14,14 +14,17 @@ IMAGE_NAME="${2:-ghcr.io/scality/zenko/zenko-e2e-ctst:ctst_codespace_setup}"
 # certain tests based on their @version tag.
 VERSION=$(cat ../../VERSION | grep -Po 'VERSION="\K[^"]*')
 POD_NAME="ctst-end2end"
+ZENKO_NAME="${ZENKO_NAME:-end2end}"
+ADMIN_ACCESS_KEY=$(kubectl get secret "${ZENKO_NAME}-management-vault-admin-creds.v1" -o jsonpath='{.data.accessKey}' | base64 -d)
+ADMIN_SECRET_KEY=$(kubectl get secret "${ZENKO_NAME}-management-vault-admin-creds.v1" -o jsonpath='{.data.secretKey}' | base64 -d)
 WORLD_PARAMETERS="$(jq -c <<EOF
 {
   "subdomain": "zenko.local",
   "ssl": false,
   "port": "80",
-  "AccountName": "zenko",
-  "AdminAccessKey": "R6JN4OJ998ZBY99DD56X",
-  "AdminSecretKey": "OEow2DytG0sr7mK3844vb/hKDBWiU+UWc+4+FVfr"
+  "AccountName": "zenko-ctst",
+  "AdminAccessKey": "$ADMIN_ACCESS_KEY",
+  "AdminSecretKey": "$ADMIN_SECRET_KEY"
 }
 EOF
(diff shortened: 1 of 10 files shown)


tests/ctst/run-ctst-locally.sh
```diff
@@ -14,14 +14,17 @@ IMAGE_NAME="${2:-ghcr.io/scality/zenko/zenko-e2e-ctst:ctst_codespace_setup}"
...
 {
   "subdomain": "zenko.local",
   "ssl": false,
   "port": "80",
-  "AccountName": "zenko",
-  "AdminAccessKey": "R6JN4OJ998ZBY99DD56X",
-  "AdminSecretKey": "OEow2DytG0sr7mK3844vb/hKDBWiU+UWc+4+FVfr"
+  "AccountName": "zenko-ctst",
+  "AdminAccessKey": "$ADMIN_ACCESS_KEY",
+  "AdminSecretKey": "$ADMIN_SECRET_KEY"
```
delthas, line 27: For this I'm not entirely sure @SylvainSenechal

SylvainSenechal: No it's good, I actually have something similar in one of my ongoing test.
Anyways, mid term goal is to not have to need this script at all

SylvainSenechal (approved):
Ok for me, just need a rebase

maeldonn (approved):

delthas:
> I'm afraid I cannot do that, @delthas:

Nice one. :clapper:

PR #2366, opened 2026-03-26
Build our own mongodb image

benzekrimaha:
Issue: ZENKO-5110


.github/workflows/build-mongodb-images.yaml
```diff
@@ -0,0 +1,106 @@
...
+
+env:
+  REGISTRY: ghcr.io
+
+jobs:
+  build-mongodb-sharded:
+    runs-on: ubuntu-24.04
+    strategy:
+      fail-fast: false
+      matrix:
```
francoisferrand: no need for multiple mongodb builds : we have multiple zenko branches, and just one mongodb per zenko branch

.github/workflows/build-mongodb-images.yaml
```diff
@@ -0,0 +1,106 @@
...
+        include:
+          - name: mongodb-sharded-8.0
+            context: solution-base/images/mongodb-sharded/8.0/debian-12
+            image: scality/zenko/mongodb-sharded
+            semver_tag: "8.0.13"
```
francoisferrand: * this is redundant with deps.yaml 
* just "version"

.github/workflows/build-mongodb-images.yaml
```diff
@@ -0,0 +1,106 @@
...
+            minor_tag: "8.0"
```
francoisferrand: this can be computed from semvar_tag

.github/actions/deploy/action.yaml
```diff
@@ -85,6 +85,53 @@ runs:
       run: ./deploy-metadata.sh
       working-directory: ./.github/scripts/end2end
       if: ${{ inputs.deploy_metadata == 'true' }}
+    - name: '[DEBUG] Dump MongoDB pod status and logs'
```
francoisferrand: to be removed

francoisferrand: for testing, you must use codespace (or a lab) instead of CI : it will be much more efficient and allow easily inspecting.

.github/workflows/build-mongodb-images.yaml
```diff
@@ -0,0 +1,106 @@
...
+
+jobs:
+  build-mongodb-sharded:
+    runs-on: ubuntu-24.04
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - name: mongodb-sharded-8.0
+            context: solution-base/images/mongodb-sharded/8.0/debian-12
```
francoisferrand: should be computed from deps.yaml : we need to build a few images. just go over the list in deps.yaml and derive the info we need (using matching path as needed)

.github/workflows/build-mongodb-images.yaml
francoisferrand: both mongo 8.0 and 8.2 are still committed : this is not needed

.github/workflows/build-mongodb-images.yaml
```diff
@@ -0,0 +1,116 @@
...
+            VERSION=$(yq eval ".${deps_key}.tag" solution-base/deps.yaml)
+            IMAGE_NAME="${IMAGE##*/}"
+
+            if [[ "${IMAGE_NAME}" == "mongodb-sharded" ]]; then
+              MAJOR_MINOR=$(echo "${VERSION}" | cut -d. -f1-2)
+              CONTEXT="solution-base/images/mongodb-sharded/${MAJOR_MINOR}/debian-12"
+            else
+              MAJOR=$(echo "${VERSION}" | cut -d. -f1)
+              CONTEXT="solution-base/images/mongodb-sharded/${IMAGE_NAME}/${MAJOR}/debian-12"
+            fi
```
francoisferrand: we don't need to store multiple versions of either of these components in the repo : so we can choose to organize in the exact way we want, and keep this much simpler.

→ choose in which folder we store them, in a consistent fashion, e.g. either store each in a folder which has the same name as the image ; or just <image>/<major_minor> if we want to avoid mismatch in case of upgrade

.github/workflows/build-mongodb-images.yaml
```diff
@@ -0,0 +1,116 @@
...
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Extract image metadata from deps.yaml
+        id: meta
+        run: |
+          resolve() {
+            local deps_key=$1
+            local prefix=$2
```
francoisferrand: make deps_key and prefix match, for simplicity?

solution-base/images/mongodb-sharded/README.md
```diff
@@ -0,0 +1,152 @@
...
+The Dockerfiles depend on them (`COPY prebuildfs /`, `COPY rootfs /`).
+
+To re-vendor after upstream changes:
+
+```bash
+git clone --depth 1 https://github.com/bitnami/containers.git /tmp/bitnami-containers
+
+# Copy for each image (example for 8.2):
+cp -r /tmp/bitnami-containers/bitnami/mongodb-sharded/8.2/debian-12/prebuildfs solution-base/images/mongodb-sharded/8.2/debian-12/
+cp -r /tmp/bitnami-containers/bitnami/mongodb-sharded/8.2/debian-12/rootfs     solution-base/images/mongodb-sharded/8.2/debian-12/
```
francoisferrand: not correct, this would keep "removed" files

solution-base/images/mongodb-sharded/README.md
```diff
@@ -0,0 +1,152 @@
...
+# Repeat for other images as needed
+
+rm -rf /tmp/bitnami-containers
+git diff   # review changes
```
francoisferrand: "copying" is intrinsically dangerous and complex, as it
- automatically discards all changes that could have been made to the charts
- starts from scratch each time, without considering the history: i.e. the common ancestor between the "old" and "new" code

These problems are already solved by Git, so may be best to use Git here.
In particular, `git subtree` [1] allows to easily merge a "remote" repository in a remote one. This kind of approach would mean we don't copy and diff, but instead perform a proper git merge.

The drawback is that this command is probably not as common for everyone ; but the specific usage can be documented here, and/or wrapped in a makefile (similar to what is done for mongo chart)

---
Going in to more details, there is an extra subtlety here because we don't want the full repo, but only a subset. There are solutions all over the internet, but it requires a few more commands: c.f. [2]

```
# Add the remote repository
$ git remote add bitnami-containers https://github.com/bitnami/containers.git
$ git fetch bitnami-containers

# Create a branch containing only the subfolders of the remote repo we are interested in
$ git checkout bitnami-containers/main
$ git subtree split -b bitnami-upstream -P bitnami/mongodb-sharded/8.2/debian-12 -P bitnami/mongodb-exporter/0/debian-12

# Back to our "dev" branch, merge the subtree
$ git checkout -b improvement/ZENKO-xxx

# First time: just "add" the subtree, without committing every change
$ git subtree add -P solution-base/images bitnami-upstream --squash

# Next times: merge upstream, to consider both our changes & bitnamis. Not sure we will want to squash changes (if we see only relevant changes, may be nice to see them individually in history), but not the most important thing anyway
$ git subtree merge -P solution-base/images bitnami-upstream --squash
```

Note:
- git subtree is part of git "contribs", but not installed by default. So need to install it, c.f. [3]
  ```
  $ cd $(brew --prefix)/share/git-core/contrib/subtree
  $ make
  $ make prefix=$(brew --prefix)/opt/git install
  ```
- since bitnami repo is large, `git subtree split` can be quite slow if the whole history was retrieved. It may be much faster if the whole repository is not fetched, not sure if that works though...
- The sub-path is making git subtree more complex to build the upstream branch or falling back to advanced git commands instead should be considered, c.f. [4] : i.e. basically we use git read-tree the first time, and followup merge upstream with git diff | git apply.

[1] https://manpages.debian.org/testing/git-man/git-subtree.1.en.html
[2] https://medium.com/@icheko/use-a-subfolder-from-a-remote-repo-in-another-repo-using-git-subtree-98046f33ca40
[3] https://codeengineered.com/blog/how-to-install-git-subtree/
[4] https://stackoverflow.com/questions/23937436/add-subdirectory-of-remote-repo-with-git-subtree

francoisferrand: Just asked Claude, seems to automate quite well (though the code will need to be reworked to also handle the list of containers to sync in the makefile/build script)

```
BITNAMI_REMOTE := bitnami-containers
BITNAMI_REPO   := https://github.com/bitnami/containers.git

create-remote:
	@git remote get-url $(BITNAMI_REMOTE) >/dev/null 2>&1 || git remote add $(BITNAMI_REMOTE) $(BITNAMI_REPO)

# Usage: make vendor-sync CHART=mongodb-sharded PREFIX=solution-base/mongodb/charts/mongodb-sharded
vendor-sync: create-remote
	git fetch $(BITNAMI_REMOTE) main --depth=1
	git subtree split --prefix=bitnami/$(CONTAINER) $(BITNAMI_REMOTE)/main -b vendor/$(CONTAINER)
	git subtree merge --prefix=solution-base/images/$(CONTAINER) vendor/$(CONTAINER) --squash

# First-time import variant
vendor-add: create-remote
	git fetch $(BITNAMI_REMOTE) main --depth=1
	git subtree split --prefix=bitnami/$(CHART) $(BITNAMI_REMOTE)/main -b vendor/$(CONTAINER)
	git subtree add --prefix=solution-base/images/$(CONTAINER) vendor/$(CONTAINER) --squash
```

Then for any container:
```
make vendor-add CONTAINER=mongodb-sharded
```

solution-base/images/mongodb-sharded/README.md
```diff
@@ -0,0 +1,152 @@
...
+## Bumping MongoDB Versions
+
+### Patch bump (e.g. 8.0.13 → 8.0.14)
+
+1. Update the Dockerfile under the version directory (e.g.
+   `8.0/debian-12/Dockerfile`):
+   - `COMPONENTS+=("mongodb-8.0.14-0-linux-...")` (update component version)
+   - `MONGODB_TARBALL="mongodb-linux-aarch64-ubuntu2204-8.0.14.tgz"` (arm64)
+   - `org.opencontainers.image.version="8.0.14"` (label)
+   - `APP_VERSION="8.0.14"` (env)
```
francoisferrand: should be a single ARG (e.g. "MONGODB_VERSION") in the Dockerfile.....whose value we retrieve from the deps.yaml, and is passed to docker build

solution-base/mongodb/how_to_upgrade.md
```diff
@@ -7,7 +7,39 @@ to the new version. Below is a structured guide to follow for a successful
...
 
-## Upgrade steps
+## Image Build (Stop-Gap)
+
+Zenko now builds its own MongoDB Sharded images from vendored Bitnami scripts.
+Image sources live under `solution-base/images/mongodb-sharded/`.
+
+### Bumping the MongoDB version in the image
+
+1. Edit the Dockerfile for the target major line
```
francoisferrand: redundant with the doc in images/mongodb-shared/README.md
either keep both here, or just use a "pointer" to the other file here

(and you see the cost of redundancy: you forgot to update this one to match your latest changes)

francoisferrand (changes requested):

solution-base/images/mongodb-sharded/README.md
francoisferrand: to clarify the intent here : however we do it, we will make changes to the image, but we need to manage it.
Such in (git) history we need to have something like:

```
* Initial mport from upstream commit XXXX
* Our own tweaks [for exemple: modify makefile to take mongodb version in argument]
* More tweaks [for exemple: bump baseline image hash]
* ...
* Merge upstream commit YYY
* ...
```

So we need to make sure we actually "merge" upstream with our own changes.
For the charts, we ended using patch files - works, but not very efficient: but we need to do something in this case as well...

.github/workflows/build-mongodb-images.yaml
```diff
@@ -0,0 +1,116 @@
...
+on:
+  push:
+    branches:
+      - 'development/**'
+      - 'feature/**'
+      - 'improvement/**'
+      - 'w/**'
+      - main
+    paths:
+      - 'solution-base/images/mongodb-sharded/**'
```
francoisferrand: also depends on deps.yaml, and possibly other build scripts

.github/workflows/build-mongodb-images.yaml
francoisferrand, line 1: why use another mongodb workflow?
- since builds are cached and the tag is computed, it does not cost anything to rebuild
- path filtering can be tricky and cause unexpected results
- it duplicates trigger conditions, which can easily get out of sync and create inconsistencies (or bugs) in the jobs actually executed
- it makes it harder to see what's going on overall (when opening a PR/...)
- it prevents using the just built image in the same build, as there is no dependency: so the mongodb image used may not have been built

.github/workflows/build-mongodb-images.yaml
```diff
@@ -0,0 +1,116 @@
...
+              CONTEXT="solution-base/images/mongodb-sharded/${MAJOR_MINOR}/debian-12"
+            else
+              MAJOR=$(echo "${VERSION}" | cut -d. -f1)
+              CONTEXT="solution-base/images/mongodb-sharded/${IMAGE_NAME}/${MAJOR}/debian-12"
+            fi
+
+            TREE_HASH=$(git rev-parse HEAD:"${CONTEXT}")
+            SHORT_HASH=${TREE_HASH:0:8}
+
+            TAGS="${{ env.REGISTRY }}/${IMAGE}:${VERSION}"
```
francoisferrand: this tags is a moving tag (may change on every commit), should probably not be used...

.github/workflows/build-mongodb-images.yaml
```diff
@@ -0,0 +1,116 @@
...
+            TAGS="${TAGS},${{ env.REGISTRY }}/${IMAGE}:${VERSION}-${SHORT_HASH}"
```
francoisferrand: the tag with the "tree-hash" that we built MUST be included in the solution-base ISO (do not depend on a floating tag!) and used when deploying mongo in tests

c.f. how this is done for kafka images (which also have such a tree hash)

.github/workflows/build-mongodb-images.yaml
benzekrimaha: I removed the version-based folder branching and standardized the Mongo image layout to a single consistent structure per image, which simplifies context resolution and avoids mismatch risks during upgrades.

solution-base/images/mongodb-sharded/README.md
benzekrimaha: Agreed and fixed. I removed copy-based vendoring instructions because they do not handle upstream deletions safely.

solution-base/images/mongodb-sharded/README.md
benzekrimaha: I switched documentation and workflow to a Git-based vendoring approach (git subtree) so upstream updates are merged properly and history is preserved.

solution-base/images/mongodb-sharded/README.md
benzekrimaha: The Dockerfile now uses one MONGODB_VERSION build arg, and CI passes its value from deps, so version updates are centralized and consistent.

solution-base/mongodb/how_to_upgrade.md
benzekrimaha: I removed redundant step-by-step image bump content and left a pointer to the canonical mongo image readme to avoid drift.

.github/workflows/build-mongodb-images.yaml
benzekrimaha: I removed the separate path-filtered Mongo workflow and moved Mongo image builds into the main pipeline, so trigger drift on deps/build script changes is no longer an issue

.github/workflows/build-mongodb-images.yaml
benzekrimaha: Mongo build is now part of the main CI flow with proper job dependencies, so tests use images built in the same run and overall CI behavior is easier to follow.

.github/workflows/build-mongodb-images.yaml
benzekrimaha: I aligned Mongo handling with the Kafka pattern: the tree-hash image tag is propagated into solution-base build inputs and test deployment paths, so both ISO content and tests use the exact built image.

.github/workflows/build-mongodb-images.yaml
benzekrimaha: The pipeline now builds and consumes immutable tree-hash-tagged images for Mongo components instead of relying on floating semver tags.

.github/scripts/end2end/install-kind-dependencies.sh
```diff
@@ -233,7 +233,19 @@ build_solution_base_manifests() {
...
+
+        case "$dep_name" in
+            mongodb-sharded|mongodb-shell|mongodb-sharded-exporter)
+                source <( "$SOLUTION_BASE_DIR/mongodb_build_vars.sh" )
+                echo "${image}-${MONGODB_BUILD_TREE_HASH}"
+                ;;
+            *)
+                echo "$image"
+                ;;
+        esac
```
francoisferrand:
```suggestion
        source <( "$SOLUTION_BASE_DIR/mongodb_build_vars.sh" )
        
        yq eval ".$dep_name | (.sourceRegistry // \"docker.io\") + \"/\" + .image + \":\" + .tag" $SOLUTION_BASE_DIR/deps.yaml |
            sed '/ghcr.io\/scality\/zenko\/mongo/ s/$/-'"${MONGODB_BUILD_TREE_HASH}"'/'
```

solution-base/mongodb/how_to_upgrade.md
francoisferrand, line 1: folder `solution-base/images/mongodb-sharded` seems useless?
it already contains folder for each image...

```suggestion
solution-base/
    images/
        mongodb-exporter/
        mongodb-sharded/
        os-shell/
        README.md
        Makefile
```

solution-base/images/Makefile
```diff
@@ -0,0 +1,26 @@
+BITNAMI_REMOTE := bitnami-containers
+BITNAMI_REPO := https://github.com/bitnami/containers.git
+
+# Required:
+# - CONTAINER: destination path under solution-base/images (example: mongodb-sharded/debian-12)
```
francoisferrand: CONTAINER is not an argument:
- we know exactly which values are possible and MUST be used
- merging/... should be done at the exact same point, we don't want to mix-and-match different versions of exporter/mongo/os-shell
- git fetch and subtree split are somewhat costly: we should avoid doing them multiple times

typically this means there should be multiple rules : creating remote, fetching/updating "vendor" branch (split subtree), then one rule for each image

solution-base/images/Makefile
```diff
@@ -0,0 +1,26 @@
...
+.PHONY: create-remote vendor-add vendor-sync
+
+create-remote:
+	@git remote get-url $(BITNAMI_REMOTE) >/dev/null 2>&1 || git remote add $(BITNAMI_REMOTE) $(BITNAMI_REPO)
+
+vendor-add: create-remote
+	@test -n "$(CONTAINER)" || (echo "CONTAINER is required"; exit 1)
+	@test -n "$(BITNAMI_PREFIX)" || (echo "BITNAMI_PREFIX is required"; exit 1)
+	git fetch $(BITNAMI_REMOTE) main --depth=1
+	git subtree split --prefix=$(BITNAMI_PREFIX) $(BITNAMI_REMOTE)/main -b vendor/$(CONTAINER)
```
francoisferrand: this does not work today : `main` does not contain `bitnami/mongodb-sharded/8.0` → should we have something else instead of `main` ?
- may be ok once we move to 8.2, but not just now...
- how does it work when we move from 8.0 to 8.2 ?

solution-base/images/Makefile
```diff
@@ -0,0 +1,26 @@
+BITNAMI_REMOTE := bitnami-containers
+BITNAMI_REPO := https://github.com/bitnami/containers.git
+
+# Required:
+# - CONTAINER: destination path under solution-base/images (example: mongodb-sharded/debian-12)
+# - BITNAMI_PREFIX: source path in bitnami/containers repository
```
francoisferrand: same BITNAMI_PREFIX is not an input, the goal of a makefile is to set these values instead of relying on some external knowledge...

solution-base/images/Makefile
```diff
@@ -0,0 +1,26 @@
...
+#   (example: bitnami/mongodb-sharded/8.0/debian-12)
+
+.PHONY: create-remote vendor-add vendor-sync
+
+create-remote:
+	@git remote get-url $(BITNAMI_REMOTE) >/dev/null 2>&1 || git remote add $(BITNAMI_REMOTE) $(BITNAMI_REPO)
+
+vendor-add: create-remote
```
francoisferrand: when/why would we use this rule? Subtrees have already been added...

solution-base/images/mongodb-sharded/README.md
```diff
@@ -0,0 +1,87 @@
...
+git fetch bitnami-containers main --depth=1
+```
+
+3. First import of a container path:
+
+```bash
+make -C solution-base/images vendor-add CONTAINER=mongodb-sharded/debian-12 BITNAMI_PREFIX=bitnami/mongodb-sharded/8.0/debian-12
+make -C solution-base/images vendor-add CONTAINER=mongodb-exporter/debian-12 BITNAMI_PREFIX=bitnami/mongodb-exporter/0/debian-12
+make -C solution-base/images vendor-add CONTAINER=os-shell/debian-12 BITNAMI_PREFIX=bitnami/os-shell/12/debian-12
+```
```
francoisferrand: already done...

solution-base/images/mongodb-sharded/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+4. Next updates:
+
+```bash
+make -C solution-base/images vendor-sync CONTAINER=mongodb-sharded/debian-12 BITNAMI_PREFIX=bitnami/mongodb-sharded/8.0/debian-12
+make -C solution-base/images vendor-sync CONTAINER=mongodb-exporter/debian-12 BITNAMI_PREFIX=bitnami/mongodb-exporter/0/debian-12
+make -C solution-base/images vendor-sync CONTAINER=os-shell/debian-12 BITNAMI_PREFIX=bitnami/os-shell/12/debian-12
```
francoisferrand: should be part of the makefile : no point keeping the values in the documentation...

solution-base/images/mongodb-sharded/README.md
```diff
@@ -0,0 +1,87 @@
...
+```
+
+This preserves a clean history: import upstream, apply our tweaks, then merge
```
francoisferrand: > apply our tweaks

should probably be made explicit : it's not about modifying the code, but really making extra _commits_ to tweak the content

solution-base/images/mongodb-sharded/README.md
```diff
@@ -0,0 +1,87 @@
...
+later upstream changes.
+
+## Bumping MongoDB Version
+
+`mongodb-sharded/debian-12/Dockerfile` exposes a single `ARG MONGODB_VERSION`.
+
+To bump MongoDB:
+1. Update `solution-base/deps.yaml` `mongodb-sharded.tag`.
+2. If needed, update `MONGODB_VERSION` default in Dockerfile for local builds.
```
francoisferrand: should not have default in Dockerfile, it is redondant (with deps.yaml) and misleading

.github/workflows/end2end.yaml
```diff
@@ -351,6 +351,61 @@ jobs:
...
+          username: "${{ github.repository_owner }}"
+          password: "${{ github.token }}"
+          registry: ghcr.io
+
+      - name: Extract environment
+        run: |-
+          solution-base/mongodb_build_vars.sh >> $GITHUB_ENV
+
+      - name: Build and push mongodb-sharded
+        uses: docker/build-push-action@v5
```
francoisferrand: this is outdated: there is a v7 already!

.github/workflows/end2end.yaml
```diff
@@ -351,6 +351,61 @@ jobs:
           cache-from: type=gha,scope=kafka-connect-${{ env.KAFKA_CONNECT_TAG }}
           cache-to: type=gha,mode=max,scope=kafka-connect-${{ env.KAFKA_CONNECT_TAG }}
 
+  build-mongodb-images:
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
```
francoisferrand, line 358: outdated

solution-base/mongodb/how_to_upgrade.md
```diff
@@ -7,7 +7,24 @@ to the new version. Below is a structured guide to follow for a successful
...
+
+### Bumping the MongoDB version in the image
+
+See `solution-base/images/mongodb-sharded/README.md` for the authoritative
+MongoDB image bump and vendoring procedure.
+
+### Re-vendoring upstream Bitnami scripts
+
+When upstream Bitnami script changes need to be pulled in,
+compare `bitnami/containers` and copy the updated `prebuildfs/` / `rootfs/` directories
```
francoisferrand:
```suggestion
compare `bitnami/containers` and merge upstream changes to `prebuildfs/` / `rootfs/` directories
```

solution-base/mongodb/how_to_upgrade.md
```diff
@@ -7,7 +7,24 @@ to the new version. Below is a structured guide to follow for a successful
...
+into the corresponding image directories. See
+`solution-base/images/mongodb-sharded/README.md` for the detailed steps.
+
```
francoisferrand, line 26: See
`solution-base/images/mongodb-sharded/README.md` for the detailed steps to upgrade the images.

solution-base/build.sh
```diff
@@ -64,7 +64,12 @@ MONGODB_MONGOS_RAM_REQUEST="MONGODB_MONGOS_RAM_REQUEST"
 
 function flatten_source_images()
 {
-    yq eval '.* | (.sourceRegistry // "docker.io") + "/" + .image + ":" + .tag' ${SOLUTION_BASE_DIR}/deps.yaml
+    source <( ${SOLUTION_BASE_DIR}/mongodb_build_vars.sh )
+
+    yq eval '.* | (.sourceRegistry // "docker.io") + "/" + .image + ":" + .tag' ${SOLUTION_BASE_DIR}/deps.yaml |
+        sed '/ghcr.io\/scality\/zenko\/mongodb-sharded/ s/$/-'"${MONGODB_BUILD_TREE_HASH}"'/' |
+        sed '/ghcr.io\/scality\/zenko\/mongodb-exporter/ s/$/-'"${MONGODB_BUILD_TREE_HASH}"'/' |
+        sed '/ghcr.io\/scality\/zenko\/os-shell/ s/$/-'"${MONGODB_BUILD_TREE_HASH}"'/'
```
francoisferrand: `sed '/ghcr.io\/scality\/zenko\/mongo/ s/$/-'"${MONGODB_BUILD_TREE_HASH}"'/'` ?

(may nee to rename `os-shell` to `mongodb-os-shell`)

francoisferrand: or -since we anyway build all images here- just drop the "mongo" filter:

```suggestion
        sed '/ghcr.io\/scality\/zenko\// s/$/-'"${MONGODB_BUILD_TREE_HASH}"'/'
```

.github/scripts/end2end/install-kind-dependencies.sh
francoisferrand: c.f. build.sh (and https://github.com/scality/Zenko/pull/2366/changes#r3074201336) : both should do the same (and indeed there is a bug somewhere: builld.sh has "os-shell" while there is mongoldb-shell here...

tests/ctst/package.json
francoisferrand: already fixed upstream, should be dropped

francoisferrand (commented):
not sure about the target branch now, checking with product.
Either way this PR will need to be rebased

solution-base/images/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+## Bumping MongoDB Version
+
+`mongodb-sharded/debian-12/Dockerfile` exposes a single `ARG MONGODB_VERSION`.
+
+To bump MongoDB:
+1. Update `solution-base/deps.yaml` `mongodb-sharded.tag`.
+2. If needed, update `MONGODB_VERSION` default in Dockerfile for local builds.
+3. CI passes `MONGODB_VERSION` from `deps.yaml` during `build-mongodb-images`.
+
```
francoisferrand, line 84: another part to update is the "base" image:
```
FROM docker.io/bitnami/minideb:bookworm@sha256:5b4e544f5d0c4669d3dbe4f4de8d3ca221d430e6a1cf3ac60007387abbbd5363
```

→ this should _at least_ be mentionned here
→ should we move this into a variable -like in the makefile or build script- so simplify updating it, if relevant? or leave it in the Dockerfile so it is in a format known by dependabot, and can get "automated" updates?

benzekrimaha: I added explicit guidance in the README:
Mongo version comes from deps.yaml and is passed as build arg in CI.
Local builds must pass --build-arg MONGODB_VERSION=<version> explicitly.
Base image digest updates are now explicitly documented (FROM ...@sha256:...).
For now I kept the base image digest in Dockerfiles (not moved to Make/build vars), so dependency tooling can detect and propose digest updates.

solution-base/mongodb/how_to_upgrade.md
benzekrimaha: Addressed. I flattened the layout under images and now keep one folder per image (mongodb-sharded/, mongodb-exporter/, os-shell/) plus README.md and Makefile.

solution-base/images/Makefile
benzekrimaha: Addressed. I removed free-form CONTAINER/BITNAMI_PREFIX usage and moved all allowed values/ref mappings into fixed Makefile rules.

solution-base/images/Makefile
benzekrimaha: mongodb-sharded is now split from a pinned upstream commit that still contains bitnami/mongodb-sharded/8.0/debian-12, while mongodb-exporter and os-shell are still split from upstream main. This is now explicit in Makefile.

solution-base/images/Makefile
benzekrimaha: Prefixes and refs are now defined in the Makefile itself, not passed from CLI arguments. See `Makefile`.

solution-base/images/Makefile
benzekrimaha: I removed first-import usage and kept the workflow focused on sync/merge. Docs now describe only vendor-sync.

solution-base/images/mongodb-sharded/README.md
benzekrimaha: I removed those first-import commands from docs. The process now uses the Makefile-managed sync flow.

solution-base/images/mongodb-sharded/README.md
benzekrimaha: Docs now use a single command (make -C solution-base/images vendor-sync) and all concrete values stay in Makefile

solution-base/images/mongodb-sharded/README.md
benzekrimaha: README now explicitly says to add dedicated local tweak commits after each upstream merge, and shows expected history shape.

solution-base/images/mongodb-sharded/README.md
benzekrimaha: ARG MONGODB_VERSION now has no default in the MongoDB Dockerfile, and value is passed from deps/CI

.github/workflows/end2end.yaml
benzekrimaha: Addressed for Mongo builds. The Mongo image build steps now use docker/build-push-action@v7.

.github/scripts/end2end/install-kind-dependencies.sh
```diff
@@ -233,7 +233,10 @@ build_solution_base_manifests() {
 
 get_image_from_deps() {
     local dep_name=$1
-    yq eval ".$dep_name | (.sourceRegistry // \"docker.io\") + \"/\" + .image + \":\" + .tag" $SOLUTION_BASE_DIR/deps.yaml
+        source <( "$SOLUTION_BASE_DIR/mongodb_build_vars.sh" )
+
```
francoisferrand: extra spaces at EOL

.github/scripts/end2end/install-kind-dependencies.sh
```diff
@@ -233,7 +233,10 @@ build_solution_base_manifests() {
 
 get_image_from_deps() {
     local dep_name=$1
-    yq eval ".$dep_name | (.sourceRegistry // \"docker.io\") + \"/\" + .image + \":\" + .tag" $SOLUTION_BASE_DIR/deps.yaml
+        source <( "$SOLUTION_BASE_DIR/mongodb_build_vars.sh" )
```
francoisferrand: bad indent

.github/workflows/end2end.yaml
francoisferrand: @benzekrimaha

solution-base/images/Makefile
```diff
@@ -0,0 +1,43 @@
+BITNAMI_REMOTE := bitnami-containers
+BITNAMI_REPO := https://github.com/bitnami/containers.git
+
+BITNAMI_MONGODB_SHARDED_REF := 48a109547d39cd8cf8a5d4058d832ecb5844829e
```
francoisferrand: kind of unexpected to have just a commit hash here : _I_ know this is needed because the 8.0 images have been removed from the upstream repo since then, but worth adding a comment here to explain this is the "latest" commit on main which has the 8.0 image

solution-base/mongodb/how_to_upgrade.md
```diff
@@ -7,7 +7,24 @@ to the new version. Below is a structured guide to follow for a successful
 upgrade (note that all along we will be taking as an example the upgrade from
 MongoDB 8.0.10 to 9.3.6):
 
-## Upgrade steps
+## Image Build (Stop-Gap)
```
francoisferrand:
```suggestion
## Image Build
```

solution-base/mongodb/how_to_upgrade.md
```diff
@@ -7,7 +7,24 @@ to the new version. Below is a structured guide to follow for a successful
...
+
+Zenko now builds its own MongoDB Sharded images from vendored Bitnami scripts.
+Image sources live under `solution-base/images/`.
+
+### Bumping the MongoDB version in the image
+
+See `solution-base/images/README.md` for the authoritative
+MongoDB image bump and vendoring procedure.
+
```
francoisferrand: That is a bit harsh, when reading this we don't even know why we should read it.

```suggestion
- Upstream images are vendored, so the code mus be synced with upstream changes.
- If a new MongoDB baseline (8.2, 9.0...) is available upstream, must switch to that baseline.
- Base images are pinned to specific image hash, and must be updated.

See `solution-base/images/README.md` for details.
```

solution-base/build.sh
```diff
@@ -64,7 +64,10 @@ MONGODB_MONGOS_RAM_REQUEST="MONGODB_MONGOS_RAM_REQUEST"
 
 function flatten_source_images()
 {
-    yq eval '.* | (.sourceRegistry // "docker.io") + "/" + .image + ":" + .tag' ${SOLUTION_BASE_DIR}/deps.yaml
+    source <( ${SOLUTION_BASE_DIR}/mongodb_build_vars.sh )
+
+    yq eval '.* | (.sourceRegistry // "docker.io") + "/" + .image + ":" + .tag' ${SOLUTION_BASE_DIR}/deps.yaml |
+    sed '/ghcr.io\/scality\/zenko\// s/$/-'"${MONGODB_BUILD_TREE_HASH}"'/'
```
francoisferrand: missing indent for readability
```suggestion
        sed '/ghcr.io\/scality\/zenko\// s/$/-'"${MONGODB_BUILD_TREE_HASH}"'/'
```

solution-base/deps.yaml
```diff
@@ -2,11 +2,14 @@
 # to sort keys, use the following command
 # yq eval 'sortKeys(.)' -i deps.yaml
 mongodb-sharded:
-  image: bitnamilegacy/mongodb-sharded
-  tag: 8.0.13-debian-12-r0
+  sourceRegistry: ghcr.io
+  image: scality/zenko/mongodb-sharded
+  tag: "8.0.13"
```
francoisferrand, line 7: nit: we are removing the debian-12 "information" from the image tag. Is it because we consider it useless/redundant, or just to simplify script?

solution-base/images/Makefile
```diff
@@ -0,0 +1,43 @@
...
+	-git branch -D $(VENDOR_OS_SHELL_BRANCH)
+	git subtree split --prefix=$(BITNAMI_MONGODB_SHARDED_PREFIX) $(BITNAMI_MONGODB_SHARDED_REF) -b $(VENDOR_MONGODB_SHARDED_BRANCH)
+	git subtree split --prefix=$(BITNAMI_MONGODB_EXPORTER_PREFIX) $(BITNAMI_MONGODB_EXPORTER_REF) -b $(VENDOR_MONGODB_EXPORTER_BRANCH)
+	git subtree split --prefix=$(BITNAMI_OS_SHELL_PREFIX) $(BITNAMI_OS_SHELL_REF) -b $(VENDOR_OS_SHELL_BRANCH)
+
+vendor-sync: update-vendor-branches vendor-sync-mongodb-sharded vendor-sync-mongodb-exporter vendor-sync-os-shell
+
+vendor-sync-mongodb-sharded:
+	git subtree merge --prefix=solution-base/images/mongodb-sharded/debian-12 $(VENDOR_MONGODB_SHARDED_BRANCH) --squash
+
```
francoisferrand: lots of redundancy with all the variables:
- merge prefix : solution-base/images/mongodb-sharded/debian-12
- BITNAMI_MONGODB_SHARDED_PREFIX : bitnami/mongodb-sharded/8.0/debian-12
- BITNAMI_MONGODB_SHARDED_REF : vendor/mongodb-sharded-debian-12

when there is only 2 useful information:
- name of the image : mongodb-sharded
- "subpath" of the image in upstream: 8.0/debian-12
- (and associated ref obviously)

to simplify maintenance, we should deduplicate this, likely using make "[pattern rules](https://www.gnu.org/software/make/manual/html_node/Pattern-Rules.html)" and [computed variable names](https://www.gnu.org/software/make/manual/html_node/Computed-Names.html)

```suggestion
BITNAMI_mongodb-sharded_PATH=8.0/debian-12
BITNAMI_mongodb-sharded_REF=48a109547d39cd8cf8a5d4058d832ecb5844829e
[...]

vendor-sync-%: fetch-remote
    -git branch -D vendor/$*/$(BITNAMI_$*_PATH)
    git subtree split --prefix=bitnami/$*/$(BITNAMI_$*_PATH) $(BITNAMI_$*_REF) -b vendor/mongodb-sharded/$(BITNAMI_$*_PATH)
    git subtree merge --prefix=solution-base/images/$* vendor/mongodb-sharded/$(BITNAMI_$*_PATH) --squash

vendor-sync: vendor-sync-mongodb-sharded vendor-sync-mongodb-exporter vendor-sync-os-shell
```

solution-base/images/Makefile
```diff
@@ -0,0 +1,43 @@
...
+BITNAMI_MONGODB_SHARDED_PREFIX := bitnami/mongodb-sharded/8.0/debian-12
+BITNAMI_MONGODB_EXPORTER_REF := $(BITNAMI_REMOTE)/main
+BITNAMI_MONGODB_EXPORTER_PREFIX := bitnami/mongodb-exporter/0/debian-12
+BITNAMI_OS_SHELL_REF := $(BITNAMI_REMOTE)/main
+BITNAMI_OS_SHELL_PREFIX := bitnami/os-shell/12/debian-12
+
+VENDOR_MONGODB_SHARDED_BRANCH := vendor/mongodb-sharded-debian-12
+VENDOR_MONGODB_EXPORTER_BRANCH := vendor/mongodb-exporter-debian-12
+VENDOR_OS_SHELL_BRANCH := vendor/os-shell-debian-12
+
```
francoisferrand: similar to how you keep the `debian-12` in the branch name, best to keep the version as well:

```suggestion
VENDOR_MONGODB_SHARDED_BRANCH := vendor/mongodb-sharded/8.0/debian-12
VENDOR_MONGODB_EXPORTER_BRANCH := vendor/mongodb-exporter/0/debian-12
VENDOR_OS_SHELL_BRANCH := vendor/os-shell/12/debian-12
```

solution-base/images/Makefile
benzekrimaha: I added a comment explaining that the pinned hash is the latest upstream commit known to still contain bitnami/mongodb-sharded/8.0/debian-12.

solution-base/deps.yaml
benzekrimaha: Intentional. We keep distro information in the source layout and Dockerfile base digest, and use clean semantic app version tags in `deps.yaml` for consistency with build args and immutable tree-hash tagging. If preferred, we can restore distro suffixes, but current flow does not rely on them.

solution-base/images/Makefile
```diff
@@ -1,43 +1,38 @@
...
-update-vendor-branches: fetch-remote
-	-git branch -D $(VENDOR_MONGODB_SHARDED_BRANCH)
-	-git branch -D $(VENDOR_MONGODB_EXPORTER_BRANCH)
-	-git branch -D $(VENDOR_OS_SHELL_BRANCH)
-	git subtree split --prefix=$(BITNAMI_MONGODB_SHARDED_PREFIX) $(BITNAMI_MONGODB_SHARDED_REF) -b $(VENDOR_MONGODB_SHARDED_BRANCH)
-	git subtree split --prefix=$(BITNAMI_MONGODB_EXPORTER_PREFIX) $(BITNAMI_MONGODB_EXPORTER_REF) -b $(VENDOR_MONGODB_EXPORTER_BRANCH)
-	git subtree split --prefix=$(BITNAMI_OS_SHELL_PREFIX) $(BITNAMI_OS_SHELL_REF) -b $(VENDOR_OS_SHELL_BRANCH)
-
-vendor-sync: update-vendor-branches vendor-sync-mongodb-sharded vendor-sync-mongodb-exporter vendor-sync-os-shell
+	git fetch $(BITNAMI_REMOTE) $(BITNAMI_mongodb_sharded_REF) --depth=1
```
francoisferrand, line 29: should we not fetch whole history?
- so the subtree split can indeed see all the changes that happened on the subtree
- this way we also don't need to fetch specifically the mongodb_sharded ref, and fully rely on the refs set in the variables

benzekrimaha: I updated Makefile accordingly:
- clarified that main refers to Bitnami upstream  by introducing BITNAMI_UPSTREAM_MAIN_REF.
- removed shallow/special-case fetches and now fetch full Bitnami main history once in fetch-remote.
- kept refs centralized in variables (BITNAMI_*_REF) so subtree split fully relies on declared refs.
This should make subtree split history-complete and remove the ad-hoc extra fetch for mongodb-sharded.

SylvainSenechal (approved):

.github/workflows/end2end.yaml
```diff
@@ -351,6 +351,61 @@ jobs:
           cache-from: type=gha,scope=kafka-connect-${{ env.KAFKA_CONNECT_TAG }}
           cache-to: type=gha,mode=max,scope=kafka-connect-${{ env.KAFKA_CONNECT_TAG }}
 
+  build-mongodb-images:
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v5
```
francoisferrand: not up to date

francoisferrand: same for a few others below

francoisferrand (approved):

francoisferrand:
@benzekrimaha This needs to be rebased on 2.14 and change target branch to 2.14
(there are quite a few PRs in queue though, may be best to wait a bit for things to clear up)

PR #2369, opened 2026-03-30
add tests for server side encryption, file backend encryption only

SylvainSenechal:
Issue: ZENKO-5239

### DIFF
--- a/tests/ctst/features/serverSideEncryption.feature
+++ b/tests/ctst/features/serverSideEncryption.feature
@@ -1,0 +1,87 @@
+Feature: Server Side Encryption
+
+    @2.14.0
+    @PreMerge
+    @ServerSideEncryption
+    @ServerSideEncryptionFileBackend
+    Scenario Outline: should encrypt object when bucket encryption is <bucketAlgo> and object encryption is <objectAlgo>
+        Given a "Non versioned" bucket
+        And bucket encryption is set to "<bucketAlgo>" with key "<bucketKeyId>"
+        Then the bucket encryption is verified for algorithm "<bucketAlgo>" and key "<bucketKeyId>"
+        When an object "<objectName>" is uploaded with SSE algorithm "<objectAlgo>" and key "<objectKeyId>"
+        Then the PutObject response should have SSE algorithm "<expectedAlgo>" and KMS key "<expectedKeyId>"
+        Then the GetObject should return the uploaded body with SSE algorithm "<expectedAlgo>" and KMS key "<expectedKeyId>"
+
+        Examples: No bucket encryption
+            | objectName     | bucketAlgo | bucketKeyId | objectAlgo | objectKeyId  | expectedAlgo | expectedKeyId |
+            | no-enc-none    |            |             |            |              |              | absent        |
+            | no-enc-aes     |            |             | AES256     |              | AES256       | absent        |
+            | no-enc-kms     |            |             | aws:kms    |              | aws:kms      | generated     |
+            | no-enc-kms-key |            |             | aws:kms    | custom-key-1 | aws:kms      | custom-key-1  |
+
+        Examples: Bucket AES256
+            | objectName      | bucketAlgo | bucketKeyId | objectAlgo | objectKeyId  | expectedAlgo | expectedKeyId |
+            | bkt-aes-none    | AES256     |             |            |              | AES256       | absent        |
+            | bkt-aes-aes     | AES256     |             | AES256     |              | AES256       | absent        |
+            | bkt-aes-kms     | AES256     |             | aws:kms    |              | aws:kms      | generated     |
+            | bkt-aes-kms-key | AES256     |             | aws:kms    | custom-key-1 | aws:kms      | custom-key-1  |
+
+        Examples: Bucket aws:kms (default key)
+            | objectName      | bucketAlgo | bucketKeyId | objectAlgo | objectKeyId  | expectedAlgo | expectedKeyId |
+            | bkt-kms-none    | aws:kms    |             |            |              | aws:kms      | generated     |
+            | bkt-kms-aes     | aws:kms    |             | AES256     |              | AES256       | absent        |
+            | bkt-kms-kms     | aws:kms    |             | aws:kms    |              | aws:kms      | generated     |
+            | bkt-kms-kms-key | aws:kms    |             | aws:kms    | custom-key-1 | aws:kms      | custom-key-1  |
+
+        Examples: Bucket aws:kms with custom key
+            | objectName         | bucketAlgo | bucketKeyId | objectAlgo | objectKeyId  | expectedAlgo | expectedKeyId |
+            | bkt-kmskey-none    | aws:kms    | bucket-key  |            |              | aws:kms      | bucket-key    |
+            | bkt-kmskey-aes     | aws:kms    | bucket-key  | AES256     |              | AES256       | absent        |
+            | bkt-kmskey-kms     | aws:kms    | bucket-key  | aws:kms    |              | aws:kms      | bucket-key    |
+            | bkt-kmskey-kms-key | aws:kms    | bucket-key  | aws:kms    | custom-key-2 | aws:kms      | custom-key-2  |
+
+    @2.14.0
+    @PreMerge
+    @ServerSideEncryption
+    @ServerSideEncryptionFileBackend
+    Scenario: DeleteBucketEncryption removes default encryption
+        Given a "Non versioned" bucket
+        And bucket encryption is set to "AES256" with key ""
+        When an object "enc-obj" is uploaded with SSE algorithm "" and key ""
+        Then the GetObject should return the uploaded body with SSE algorithm "AES256" and KMS key "absent"
+        When the user deletes bucket encryption
+        Then the GetObject should return the uploaded body with SSE algorithm "AES256" and KMS key "absent"
+        When an object "plain-obj" is uploaded with SSE algorithm "" and key ""
+        Then the GetObject should return the uploaded body with SSE algorithm "" and KMS key "absent"
+
+    @2.14.0
+    @PreMerge
+    @ServerSideEncryption
+    @ServerSideEncryptionFileBackend
+    Scenario Outline: PutObject with invalid SSE parameters returns an error: <objectName>
+        Given a "Non versioned" bucket
+        When an object "<objectName>" is uploaded with SSE algorithm "<algo>" and key "<keyId>"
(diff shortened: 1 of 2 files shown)


tests/ctst/features/serverSideEncryption.feature
SylvainSenechal, line 1: Here, its not obvious from the name of the files and tests, but we are only testing encryption for the "file" backend. So more tests coming later in other prs.
Later we will need to test pretty much the same thing, but for KMIP and AWS-KMS backends. I suspect the test will be similar, but might have a few differences, and more importantly, the setup will be heavier : 

For KMIP, will need to setup a PyKMIP pod inside the cluster, for AWS-KMS, I found a mock on github already used in cloudserver, so i will check this.
And these will also trigger reconciliation on cloudserver because they need their own config/env variables to work. 

When we get to that point, I will retouch the filename/test names to make all that clearer, and probably put everything under a folder

tests/ctst/features/serverSideEncryption.feature
```diff
@@ -0,0 +1,87 @@
...
+    @2.14.0
+    @PreMerge
+    @ServerSideEncryption
+    @ServerSideEncryptionFileBackend
+    Scenario Outline: Encryption behavior matrix: bucket=<bucketAlgo> key=<bucketKeyId>, object=<objectAlgo> key=<objectKeyId>
+        Given a "Non versioned" bucket
+        And bucket encryption is set to "<bucketAlgo>" with key "<bucketKeyId>"
+        Then the bucket encryption is verified for algorithm "<bucketAlgo>" and key "<bucketKeyId>"
+        When an object "<objectName>" is uploaded with SSE algorithm "<objectAlgo>" and key "<objectKeyId>"
+        # TODO: uncomment when CloudServer returns SSE headers in PutObject response https://github.com/scality/cloudserver/pull/6122
```
SylvainSenechal: I have other pr coming later, I can wait now, or uncomment later in another pr, because we need to wait for a cloudserver release

tests/ctst/features/serverSideEncryption.feature
```diff
@@ -0,0 +1,87 @@
...
+        Then the GetObject should return the uploaded body with SSE algorithm "AES256" and KMS key "absent"
+        When the user deletes bucket encryption
+        And an object "plain-obj" is uploaded with SSE algorithm "" and key ""
+        Then the GetObject should return the uploaded body with SSE algorithm "" and KMS key "absent"
+
+    @2.14.0
+    @PreMerge
+    @ServerSideEncryption
+    @ServerSideEncryptionFileBackend
+    Scenario Outline: PutObject with invalid SSE parameters returns an error: <objectName>
```
SylvainSenechal, line 61: These last 3 scenarios are maybe not so relevant for functional tests, test in cloudserver would probably be enough but I guess its fine

tests/ctst/steps/serverSideEncryption.ts
```diff
@@ -0,0 +1,215 @@
...
+import {
+    S3Client,
+    GetObjectCommand,
+    PutObjectCommand,
+    type ServerSideEncryption,
+} from '@aws-sdk/client-s3';
+import Zenko from 'world/Zenko';
+import assert from 'assert';
+
+// We use the AWS SDK directly instead of cli-testing for PutObject and GetObject
```
SylvainSenechal, line 12: This is more annoying than I thought, even later if we wanna parallelize **coldStorage** test, one of the test is using getObject..

tests/ctst/features/serverSideEncryption.feature
francoisferrand: when it comes to that : if we already cover every (corner) case in Cloudserver, do we really need to test everything here?
i.e. in face of the reconciliation impact, maybe just validating that we can handle the "nominal" case (kmip?) end-2-end (i.e. system config, deployment, activation, etc...) could be enough? what do you think?

(just food for thought though, no impact on this specific PR indeed)

tests/ctst/features/serverSideEncryption.feature
francoisferrand: should this PR before before cloudserver PR, or does it need to be updated?

tests/ctst/features/serverSideEncryption.feature
```diff
@@ -0,0 +1,87 @@
...
+    @ServerSideEncryption
+    @ServerSideEncryptionFileBackend
+    Scenario: DeleteBucketEncryption removes default encryption
+        Given a "Non versioned" bucket
+        And bucket encryption is set to "AES256" with key ""
+        When an object "enc-obj" is uploaded with SSE algorithm "" and key ""
+        Then the GetObject should return the uploaded body with SSE algorithm "AES256" and KMS key "absent"
+        When the user deletes bucket encryption
+        And an object "plain-obj" is uploaded with SSE algorithm "" and key ""
+        Then the GetObject should return the uploaded body with SSE algorithm "" and KMS key "absent"
```
francoisferrand, line 55: shouldn't we also check that we can still properly retrieve the "enc-obj" object?

tests/ctst/features/serverSideEncryption.feature
francoisferrand: Can you double check we have the tests in Cloudserver, though?

tests/ctst/steps/serverSideEncryption.ts
```diff
@@ -0,0 +1,215 @@
...
+// because:
+// - cli-testing has a casing bug: --ssekms-key-id → SsekmsKeyId (should be
+//   SSEKMSKeyId), so the KMS key id is silently dropped on PutObject.
+// - cli-testing's S3.getObject writes the body to a shared file (out.loc)
+//   which races under parallel execution.
```
francoisferrand, line 17: shouldn't we fix these (in CTST): seems easy enough to fix the case and use a dynamically allocated temp file name?
(even if there is another long term solution, nothing settled there it seems: "consider dropping"...)

tests/ctst/steps/serverSideEncryption.ts
```diff
@@ -0,0 +1,215 @@
...
+        // } else {
+        //     assert.strictEqual(result.serverSideEncryption, undefined,
+        //         `PutObject SSE: expected absent, got "${result.serverSideEncryption}"`);
+        // }
+        // if (expectedKey === 'present') {
+        //     assert.ok(result.sseKmsKeyId, 'PutObject: SSEKMSKeyId should be present');
+        // } else {
+        //     assert.strictEqual(result.sseKmsKeyId, undefined,
+        //         `PutObject: SSEKMSKeyId should be absent, got "${result.sseKmsKeyId}"`);
+        // }
```
delthas: Opening a review thread so we remember to uncomment before merging.

delthas (approved):

tests/ctst/features/serverSideEncryption.feature
SylvainSenechal: Yes it makes sense, for next time need to think more about what we put in the tests here, as we already have tests in other codebase

tests/ctst/features/serverSideEncryption.feature
SylvainSenechal: the cloudserver pr is merged, but cloudserver is not bumped yet, we need a new version of cloudserver

tests/ctst/features/serverSideEncryption.feature
SylvainSenechal: half of them are covered, so yeah it's very moderately useful

tests/ctst/steps/serverSideEncryption.ts
SylvainSenechal: The issue I have is cli testing is still stuck on an older version, bumping it in zenko is what I did in another pr but it's a lot of change and not merged yet
I can't even make the change on a lower version of cli testing because the branch 1.2 of cli testing also has changes that requires lots of changes in zenko regarding keycloak seeding and probably other things
I don't really like this coupling where we have to make updates to cli testing whenever we want to add a test in zenko

tests/ctst/features/serverSideEncryption.feature
SylvainSenechal: yes, added

tests/ctst/features/serverSideEncryption.feature
```diff
@@ -0,0 +1,88 @@
+Feature: Server Side Encryption
+
+    @2.14.0
+    @PreMerge
+    @ServerSideEncryption
+    @ServerSideEncryptionFileBackend
+    Scenario Outline: Encryption behavior matrix: bucket=<bucketAlgo> key=<bucketKeyId>, object=<objectAlgo> key=<objectKeyId>
```
francoisferrand: should have a more "human friendly" name : BDD is not about having a table, but about _describing_ the actual scenario...

something like:
```suggestion
    Scenario Outline: should encrypt object when bucket encryption is <bucketAlgo> and object encryption is <objectAlgo>
```

tests/ctst/steps/serverSideEncryption.ts
francoisferrand: > The issue I have is cli testing is still stuck on an older version, bumping it in zenko is what I did in another pr but it's a lot of change and not merged yet

should create followup tickets?
(esp. instead of comments, which is not a good way to track remaining work)

> I don't really like this coupling where we have to make updates to cli testing whenever we want to add a test in zenko
👍

Though it was not _supposed_ to be the case : CTST s3 client was supposed to be auto-generated, so we always had everything available... not sure why/how it broke (or if this s3 layer is still relevant...)

SylvainSenechal: I linked this ticket line 18 https://scality.atlassian.net/browse/ZENKO-5247

tests/ctst/features/serverSideEncryption.feature
SylvainSenechal: cloudserver is released, no need to comment this anymore

tests/ctst/features/serverSideEncryption.feature
francoisferrand: so half are **not** covered : can you create a ticket in cloudserver?

francoisferrand (approved):

tests/ctst/features/serverSideEncryption.feature
francoisferrand: then you need to rebase on top of https://github.com/scality/Zenko/tree/refs/heads/w/2.14/improvement/ZENKO-5242, otherwise you don't have the bump yet...

PR #2372, opened 2026-04-07
Run cucumber tests out of cluster

SylvainSenechal:
Issue: ZENKO-5209

-Cucumber tests are run directly in the github runner, no need for a pod anymore. This is quite nice for codespace testing where I got the opportunity to test it as I was making the pr and fixing the few bugs that didn't work
- Cleanup a few unused files, related to this pr and also a few unrelated
- updated readme to explain how to run the tests
- fixed a bug in one of the bash setup file that was breaking one of the zenko tests
- fixed an issue with issue location creation, stopping the codespace from being setup properly because one of the gcp location env var isn't defined in codespace, so we were creating an empty location

Future work : 
- Some cleanup around env variables vs world parameters, its a bit messy now, we could get rid of ctst world parameters and just use env variables everywhere
- Unify tests (zenko mocha, cucumber) setup, unify tests package under the same js project.

Once these 2 are done, I believe it will be much easier to work on other tasks, especially everything dealing with the setup in the bash and python files


.github/scripts/end2end/setup-e2e-env.sh
SylvainSenechal, line 1: Not a big fan of this file, quite long, with a flag depending if it runs for ctst or zenko tests.
In a next pr i wanna refine it to make it work for all tests without flags

I also wanna do something about world parameters vs env variables. 

Currently for cucumber tests, what we have is :
- Env var XXX is defined in end2end.yaml
- In this file, we create a world parameters thats Xxx=XXX
- Cucumber tests uses this.parameters.Xxx

We could remove one level of indirection and directly use env variables in cucumber tests.
This would also be helpful to track env variable utilisation better, because here the world parameters are sometimes renamed, example : 

KeycloakUsername = $OIDC_USERNAME

So when grepping OIDC_USERNAME, you can't directly see that it's used because the test uses this.parameters.KeycloakUsername

.github/scripts/mocks/azure/docker-entrypoint.sh
SylvainSenechal, line 1: Not directly related to this pr but didn't want to create a ticket just for that : 
azure/docker-entrypoint and the following dockerfile aren't used, what's used is the azure-mock.yaml file where the image is directly defined inside of it.

Btw, in the deleted dockerfile, there is a typo in the azurite version with an extra whitespace, so not even sure it was working before 🧐

.github/scripts/mocks/aws-mock.yaml
```diff
@@ -69,8 +69,6 @@ spec:
   containers:
   - image: ghcr.io/scality/cloudserver:8.8.59
     name: aws-mock
-    ports:
```
SylvainSenechal, line 72: defined again line 83

.github/workflows/end2end.yaml
```diff
@@ -407,42 +405,9 @@ jobs:
...
-            SORBET_IMAGE=${{ env.SORBET_IMAGE }}
-            SORBET_TAG=${{ env.SORBET_TAG }}
-            DRCTL_IMAGE=${{ env.DRCTL_IMAGE }}
-            DRCTL_TAG=${{ env.DRCTL_TAG }}
-          tags: "${{ env.E2E_CTST_IMAGE_NAME }}:${{ env.E2E_IMAGE_TAG }}"
-          cache-from: type=gha,scope=end2end-ctst
-          cache-to: type=gha,mode=max,scope=end2end-ctst
 
   end2end-pra:
-    needs: [build-kafka, lint-and-build-ctst]
```
SylvainSenechal, line 446: Can discuss this, I removed the lint dependency to start running the test, still needed for final ci result though

My argument is, when i push some code before lunch break to have it run for one hour while i eat, i don't wanna come back only to realize the tests never started because i forgot a comma

tests/ctst/steps/dr/drctl.ts
```diff
@@ -94,8 +94,8 @@ type BootstrapLoadConfig = {
 };
 
 type VolumeGetConfig = {
-    targetZenkoKubeconfigPath?: string;
```
SylvainSenechal, line 97: These are translated into cli commands for drctlt, and they were misnamed

.devcontainer/README.md
```diff
@@ -39,47 +39,36 @@ yarn mocha --exit -t 10000 --grep "should list objects" --recursive cloudserver/
...
 ```
 
 ## Accessing s3 service
 
 This devcontainer is a full Zenko development environment.
 Once you start a new Codespace, you will have a full Zenko stack running in Kubernetes after a few minutes.
 
-After deployment is done, which you can follow by opening another terminal, you will be able to access S3 service through a port-forward.
-First find a cloudserver connector using the following command:
+After deployment is done, the devcontainer setup configures ingress endpoints and `/etc/hosts` entries
```
francoisferrand, line 50: nit: not for now, but using `/etc/hosts` is probably a nogo for local development... so we'll eventually need to come back to this

.devcontainer/README.md
```diff
@@ -39,47 +39,36 @@ yarn mocha --exit -t 10000 --grep "should list objects" --recursive cloudserver/
...
+aws configure set aws_secret_access_key $SECRET_KEY
+aws configure set region us-east-1
 ```
 
-Now you can use aws cli to interact with the S3 service
+Use it:
 
 ```bash
-    aws s3 ls --endpoint http://localhost:8080
+aws s3 ls --endpoint http://s3.zenko.local
```
francoisferrand: endpoint can also be set in configuration. For exemple in my .aws/config I have defined a scaleway profile:

```
[profile scaleway]
endpoint_url = https://s3.fr-par.scw.cloud
region = fr-par
```

.devcontainer/README.md
```diff
@@ -39,47 +39,36 @@ yarn mocha --exit -t 10000 --grep "should list objects" --recursive cloudserver/
...
+Configure the AWS CLI:
 
 ```bash
-    aws configure set aws_access_key_id $ACCESS_KEY
-    aws configure set aws_secret_access_key $SECRET_KEY
-    aws configure set region us-east-1
+aws configure set aws_access_key_id $ACCESS_KEY
+aws configure set aws_secret_access_key $SECRET_KEY
+aws configure set region us-east-1
 ```
```
francoisferrand, line 67: nit: should this be done by setup-e2e-env ?

.devcontainer/setup.sh
```diff
@@ -67,10 +67,24 @@ done
     bash configure-e2e-ctst.sh
 )
 
-# Build CTST image from current branch
+docker image prune -af
```
francoisferrand: should not be done:
- certainly not appropriate if running locally
- even if running in codespace, either images are preloaded (and we should not drop them / or avoid having them) or this is useless...

.devcontainer/setup.sh
```diff
@@ -67,10 +67,24 @@ done
...
 DRCTL_TAG=$(yq eval '.drctl.tag' solution/deps.yaml)
-TAG_NAME=ctst_codespace_setup
-GIT_AUTH_TOKEN=$GITHUB_TOKEN docker build --secret id=GIT_AUTH_TOKEN --build-arg SORBET_TAG=$SORBET_TAG --build-arg DRCTL_TAG=$DRCTL_TAG -t $E2E_CTST_IMAGE_NAME:$TAG_NAME ./tests/ctst
-kind load docker-image ${E2E_CTST_IMAGE_NAME}:$TAG_NAME
-docker rmi ${E2E_CTST_IMAGE_NAME}:$TAG_NAME
+
+_cid=$(docker create "${SORBET_IMAGE}:${SORBET_TAG}" true)
+docker cp "${_cid}:/sorbetctl" "${CTST_DIR}/sorbetctl"
+docker rm "${_cid}" >/dev/null
+chmod +x "${CTST_DIR}/sorbetctl"
```
francoisferrand, line 81: `sorbetctl` should not be needed: the recommended way use it is to "exec" in sorbet "forwarder" pod (where it already has many parameters, thanks to the config file already present for the forwarder)

(but this may be for later, as it may require more changes to the tests)

.devcontainer/setup.sh
```diff
@@ -67,10 +67,24 @@ done
...
+
+_cid=$(docker create "${DRCTL_IMAGE}:${DRCTL_TAG}" true)
+docker cp "${_cid}:/zenko-drctl" "${CTST_DIR}/zenko-drctl"
+docker rm "${_cid}" >/dev/null
+chmod +x "${CTST_DIR}/zenko-drctl"
+
+# Install CTST test dependencies
+(cd "${CTST_DIR}" && yarn install)
```
francoisferrand: * this should be done just one, during _build_ : e.g. in devcontariner's Dockerfile. this way we don't need to fetch packages every time
* must use `--frozen-lock-file` flag, to ensure we use the packages that were already resolved

.github/scripts/end2end/setup-e2e-env.sh
```diff
@@ -38,12 +39,7 @@ MONGO_NS=$(echo "${MONGO_FQDN}" | cut -d. -f2)
 # Only start port-forward if not already listening
 if ! ss -tlnp 2>/dev/null | grep -q ":${MONGO_PORT}" && \
    ! lsof -i ":${MONGO_PORT}" &>/dev/null; then
-    kubectl port-forward -n "${MONGO_NS}" "svc/${MONGO_SVC}" "${MONGO_PORT}:${MONGO_PORT}" &
-    _MONGO_PF_PID=$!
-    if [ -z "${_SETUP_E2E_CLEANUP_SET:-}" ]; then
-        trap "kill ${_MONGO_PF_PID} 2>/dev/null || true" EXIT
-        export _SETUP_E2E_CLEANUP_SET=1
-    fi
+    kubectl port-forward -n "${MONGO_NS}" "svc/${MONGO_SVC}" "${MONGO_PORT}:${MONGO_PORT}" &>/dev/null &
```
francoisferrand, line 42: this one is using port-forward, while ingresses are used elsewhere : any reason not to do the same everywhere?

.github/scripts/end2end/setup-e2e-env.sh
```diff
@@ -119,12 +116,235 @@ if kubectl get namespace metadata &>/dev/null; then
...
+    fi
+    export KAFKA_AUTH_HOST_PORT="${KAFKA_AUTH_LOOPBACK}:${KAFKA_AUTH_PORT}"
+
+    # Point the auth broker's advertised hostname to the dedicated loopback
+    KAFKA_AUTH_FQDN="${KAFKA_AUTH_HOST}.${NAMESPACE}.svc.cluster.local"
+    if ! grep -qw "${KAFKA_AUTH_FQDN}" /etc/hosts 2>/dev/null; then
+        echo "${KAFKA_AUTH_LOOPBACK} ${KAFKA_AUTH_HOST} ${KAFKA_AUTH_FQDN}" | sudo tee -a /etc/hosts
+    fi
+
+    # Prometheus — port-forward for CTST PRA tests
```
francoisferrand, line 227: same: why port-forward instead of ingress like other services?

.github/scripts/end2end/setup-e2e-env.sh
```diff
@@ -119,12 +116,235 @@ if kubectl get namespace metadata &>/dev/null; then
...
+        docker rm "${_cid}" >/dev/null
+        chmod +x "${CTST_DIR}/sorbetctl"
+    fi
+    if [ ! -f "${CTST_DIR}/zenko-drctl" ]; then
+        _cid=$(docker create "${DRCTL_IMAGE}:${DRCTL_TAG}" true)
+        docker cp "${_cid}:/zenko-drctl" "${CTST_DIR}/zenko-drctl"
+        docker rm "${_cid}" >/dev/null
+        chmod +x "${CTST_DIR}/zenko-drctl"
+    fi
+
```
francoisferrand, line 282: redundant: already done in `setup.sh` : should do it just once

.github/scripts/end2end/setup-e2e-env.sh
```diff
@@ -119,12 +116,235 @@ if kubectl get namespace metadata &>/dev/null; then
...
+# --- 10. CTST-specific setup (Kafka, service users, Zenko CR data) ---
+# =====================================================================
+# This block is skipped when SKIP_CTST=1 (e.g., mocha-only jobs).
+# CTST setup runs by default; non-CTST callers must set SKIP_CTST=1.
+
+ZENKO_ROOT="$(cd "$_SETUP_DIR/../../.." && pwd)"
+export TARGET_VERSION=$(sed -n 's/^VERSION="\([^"]*\)"/\1/p' "${ZENKO_ROOT}/VERSION")
+
+if [ "${SKIP_CTST:-}" = "1" ]; then
+    echo "SKIP_CTST=1 set, skipping CTST-specific setup"
```
francoisferrand, line 132: do we actually need to skip?
more cleanup can be done in a followup PR, but it seems that doing this is harmless and can be done as well to run zenko tests?

.github/workflows/end2end.yaml
francoisferrand: the dependency was really on the _build_ part of `lint-and-build-ctst` : since we don't use that docker image, not really needed anymore...

.github/workflows/end2end.yaml
francoisferrand, line 402: should we also do a `yarn build`, to valide typescript code?

.github/workflows/end2end.yaml
```diff
@@ -493,9 +462,16 @@ jobs:
...
         run: bash configure-e2e-ctst.sh
         working-directory: ./.github/scripts/end2end
+      - uses: actions/setup-node@v6
+        with:
+          node-version: '24'
+          cache: yarn
+          cache-dependency-path: tests/ctst/yarn.lock
+      - name: Install ctst test dependencies
+        working-directory: tests/ctst
+        run: yarn install --network-concurrency 1 || (yarn cache clean && yarn install --network-concurrency 1)
```
francoisferrand, line 476: not needed, done already in run-e2e-ctst (or better yet, remove from there ?)

tests/ctst/steps/dmf.ts
```diff
@@ -3,8 +3,33 @@ import assert from 'assert';
...
+                command: ['sh', '-c', cmd],
+                volumeMounts: [{ name: 'vol', mountPath: '/cold-data' }],
+            }],
+        },
+    }).replace(/'/g, "'\\''");
+    // Use --attach (not -i/--stdin) so kubectl does not wait for stdin EOF,
+    // which would hang when called from Node.js exec() whose stdin pipe is
+    // never closed by the caller.
+    return execShellCommand(
+        `kubectl run ${podName} --image=busybox:1 --attach --rm --restart=Never --overrides='${spec}'`,
```
francoisferrand, line 27: not the way to do this: should be done with `kubectl debug` instead

.github/scripts/mocks/azure/Dockerfile
```diff
@@ -1,19 +0,0 @@
...
-    && apk add -U --no-cache --virtual .build-deps \
-        python3-dev \
-        linux-headers \
-        musl-dev \
-        libffi-dev \
-        gcc \
-        make \
-        openssl-dev \
-    && python3 -m pip install --upgrade pip \
-    && python3 -m pip install azure-cli \
```
francoisferrand, line 13: this was done to give us access to azure-cli ? So not needed anyway since it is present directly on the host/runner?
(is it also available in devcontainer/codespace? I don't see it in either the "features" or Dockerfile)

.github/scripts/mocks/azure/docker-entrypoint.sh
```diff
@@ -1,22 +0,0 @@
...
-echo "Starting Azurite..."
-azurite -l /data --blobHost 0.0.0.0 --blobPort 80 &
-AZURE_PID="$!"
-echo "Done!"
-
-sleep 2 # sleep for good measure
-
-echo "Creating buckets..."
-makebkt 'ci-zenko-azure-target-bucket'
-makebkt 'ci-zenko-azure-crr-target-bucket'
```
francoisferrand, line 19: is this not needed? these buckets were created uselessly?

.devcontainer/README.md
SylvainSenechal: I moved the endpoint to the "configure set" commands. Anyways i dont think this is used much, i would hope that with codespace we can write a test and run it to test it, including sdk commands

.devcontainer/README.md
SylvainSenechal: Could be but I think no need, I talked about it a bit in the other comment, but I think this should barely be used, as I would expect people to just develop a test and run the test directly, so don't really need to run s3 cli commands here. And setup-e2e-env is more for things that could be used both for codespace and github ci, than only for codespace

.devcontainer/setup.sh
SylvainSenechal: Okay let's leave it as it here in this pr and I created this https://scality.atlassian.net/jira/software/c/projects/OS/boards/268?assignee=712020%3A0e40d155-bc02-4e11-b6ac-64a79378e81f&selectedIssue=ZENKO-5256

.devcontainer/setup.sh
SylvainSenechal: If we run it in the dockerfile, we still need to do a cp in setup.sh no ? I think it's easier to keep it here, setup.sh is run once 🤔 We could even remove this entirely and let user do a yarn install in their codespace when needed

.github/scripts/end2end/setup-e2e-env.sh
SylvainSenechal: For this one, unlike for mongo, we could use ingress, I guess its a bit more work but maybe more stable, should I do it ?

.github/scripts/end2end/setup-e2e-env.sh
SylvainSenechal: this file is used both for github ci and codespace, setup.sh is codespace only so its needed here. Maybe this should be removed from codespace, but if it takes some time, doing it in codespace setup makes it faster to run tests after setup

.github/scripts/end2end/setup-e2e-env.sh
SylvainSenechal: Yeah I wanna get rid of it in future pr, it's a bit ugly, I doubt we can remove it now, some ctst specific command may fail in zenko_tests context so I will spend some time on it in another pr. It's one of the 2 next tickets I wanna work on for the ci

.github/workflows/end2end.yaml
SylvainSenechal: yes, and i fixed 3 small errors found by it

.github/workflows/end2end.yaml
SylvainSenechal: I don't see it in run-e2e-ctst. run-e2e-ctst sources setup-e2e-env which runs it, but for the zenko tests only.
Imo better to leave it like this right now, one of the ticket i wanna do is put zenko tests and ctst tests under the same package.json, maybe there will be some refactoring there

tests/ctst/steps/dmf.ts
SylvainSenechal: what's the problem with kubectl run 🤔 The pod stays there after the test is run ?

.github/scripts/mocks/azure/docker-entrypoint.sh
SylvainSenechal: the first bucket isnt used at all
The second is used, but its created in the python file create_buckets.py, function create_azure_containers

So at some point, the codebase used to need the azure cli to create these, and then I guess the logic was migrated into a python file

.github/scripts/mocks/azure/Dockerfile
SylvainSenechal: addressed in the comment on docker-entrypoint.sh file

.github/scripts/end2end/configure-e2e-endpoints.sh
```diff
@@ -14,6 +14,7 @@ apply_ingress() {
     local name="$1"
     local host="$2"
     local service="$3"
+    local port="${4:-name: http}"
```
delthas: Would it perhaps be nicer to have this be always a number and have `number: ${port}` below? Then we could just use `80` as the default instead of  `name: http`

delthas (approved):

.github/workflows/end2end.yaml
francoisferrand: indeed, I don't see it : probably miss-read, actually done in .devcontainer/setup.sh

tests/ctst/steps/dmf.ts
francoisferrand: not a "problem" with kubectl run, just not the same use : kubectl run is designed to run a "workload" (i.e. some production stuff) while kubectl debug is designed to do exactly what you want (i.e. like running a command in another pod -even if it does not have a shell-, ...)

sure it can work -especially in CI- with `kubectl run`, but the future is to use `kubectl debug` (instead of kubectl run/kuebctl exec) for most troubleshooting procedures for exemple

.devcontainer/setup.sh
francoisferrand: the goal of doing it in the Dockerfile is necessarily to install, but to fetch the packages to the yarn cache (which is the slow part of install, most often). maybe we don't need a full install, maybe there is better command; or even just install in a separate "layer", and just copy the cache.

.github/workflows/end2end.yaml
```diff
@@ -124,7 +122,7 @@ env:
   TMATE_SERVER_RSA_FINGERPRINT: ${{ secrets.TMATE_SERVER_RSA_FINGERPRINT }}
   TMATE_SERVER_ED25519_FINGERPRINT: ${{ secrets.TMATE_SERVER_ED25519_FINGERPRINT }}
   # Mocha reporter configuration
-  MOCHA_FILE: /reports/test-results-[hash].xml
+  MOCHA_FILE: reports/test-results-[hash].xml
```
francoisferrand: no correct, properly fixed in https://github.com/scality/Zenko/pull/2375 → to be removed from here
(and rebased on top of it, hopefully it will merge in a few minutes...)

.github/scripts/end2end/setup-e2e-env.sh
francoisferrand: could be moved to a separate script, which can be reused easily...

.github/scripts/end2end/setup-e2e-env.sh
```diff
@@ -119,12 +116,237 @@ if kubectl get namespace metadata &>/dev/null; then
...
+    SORBET_FWD_2_SECRETKEY=$(kubectl get secret -l app.kubernetes.io/name=sorbet-fwd-creds,app.kubernetes.io/instance=${ZENKO_NAME} -o jsonpath='{.items[0].data.secretKey}' | base64 -d)
+    export SERVICE_USERS_CREDENTIALS=$(echo '{"backbeat-lifecycle-bp-1":'"${BACKBEAT_LCBP_1_CREDS}"',"backbeat-lifecycle-conductor-1":'"${BACKBEAT_LCC_1_CREDS}"',"backbeat-lifecycle-op-1":'"${BACKBEAT_LCOP_1_CREDS}"',"backbeat-qp-1":'"${BACKBEAT_QP_1_CREDS}"',"sorbet-fwd-2":{"accessKey":"'"${SORBET_FWD_2_ACCESSKEY}"'","secretKey":"'"${SORBET_FWD_2_SECRETKEY}"'"}}' | jq -R)
+
+    # --- 12. Kafka topics for sorbet ---
+    SORBET_CONFIG=$(kubectl get secret -l app.kubernetes.io/name=cold-sorbet-config-e2e-azure-archive,app.kubernetes.io/instance=${ZENKO_NAME} \
+        -o jsonpath='{.items[0].data.config\.json}' | base64 -di)
+    export KAFKA_DEAD_LETTER_TOPIC=$(echo "${SORBET_CONFIG}" | jq -r '."kafka-dead-letter-topic"')
+    export KAFKA_OBJECT_TASK_TOPIC=$(echo "${SORBET_CONFIG}" | jq -r '."kafka-object-task-topic"')
+    export KAFKA_GC_REQUEST_TOPIC=$(echo "${SORBET_CONFIG}" | jq -r '."kafka-gc-request-topic"')
```
francoisferrand, line 167: for followup: everything we retrieve from secrets/k8s can be retrieved directly from CTST. Just needs creds and ZENKO_NAME

.github/scripts/end2end/setup-e2e-env.sh
```diff
@@ -119,12 +116,237 @@ if kubectl get namespace metadata &>/dev/null; then
...
+    KAFKA_PORT=${KAFKA_HOST_PORT_ORIG#*:}
+
+    # Port-forward for Kafka broker (TCP — no ingress possible)
+    if ! ss -tlnp 2>/dev/null | grep -q ":${KAFKA_PORT}" && \
+       ! lsof -i ":${KAFKA_PORT}" &>/dev/null; then
+        kubectl port-forward "svc/${KAFKA_SVC}" "${KAFKA_PORT}:${KAFKA_PORT}" &>/dev/null &
+        _KAFKA_PF_PID=$!
+        timeout 10 bash -c "until ss -tlnp 2>/dev/null | grep -q ':${KAFKA_PORT}'; do sleep 0.2; done"
+    fi
+    export KAFKA_HOST_PORT="localhost:${KAFKA_PORT}"
```
francoisferrand, line 182: another idea for followup: instead of port-forwarding here, maybe we can port-forward directly in CTST or run a command in the k8s cluster (depending on the case)

this would have the same benefit as above: less setup, tests can autonomously do what they need, as long as they know how to access zenko running in k8s → fewer variables

.github/scripts/end2end/setup-e2e-env.sh
```diff
@@ -175,4 +399,29 @@ if [ -n "${GITHUB_ENV:-}" ]; then # Don't do it for Codespace
     # JSON vars need the heredoc delimiter syntax (values contain '=')
     printf 'CRR_SOURCE_INFO<<EOF\n%s\nEOF\n' "$CRR_SOURCE_INFO" >> "$GITHUB_ENV"
     printf 'CRR_DESTINATION_INFO<<EOF\n%s\nEOF\n' "$CRR_DESTINATION_INFO" >> "$GITHUB_ENV"
+    # CTST-specific vars (only if CTST setup ran)
+    if [ "${SKIP_CTST:-}" != "1" ]; then
```
francoisferrand: can't we move this inside of the existing SKIP_CTST block ?

(and wondering why we need these, even though we generate a "world" file which should contain everything needed by ctst... probably something to further cleanup there, eventually)

tests/ctst/steps/pra.ts
```diff
@@ -317,9 +317,9 @@ Then('the kafka DR volume exists', { timeout: volumeTimeout + 2000 }, async func
     assert(volumeParsed.result!['volume phase'] === 'Bound');
 });
 
-Then('prometheus should scrap federated metrics from DR sink', { timeout: 180000 }, async function (this: Zenko) {
+Then('prometheus should scrap federated metrics from DR sink', { timeout: 180000 }, async () => {
     const prom = new PrometheusDriver({
-        endpoint: `http://${this.parameters.PrometheusService}:9090`,
+        endpoint: 'http://localhost:9090',
```
francoisferrand: should keep `this.parameters.PrometheusService` here : and just set the parameter to localhost.

tests/ctst/steps/dmf.ts
francoisferrand: in particular kubectl debug knows how to "share" volumes already, no need to add coupling on the way the pod is started

tests/ctst/steps/dmf.ts
```diff
@@ -3,8 +3,33 @@ import assert from 'assert';
 import { execShellCommand } from 'common/utils';
 import Zenko from 'world/Zenko';
 
+// The mock-sorbet pod uses a distroless image (no shell, rm, find, etc.),
+// so we cannot kubectl exec into it.  Instead, we spin up a short-lived
```
francoisferrand: more precisely, we _can_ kubectl exec (for exemple `kubectl exec -- sorbetctl` ...): but we cannot exec to run a shell or shell commands indeed

francoisferrand (approved):

.github/scripts/end2end/setup-e2e-env.sh
SylvainSenechal: edit april 15, meant to write this comment 1 day earlier but i forgot to click "comment" : 

I think ingress can only be used for http(s) traffic, and mongo is TCP

.github/scripts/end2end/configure-e2e-endpoints.sh
SylvainSenechal: ah ok yeah, i didnt know we could use name: http syntax when the port is 80

.github/scripts/end2end/setup-e2e-env.sh
SylvainSenechal: ah you mean directly in the code 🤔 I haven't thought of that, it's probably possible to do

.github/scripts/end2end/setup-e2e-env.sh
SylvainSenechal: Actually yeah since its already in world parameters, no need to keep it

benzekrimaha (approved):

PR #2378, opened 2026-04-17
Add E2E test verifying OTEL tracing across Zenko services

delthas:
## Summary

- Deploy Jaeger all-in-one (memory-only, OTLP-enabled, pinned to `1.76.0`)
  in the kind CI cluster alongside existing dependencies (Keycloak,
  Prometheus, etc.)
- Enable tracing on the Zenko CR (`spec.tracing`, ratio `"0"`) so the
  cluster runs in production-mode (no ambient traces). The test sends
  a single PUT through the existing S3 ingress with a W3C `traceparent`
  header injected via SDK middleware. Cloudserver's parent-based sampler
  honors the header → only that one request is traced.
- Add a `@PreMerge` CTST scenario that asserts the trace appears in
  Jaeger and contains spans from both **connector-cloudserver** and
  **connector-vault**

## What changed

| File | What |
|------|------|
| `solution/deps.yaml` | Bump the component images this test needs to run against OTEL/CRR-capable builds: `vault` `8.11.6 → 8.11.7` and `cloudserver` `9.3.9 → 9.4.0-preview.2` (OTEL support — parent-based sampler), `backbeat` `9.4.0 → 9.5.0-preview.1`, and `s3utils` `1.17.12 → 1.19.0`. **The s3utils bump is required transitively, not unrelated:** the preview cloudserver stores CRR bucket metadata in the new arsenal-8.5 (multi-CRR) format, which s3utils 1.17.12 (arsenal 8.2.36) cannot deserialize — `count-items` crashes in `getBucketAttributes`, which hangs the `@Quotas` CTST scenarios and blocks this suite from completing. 1.19.0 pins arsenal 8.5.0 to match cloudserver. |
| `configs/zenko.yaml` | Add `spec.tracing` (enabled, ratio `"0"`, jaeger endpoint) |
| `configs/jaeger.yaml` | Jaeger Deployment + Service (pinned `1.76.0`, memory-only, requests/limits) |
| `install-kind-dependencies.sh` | `kubectl apply -f configs/jaeger.yaml` |
| `setup-e2e-env.sh` | Port-forward Jaeger query API + `JaegerQueryEndpoint` in world params |
| `world/Zenko.ts` | `JaegerQueryEndpoint` world parameter |
| `features/otel-tracing.feature` | New `@PreMerge` scenario |
| `steps/otel-tracing.ts` | Generates W3C `traceparent`, adds an SDK middleware that injects the header on the CTST S3 client, polls `GET /api/traces/<id>` until the trace appears, asserts cloudserver + vault spans |

## Why

Parent ticket [OS-1072](https://scality.atlassian.net/browse/OS-1072)
tracks adding OpenTelemetry tracing across the Zenko stack. This PR
adds CI coverage to verify that traces actually propagate end-to-end
(cloudserver → vault) once tracing is enabled.

Note on trust boundary: production runs metalk8s nginx which strips
incoming `traceparent` headers; Zenko CI uses upstream
`kubernetes/ingress-nginx` which forwards them. So in CI the test
injects via the regular ingress; the production trust-boundary
(external `traceparent` is dropped at the edge) is metalk8s-specific
and out of scope for this CI test.

Issue: ZENKO-5258

[OS-1072]: https://scality.atlassian.net/browse/OS-1072?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/scripts/end2end/configs/jaeger.yaml
+++ b/.github/scripts/end2end/configs/jaeger.yaml
@@ -1,0 +1,64 @@
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: jaeger
+  namespace: default
+  labels:
+    app: jaeger
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: jaeger
+  template:
+    metadata:
+      labels:
+        app: jaeger
+    spec:
+      containers:
+      - name: jaeger
+        image: jaegertracing/all-in-one:1.76.0
+        resources:
+          requests:
+            cpu: 100m
+            memory: 128Mi
+          limits:
+            memory: 512Mi
+        env:
+        - name: COLLECTOR_OTLP_ENABLED
+          value: "true"
+        - name: MEMORY_MAX_TRACES
+          value: "10000"
+        ports:
+        - containerPort: 16686
+          name: query
+        - containerPort: 4317
+          name: otlp-grpc
+        - containerPort: 4318
+          name: otlp-http
+        readinessProbe:
+          httpGet:
+            path: /
+            port: 16686
+          initialDelaySeconds: 5
+          periodSeconds: 5
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: jaeger
+  namespace: default
+spec:
+  selector:
+    app: jaeger
+  ports:
+  - name: query
+    port: 16686
+    targetPort: 16686
+  - name: otlp-grpc
+    port: 4317
+    targetPort: 4317
+  - name: otlp-http
+    port: 4318
+    targetPort: 4318
--- a/.github/scripts/end2end/configure-e2e-ctst.sh
+++ b/.github/scripts/end2end/configure-e2e-ctst.sh
@@ -111,3 +111,4 @@ kubectl run kafka-topics \
 # Deploy PyKMIP server (infra only, does NOT patch the CR).
 # The CR is patched later, after file-backend SSE tests have run.
 bash "$(dirname "$0")/../mocks/setup-kmip.sh"
+
--- a/.github/scripts/end2end/install-kind-dependencies.sh
+++ b/.github/scripts/end2end/install-kind-dependencies.sh
@@ -153,6 +153,9 @@ helm upgrade --install --version ${KEYCLOAK_VERSION} keycloak codecentric/keyclo
 
 kubectl rollout status sts/keycloak --timeout=10m
 
+# jaeger all-in-one (OTLP collector + query UI, memory-only)
+kubectl apply -f "$(dirname "$0")/configs/jaeger.yaml"
+kubectl rollout status deployment/jaeger --timeout=5m
 
 # TODO: use zenko-operator install-deps
 kubectl apply -f - <<EOF
--- a/.github/scripts/end2end/setup-e2e-env.sh
+++ b/.github/scripts/end2end/setup-e2e-env.sh
@@ -231,6 +231,15 @@ else
     fi
     export PROMETHEUS_SERVICE="${PROMETHEUS_SVC}.${NAMESPACE}.svc.cluster.local"
 
+    # Jaeger query API — port-forward for OTEL tracing tests
+    JAEGER_QUERY_PORT=16686
+    if ! ss -tlnp 2>/dev/null | grep -q ":${JAEGER_QUERY_PORT}" && \
+       ! lsof -i ":${JAEGER_QUERY_PORT}" &>/dev/null; then
+        kubectl port-forward "svc/jaeger" "${JAEGER_QUERY_PORT}:${JAEGER_QUERY_PORT}" &>/dev/null &
+        timeout 10 bash -c "until ss -tlnp 2>/dev/null | grep -q ':${JAEGER_QUERY_PORT}'; do sleep 0.2; done"
+    fi
+    export JAEGER_QUERY_ENDPOINT="http://localhost:${JAEGER_QUERY_PORT}"
+
     # --- 14. Zenko CR metadata ---
     export TIME_PROGRESSION_FACTOR=$(kubectl get zenko ${ZENKO_NAME} -o jsonpath="{.metadata.annotations.zenko\.io/time-progression-factor}")
     export INSTANCE_ID=$(kubectl get zenko ${ZENKO_NAME} -o jsonpath='{.status.instanceID}')
@@ -334,7 +343,8 @@ else
       "DRAdminSecretKey":"${ADMIN_PRA_SECRET_ACCESS_KEY}",
       "UtilizationServiceHost":"${UTILIZATION_SERVICE_HOST}",
       "UtilizationServicePort":"${UTILIZATION_SERVICE_PORT}",
-      "KubeconfigPath":"${KUBECONFIG:-${HOME}/.kube/config}"
+      "KubeconfigPath":"${KUBECONFIG:-${HOME}/.kube/config}",
+      "JaegerQueryEndpoint":"${JAEGER_QUERY_ENDPOINT}"
     }
 EOF
     )"
--- a/solution/deps.yaml
+++ b/solution/deps.yaml
@@ -100,7 +100,7 @@ s3utils:
   sourceRegistry: ghcr.io/scality
   dashboard: s3utils/s3utils-dashboards
   image: s3utils
-  tag: 1.17.12
+  tag: 1.19.0
   envsubst: S3UTILS_TAG
 scuba:
   sourceRegistry: ghcr.io/scality
--- a/solution/deps.yaml
+++ b/solution/deps.yaml
@@ -16,7 +16,7 @@ cloudserver:
   sourceRegistry: ghcr.io/scality
   dashboard: cloudserver/cloudserver-dashboards
   image: cloudserver
-  tag: 9.3.9
+  tag: 9.4.0-preview.2
   envsubst: CLOUDSERVER_TAG
 drctl:
   sourceRegistry: ghcr.io/scality
--- a/solution/deps.yaml
+++ b/solution/deps.yaml
@@ -125,7 +125,7 @@ vault:
   dashboard: vault2/vault-dashboards
   policy: vault2/vault-policies
   image: vault2
-  tag: 8.11.6
+  tag: 8.11.7
   envsubst: VAULT_TAG
 zenko-operator:
   sourceRegistry: ghcr.io/scality
--- a/solution/deps.yaml
+++ b/solution/deps.yaml
@@ -6,7 +6,7 @@ backbeat:
   dashboard: backbeat/backbeat-dashboards
   image: backbeat
   policy: backbeat/backbeat-policies
-  tag: 9.4.0
+  tag: 9.5.0-preview.1
   envsubst: BACKBEAT_TAG
 busybox:
   image: busybox
--- a/tests/functional/ctst/features/otel-tracing.feature
+++ b/tests/functional/ctst/features/otel-tracing.feature
@@ -1,0 +1,13 @@
+@2.16.0
+@PreMerge
+Feature: OpenTelemetry Tracing
+  Even when global sampling is disabled, a request carrying a W3C
+  traceparent header from a trusted (in-cluster) source must produce
+  a trace spanning all the Zenko services it touches.
+
+  Scenario: PutObject with injected traceparent produces a trace spanning cloudserver and vault
+    Given a "Non versioned" bucket
+    When I put an object with an injected traceparent
+    Then the injected trace should be found in Jaeger
+    And the trace should contain spans from service "connector-cloudserver"
+    And the trace should contain spans from service "connector-vault"
--- a/tests/functional/ctst/steps/otel-tracing.ts
+++ b/tests/functional/ctst/steps/otel-tracing.ts
@@ -1,0 +1,130 @@
+import { Then, When } from '@cucumber/cucumber';
+import { strict as assert } from 'assert';
+import { PutObjectCommand, S3Client } from '@aws-sdk/client-s3';
+import { randomBytes } from 'crypto';
+import { Utils } from 'cli-testing';
+import Zenko from 'world/Zenko';
+
+const JAEGER_POLL_TIMEOUT = 30000;
+const JAEGER_POLL_INTERVAL = 2000;
+const TRACED_OBJECT_KEY = 'otel-trace-test-object';
+
+interface JaegerTrace {
+    traceID: string;
+    spans: { processID: string }[];
+    processes: Record<string, { serviceName: string }>;
+}
+
+function generateTraceContext(): { traceparent: string; traceId: string } {
+    const traceId = randomBytes(16).toString('hex');
+    const spanId = randomBytes(8).toString('hex');
+    return { traceparent: `00-${traceId}-${spanId}-01`, traceId };
+}
+
+function injectTraceparent(client: S3Client, traceparent: string): void {
+    // traceparent is not part of any SigV4 signed-header set, so injecting at
+    // the 'build' step (pre-signing) does not invalidate the signature.
+    client.middlewareStack.add(
+        next => async args => {
+            const request = args.request as { headers: Record<string, string> };
+            request.headers.traceparent = traceparent;
+            return next(args);
+        },
+        { step: 'build', name: 'injectTraceparent' },
+    );
+}
+
+async function fetchTraceById(endpoint: string, traceId: string): Promise<JaegerTrace | null> {
+    const response = await fetch(`${endpoint}/api/traces/${traceId}`, {
+        signal: AbortSignal.timeout(5000),
+    });
+    if (response.status === 404) {
+        return null;
+    }
+    if (!response.ok) {
+        throw new Error(`Jaeger query returned HTTP ${response.status}`);
+    }
+    const body = await response.json() as { data: JaegerTrace[] };
+    return body.data?.[0] ?? null;
+}
+
+async function pollJaegerForTrace(
+    endpoint: string,
+    traceId: string,
+    timeoutMs = JAEGER_POLL_TIMEOUT,
+    intervalMs = JAEGER_POLL_INTERVAL,
+): Promise<JaegerTrace> {
+    const deadline = Date.now() + timeoutMs;
+    let lastError: Error | null = null;
+
+    while (Date.now() < deadline) {
+        try {
+            const trace = await fetchTraceById(endpoint, traceId);
+            if (trace) {
+                return trace;
+            }
+        } catch (err) {
+            lastError = err as Error;
+        }
+        await Utils.sleep(intervalMs);
+    }
+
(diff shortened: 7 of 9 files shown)


.github/scripts/end2end/configure-e2e-ctst.sh
```diff
@@ -111,3 +111,19 @@ kubectl run kafka-topics \
 # Deploy PyKMIP server (infra only, does NOT patch the CR).
 # The CR is patched later, after file-backend SSE tests have run.
 bash "$(dirname "$0")/../mocks/setup-kmip.sh"
+
+# Enable OTEL tracing on the Zenko CR (always-on in CI, acts as a smoke test)
```
francoisferrand: how long does it add to CI (deployment time) ?
is there a performance impact ? (esp. if performance degradation is preventing from hitting race conditions, not sure it is better :-/ )

.github/scripts/end2end/configure-e2e-ctst.sh
```diff
@@ -111,3 +111,19 @@ kubectl run kafka-topics \
...
+NAMESPACE="${NAMESPACE:-default}"
+kubectl patch zenko ${ZENKO_NAME} -n ${NAMESPACE} --type merge -p '{
+  "spec": {"tracing": {
+    "enabled": true,
+    "samplingRatio": "1.0",
```
francoisferrand: trace ***every*** ?

.github/scripts/end2end/configure-e2e-ctst.sh
```diff
@@ -111,3 +111,19 @@ kubectl run kafka-topics \
 # Deploy PyKMIP server (infra only, does NOT patch the CR).
 # The CR is patched later, after file-backend SSE tests have run.
 bash "$(dirname "$0")/../mocks/setup-kmip.sh"
+
+# Enable OTEL tracing on the Zenko CR (always-on in CI, acts as a smoke test)
+NAMESPACE="${NAMESPACE:-default}"
+kubectl patch zenko ${ZENKO_NAME} -n ${NAMESPACE} --type merge -p '{
```
francoisferrand: instead of patching Zenko (i.e re-reconcile), should be added to the Zenko CR before it is installed on the cluster.

.github/scripts/end2end/install-kind-dependencies.sh
```diff
@@ -153,6 +153,67 @@ helm upgrade --install --version ${KEYCLOAK_VERSION} keycloak codecentric/keyclo
 
 kubectl rollout status sts/keycloak --timeout=10m
 
+# jaeger all-in-one (OTLP collector + query UI, memory-only)
+kubectl apply -f - <<'EOF'
```
francoisferrand: best to use a separate CR, and keep the bash script simple?

tests/functional/ctst/features/otel-tracing.feature
```diff
@@ -0,0 +1,10 @@
+@2.15.0
+@PreMerge
+Feature: OpenTelemetry Tracing
+  Traces should propagate across Zenko services
```
francoisferrand: instead of tracing every request, could we "trust" localhost : and let the test initiate a request with an Otel trace Id ?
This may limit impact of the change....

tests/functional/ctst/features/otel-tracing.feature
```diff
@@ -0,0 +1,10 @@
...
+
+  Scenario: S3 PutObject produces a trace spanning cloudserver and vault
```
francoisferrand: enabling trace all the time means we are not testing the "production" case anymore...

should we instead enable/disable tracing at the beginning and end of this test? (it would reconcile and possibly trigger instability -if we are not resilient-, but would ensure we test both cases...)

tests/functional/ctst/steps/otel-tracing.ts
```diff
@@ -0,0 +1,124 @@
...
+
+interface JaegerTrace {
+    traceID: string;
+    spans: JaegerSpan[];
+    processes: Record<string, JaegerProcess>;
+}
+
+interface JaegerSearchResponse {
+    data: JaegerTrace[];
+}
```
francoisferrand: - are these really **_Jeager_** specific? or really standard OpenTelemetry?
- can't we use types from some sdk, instead of re-defining these?

.github/scripts/end2end/setup-e2e-env.sh
```diff
@@ -338,7 +348,8 @@ else
       "DRAdminSecretKey":"${ADMIN_PRA_SECRET_ACCESS_KEY}",
       "UtilizationServiceHost":"${UTILIZATION_SERVICE_HOST}",
       "UtilizationServicePort":"${UTILIZATION_SERVICE_PORT}",
-      "KubeconfigPath":"${KUBECONFIG:-${HOME}/.kube/config}"
+      "KubeconfigPath":"${KUBECONFIG:-${HOME}/.kube/config}",
+      "JaegerQueryEndpoint":"${JAEGER_QUERY_ENDPOINT}"
```
francoisferrand, line 347: can we add trailing comma already?
so we can add new lines without modifying the previous/last one?

francoisferrand (commented):
overall not sure where to stand here: there is a fundamental compromise: do we want to test the "production" case first and foremost (i.e. without tracing), or is it safe-enough to enable Otel all the time (i.e. not testing production anymore, which could hide crashes or introduce subtle delays...)

.github/scripts/end2end/configure-e2e-ctst.sh
SylvainSenechal: I was gonna ask the same thing but so far no cucumber run was successful

.github/scripts/end2end/configure-e2e-ctst.sh
SylvainSenechal: yeah it can go directly in configs/zenko.yaml 🤔

tests/functional/ctst/features/otel-tracing.feature
SylvainSenechal: I don't understand what you mean by "not testing the production case".
Otel enabled or not, we are testing the same thing, just with extra telemetry when its enabled 🤔 

Maybe what we could have though is a sampling ratio lower than 100%, and do many putobject, to we have traces rate close to the ratio

SylvainSenechal (commented):
side question, here the jaeger server will be killed at the end the ci run, but would it be possible to have the traces data retained, so that maybe we could export these traces data in the artifact, and spin up a server to explore them with a ui ?

tests/functional/ctst/steps/otel-tracing.ts
delthas: These come from Jaeger's REST query API (`GET /api/traces`, `GET /api/traces/<id>`), not the OpenTelemetry wire format — so they're Jaeger-specific. There's no maintained npm package for the Jaeger query API: `jaeger-client` exists but is deprecated and only emits traces. OpenTelemetry defines no standard trace query API either (Tempo has TraceQL but Jaeger doesn't implement it). Jaeger publishes a gRPC `query.proto` and an OpenAPI swagger for `api_v3`, but the REST JSON endpoint we use here is explicitly undocumented and marked subject to change. Given the only consumer is this one test, hand-rolled minimal interfaces seem like the right tradeoff — happy to revisit if we end up needing a real client.

.github/scripts/end2end/setup-e2e-env.sh
delthas: The block already passes through `jq -c <<EOF` to compact, but the heredoc body is parsed as strict JSON — jq's parser rejects trailing commas there too. To get the "new lines don't touch prior ones" property we'd need to either rewrite as `jq -n '{...}'` object construction (still strict, no trailing comma allowed) or build the object up with a chain of `jq '. + {...}'` calls, both of which are uglier than the status quo. Leaving as-is unless we want to do a broader refactor.

.github/scripts/end2end/configure-e2e-ctst.sh
delthas: The redesign in the latest push (`8f05574`) drops sampling to `0` globally — only the test's own injected request emits spans, so the runtime cost on every other PreMerge scenario is essentially zero. The remaining cluster overhead is just the Jaeger all-in-one pod (memory-only, capped at 10k traces).

@SylvainSenechal — the cucumber failures aren't related to perf: this PR depends on zenko-operator [PR #607 (ZKOP-539)](https://github.com/scality/zenko-operator/pull/607), which adds the `spec.tracing` CRD field. Until that lands + the operator dep is bumped, the CR rejects our config block and the test can't run.

.github/scripts/end2end/configure-e2e-ctst.sh
delthas: Latest push (`8f05574`) drops sampling to `"0"` — only the test's own request, which carries a W3C `traceparent` with `sampled=1`, is traced. Cloudserver's parent-based sampler honors that flag, so other traffic on the same PreMerge run stays unsampled.

.github/scripts/end2end/configure-e2e-ctst.sh
delthas: Done in `8f05574` — `spec.tracing` is now part of `configs/zenko.yaml`; the runtime `kubectl patch` and rollout-wait block are gone.

.github/scripts/end2end/install-kind-dependencies.sh
delthas: Done in `8f05574` — extracted to `configs/jaeger.yaml`, the install script just does `kubectl apply -f` on it.

tests/functional/ctst/features/otel-tracing.feature
delthas: Adopted in `8f05574` — sampling stays at `0`, the test sends its PutObject to a port-forwarded `connector-cloudserver` svc (bypassing nginx, which strips `traceparent`), and injects a W3C `traceparent` with `sampled=1`. Cloudserver's parent-based sampler honors that flag, so only the test's own request emits spans. We then look up the trace by ID directly via `GET /api/traces/<id>`.

tests/functional/ctst/features/otel-tracing.feature
delthas: Latest push (`8f05574`) takes the production case directly: `samplingRatio: "0"` is the production setting. Only the test's own request — carrying an injected `traceparent` with `sampled=1` from a trusted (in-cluster, port-forwarded) source — gets traced. Other traffic on the same PreMerge run remains unsampled, exactly like prod.

.github/scripts/end2end/configs/jaeger.yaml
```diff
@@ -0,0 +1,58 @@
...
+      app: jaeger
+  template:
+    metadata:
+      labels:
+        app: jaeger
+    spec:
+      containers:
+      - name: jaeger
+        image: jaegertracing/all-in-one:1.76.0
+        env:
```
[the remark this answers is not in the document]
delthas: Applied as-suggested in `516077421c`.

delthas:
Moved to a ratio of 0% (no traces, but enabled); and injecting a forcefully enabled trace using a `traceparent` header directly made against cloudserver (bypassing the Ingress that strips it)

This should bring no performance overhead, no overlay reconfiguration, and nicely isolate our test trace.

.github/scripts/end2end/setup-e2e-env.sh
```diff
@@ -235,6 +235,27 @@ else
     fi
     export PROMETHEUS_SERVICE="${PROMETHEUS_SVC}.${NAMESPACE}.svc.cluster.local"
 
+    # Jaeger query API — port-forward for OTEL tracing tests
+    JAEGER_QUERY_PORT=16686
+    if ! ss -tlnp 2>/dev/null | grep -q ":${JAEGER_QUERY_PORT}" && \
+       ! lsof -i ":${JAEGER_QUERY_PORT}" &>/dev/null; then
+        kubectl port-forward "svc/jaeger" "${JAEGER_QUERY_PORT}:${JAEGER_QUERY_PORT}" &>/dev/null &
+        _JAEGER_PF_PID=$!
```
maeldonn: `_JAEGER_PF_PID` and `_CLOUDSERVER_INTERNAL_PF_PID` are captured but never used.

delthas: Good catch — dropped both. (For context: every other port-forward in this file does the same capture-but-never-use pattern — `_KAFKA_PF_PID`, `_KAFKA_AUTH_PF_PID`, `_PROM_PF_PID` — but those are pre-existing and out of scope here. Cleanup tracked in #2446.)

solution/deps.yaml
```diff
@@ -125,7 +125,7 @@ vault:
   dashboard: vault2/vault-dashboards
   policy: vault2/vault-policies
   image: vault2
-  tag: 8.11.6
+  tag: 8.11.7
```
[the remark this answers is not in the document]
delthas: PR body updated — the "What changed" table now lists `vault 8.11.6 → 8.11.7` and `cloudserver 9.3.9 → 9.4.0-preview.1` as bumps required for OTEL instrumentation (vault span emission, cloudserver parent-based sampler).

delthas:
CI failing for unrelated reasons; OTEL test succeeding.

tests/functional/ctst/steps/otel-tracing.ts
```diff
@@ -0,0 +1,130 @@
...
+
+When('I put an object with an injected traceparent',
+    async function (this: Zenko) {
+        const bucketName = this.getSaved<string>('bucketName');
+        assert.ok(bucketName, 'No bucketName saved from a previous step');
+
+        const { traceparent, traceId } = generateTraceContext();
+        const client = this.createS3Client();
+        injectTraceparent(client, traceparent);
```
francoisferrand, line 93: how come CTST can inject a trace (and esp. cloudserver/... accepts it) ?
CTST is not running k8s in-cluster AFAIK : it runs on the "host" now, while zenko runs in a "kind" k8s cluster

→ is the "trusted hosts" protection not working? I would expect we need to explicitely fill some field in zenko CR to end up trusting ctst?

francoisferrand (approved):

solution/deps.yaml
```diff
@@ -3,20 +3,20 @@
 # yq eval 'sortKeys(.)' -i deps.yaml
 backbeat:
   sourceRegistry: ghcr.io/scality
-  dashboard: backbeat/backbeat-dashboards
-  image: backbeat
-  policy: backbeat/backbeat-policies
-  tag: 9.4.0
+  dashboard: playground/delthas/backbeat/backbeat-dashboards
+  image: playground/delthas/backbeat
```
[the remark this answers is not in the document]
SylvainSenechal: up

tests/functional/ctst/features/otel-tracing.feature
SylvainSenechal, line 1: Wait so, the CR is configured with samplingRatio = 0, so no traces are generated by the services.
Then you inject the trace yourself by manipulating the header of the request 🤔 

Wouldn't we want to test it with a non zero samplingRatio ? 
Or I guess you did it because you don't wanna enable tracing for every tests, and you don't want to patch the cr during the tests ?

SylvainSenechal (approved):
Just 1 question + update the /playground deps

maeldonn (approved):

tests/functional/ctst/features/otel-tracing.feature
delthas: Exactly:
- if we enable it for all the tests, we have to set a sampling ratio low enough to avoid generating too much load (which wer ethe PR comments in the first place), but also high enough such that we avoid having to retry so many times in order to get a sampled request
- we could also patch it, but it needs restarts of all services; sounded a bit heavy to me

So just setting it to 0 and forcing a sample with a traceparent header is lowest impact and good enough for a simple happy case test IMO.

tests/functional/ctst/steps/otel-tracing.ts
delthas: Trusted hosts is for outbound; stripping incoming headers is done by metalk8s nginx-ingress configuration cluster-wide. But here we just use a vanilla nginx-ingress so no traceparent header stripping mechanism.

(I initially thought I had to port forward to cloudserver directly in order to bypass our nginx-ingress but given the deployment, there's no need to do this at all.)

.github/scripts/end2end/configs/jaeger.yaml
```diff
@@ -0,0 +1,64 @@
...
+  replicas: 1
+  selector:
+    matchLabels:
+      app: jaeger
+  template:
+    metadata:
+      labels:
+        app: jaeger
+    spec:
+      containers:
```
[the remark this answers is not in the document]
delthas: This is an ephemeral CI-only test fixture (Jaeger all-in-one, used by the OTEL e2e test), not a production or Helm-managed manifest. None of the 9 sibling manifests in `.github/scripts/end2end/configs/` set a `securityContext` (fluentbit, prometheus, keycloak, zenko, notification_destinations, …), and the review criterion's stated scope is production charts (`solution-base/mongodb/charts/`) and `monitoring/`. Keeping this consistent with its neighbors — resolving as won't-fix.

solution/deps.yaml
```diff
@@ -100,7 +100,7 @@ s3utils:
   sourceRegistry: ghcr.io/scality
   dashboard: s3utils/s3utils-dashboards
   image: s3utils
-  tag: 1.17.12
+  tag: 1.19.0
```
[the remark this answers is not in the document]
delthas: It's required transitively by this PR, not unrelated — now documented in the description's "What changed" table. Bumping cloudserver to the OTEL-capable preview (9.4.0-preview.2) also brings the new multi-CRR support, which stores bucket replication metadata in the arsenal-8.5 format. s3utils 1.17.12 (arsenal 8.2.36) can't deserialize that — `count-items` throws in `getBucketAttributes` (`BucketInfo`/`ReplicationConfiguration.validateConfig`, a field expected to be a string is `undefined`), which fails the `ops-count-items` job and hangs the `@Quotas` CTST scenarios, blocking this suite. s3utils 1.19.0 pins arsenal 8.5.0 to match cloudserver. Keeping it in this PR since the test can't pass without it.

tests/functional/ctst/features/otel-tracing.feature
```diff
@@ -0,0 +1,13 @@
+@2.15.0
```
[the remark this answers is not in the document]
delthas: Done — updated to `@2.16.0` (this PR now targets `development/2.16`, and `spec.tracing` needs the 2.16-cycle operator, so the test must be gated to 2.16.0+).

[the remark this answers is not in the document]
delthas: Done — updated to `@2.16.0` (this PR now targets `development/2.16`, and `spec.tracing` needs the 2.16-cycle operator, so the test must be gated to 2.16.0+).

PR #2379, opened 2026-04-20
added kmip server side encryption tests

SylvainSenechal:
Issue: ZENKO-5241

Adding more encryption tests but with KMIP mode
A bit annoying as we need to patch the CR, and when kmip mode is enabled, the base encryption tests don't work anymore

### DIFF
--- a/.github/scripts/end2end/configure-e2e-ctst.sh
+++ b/.github/scripts/end2end/configure-e2e-ctst.sh
@@ -106,3 +106,8 @@ kubectl run kafka-topics \
         kafka-topics.sh --create --topic $AZURE_ARCHIVE_STATUS_TOPIC_2_NV --partitions 10 --bootstrap-server $KAFKA_HOST_PORT --if-not-exists ; \
         kafka-topics.sh --create --topic $AZURE_ARCHIVE_STATUS_TOPIC_2_V --partitions 10 --bootstrap-server $KAFKA_HOST_PORT --if-not-exists ; \
         kafka-topics.sh --create --topic $AZURE_ARCHIVE_STATUS_TOPIC_2_S --partitions 10 --bootstrap-server $KAFKA_HOST_PORT --if-not-exists"
+
+# KMIP mock setup
+# Deploy PyKMIP server (infra only, does NOT patch the CR).
+# The CR is patched later, after file-backend SSE tests have run.
+bash "$(dirname "$0")/../mocks/setup-kmip.sh"
--- a/solution/deps.yaml
+++ b/solution/deps.yaml
@@ -130,7 +130,7 @@ vault:
 zenko-operator:
   sourceRegistry: ghcr.io/scality
   image: zenko-operator
-  tag: v1.8.9
+  tag: v1.8.10
   envsubst: ZENKO_OPERATOR_TAG
 zookeeper:
   sourceRegistry: ghcr.io/adobe/zookeeper-operator
--- a/tests/ctst/common/hooks.ts
+++ b/tests/ctst/common/hooks.ts
@@ -6,7 +6,7 @@ import {
     ITestCaseHookParameter,
 } from '@cucumber/cucumber';
 import Zenko from '../world/Zenko';
-import { CacheHelper, Identity } from 'cli-testing';
+import { CacheHelper, Identity, WorkCoordination } from 'cli-testing';
 import { prepareQuotaScenarios, teardownQuotaScenarios } from 'steps/quotas/quotas';
 import { prepareUtilizationScenarios } from 'steps/utilization/utilizationAPI';
 import { prepareMetricsScenarios } from './utils';
@@ -16,6 +16,7 @@ import { displayDebuggingInformation, preparePRA } from 'steps/pra';
 import {
     cleanupAccount,
 } from './utils';
+import { createKubeCustomObjectClient, waitForZenkoToStabilize } from 'steps/utils/kubernetes';
 
 import 'cli-testing/hooks/KeycloakSetup';
 import 'cli-testing/hooks/Logger';
@@ -33,6 +34,7 @@ const noParallelRun = atMostOnePicklePerTag([
     '@AfterAll',
     '@PRA',
     '@ColdStorage',
+    '@ServerSideEncryption',
     ...replicationLockTags
 ]);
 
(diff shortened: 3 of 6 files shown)


solution/deps.yaml
```diff
@@ -130,7 +130,7 @@ vault:
 zenko-operator:
   sourceRegistry: ghcr.io/scality
   image: zenko-operator
-  tag: v1.8.9
+  tag: 15b35ecc60bc8b83e855e0cb17091b10e592aa07
```
SylvainSenechal: TO BE DONE 
!

tests/ctst/common/hooks.ts
```diff
@@ -75,6 +77,65 @@ Before({ tags: '@PrepareStorageUsageReportingScenarios', timeout: 1200000 }, asy
     });
 });
 
+Before({ tags: '@ServerSideEncryptionKmip', timeout: 15 * 60 * 1000 },
```
SylvainSenechal, line 80: best i can come up with in CUCUMBER TESTS to have certains tests ran after others

tests/ctst/features/serverSideEncryption.feature
SylvainSenechal, line 1: again, might have added too many tests that aren't super useful for FUNCTIONAL testing 🤔

tests/ctst/common/hooks.ts
delthas, line 80: Do we need an After/cleanup for this? Like I guess we make changes to the Zenko CR while the tests are running because we don't want to affect other tests; but then we should clear our changes before the next tests run, no?

delthas (approved):

tests/ctst/common/hooks.ts
SylvainSenechal: I left it without a cleanup. The cr can stay with a kmip encryption mode, anyways it won't affect the other tests as you still need to enable the encryption setting on the buckets. Changing cr again is gonna re trigger reconciliation and add some 50 lines of code function, not super useful

tests/ctst/steps/serverSideEncryption.ts
```diff
@@ -208,3 +222,24 @@ Then('it should fail with error {string}',
             `Expected error "${expectedError}" but got: ${result.err}`);
     },
 );
+
+Then('objects {string} and {string} share the same KMS key',
+    async function (this: Zenko, objA: string, objB: string) {
+        const bucket = this.getSaved<string>('bucketName');
+        const client = buildS3Client();
```
maeldonn: There is already a `createS3Client()` method on the world. Can it be used ?

tests/ctst/steps/serverSideEncryption.ts
```diff
@@ -208,3 +222,24 @@ Then('it should fail with error {string}',
...
+        try {
+            const [respA, respB] = await Promise.all([
```
maeldonn, line 198: Cannot we use `cli-testing` ?

tests/ctst/steps/serverSideEncryption.ts
```diff
@@ -189,8 +198,13 @@ Then('the GetObject should return the uploaded body with SSE algorithm {string}
                     `GetObject: SSEKMSKeyId should be absent, got "${resp.SSEKMSKeyId}"`);
             } else if (expectedKey === 'generated') {
                 assert.ok(resp.SSEKMSKeyId, 'GetObject: SSEKMSKeyId should be present');
-                assert.match(resp.SSEKMSKeyId, /^[a-f0-9]{64}$/,
-                    `GetObject: expected a generated hex key, got "${resp.SSEKMSKeyId}"`);
+                // Accept either:
+                //   - 64-char hex (file/internal KMS backend)
+                //   - KMIP numeric ID or ARN: arn:scality:kms:external:kmip:<provider>:key/<n>
+                const isFileBackendKey = /^[a-f0-9]{64}$/.test(resp.SSEKMSKeyId);
```
maeldonn: Could you extract duplicated regex to a new function?

```js
/**
 * Validates if the provided SSE KMS Key ID matches supported backend formats:
 * 1. File Backend: A 64-character hex string.
 * 2. KMIP Backend: A numeric string OR a specific Scality KMIP ARN.
 * @param sseKmsKeyId - The key ID string to validate
 * @returns boolean
 */
function isValidSseKmsKeyId(sseKmsKeyId: string): boolean {
    if (!sseKmsKeyId) return false;

    const isFileBackendKey = /^[a-f0-9]{64}$/.test(sseKmsKeyId);
    const isKmipKey = /^(\d+|arn:scality:kms:external:kmip:[a-z0-9]+:key\/\d+)$/.test(sseKmsKeyId);

    return isFileBackendKey || isKmipKey;
}
```

.github/scripts/end2end/configure-e2e-ctst.sh
```diff
@@ -106,3 +106,12 @@ kubectl run kafka-topics \
         kafka-topics.sh --create --topic $AZURE_ARCHIVE_STATUS_TOPIC_2_NV --partitions 10 --bootstrap-server $KAFKA_HOST_PORT --if-not-exists ; \
         kafka-topics.sh --create --topic $AZURE_ARCHIVE_STATUS_TOPIC_2_V --partitions 10 --bootstrap-server $KAFKA_HOST_PORT --if-not-exists ; \
         kafka-topics.sh --create --topic $AZURE_ARCHIVE_STATUS_TOPIC_2_S --partitions 10 --bootstrap-server $KAFKA_HOST_PORT --if-not-exists"
+
+# KMIP mock setup
+# Deploy PyKMIP server (infra only, does NOT patch the CR).
+# The CR is patched later, after file-backend SSE tests have run.
+if ! kubectl get deployment pykmip &>/dev/null; then
```
maeldonn: Should be checked inside the script imo

tests/ctst/steps/serverSideEncryption.ts
SylvainSenechal: yes I'm gonna use it

tests/ctst/steps/serverSideEncryption.ts
SylvainSenechal: Copying the comment I left on the code somewhere :

// We use the AWS SDK directly instead of cli-testing for PutObject and GetObject
// because:
// - cli-testing has a casing bug: --ssekms-key-id → SsekmsKeyId (should be
//   SSEKMSKeyId), so the KMS key id is silently dropped on PutObject.
// - cli-testing's S3.getObject writes the body to a shared file (out.loc)
//   which races under parallel execution.

.github/scripts/end2end/configure-e2e-ctst.sh
SylvainSenechal: you're absolutely right

tests/ctst/steps/serverSideEncryption.ts
maeldonn: Sorry, I missed that 👀

maeldonn (approved):
LGTM

PR #2383, opened 2026-04-21
refactor test organisation

SylvainSenechal:
Issue: ZENKO-5251

This pr is a bit savage, but it's mostly renames and changes in folder structure. Actually really not that long to review.

I recommend just pulling the branch and having a look at the /tests directory

I like this structure, I think it will be easier to maintain packages, we now have a single node 24 version to deal with accross all tests.
With this we move toward have a unique setup for all tests, this will also enable us to have utility functions that can be used across all tests which is what I need for another pr

This is one of the last pr I wanted to do before we can try to have a look (if we want to invest the time 👀 ) at re organizing the setup and installation of Zenko (all the .sh and .py files that are a bit all over the place)  

**EDIT FRIDAY 24TH APRIL :**

Thanks for the review everyone, I apologize I re did some big force push so the comments might be all over the place
TLDR : 
- Won't include /workflow tests : this pr does 
  - Reorganise functional tests in a folder, merging cucumber and mocha tests into their own javascript project
  - Moving some python setup into their own folder as they are used for both mocha and cucumber tests
  - I wanna get rid of the dependabots prs and update dependencies so : Big cleanup of the requirements.txt for the python setup + many javascript package upgrade too
  - Some refactoring of the node installation : everything now uses node 24

Why this pr ? 
I think previous comment and diffs speak for themselves but :
- with this new tests organisation we will be able to develop utility functions that can be used by any test suite. 
- One less package json to maintain, the more dependencies to maintain, the less we do it (old mocha test were using Vaultclient with a hash from 3 or 4 years ago, not compatible with more recent node version for example..)
- Better test structure for newcomers
- (probably will never do this, BUT) : Easier to bridge the gap between old mocha test and cucumber tests if we ever wanna do some migrations
- New test structure will make it easier to to setup the github ci : Basically, I wanna avoid having .sh file that are like "e2e-ctst.sh, e2e-mocha.sh" : We should try to have one setup that works for everything as mush as possible to reduce complexity


.github/actions/setup-node-env/action.yaml
SylvainSenechal, line 1: end2end.yaml file was like 700 lines, with like ~6 nodes setup so I think this is better

.github/workflows/end2end.yaml
```diff
@@ -348,66 +347,26 @@ jobs:
...
-          cache-dependency-path: tests/zenko_tests/node_tests/yarn.lock
-      - name: Install dependencies
-        working-directory: tests/zenko_tests/node_tests
-        run: yarn install --frozen-lockfile
-      - name: Lint e2e tests
-        working-directory: tests/zenko_tests/node_tests
-        run: yarn lint
-
-  lint-ctst:
+  lint-tests:
```
SylvainSenechal: One linter for everything now, since all tests fits in a single js project

tests/functional/configuration_setup/e2e_config/accounts.py
SylvainSenechal, line 1: We had some python scripts sitting in the /zenko_tests (old mocha tests) folder.
Yet these scripts are actually used for both cucumber and mocha tests. I moved these python script to their own folder. I was even thinking about putting them in .github/something but for now they are still in /tests, just in a separate subfolder

tests/mocha/backbeat/Using.md
```diff
@@ -36,7 +36,7 @@
     ```
 
 5. Install node and npm.
-6. Navigate to `Zenko/tests/node_tests/backbeat`.
+6. Navigate to `Zenko/tests/mocha/backbeat`.
 7. Install node modules: `npm i`.
```
SylvainSenechal, line 40: might need another pass later for these readmeS and .md
I did some update, but here I see for example "npm i" but we generally use yarn. I think some stuffs are stale but not gonna address in this pr

tests/functional/mocha/.mocharc.js
```diff
@@ -0,0 +1,8 @@
+// Allow CI to make mocha exit 0 on test failures, so that infrastructure
+// errors still propagate while test results are evaluated via JUnit reports.
+module.exports = {
+    'pass-on-failing-test-suite': process.env.CI_PASS_ON_TEST_FAILURE === 'true',
+    'exit': true,
+    'reporter': 'mocha-multi-reporters',
+    'reporter-options': 'configFile=mocha/mocha-reporter.json',
```
SylvainSenechal, line 7: this used to be embedded in the package.json yarn command of the mocha test

tests/functional/eslint.config.mjs
```diff
@@ -0,0 +1,78 @@
+import path from 'node:path';
```
SylvainSenechal, line 1: This file, as well as the as the new package.json and tsconfig.json are really just some kind of merge of what we already had before, not much change

francoisferrand:
Few remarks (did not review the full PR, just had a quick look):
- kind of orthogonal to your actual change, but not sure we should rename the folders (ctst → cucumber, Zenko-tests → mocha, etc) : best IMHO to name them by concept/scope than technology (which can change...), and keeping usage is simpler for people to get around (muscle memory)
- I trust you that it works, but I fear mixing different frameworks (act.js, mocha/jest/cucumber...) will eventually lead to dependency hell and lots of conflicts... or require changing everything at once (think AWS Sdk migration, it was kind of convenient to be able to migrate each test suite separately)
- There are more tests (recently introduced), e.g. for scripts
- As far as devx, it's a mixed bag: install all deps (node packages) in a single call and run all tests from one place; but longer download and esp. build times.... And yet we will still not be able to run all tests in one command, need to run each suite separately so it is manageable...
- Generally, I think it is better to have multiple "small" packages than a single "large" package (kind of distributed vs mono-repo debate). We need to share a package -which is great-, but I wonder if merging everything in really what we _want_ (and why) or just the solution we found to allow sharing... (Btw, how did William manage to introduce a common package , in his old PR?)

tests/functional/ctst/common/common.ts
maeldonn, line 1: In all our tickets, documentation, and other materials, we refer to these tests as 'ctst.' Do you think it's a good idea to rename them ?

maeldonn (commented):
Merging cucumber and mocha makes sense as they share most dependencies. But workflows shares nothing with them, and its postinstall (downloads the act binary) now runs for every yarn install. Could we keep it as its own package.json like before?

tests/functional/ctst/common/common.ts
SylvainSenechal: im gonna keep workflow separated, especially now that other python tests are coming inside
and also ctst instead of cucumber

tests/package.json
```diff
@@ -0,0 +1,91 @@
...
+    "test:crr_pause_resume": "mocha --config mocha/.mocharc.js -t 10000 --recursive mocha/backbeat/tests/crr-pause-resume",
+    "test:expiration": "mocha --config mocha/.mocharc.js -t 900000 mocha/backbeat/tests/lifecycle/expiration.js",
+    "test:transition": "mocha --config mocha/.mocharc.js -t 900000 mocha/backbeat/tests/lifecycle/transition.js",
+    "test:lifecycle": "mocha --config mocha/.mocharc.js -t 1800000 --recursive mocha/backbeat/tests/lifecycle",
+    "test:ingestion_oob_s3c": "mocha --config mocha/.mocharc.js -t 180000 --recursive mocha/backbeat/tests/ingestion",
+    "test:location_quota": "mocha --config mocha/.mocharc.js -t 10000 --recursive mocha/cloudserver/locationQuota/tests",
+    "test:bucket_get_v2": "mocha --config mocha/.mocharc.js -t 10000 --recursive mocha/cloudserver/bucketGetV2/tests",
+    "test:bucket_policy": "mocha --config mocha/.mocharc.js -t 10000 --recursive mocha/cloudserver/bucketPolicy/tests",
+    "test:all_extensions": "run-p --aggregate-output test:crr test:aws_crr test:expiration test:transition test:ingestion_oob_s3c",
+    "postinstall": "node node_modules/@kie/act-js/scripts/postinstall.js 0.2.75"
```
benzekrimaha: As long as @kie/act-js lives in the root tests/package.json, every job using the new setup-node-env composite action pays the download cost — including lint-tests which has no use for it. And postinstall is silent so nobody will notice it growing.
I think that you can move workflows back into its own package (as you said), keep tests/package.json for cucumber + mocha only.

.github/actions/setup-node-env/action.yaml
```diff
@@ -0,0 +1,57 @@
...
+
+inputs:
+  app-id:
+    description: GitHub App ID
+    required: true
+  private-key:
+    description: GitHub App private key
+    required: true
+  repositories:
+    description: Newline-separated list of private repositories to grant access to
```
benzekrimaha, line 15: before the refactor, each job explicitly requested only the repos it consumes:
lint-ctst → cli-testing
end2end-2-shards-http → zenko-operator
end2end-sharded → metadata, zenko-operator
now, unless explicitly overridden (and no caller overrides), every job receives a token with access to all 3 repos. And check-workflows ends up with that token while it previously needed no private repo at all. I don't know if that is okay but rather point it as it's hidden inside this composite action.
I think we should remove the default, force each caller to pass the exact list (like the original code). If you want, keep a minimal default of cli-testing and we can document the convention clearly.

tests/functional/mocha/.mocharc.js
```diff
@@ -0,0 +1,8 @@
+// Allow CI to make mocha exit 0 on test failures, so that infrastructure
+// errors still propagate while test results are evaluated via JUnit reports.
+module.exports = {
+    'pass-on-failing-test-suite': process.env.CI_PASS_ON_TEST_FAILURE === 'true',
+    'exit': true,
```
benzekrimaha, line 5: `pass-on-failing-test-suite` makes the process exit 0 even when tests fail. 
Meaning green CI ≠ tests pass; the real verdict depends on the downstream JUnit aggregator. This is fragile: if the JUnit reporting step does not run (e.g. infra failure), the job appears green while everything broke. I know this is not new (already in the mocha package.json before), but moving it into .mocharc.js makes it the default, almost invisible to new contributors. At a minimum we need a comment explaining why.
`'exit': true`:  Mocha explicitly does not recommend it (hides leaking handles → real bugs). Since we are touching this setup anyway, let's create a "drop --exit + fix the leaking handles" ticket and link it in the comment.

tests/package.json
```diff
@@ -0,0 +1,91 @@
+{
```
benzekrimaha: => mocha, mocha-junit-reporter, mocha-multi-reporters are in dependencies → these are test-only, should be devDependencies.
=> @aws-sdk/client-s3, @aws-sdk/client-iam, @aws-sdk/client-sts in devDependencies → but used at runtime by the cucumber/mocha tests. It is private: true so practically no consequence, but semantically inconsistent.
request: ^2.87.0 → deprecated since 2020.
uuid: ^3.0.1 → uuid 3 is EOL (uuid is at v11).
async: 2.1.2 → is pinned to a very old patch (v4 is current)- _I don't know if this will break something_-
=> werelogs: github:scality/werelogs → no tag/ref pinned. Any push on werelogs main can break the tests overnight. Pin a tag (cf. how arsenal is pinned).

tests/package.json
```diff
@@ -0,0 +1,91 @@
...
+    "test:transition": "mocha --config mocha/.mocharc.js -t 900000 mocha/backbeat/tests/lifecycle/transition.js",
+    "test:lifecycle": "mocha --config mocha/.mocharc.js -t 1800000 --recursive mocha/backbeat/tests/lifecycle",
+    "test:ingestion_oob_s3c": "mocha --config mocha/.mocharc.js -t 180000 --recursive mocha/backbeat/tests/ingestion",
+    "test:location_quota": "mocha --config mocha/.mocharc.js -t 10000 --recursive mocha/cloudserver/locationQuota/tests",
+    "test:bucket_get_v2": "mocha --config mocha/.mocharc.js -t 10000 --recursive mocha/cloudserver/bucketGetV2/tests",
+    "test:bucket_policy": "mocha --config mocha/.mocharc.js -t 10000 --recursive mocha/cloudserver/bucketPolicy/tests",
+    "test:all_extensions": "run-p --aggregate-output test:crr test:aws_crr test:expiration test:transition test:ingestion_oob_s3c",
+    "postinstall": "node node_modules/@kie/act-js/scripts/postinstall.js 0.2.75"
+  },
+  "resolutions": {
```
benzekrimaha: Why do we need this? Usually a resolutions entry signals an upstream bad pin or a version conflict. Can we document the reason in a comment, or we fix the issue it at the source if possible?

tests/functional/tsconfig.json
```diff
@@ -0,0 +1,34 @@
...
+        "strict": true,
+        "noImplicitAny": true,
+        "noImplicitThis": true,
+        "skipLibCheck": true,
+        "noImplicitReturns": true,
+        "noFallthroughCasesInSwitch": true,
+        "resolveJsonModule": true,
+        "baseUrl": "./",
+        "paths": {
+            "cli-testing": ["./node_modules/cli-testing"],
```
benzekrimaha, line 17: with this root tsconfig + allowJs: false + strict: true + cucumber-only paths, I am not sure how the .ts files under workflows/ get type-checked
`tsc --build` is in `build:cucumber` but without include/references it will pick up every .ts file under baseUrl, including workflows/**/*.ts
can we confirm yarn build:cucumber and yarn test:workflows pass ?

tests/functional/eslint.config.mjs
```diff
@@ -0,0 +1,78 @@
...
+        },
+    },
+
+    // --- JavaScript files (mocha) ---
+    {
+        files: ['mocha/**/*.js'],
+        extends: [
+            ...compat.extends('scality'),
+        ],
+        rules: {
```
benzekrimaha, line 47: I know it is "what we had before", but by merging the configs we formally inherit all this debt in a single place. Either:
- we accept this debt consciously and open a ticket to reduce it
or we use this opportunity to drop the ones that are not really relevant anymore
- and for workflows: '@typescript-eslint/no-explicit-any': 'off' — since we are consolidating, can we at least plan to remove it? otherwise we lose one of the benefits of unifying...

tests/package.json
SylvainSenechal: I'm re writting the pr, removed tests/workflow from this new package because yeah these tests are a bit appart from the functional tests, so we won't have this problem

.github/actions/setup-node-env/action.yaml
SylvainSenechal: Yeah I noticed this stuff, not sure it's super important, but yeah I removed the default value and re-add explicit repo list

tests/functional/mocha/.mocharc.js
SylvainSenechal: I mean as you said the pass on failing test suite was already there before, with a comment above explaining why its done, i dont really like that its a bit implicit and can be missed but not much to do here

The exit thing and the rest were also already there before but they were in the package json commands directly.

Now we have this 
    "test:operator": "mocha --config mocha/.mocharc.js -t 10000 mocha/init_test.js",
Before we had
    "test_operator": "mocha --exit -t 10000 --reporter mocha-multi-reporters --reporter-options configFile=config.json,cmrOutput=mocha-junit-reporter+testsuitesTitle+test_operator ./init_test.js",


I don't think we should bother too much with these tests, but I can remove the args in the package.json if you think it's more visible

tests/package.json
SylvainSenechal: Yes exactly there is an issue with azure client, some weird stuff with azurite moch being unbumpable because microsoft keeps on doing layoffs so their engineers are unable to update Azurite in time (https://github.com/Azure/Azurite/issues/2623), making it incompatible with their latest sdk versions

Anyways, I think I have created a ticket, or at least we have this in mind, even Francois had a similar issue with Sorbet around this

tests/functional/tsconfig.json
SylvainSenechal: I guess no need to answer this question since I removed workflow from this folder 🌚 
Which kinda raises the question about linting/building the workflow folder, but this is not the concern of this pr anymore

tests/functional/eslint.config.mjs
SylvainSenechal: I think for the second point no need to look at it anymore for now.

For the first point, so if I understand correctly, the best thing we want to do now is to inherit everything from scality without adding any other rules ?

SylvainSenechal:
@claude[agent] hello ?

SylvainSenechal:
@claude[agent] review this stuff, be VERY MEAN with me

tests/functional/package.json
```diff
@@ -0,0 +1,78 @@
...
+    "cli-testing": "git+https://github.com/scality/cli-testing.git#v1.3.0",
+    "mocha": "^11.7.5",
+    "mocha-junit-reporter": "^2.2.1",
+    "mocha-multi-reporters": "^1.1.7",
+    "prometheus-query": "^3.4.0",
+    "proper-lockfile": "^4.1.2",
+    "scubaclient": "git+https://github.com/scality/scubaclient.git#1.1.4",
+    "uuid": "^14.0.0",
+    "vaultclient": "scality/vaultclient#8.5.6",
+    "werelogs": "github:scality/werelogs"
```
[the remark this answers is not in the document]
SylvainSenechal: aint touching this now, i noticed many repo do the same thing

tests/functional/package.json
```diff
@@ -0,0 +1,78 @@
...
+  "dependencies": {
+    "@cucumber/cucumber": "^12.7.0",
+    "@kubernetes/client-node": "^1.4.0",
+    "@platformatic/kafka": "^1.30.0",
+    "@smithy/node-http-handler": "^4.0.0",
+    "@azure/storage-blob": "^12.12.0",
+    "@google-cloud/storage": "^7.12.1",
+    "arsenal": "scality/Arsenal#8.1.38",
+    "assert": "^2.1.0",
+    "async": "2.1.2",
```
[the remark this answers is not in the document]
SylvainSenechal: breaking changes are likely, not touching for now

maeldonn (approved):

CLAUDE.md
```diff
@@ -8,7 +8,7 @@
 - `solution/zenkoversion.yaml` — `ZenkoVersion` CR template (dashboards, policies, feature flags, capabilities, location types)
 - `solution/kafka/Dockerfile`, `solution/kafka-connect/Dockerfile` — Scality-built Kafka + Kafka Connect images
 - `solution-base/mongodb/` — MongoDB Helm charts and patches
-- `tests/ctst/` — TypeScript Cucumber end-to-end tests
+- `tests/functional/ctst/` — TypeScript Cucumber end-to-end tests
 - `tests/workflows/` — TypeScript Jest tests for CI tooling
 - `tests/zenko_tests/` — Python + Node.js integration tests
```
benzekrimaha, line 13: this still describes `tests/zenko_tests/` as Python + Node.js integration tests, but the PR moved those pieces under `tests/functional`.
Can we update this doc while touching the test layout?

.github/copilot-instructions.md
```diff
@@ -10,7 +10,7 @@ Key paths:
 - `solution/zenkoversion.yaml` — `ZenkoVersion` CR template (dashboards, policies, feature flags, capabilities, location types)
 - `solution/kafka/Dockerfile`, `solution/kafka-connect/Dockerfile` — Scality Kafka + Kafka Connect images
 - `solution-base/mongodb/` — MongoDB Helm charts and patches
-- `tests/ctst/` — TypeScript Cucumber end-to-end tests
+- `tests/functional/ctst/` — TypeScript Cucumber end-to-end tests
 - `tests/workflows/` — TypeScript Jest tests for CI tooling
 - `tests/zenko_tests/` — Python + Node.js integration tests
```
benzekrimaha, line 15: same here the path is stale after the moves made in the PR

benzekrimaha (approved):
LGTM overall

CLAUDE.md
SylvainSenechal: thank you, I'm gonna make the change in the next pr so i don't rerun everything here

PR #2384, opened 2026-04-22
Add claude code review

francoisferrand:
Introduce claude code review & dependabot reviews.

Issue: ZENKO-5260

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1,0 +1,43 @@
+# Zenko — Copilot / AI Reviewer Instructions
+
+## Repo context
+
+This is **Scality's open-source multi-cloud data controller** — a Kubernetes-based deployment of the Zenko stack. The actual service implementations (cloudserver, backbeat, vault2, sorbet, scuba, s3utils, pensieve-api, kafka-cleaner, drctl, zenko-operator) live in separate repos; this repo packages, versions, and tests them together.
+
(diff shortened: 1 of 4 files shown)


.github/copilot-instructions.md
```diff
@@ -0,0 +1,43 @@
+# Zenko — Copilot / AI Reviewer Instructions
+
+## Repo context
```
maeldonn, line 3: Is `CLAUDE.md` not enough? Do we really need to duplicate key paths ?

maeldonn (approved):

.github/copilot-instructions.md
SylvainSenechal: I guess its the claude.md that's pointing to copilot.md.
But wonder why we have to add in claude.md the "what lives there" and "toolchain" section, when everything could be on this file, and claude.md just pointing at this file 🤔

SylvainSenechal (approved):

.github/copilot-instructions.md
francoisferrand: copilot-instructions is designed to provide _review instructions_: it is read by /review-pr and copilot when doing reviews, and that's it.
- these tools will not read CLAUDE.md, so some context may be needed here to help
- it provides specific _review_ instructions/criterias, which are not relevant for generic Claude Code

PR #2391, opened 2026-04-22
Add missing tests

francoisferrand:
- Add tests for archive-artifacts
- Add merge-junit-reports tests
- Move workflow test resources to folder

Issue: ZENKO-5222


delthas (approved):

.github/workflows/end2end.yaml
```diff
@@ -144,6 +144,20 @@ jobs:
     uses: ./.github/workflows/alerts.yaml
     secrets: inherit
 
+  check-scripts:
```
SylvainSenechal, line 147: Is it not possible to have these tests be javascript and merged with check-workflows ? haven't looked exactly at what they do but it seems related to checking github ci stuff

francoisferrand: I would have preferred, but the tests directly import the python script, so that individual functions can be tested unitary.

And the script was written in python as it is -afaik- the only tool available (in gha) which can process xml natively, without installing extra package.

SylvainSenechal (approved):

benzekrimaha (approved):

PR #2392, opened 2026-04-23
✨(ZENKO-5262) Define new alerts and enable monitor for Zenko Operator

DarkIsDude:
Issue: ZENKO-5262

Define new alerts and enable monitoring of the operator.

https://github.com/scality/zenko-operator/pull/608/changes

Fixes ZENKO-5262.

### DIFF
--- a/solution/build.sh
+++ b/solution/build.sh
@@ -77,6 +77,11 @@ spec:
     image:
       name: zenko-operator
       tag: '$(zenko_operator_tag)'
+    metrics:
+      enabled: true
+      scheme: http
+      port: 8080
+      path: /metrics
 EOF
 }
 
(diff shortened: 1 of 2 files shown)


monitoring/zenko-operator/alerts.test.yaml
francoisferrand, line 1: usually we keep the alerts along with the components which "owns" these metrics, to reduce coupling and ensure we rename the metrics and alerts/dashboards at the same time.

That would mean both alerts.yaml and building the Oras image would be in Zenko-operator; and in Zenko we would simply need to change the content of zenkoversion to reference zkop's dashboard image (along with the manifest change to enable scraping)

any reason not to do this for these?

SylvainSenechal (approved):

delthas (approved):
LGTM. Could you squash commits before merging? I see you've added then removed a large chunk of code.

francoisferrand (approved):

PR #2394, opened 2026-04-24
refactor env parameters

SylvainSenechal:
Issue: ZENKO-5259

Mostly refactoring around tests configuration.

- Lots of tests parameters are shared between mocha and cucumber tests => setup a configuration folder that can be used by both test suite to load tests parameters
- Trying to offload the work from bash script directly into the tests
- Many parameters are still not added in this pr, because :
  - the pr would become too big
  - some parameters are still used in bash or python scripts => Putting them in the js files would not allow us to remove them from where they currently sit, so we would end up with duplicated declarations which is worse. Longer term we have tickets to migrate python setup scripts into the tests, so this migration can be done at that time
  - We have a fully typed configuration clean object holding all the tests variable/configuration


tests/functional/tests_common/configuration.ts
```diff
@@ -0,0 +1,225 @@
...
+    };
+}
+
+export const config = {} as TestsConfiguration;
+
+export const initConfig = async (): Promise<void> => {
+    if (Object.keys(config).length > 0) return;
+    const zenkoAccountCredentials = await loadZenkoAccount();
+    const clients = createClients(zenkoAccountCredentials);
+    Object.assign(config, {
```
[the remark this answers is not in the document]
SylvainSenechal: yep and im fine with it. 
wrapping each loader would means that every time someone wants to use a field in this config, they first have to wonder "is this gonna be undefined because im in this particular test or not", and have some weird failures later down the line because some variable is undefined
I don't want that. Good long term goal is one zenko setup for all tests

basically fail fast and be explicit

tests/functional/tests_common/configuration.ts
```diff
@@ -0,0 +1,225 @@
+import { S3Client } from '@aws-sdk/client-s3';
+import { IAMClient } from '@aws-sdk/client-iam';
+import { NodeHttpHandler } from '@smithy/node-http-handler';
+import { getSecretByLabel, getSecretByName, getCustomObject } from "./kubernetes";
+
+const ZENKO_NAME = 'end2end';
```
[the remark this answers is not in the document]
SylvainSenechal: kinda good catch but actually what im gonna change here is the positional argument in setup e2e, which is never used, so im removing it

.devcontainer/devcontainer.json
SylvainSenechal, line 1: Review the /tests_common folder first, at the end of this pr

.github/scripts/end2end/setup-e2e-env.sh
```diff
@@ -7,7 +7,7 @@ _SETUP_DIR="$(cd "$(dirname "${BASH_SOURCE[0]:-$0}")" && pwd)"
 
 . "$_SETUP_DIR/common.sh"
 
-ZENKO_NAME="${ZENKO_NAME:-end2end}"
```
SylvainSenechal, line 10: argument never provided so removing ambiguity

.github/scripts/end2end/setup-e2e-env.sh
```diff
@@ -131,40 +125,8 @@ export TARGET_VERSION=$(sed -n 's/^VERSION="\([^"]*\)"/\1/p' "${ZENKO_ROOT}/VERS
...
-    export STORAGE_ACCOUNT_OWNER_USER_NAME="storage_account_owner"
-    export DATA_CONSUMER_USER_NAME="data_consumer"
-    export DATA_ACCESSOR_USER_NAME="data_accessor"
-    # env vars used by cli-testing's Keycloak.ts seeder
-    export ACCOUNT="${ZENKO_ACCOUNT_NAME}"
-    export STORAGE_MANAGER="${STORAGE_MANAGER_USER_NAME}"
-    export STORAGE_ACCOUNT_OWNER="${STORAGE_ACCOUNT_OWNER_USER_NAME}"
-    export DATA_CONSUMER="${DATA_CONSUMER_USER_NAME}"
-    export DATA_ACCESSOR="${DATA_ACCESSOR_USER_NAME}"
+    export ACCOUNT="zenko-ctst"
```
SylvainSenechal, line 128: some stuffs like this needs to stay here, they are used by cli-testing, or are coupled with other script/logic and can't be migrated easily without some other preliminary work

tests/functional/ctst/common/common.ts
```diff
@@ -39,8 +40,7 @@ export async function cleanS3Bucket(
         // Do not try to clean a bucket with compliance retention
         return;
     }
-    Identity.useIdentity(IdentityEnum.ACCOUNT, world.getSaved<string>('accountName') ||
-        world.parameters.AccountName);
+    Identity.useIdentity(IdentityEnum.ACCOUNT, world.getSaved<string>('accountName') || ZENKO_ACCOUNT_NAME);
```
SylvainSenechal, line 43: you'll see a lot of similar changes like this in the pr : 
I'm removing ambiguity as much as possible when it is not used and not useful

tests/functional/ctst/steps/utilization/utilizationAPI.ts
```diff
@@ -27,8 +28,8 @@ When('the user retrieves utilization metrics using scubaclient for metric type {
         this.addToSaved('metricType', metricType);
 
         const client = new ScubaClient({
-            port: parseInt(this.parameters.UtilizationServicePort),
-            host: this.parameters.UtilizationServiceHost,
+            port: 80,
```
SylvainSenechal, line 31: could be stored in some UTILIZATION_SERVICE_PORT = 80 const, I did it for a few other variables.
Not super important, its just always 80 defautl http port

tests/functional/ctst/world/Zenko.ts
```diff
@@ -92,18 +79,7 @@ export interface ZenkoWorldParameters extends ClientOptions {
...
-    KafkaGCRequestTopic: string;
-    InstanceID: string;
-    BackbeatApiHost: string;
-    BackbeatApiPort: string;
-    KafkaCleanerInterval: string;
-    SorbetdRestoreTimeout: string;
-    UtilizationServiceHost: string;
-    UtilizationServicePort: string;
-    [key: string]: unknown;
+    KubeconfigPath?: string;
```
SylvainSenechal, line 79: this was the only [key: string]: unknown
It's not unknown at all 🧐

tests/functional/ctst/world/Zenko.ts
```diff
@@ -175,44 +157,37 @@ export default class Zenko extends World<ZenkoWorldParameters> {
 
         CacheHelper.savedAcrossTests[Zenko.PRA_INSTALL_COUNT_KEY] = 0;
 
-
-        if (this.parameters.AccountName && !Identity.hasIdentity(IdentityEnum.ACCOUNT, this.parameters.AccountName)) {
-            Identity.addIdentity(IdentityEnum.ACCOUNT, this.parameters.AccountName, {
-                accessKeyId: this.parameters.AccountAccessKey,
-                secretAccessKey: this.parameters.AccountSecretKey,
-            });
+        Identity.defaultAccountName = ZENKO_ACCOUNT_NAME;
```
SylvainSenechal, line 149: Side note on this whole long Identity thing part : 
I'm thinking this is gonna change again in another pr, I don't find it super clear, lot's of ifs(somethingDefined), very smelly. It works but not very clean

tests/functional/ctst/world/Zenko.ts
```diff
@@ -276,19 +251,19 @@ export default class Zenko extends World<ZenkoWorldParameters> {
             await this.prepareIamUser();
             break;
         case EntityType.STORAGE_MANAGER:
-            await this.prepareARWWI(this.parameters.StorageManagerUsername || 'storage_manager',
+            await this.prepareARWWI('storage_manager',
```
SylvainSenechal, line 243: I will probably redefine constants later, but same thing : this is also defined in cli-testing, there is some annoying coupling, don't wanna touch it too much if its gonna change again later

tests/functional/mocha/backbeat/tests/api/metrics.js
```diff
@@ -116,11 +116,14 @@ describe('Backbeat replication metrics route validation', function dF() {
 });
 
 describe('Backbeat replication metrics data', function dF() {
+    let scalityUtils;
```
SylvainSenechal, line 119: Lots of upcoming pattern like this one, moving top level definition into before hooks : 
Can't define these at the top level because the config is not loaded yet at that time (clients are created a in mocha before all setup)

tests/functional/mocha/s3SDK.js
SylvainSenechal, line 1: this whole file should disappear, and we should just initialize clients from the widely available config.
then tests could simply call config.vaultClient, config.s3Client etc

tests/functional/mocha/setup.ts
SylvainSenechal, line 1: this is picked up & run automatically by mocha on startup

tests/functional/mocha/VaultClient.js
```diff
@@ -129,32 +129,16 @@ class VaultClient {
...
-                cert = fs.readFileSync(
-                    process.env.VAULT_SSL_CERT || '/conf/test.crt',
-                    'ascii',
-                );
-                key = fs.readFileSync(
-                    process.env.VAULT_SSL_KEY || '/conf/test.key',
-                    'ascii',
-                );
-            }
+        if (VAULT_ENDPOINT.startsWith('https://')) {
```
SylvainSenechal, line 138: vaultendpoint is hardcoded as http so this is never run 👀 

Keeping it for now because I'm thinking at some there was probably a good reason to have these, maybe will remove later, not really the role of this pr

tests/functional/tests_common/configuration.ts
```diff
@@ -0,0 +1,225 @@
...
+    ServiceUsers: {
+        BackbeatLifecycleBp1: ServiceUserCredentials;
+        BackbeatLifecycleConductor1: ServiceUserCredentials;
+        BackbeatLifecycleOp1: ServiceUserCredentials;
+        BackbeatQp1: ServiceUserCredentials;
+        SorbetFwd2: ServiceUserCredentials;
+    };
+    PRAAdmin?: ServiceUserCredentials;
+    AdminCredentials: ServiceUserCredentials;
+    ZenkoAccount: {
```
SylvainSenechal, line 83: Might deserve some renaming later or re organisation later, or structure/substrucutre, but it will come more naturally when more variables are migrated here

.github/scripts/end2end/setup-e2e-env.sh
francoisferrand, line 144: not for this PR, but kafka port forward can be setup from tests as well.

(also, FYI: in Golang you can (in the kafka client) handle the "resolve" part, so you don't need to modify /etc/hosts ; it would be great if we can do the same in node...)

.github/scripts/end2end/setup-e2e-env.sh
```diff
@@ -235,20 +197,6 @@ else
     fi
     export PROMETHEUS_SERVICE="${PROMETHEUS_SVC}.${NAMESPACE}.svc.cluster.local"
```
francoisferrand, line 198: could be dynamically retrieved as well?
there is most likely a single prometheus service in the cluster: we could find it by labels ("managed by prometheus operator") or checking the prometheus CR...

.github/scripts/end2end/setup-e2e-env.sh
```diff
@@ -235,20 +197,6 @@ else
...
-    export BACKBEAT_API_HOST="backbeat-api.zenko.local"
-    export BACKBEAT_API_PORT="80"
-
-    # Utilization service
-    export UTILIZATION_SERVICE_HOST=$(kubectl get zenko ${ZENKO_NAME} -o jsonpath='{.spec.scuba.api.ingress.hostname}')
-    export UTILIZATION_SERVICE_PORT="80"
-
     # Azure archive settings
     export AZURE_ARCHIVE_ACCESS_TIER="Hot"
     export AZURE_ARCHIVE_MANIFEST_ACCESS_TIER="Hot"
```
francoisferrand: nothing specific here (I think), should be in tests or read from cluster...

tests/functional/ctst/common/common.ts
francoisferrand: instead of updating code all over the place, should we not initialize the world (why drop support for `world.parameters`) once on startup?

tests/functional/ctst/common/hooks.ts
```diff
@@ -21,6 +22,12 @@ import {
 import { createKubeCustomObjectClient, waitForZenkoToStabilize } from 'steps/utils/kubernetes';
 import { startDLQConsumer, stopDLQConsumer } from 'steps/utils/kafka';
 
+BeforeAll(async function () {
+    // Some hooks are defined in cli-testing and use the configuration,
+    // we need to have this run before anything else
+    await initConfig();
```
francoisferrand, line 28: since we need to init config anyway : best initialize the `world`, which is the CTST (or even cucumber?) concept to hold such configuration

tests/functional/ctst/world/Zenko.ts
francoisferrand: no reason (anymore) to be there I think; at the beginning world was not typed, then proper fields were added but still failed to compile (some missing): must have been fixed now.

tests/functional/ctst/world/Zenko.ts
```diff
@@ -153,17 +129,23 @@ export default class Zenko extends World<ZenkoWorldParameters> {
...
-                JSON.parse(this.parameters.ServiceUsersCredentials) as Record<string, ServiceUsersCredentials>;
-            for (const serviceUserName in serviceUserCredentials) {
-                if (!Identity.hasIdentity(IdentityEnum.SERVICE_USER, serviceUserName, this.parameters.AccountName)) {
-                    Identity.addIdentity(IdentityEnum.SERVICE_USER, serviceUserName, {
-                        accessKeyId: serviceUserCredentials[serviceUserName].accessKey,
-                        secretAccessKey: serviceUserCredentials[serviceUserName].secretKey,
-                    }, this.parameters.AccountName);
+        if (config?.ServiceUsers) {
+            const serviceUsers = config.ServiceUsers;
+            const entries: [string, ServiceUserCredentials][] = [
```
francoisferrand: nit: maybe we could remove this list by looking at the secrets in the cluster instead

tests/functional/ctst/world/Zenko.ts
```diff
@@ -221,7 +196,7 @@ export default class Zenko extends World<ZenkoWorldParameters> {
     }
 
     private needsSecondarySite() {
-        return this.parameters.DRAdminAccessKey && this.parameters.DRAdminSecretKey && this.parameters.DRSubdomain;
+        return config.PRAAdmin && this.parameters.DRSubdomain;
```
francoisferrand: "DR" is more appropriate, this is the name of the feature (and component) in Zenko (PRA is customer-specific branding)

tests/functional/mocha/backbeat/tests/api/metrics.js
francoisferrand: not sure it's better, but could be done if the "config" exported by 'tests_common/configuration' is the _result_ of initConfig (same pattern as `cloudserver`'s Config)

tests/functional/tests_common/configuration.ts
```diff
@@ -0,0 +1,225 @@
...
+        GcRequest: string;
+    }
+    ServiceUsers: {
+        BackbeatLifecycleBp1: ServiceUserCredentials;
+        BackbeatLifecycleConductor1: ServiceUserCredentials;
+        BackbeatLifecycleOp1: ServiceUserCredentials;
+        BackbeatQp1: ServiceUserCredentials;
+        SorbetFwd2: ServiceUserCredentials;
+    };
+    PRAAdmin?: ServiceUserCredentials;
```
francoisferrand: keep "DR" instead of "PRA"

tests/functional/tests_common/configuration.ts
```diff
@@ -0,0 +1,225 @@
...
+        KafkaCleanerInterval: string;
+        SorbetdRestoreTimeout: string;
+        UtilizationServiceHost: string;
+    };
+}
+
+export const config = {} as TestsConfiguration;
+
+export const initConfig = async (): Promise<void> => {
+    if (Object.keys(config).length > 0) return;
```
francoisferrand:
```suggestion
    if (Object.keys(config).length > 0) {
        return;
    }
```

tests/functional/tests_common/configuration.ts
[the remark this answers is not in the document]
francoisferrand: we should keep it.
goal (eventually) is to be able to run against any zenko instance (e.g. artesca), so best to keep it: it does not hurt.

tests/functional/tests_common/configuration.ts
```diff
@@ -0,0 +1,225 @@
...
+    const lcBp = 'backbeat-lifecycle-bp-1';
+    const lcConductor = 'backbeat-lifecycle-conductor-1';
+    const lcOp = 'backbeat-lifecycle-op-1';
+    const qp = 'backbeat-qp-1';
+    const serviceUsers: Record<string, string> = {
+        'backbeat-lcbp-user-creds': lcBp,
+        'backbeat-lcc-user-creds': lcConductor,
+        'backbeat-lcop-user-creds': lcOp,
+        'backbeat-qp-user-creds': qp,
+    };
```
francoisferrand: there is a similar map in another place, should be dedup'ed (or even read from Zenko's secrets...)

francoisferrand (commented):
Great improvement to move from setup script to TS module;

However lots of the changes seems to be here just for the sake of it: the `config` object is really  just being used as `world` was, so it would all more naturally fit to split responsibility:
- tests_common handles reading all parameters from the environment, (eventually) helpers to setup tunnels and clients...
- [worldParameters](https://github.com/cucumber/cucumber-js/blob/main/docs/support_files/world.md#world-parameters) is THE standard way to pass parameters in cucumber.js (and already used), so it should stay : the change should just update these world parameters using tests_common

This would greatly reduce the scope of the change, while separating responsibility (reading env VS adapting config for cucumber).

tests/functional/tests_common/configuration.ts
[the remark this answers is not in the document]
SylvainSenechal: Ok I kept it but current state is not super nice : grep 'end2end' returns 164 results so if we want to use that variable zenko name there are many places that will need to change

tests/functional/ctst/common/common.ts
SylvainSenechal: The current world parameters is currently automatically loaded from the --world-parameters json value.
The new type I added loads value with a kubernete client

Do you mean that you would prefer all the parameters staying in the ZenkoWorldParameters ? I think it's possible but may introduce some other complexities (the world is recreated before each scenario in a Before hook, with Zenko.init), + I want the new TestParameters to be usable with any tests suite, and I don't like the --world-parameters so much (only strings allowed, need to define values in bash script) so i want to avoid ultimately have everything in the new TestParameters interface

tests/functional/ctst/world/Zenko.ts
SylvainSenechal: yes the [key: string]: unknow is removed, and kubeconfigPath added as it was not defined, but its needed

tests/functional/mocha/backbeat/tests/api/metrics.js
SylvainSenechal: I'm not sure which pattern you are referencing from cloudserver

.github/scripts/end2end/setup-e2e-env.sh
SylvainSenechal: Ok I didn't know we could do that from the tests directly, we can do it later yeah

.github/scripts/end2end/setup-e2e-env.sh
SylvainSenechal: There are multiple values we can load that I didn't add yet
My problem is, for PROMETHEUS_SERVICE for example, its built from PROMETHEUS_SVC, but PROMETHEUS_SVC is used right before that to do the port forward. So If I move PROMETHEUS_SERVICE to the ts code, I still need to have it defined in the sh script and then we have the same variable defined in 2 places. Best is to keep it here for now, as you said later we should be able to do the port forward from the node code so we can do it here

.github/scripts/end2end/setup-e2e-env.sh
SylvainSenechal: Yeah i can move it now since you mention it, but the reason I didnt do it is just that I didnt fully review the whole file as the pr is getting big

tests/functional/ctst/common/hooks.ts
SylvainSenechal: I explored a bit the best strategy to deal with this config initialization, but I think here is the best place to do it.
The World class is created for each scenario, the constructor of the world can't be asynchronous, so its a bit tricky to do this initialization in the world constructor or as a static config. 
BeforeAll is run once per worker, before the world creation, and before pretty much everything since its the first beforeAll, so the config is loaded once and available everywhere

tests/functional/ctst/world/Zenko.ts
SylvainSenechal: yes we can, and i just did some refactoring around it

tests/functional/tests_common/configuration.ts
SylvainSenechal: yes did some deduplication and refactoring around this

tests/functional/ctst/common/common.ts
francoisferrand: I mean parameters should be a `World` -i.e. `Zenko`-, as it is the standard way to define the environment for cucumber tests.

We may keep parameters in `ZenkoWorldParameters`, or possibly remove/tweak some; and use the new "config" lib to initialize the Zenko world instance. But IMHO there should not be coupling between tests and the config lib: should be handled through the world.

tests/functional/mocha/backbeat/tests/api/metrics.js
francoisferrand: in cloudserver, config.js it looks like this:

```
class Config {
    constructor() {
       // load config file....
    }
};

export {
  config: new Config(),
  Config,
};
```
so when you import `config` from cloudserver you immediately get the actual config which was loaded

.github/scripts/end2end/setup-e2e-env.sh
francoisferrand: please create followup ticket

tests/functional/ctst/common/hooks.ts
francoisferrand: so World constructor will be called after config is init, and can used the config directly?

tests/functional/tests_common/configuration.ts
francoisferrand: even if it ends up being changed later, "ZenkoAccount" struct is really weird : an s3Client and IamClient are not part of an account ; and credentials is the actual identifiers of the account...
→ better to just remove the nesting?

.github/scripts/end2end/setup-e2e-env.sh
francoisferrand: please create followup ticket

francoisferrand (commented):
as discussed, pending some rework before reviewing

PR #2399, opened 2026-05-04
Increase robustness of transition tests by watching the dead letter queue

SylvainSenechal:
Increase robustness of transition tests by watching the dead letter queue : 

Checking DLQ on 2 situations :
- One for the very generic "should be transitioning" step that's used in many tests
- Another one more niche for restoring objects

This enable earlier returns in case of tests issues, and allow to rely on real logic instead of timeout for the second test.

Ran in a broken codespace where the python setup wasn't properly working, I did found the DLQ errors this way, then fix the codespace and the dlq errors were gone.

Also reviewed a test which logic wasn't very clear.

Issue: ZENKO-4906

### DIFF
--- a/tests/functional/ctst/features/azureArchive.feature
+++ b/tests/functional/ctst/features/azureArchive.feature
@@ -238,8 +238,7 @@ Feature: Azure Archive
     And object "retry-obj-2" should be "transitioning" and have the storage class "e2e-azure-archive"
     And manifest containing object "retry-obj-1" should "contain" object "retry-obj-2"
     When i restore object "retry-obj-1" for <restoreDays> days
-    Then blob for object "retry-obj-1" fails to rehydrate
-    And blob for object "retry-obj-2" fails to rehydrate
+    Then restoration of object "retry-obj-1" failed and ends up in DLQ
     Then object "retry-obj-1" should be "transitioning" and have the storage class "e2e-azure-archive"
     When i run sorbetctl to retry failed restore for "e2e-azure-archive" location
     Then object "retry-obj-1" should be "restored" and have the storage class "e2e-azure-archive"
--- a/tests/functional/ctst/steps/azureArchive.ts
+++ b/tests/functional/ctst/steps/azureArchive.ts
@@ -8,6 +8,7 @@ import util from 'util';
 import { exec } from 'child_process';
 import Zenko from 'world/Zenko';
 import { waitForDataServicesToStabilize, waitForZenkoToStabilize } from './utils/kubernetes';
+import { waitForDLQMessage } from './utils/kafka';
 
 type manifestEntry = {
     'archive-id': string,
@@ -308,22 +309,29 @@ Then('blob for object {string} must be rehydrated',
         assert.strictEqual(sent, true, `Failed to send BlobCreatedEvent for ${tarName}, object ${objectName}`);
     });
 
-/**
- * This is used to intentionally fail rehydration
- * To do that, we verify that the blob is rehydrated in azure
- * But we don't send the event to the queue so that
- * zenko is not aware of the rehydration and mark it as failed
- */
-Then('blob for object {string} fails to rehydrate',
+Then('restoration of object {string} failed and ends up in DLQ',
     async function (this: Zenko, objectName: string) {
         const tarName = await isObjectRehydrated(this, objectName);
+        assert(tarName);
+        // Skipping sending BlobCreatedEvent to queue,
+        // which fails the restore and eventually ends up in DLQ
+
+        const objName = objectName || this.getSaved<string>('objectName');
+        const bucketName = this.getSaved<string>('bucketName');
 
         // wait for restore to fail and end up in dead letter queue
-        const restoreTimeoutSeconds = parseInt(this.parameters.SorbetdRestoreTimeout);
-        await Utils.sleep(restoreTimeoutSeconds * 1000 + 1000);
-        assert(tarName);
-        // restoreTimeout is set to 30s in the config
-        await Utils.sleep(30000);
+        const restoreTimeoutMs = parseInt(this.parameters.SorbetdRestoreTimeout) * 1000;
+        if (isNaN(restoreTimeoutMs)) {
+            throw new Error('SorbetdRestoreTimeout world parameter is missing or not a valid integer');
+        }
+        const timeoutMs = restoreTimeoutMs + 30_000 + 1_000;
+
+        const seenDLQ = this.getSaved<Set<string>>('seenDLQMessages') ?? new Set<string>();
+        this.addToSaved('seenDLQMessages', seenDLQ);
+        const dlqMsg = await waitForDLQMessage(Zenko.dlqBuffer, 'restore', objName, bucketName, seenDLQ, timeoutMs);
+        this.logger.info('Found rehydration failure in DLQ', {
+            objectName: objName, reason: dlqMsg.reason,
+        });
     });
 
 Then('the storage class of object {string} must stay {string} for {int} seconds',
--- a/tests/functional/ctst/steps/utils/kafka.ts
+++ b/tests/functional/ctst/steps/utils/kafka.ts
@@ -1,4 +1,100 @@
 import { Utils } from 'cli-testing';
+import { Consumer, stringDeserializers } from '@platformatic/kafka';
+
+export interface DLQMessage {
+    op: string;
+    bucketName: string;
+    objectKey: string;
+    accountId: string;
+    eTag: string;
+    archiveInfo: Record<string, unknown>;
+    requestId: string;
+    originalMessage: string;
+    reason: string;
+    date: string;
+}
+
+export function dlqKey(op: string, bucketName: string, objectKey: string): string {
+    return `${op}:${bucketName}:${objectKey}`;
+}
+
+let dlqConsumerCleanup: (() => Promise<void>) | undefined;
+
+/**
+ * Starts a long-lived background Kafka consumer. Each arriving DLQ message is
+ * passed to `onMessage` — the caller owns the buffer. Call once from BeforeAll.
+ */
+export async function startDLQConsumer(
+    kafkaHosts: string, topic: string,
+    onMessage: (msg: DLQMessage) => void,
+): Promise<void> {
+    const consumer = new Consumer({
+        clientId: `zenko-e2e-dlq-${Utils.randomString()}`,
+        groupId: `zenko-e2e-dlq-${Utils.randomString()}`,
+        bootstrapBrokers: [kafkaHosts],
+        deserializers: stringDeserializers,
+    });
+    const stream = await consumer.consume({ topics: [topic], mode: 'earliest' });
+    void (async () => {
+        try {
+            for await (const msg of stream) {
+                try { onMessage(JSON.parse(msg.value) as DLQMessage); } catch { /* ignore malformed messages */ }
+            }
+        } catch { /* consumer disconnected — stop silently */ }
+    })();
+    dlqConsumerCleanup = async () => {
+        await stream.close().catch(() => {});
+        await consumer.close().catch(() => {});
+    };
+}
+
+/** Stops the background DLQ consumer. Call from AfterAll. */
+export async function stopDLQConsumer(): Promise<void> {
+    await dlqConsumerCleanup?.();
+}
+
+/**
+ * Synchronous, non-blocking check of a DLQ buffer.
+ * Returns the first unseen message matching (op, objectName, bucketName),
+ * or undefined if none has arrived yet. The caller is responsible for
+ * sleeping and retrying if needed.
+ *
+ * seenMessages prevents the same Kafka message being matched twice across
+ * retries — e.g. a failed-restore DLQ entry must not re-trigger a later
+ * "wait for successful restore" check on the same object.
+ */
+export function pollDLQBuffer(
+    buffer: Map<string, DLQMessage[]>,
+    op: string, objectName: string, bucketName: string,
+    seenMessages: Set<string>,
+): DLQMessage | undefined {
+    return buffer.get(dlqKey(op, bucketName, objectName))
+        ?.find(m => !seenMessages.has(m.requestId));
+}
+
+/**
+ * Blocks until a matching DLQ message appears or timeoutMs elapses.
+ * Used when a test step explicitly expects a failure to land in the DLQ.
+ */
+export async function waitForDLQMessage(
+    buffer: Map<string, DLQMessage[]>,
+    op: string, objectName: string, bucketName: string,
+    seenMessages: Set<string>, timeoutMs: number,
+): Promise<DLQMessage> {
+    const deadline = Date.now() + timeoutMs;
+    while (Date.now() < deadline) {
+        const msg = pollDLQBuffer(buffer, op, objectName, bucketName, seenMessages);
+        if (msg) {
+            seenMessages.add(msg.requestId);
+            return msg;
+        }
+        await Utils.sleep(500);
+    }
+    throw new Error(
+        `DLQ: no "${op}" message for object "${objectName}" in bucket "${bucketName}" ` +
+        `within ${timeoutMs / 1000}s`,
+    );
+}
 
 interface ConnectorInfo {
     info: {
--- a/tests/functional/ctst/steps/utils/utils.ts
+++ b/tests/functional/ctst/steps/utils/utils.ts
@@ -16,6 +16,7 @@ import assert from 'assert';
 import constants from 'common/constants';
 import { getLocationConfigs } from './kubernetes';
 import { S3Client } from '@aws-sdk/client-s3';
+import { pollDLQBuffer } from './kafka';
 
 enum AuthorizationType {
     ALLOW = 'Allow',
@@ -473,6 +474,8 @@ async function putBucketReplication(
     }
 }
 
+// Polls for transition status and checks the DLQ buffer on every iteration for early failure.
+// Relies on bucket names being randomized so a DLQ entry from a previous run cannot match.
 async function verifyObjectLocation(this: Zenko, objectName: string,
     objectTransitionStatus: string, storageClass: string) {
     const objName =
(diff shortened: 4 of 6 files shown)


tests/functional/ctst/features/azureArchive.feature
```diff
@@ -238,8 +238,7 @@ Feature: Azure Archive
     And object "retry-obj-2" should be "transitioning" and have the storage class "e2e-azure-archive"
     And manifest containing object "retry-obj-1" should "contain" object "retry-obj-2"
     When i restore object "retry-obj-1" for <restoreDays> days
-    Then blob for object "retry-obj-1" fails to rehydrate
-    And blob for object "retry-obj-2" fails to rehydrate
```
SylvainSenechal, line 242: This is not needed and was silently failing before, but passing.
- It was passing because this step was ONLY waiting for some time
- Now that it is not just sleeping but checking the DLQ, I found that the entry for obj-2 never appear in the DLQ, only for obj-1, and this is completely logic because line 240 we only try to restore obj-1

Renamed the step because the phrasing was very confusing, made it look like the obj fails to rehydrate because of what we did in the previous step line 240, when in reality we fail the rehydration voluntarily from within that step

tests/functional/ctst/steps/utils/utils.ts
```diff
@@ -473,6 +474,9 @@ async function putBucketReplication(
     }
 }
 
+// This function polls for transition status while watching the Kafka DLQ for early failure.
+// It relies on the bucket name being randomized as otherwise it may match
+// a DLQ message from a previous run when retrying a test.
 async function verifyObjectLocation(this: Zenko, objectName: string,
```
SylvainSenechal, line 479: ugly diff but really not that complicated if you pull the branch.
Same logic as before, but instead of just 
- while looping until we get a result || timeout
we do :
- while looping until we get a result || timeout || entry in DLQ

tests/functional/ctst/steps/utils/utils.ts
```diff
@@ -484,45 +488,122 @@ async function verifyObjectLocation(this: Zenko, objectName: string,
...
+        deserializers: stringDeserializers,
+    });
+    const dlqStream = await consumer.consume({
+        topics: [this.parameters.KafkaDeadLetterQueueTopic],
+        mode: 'earliest',
+        sessionTimeout: 10000,
+        heartbeatInterval: 500,
+    });
+
+    // Example DLQ msg:
```
SylvainSenechal: wanted to add this because I suspect most people wouldn't know whats exactly in a DLQ message, here I literaly consoled log a dlq message to see whats inside and document it directly, as most people won't bother setting up a failed test in codespace to see it. I think it's nice to have.

Alternative, u tell me if u wanna have this : I can setup a proper typescript structure that mirrors this

tests/functional/ctst/steps/azureArchive.ts
```diff
@@ -308,22 +309,58 @@ Then('blob for object {string} must be rehydrated',
...
+                    const ignored = this.getSaved<string[]>('ignoreDLQOps') || [];
+                    this.addToSaved('ignoreDLQOps', [...ignored, 'restore']);
+                    return; // success
+                }
+            }
+            throw new Error(
+                `Restore failure for "${objName}" not found in DLQ within ${timeoutMs / 1000}s`,
+            );
+        } finally {
+            await dlqStream.close();
```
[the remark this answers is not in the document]
SylvainSenechal: alright dud

tests/functional/ctst/steps/azureArchive.ts
```diff
@@ -308,22 +309,59 @@ Then('blob for object {string} must be rehydrated',
...
+        const restoreTimeoutMs = parseInt(this.parameters.SorbetdRestoreTimeout) * 1000;
+        const timeoutMs = restoreTimeoutMs + 30_000 + 1_000;
+        const deadline = Date.now() + timeoutMs;
+
+        try {
+            for await (const msg of dlqStream) {
+                if (Date.now() > deadline) {
+                    break;
+                }
+                const value = JSON.parse(msg.value);
```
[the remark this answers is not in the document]
SylvainSenechal: addressed in other comment

tests/functional/ctst/steps/utils/utils.ts
```diff
@@ -484,45 +488,123 @@ async function verifyObjectLocation(this: Zenko, objectName: string,
...
+    //     'op' => 'archive',
+    //     'js' => 'pending',
+    //   },
+    //   topic: '55f00ebb-...-e2e-azure-archive',
+    //   partition: 0, offset: 18n, timestamp: 1777909037564n,
+    // }
+    const watchDLQ = async () => {
+        const ignoreDLQOps = this.getSaved<string[]>('ignoreDLQOps') || [];
+        for await (const msg of dlqStream) {
+            const value = JSON.parse(msg.value);
```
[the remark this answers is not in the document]
SylvainSenechal: What do you think ? Should i implement a typescript interface for a dql message to parse properly ?

tests/functional/ctst/features/azureArchive.feature
francoisferrand: > Renamed the step because the phrasing was very confusing, made it look like the obj fails to rehydrate because of what we did in the previous step line 240, when in reality we fail the rehydration voluntarily from within that step

The object _really_ fails to rehydrate, and this is really the scenario we want to test.

The failure is voluntary indeed, but withholding the notification is the way we found to introduce the failure. But the functionality we test (and scenario) is really "object fails to rehydrate" (and all related code path in forwarder and backbeat), while the implementation relies on dedicated code in mock sorbet server to withhold the notification.

The titles of the step are expected to be customer-focused, i.e. they provide an abstraction from the implementation detail (this is the main goal/benefit of cucumber): so we should stick to product-level notions, and there is no notion of "withholding notification" in the product.

→ either stick to `object "xxx" fails to rehydrate` / `fails to restore`, or maybe make DLQ more explicit `object "xxx" failed to restore and ends up in DLQ` but keep implementation details abstracted away

tests/functional/ctst/steps/utils/utils.ts
```diff
@@ -484,45 +488,126 @@ async function verifyObjectLocation(this: Zenko, objectName: string,
...
+    //     'messageId' => 'ced79367-e9c1-4a0b-8c9e-8b10207b4e90',
+    //     'jobId' => 'put-http://.../.../zerobyte-obj-1-',
+    //     'op' => 'archive',
+    //     'js' => 'pending',
+    //   },
+    //   topic: '55f00ebb-...-e2e-azure-archive',
+    //   partition: 0, offset: 18n, timestamp: 1777909037564n,
+    // }
+    const watchDLQ = async () => {
+        const ignoreDLQOps = this.getSaved<string[]>('ignoreDLQOps') || [];
```
francoisferrand: no need for using `saved` : either we can deduce it from existing parameters (objectTransitionStatus), or should be a parameter of the function.

(would be more logical to specify what we are intested in, instead of what we ignore : basically we are interested in just a single DLQ message, i.e. the archive/restore/delete failure of the current object)

tests/functional/ctst/steps/utils/utils.ts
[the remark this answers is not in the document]
francoisferrand: a typescript interface would not protect you here : claude reported an issue with invalid JSON... Either way you need to handle that case in some way...

tests/functional/ctst/steps/utils/utils.ts
francoisferrand: comment is not manageable.

Either
- just parse as you did (and let people look up format of messages in Sorbet)
- or add a typescript if you want to formalize the format
- or use sorbetctl directly if you don't want to care (but may be more complicated to skip unrelated events)

tests/functional/ctst/steps/utils/utils.ts
```diff
@@ -484,45 +488,126 @@ async function verifyObjectLocation(this: Zenko, objectName: string,
...
         }
-        if (objectTransitionStatus == 'restored') {
-            const isRestored = !!parsed.result?.Restore &&
-                parsed.result.Restore.includes('ongoing-request="false", expiry-date=');
-            conditionOk = conditionOk && isRestored;
-        } else if (objectTransitionStatus == 'cold') {
-            conditionOk = conditionOk && !parsed.result?.Restore;
+    };
+
+    const groupId = `zenko-e2e-dlq-watcher-${Date.now()}`;
```
francoisferrand: can't we have multiple calls with the same timestamp?
should we add some randomness or test-case name/id?

tests/functional/ctst/steps/utils/utils.ts
```diff
@@ -484,45 +488,126 @@ async function verifyObjectLocation(this: Zenko, objectName: string,
...
+    //   headers: Map {
+    //     'messageId' => 'ced79367-e9c1-4a0b-8c9e-8b10207b4e90',
+    //     'jobId' => 'put-http://.../.../zerobyte-obj-1-',
+    //     'op' => 'archive',
+    //     'js' => 'pending',
+    //   },
+    //   topic: '55f00ebb-...-e2e-azure-archive',
+    //   partition: 0, offset: 18n, timestamp: 1777909037564n,
+    // }
+    const watchDLQ = async () => {
```
francoisferrand: should probably introduce a helper function, the same code (or almost) is duplicated

francoisferrand: that function could also handle the setup to kafka consumer, so it would work with a single call `watchDLQ(...)` (possibly with a callback to 'checking' condition, if needed for flexibility)

tests/functional/ctst/steps/utils/utils.ts
```diff
@@ -484,45 +488,126 @@ async function verifyObjectLocation(this: Zenko, objectName: string,
...
+                    + ` (op: ${value.op},`
+                    + ` reason: ${value.reason})`,
+                );
+            }
         }
-        await Utils.sleep(1000);
+    };
+
+    const poll = pollTransition();
+    const watch = watchDLQ();
```
francoisferrand: nit: maybe we should try delete the consumer group (from kafka!) at the end of the function, to avoid accumulating stuff?

(but not sure if really relevant, should not really have any impact if we have a few extra consumer group ; especially on this low-traffic topic...)

tests/functional/ctst/steps/utils/utils.ts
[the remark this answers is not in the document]
SylvainSenechal: I'm adding both

tests/functional/ctst/steps/utils/utils.ts
SylvainSenechal: yeah implementing the first 2 points

tests/functional/ctst/steps/utils/utils.ts
SylvainSenechal: I think its nit but yes I replaced it with our randomString utility

tests/functional/ctst/steps/utils/utils.ts
SylvainSenechal: Yeah I changed it to work the other way : instead of having an ignore list, now we will only be looking for specific dlq op message, no need to store anything because the op we are interested in is available in the function that calls watchDLQ

tests/functional/ctst/steps/utils/kafka.ts
```diff
@@ -1,4 +1,103 @@
...
+                options.logger.warn('Failed to parse DLQ message, skipping', {
+                    raw: msg, error: String(err),
+                });
+                continue;
+            }
+            if (value.op !== options.op) {
+                continue;
+            }
+            const objectMatch = value.objectKey?.includes(options.objectName);
+            const bucketMatch = value.bucketName?.includes(options.bucketName);
```
benzekrimaha: why includes? we know the bucket and object key we are waiting for, no point accepting partial matches. This could match obj-1 with obj-10 or another stale event.

SylvainSenechal: Yeah let's do strict comparison, I was worried maybe the dlq format would be a bit loose but probably not

tests/functional/ctst/features/azureArchive.feature
```diff
@@ -238,8 +238,7 @@ Feature: Azure Archive
...
+    Then the restoration of object "retry-obj-1" is failed and ends up in DLQ
```
francoisferrand: nit:
```suggestion
    Then restoration of object "retry-obj-1" failed and ends up in DLQ
```

tests/functional/ctst/steps/utils/kafka.ts
```diff
@@ -1,4 +1,112 @@
...
+    });
+
+    const deadline = Date.now() + options.timeoutMs;
+
+    const timeoutHandle = setTimeout(() => {
+        dlqStream.close().catch(() => {});
+    }, options.timeoutMs);
+
+    const promise = (async () => {
+        try {
```
francoisferrand, line 39: can't we create the dlqstreams and consumer in this function?

this was we could close everything in the `finally` block as well, return a single promise, and be perfectly sure there is no leak (either the `promise` function is called: it will create stuff and _finally_ close them; or it is not called and nothing happens)

francoisferrand: would also avoid the 2 "nested" promises : we may end up a (non-async) function returning the async callback.... or even just a single async function: just "inlining" the `const promise = (async () => {` (i.e. removing line 67 and 101, plus calling close() inside the finally block, should do the trick?)

tests/functional/ctst/steps/utils/kafka.ts
```diff
@@ -1,4 +1,112 @@
...
+            for await (const msg of dlqStream) {
+                if (Date.now() > deadline) {
```
francoisferrand: seems we have 2 mechanisms to handle the timeout : both this test here, and setTimeout above.
→ should we simply drop this one?

(if we receive a message after the timeout but still process it, it should actually still "count" as getting a message from DLQ : if were expecting it, it would avoid a flaky due to timeout being slightly too short; and otherwise it means the transition has failed -although slowly- but still failed and this should be reported)

tests/functional/ctst/steps/utils/utils.ts
francoisferrand: not so much anymore, mostly about indent...
(if we extract the function, the whole block will be _moved_ : which technically increases the diff, but may look better if the viewer handles _move_ detection)

DarkIsDude (approved):

tests/functional/ctst/steps/utils/kafka.ts
```diff
@@ -1,4 +1,112 @@
...
+    const groupId = `zenko-e2e-dlq-${Utils.randomString()}`;
+    const consumer = new Consumer({
+        clientId: groupId,
+        groupId,
+        bootstrapBrokers: [options.kafkaHosts],
+        deserializers: stringDeserializers,
+    });
+    const dlqStream = await consumer.consume({
+        topics: [options.topic],
+        mode: 'earliest',
```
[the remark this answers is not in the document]
SylvainSenechal: yeyeye spotted, but didn't add a Date.now() because depending on when the Date.now() is called, there are still possible race condition if date.now() is called AFTER the dlq is already created 🧐 

So added some exclusion list by dlq requestID instead

tests/functional/ctst/steps/utils/utils.ts
SylvainSenechal: yeah I found the admin api to close the consumer

tests/functional/ctst/steps/utils/kafka.ts
SylvainSenechal: I did some refactoring around it, no nested promise and timeout is cleaner

tests/functional/ctst/steps/utils/kafka.ts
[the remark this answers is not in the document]
francoisferrand: - even we date, since we share a location, we may get unrelated events. So date is indeed not a protection here: at most an optimization (to avoid skipping entries), but probably not significant (not many messages overall)
- do we have "real" (not bugs) cases where we transition the same object multiple times? If not, should not need exclusion
- I find _exclusion_ a bit difficult to work with, it introduces coupling with "all other tests" (which may conflict) ; is there maybe another approach, i.e. with a way to better identify precisely the message we want? (maybe matching versionId as well?)

tests/functional/ctst/features/azureArchive.feature
francoisferrand: not applied?

tests/functional/ctst/steps/utils/kafka.ts
```diff
@@ -1,4 +1,107 @@
...
+    } finally {
+        clearTimeout(timeoutHandle);
+        await dlqStream.close().catch(() => {});
+        await consumer.close();
+        const admin = new Admin({ clientId: groupId, bootstrapBrokers: [options.kafkaHosts] });
+        try {
+            await admin.deleteGroups({ groups: [groupId] });
+        } finally {
+            await admin.close();
+        }
```
francoisferrand: - don't we want to ignore deleteGroups failures, to avoid flakiness?
- can admin.close() fail as well, should we ignore the failure?

```suggestion
try {
   await admin.deleteGroups({ groups: [groupId] });
} catch() {}
try {
   await admin.close();
} catch() {}
```

tests/functional/ctst/steps/utils/utils.ts
```diff
@@ -484,45 +488,84 @@ async function verifyObjectLocation(this: Zenko, objectName: string,
...
+            + ` (op: ${dlqMsg.op},`
+            + ` reason: ${dlqMsg.reason})`,
+        );
+    });
+
+    const poll = pollTransition();
+    try {
+        await Promise.race([poll, watch]);
+    } finally {
+        raceSettled = true;
```
francoisferrand: nit: could we not use abortController as well in `pollTransition`, instead of an extra bool flag ?

tests/functional/ctst/steps/utils/utils.ts
```diff
@@ -484,45 +488,84 @@ async function verifyObjectLocation(this: Zenko, objectName: string,
...
+        abortController.abort();
+        poll.catch(() => {});
+        watch.catch(() => {});
```
francoisferrand: should we wait ?

```suggestion
        await poll.catch(() => {});
        await watch.catch(() => {});
```

tests/functional/ctst/steps/utils/utils.ts
```diff
@@ -484,45 +488,84 @@ async function verifyObjectLocation(this: Zenko, objectName: string,
...
-            conditionOk = conditionOk && isRestored;
-        } else if (objectTransitionStatus == 'cold') {
-            conditionOk = conditionOk && !parsed.result?.Restore;
-        }
-        await Utils.sleep(1000);
+    };
+
+    const op = objectTransitionStatus === 'restored' ? 'restore' : 'archive';
+    const seenDLQ = this.getSaved<Set<string>>('seenDLQMessages') || new Set<string>();
+    this.addToSaved('seenDLQMessages', seenDLQ);
```
francoisferrand: - the `seenDLQMessages` "saved" is never set to anything else (in 2 places, we just set it to its previous value or an empty set) : so we will always have an empty set here, and this is actually useless?
- not sure what this approach would give : even if we already "consumed" a message, it may be that we skipped it ; and since that kind of "conflict" is created by concurrent tests, we cannot guarantee the other one would not be created after we compute the existing messages to ignore...

francoisferrand (approved):
LGTM, except the seamingly useless/risky `seenDLQMessages` and associated `excludedMessages`

tests/functional/ctst/features/azureArchive.feature
SylvainSenechal: applied now, I forgot to change the step name here

tests/functional/ctst/steps/utils/kafka.ts
SylvainSenechal: after a few reruns on practice it doesn't seem to fail 🤔

tests/functional/ctst/steps/utils/utils.ts
SylvainSenechal: - line 85 in kafka.ts, the seen list is updated : 
options.seenMessages.add(value.requestId);

- There is no problem with concurrent tests, as they each have their own world, this is addressing issues within the same scenario where different step need to watch the dlq but we don't want the second watcher to look at messages seen in the first watcher. 

I agree the pattern in general is not super nice but that's more an issue with cucumber in general being stateful and not explicit with function parameters

PR #2404, opened 2026-05-08
Improve kafka cleaner test

SylvainSenechal:
Issue: ZENKO-5270

First work on increasing execution speed and reducing flakiness : 

- Rework Kafka cleaner that had a bug (previously was taking exactly 36 minutes all the time when it can work in seconds since anyways its cleaning progressively during the whole test run every minute) + add a new quick step that verify a deletion metrics incremented by kafka cleaner 
image

### DIFF
--- a/tests/functional/ctst/common/common.ts
+++ b/tests/functional/ctst/common/common.ts
@@ -311,50 +311,50 @@ Then('i {string} be able to add user metadata to object {string}',
 Then('kafka consumed messages should not take too much place on disk', { timeout: -1 },
     async function (this: Zenko) {
         const kfkcIntervalSeconds = parseGoDuration(this.parameters.KafkaCleanerInterval);
-        const checkInterval = kfkcIntervalSeconds * (1000 + 5000);
-
-        const timeoutID = setTimeout(() => {
-            assert.fail('Kafka cleaner did not clean the topics within the expected time');
-        }, checkInterval * 10); // Timeout after 10 Kafka cleaner intervals
+        const checkInterval = kfkcIntervalSeconds * 1000;
+        const deadline = Date.now() + checkInterval * 3;
 
         const kafkaAdmin = new Admin({
             clientId: 'ctst-kafka-cleaner-check',
             bootstrapBrokers: [this.parameters.KafkaHosts],
         });
-        
+
         try {
-            const ignoredTopics = ['dead-letter'];
+            const excludedTopics = ['dead-letter', 'backbeat-metrics'];
+            const prefix = `${this.parameters.InstanceID}.`;
             const allTopics = await kafkaAdmin.listTopics();
             const topics: string[] = allTopics
-                .filter(t => (t.includes(this.parameters.InstanceID) &&
-                    !ignoredTopics.some(e => t.includes(e))));
+                .filter(t => t.startsWith(prefix) &&
+                    !excludedTopics.some(excluded => t.includes(excluded)));
 
-            const previousOffsets = await getTopicsOffsets(topics, kafkaAdmin);
+            const previousOffsets = new Map<string, { low: string, high: string }[]>();
+            for (const { topic, partitions } of await getTopicsOffsets(topics, kafkaAdmin)) {
+                previousOffsets.set(topic, partitions);
+            }
 
-            while (topics.length > 0) {
-                // Checking topics offsets before kafkacleaner passes to be sure kafkacleaner works
-                // This function can be improved by consuming messages and
-                // verify that the timestamp is not older than last kafkacleaner run
-                // Instead of waiting for a fixed amount of time,
-                // we could also check for metrics to see last kafkacleaner run
+            let remainingTopics = topics;
 
-                // 3 seconds added to be sure kafkacleaner had time to process
-                await Utils.sleep(checkInterval);
+            while (true) {
+                assert(Date.now() < deadline,
+                    'Kafka cleaner did not clean topics within the expected time. ' +
+                    `Remaining: ${remainingTopics.join(', ')}`);
 
-                const newOffsets = await getTopicsOffsets(topics, kafkaAdmin);
+                const newOffsets = await getTopicsOffsets(remainingTopics, kafkaAdmin);
+                const nextTopics: string[] = [];
 
-                for (let i = 0; i < topics.length; i++) {
-                    this.logger.debug('Checking topic', { topic: topics[i] });
+                for (let i = 0; i < remainingTopics.length; i++) {
+                    const topic = remainingTopics[i];
+                    const previousTopicOffsets = previousOffsets.get(topic)!;
+                    this.logger.debug('Checking topic', { topic });
                     let topicCleaned = false;
                     for (let j = 0; j < newOffsets[i].partitions.length; j++) {
                         const newOffsetPartition = newOffsets[i].partitions[j];
-                        const oldOffsetPartition = previousOffsets[i].partitions[j];
+                        const oldOffsetPartition = previousTopicOffsets[j];
 
                         if (!oldOffsetPartition) {
                             continue;
                         }
 
-                        // Ensure we're accessing the correct partition details
                         const lowOffsetIncreased = parseInt(newOffsetPartition.low) >
                             parseInt(oldOffsetPartition.low);
                         // We tolerate one message not being cleaned, as it can be due to the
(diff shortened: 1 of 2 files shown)


tests/functional/ctst/common/common.ts
```diff
@@ -311,84 +311,111 @@ Then('i {string} be able to add user metadata to object {string}',
 Then('kafka consumed messages should not take too much place on disk', { timeout: -1 },
     async function (this: Zenko) {
         const kfkcIntervalSeconds = parseGoDuration(this.parameters.KafkaCleanerInterval);
-        const checkInterval = kfkcIntervalSeconds * (1000 + 5000);
-
-        const timeoutID = setTimeout(() => {
-            assert.fail('Kafka cleaner did not clean the topics within the expected time');
-        }, checkInterval * 10); // Timeout after 10 Kafka cleaner intervals
+        const checkInterval = kfkcIntervalSeconds * 1000 + 5000;
```
SylvainSenechal: In practice, this test is basically running instantly : 
Kafka cleaner runs every minute from the beginning of the tests, and in total there are ~1000 kakfa messages created, so cleanup is instant

tests/functional/ctst/common/common.ts
```diff
@@ -311,84 +311,111 @@ Then('i {string} be able to add user metadata to object {string}',
...
     });
 
+Then('kafka cleaner has successfully cleaned topic messages', async function (this: Zenko) {
+    const prometheusQuery = encodeURIComponent(
+        'kafka_cleaner_kafka_requests{type="DeleteRecords",status="success"}'
+    );
+    const metricsResponse = await fetch(
+        `${this.parameters.PrometheusEndpoint}/api/v1/query?query=${prometheusQuery}`
+    );
+    const metricsData = await metricsResponse.json() as {
```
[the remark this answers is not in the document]
SylvainSenechal: kay

delthas (approved):

tests/functional/ctst/common/common.ts
```diff
@@ -311,84 +311,116 @@ Then('i {string} be able to add user metadata to object {string}',
 Then('kafka consumed messages should not take too much place on disk', { timeout: -1 },
     async function (this: Zenko) {
         const kfkcIntervalSeconds = parseGoDuration(this.parameters.KafkaCleanerInterval);
-        const checkInterval = kfkcIntervalSeconds * (1000 + 5000);
-
-        const timeoutID = setTimeout(() => {
-            assert.fail('Kafka cleaner did not clean the topics within the expected time');
-        }, checkInterval * 10); // Timeout after 10 Kafka cleaner intervals
+        const checkInterval = kfkcIntervalSeconds * 1000 + 5000;
```
francoisferrand: what is this computation/value?

we first get the "real" kfkc interval (as configured), then converst to millisecond ; but then
- add 5 seconds which I don't understand → this would make the intervals completely desynchronized, I don't see how this would be what we want
- later multiply by 3 → so `checkInterval` does not represent the interval between checks (in the test), but really the total time

tests/functional/ctst/common/common.ts
```diff
@@ -311,84 +311,116 @@ Then('i {string} be able to add user metadata to object {string}',
...
 
         const kafkaAdmin = new Admin({
             clientId: 'ctst-kafka-cleaner-check',
             bootstrapBrokers: [this.parameters.KafkaHosts],
         });
-        
+
         try {
-            const ignoredTopics = ['dead-letter'];
+            // List from Zenko Operator
```
francoisferrand: having a *back-list* was done on purpose : we want to check that *ALL* topics are cleaned, even if we add a topic (to the product) and forget to setup cleanup...

some topics may be missing from blacklist (or just be added to cleanup in ZKOP, or removed if not used), but the logic should not be converted to white-list.

tests/functional/ctst/common/common.ts
```diff
@@ -311,84 +311,116 @@ Then('i {string} be able to add user metadata to object {string}',
 Then('kafka consumed messages should not take too much place on disk', { timeout: -1 },
     async function (this: Zenko) {
         const kfkcIntervalSeconds = parseGoDuration(this.parameters.KafkaCleanerInterval);
-        const checkInterval = kfkcIntervalSeconds * (1000 + 5000);
-
-        const timeoutID = setTimeout(() => {
-            assert.fail('Kafka cleaner did not clean the topics within the expected time');
-        }, checkInterval * 10); // Timeout after 10 Kafka cleaner intervals
+        const checkInterval = kfkcIntervalSeconds * 1000 + 5000;
+        const deadline = Date.now() + checkInterval * 3;
```
francoisferrand, line 315: 10 times was added (some times back) to reduce flakiness : are we confident it works with 3 attempts now?

tests/functional/ctst/common/common.ts
```diff
@@ -311,84 +311,116 @@ Then('i {string} be able to add user metadata to object {string}',
...
+                        remaining.delete(topic);
                     }
                 }
-            }
 
-            // If a topic remains in this array, it means it has not been cleaned
-            assert(topics.length === 0, `Topics ${topics.join(', ')} still have not been cleaned`);
+                if (remaining.size > 0) {
+                    await Utils.sleep(5000);
+                }
```
francoisferrand: the condition is redundant with the `while` condition. Seems it should be like this instead:

```
while(true) {
   [...]

   if (remaining.size == 0) {
      break;
   }
   
   await Utils.sleep(5000);
}
```

(though I wonder if the initial delay was not on purpose, to make sure we wait for KFKC to run after the "last" tests ? is there no race condition there?)

tests/functional/ctst/common/common.ts
```diff
@@ -311,84 +311,116 @@ Then('i {string} be able to add user metadata to object {string}',
...
 
-                // 3 seconds added to be sure kafkacleaner had time to process
-                await Utils.sleep(checkInterval);
+            while (remaining.size > 0) {
+                assert(Date.now() < deadline,
+                    'Kafka cleaner did not clean topics within the expected time. ' +
+                    `Remaining: ${[...remaining].join(', ')}`);
 
-                const newOffsets = await getTopicsOffsets(topics, kafkaAdmin);
+                const remainingTopics = [...remaining];
```
francoisferrand: not sure this is relevant for this test, but this is a somewhat costly operation.

instead of unpacking Set to List, and later removing from Set : can't we just build "next" list at iteration with all the topics which did not pass?

tests/functional/ctst/common/common.ts
```diff
@@ -311,84 +311,116 @@ Then('i {string} be able to add user metadata to object {string}',
...
-                    this.logger.debug('Checking topic', { topic: topics[i] });
+                for (let i = 0; i < remainingTopics.length; i++) {
+                    const topic = remainingTopics[i];
+                    const oldPartitions = previousOffsets.get(topic)!;
+                    this.logger.debug('Checking topic', { topic });
                     let topicCleaned = false;
                     for (let j = 0; j < newOffsets[i].partitions.length; j++) {
                         const newOffsetPartition = newOffsets[i].partitions[j];
-                        const oldOffsetPartition = previousOffsets[i].partitions[j];
+                        const oldOffsetPartition = oldPartitions[j];
```
francoisferrand: except for the timer/sleep handling, I don't really see how this code improves or even behaves differently from what was there already?
is it that there could a mixup between topics, and we would compare wrong new/old offset because kafka would not return the list in the same order?

tests/functional/ctst/common/common.ts
```diff
@@ -311,84 +311,116 @@ Then('i {string} be able to add user metadata to object {string}',
...
+                    await Utils.sleep(5000);
+                }
+            }
         } finally {
-            clearTimeout(timeoutID);
             await kafkaAdmin.close();
         }
     });
 
+Then('kafka cleaner has successfully cleaned topic messages', async function (this: Zenko) {
```
francoisferrand:
```suggestion
Then('kafka cleaner has successfully deleted some messages', async function (this: Zenko) {
```

this does not check the topic is cleaned, only that some messages were successfully deleted...

tests/functional/ctst/common/common.ts
```diff
@@ -311,84 +311,116 @@ Then('i {string} be able to add user metadata to object {string}',
...
+    assert(metricsResponse.ok,
+        `Prometheus query failed with status ${metricsResponse.status}: ${metricsResponse.statusText}`);
+    const metricsData = await metricsResponse.json() as {
+        status: string;
+        data: { result: { value: [number, string] }[] };
+    };
+    assert.strictEqual(metricsData.status, 'success',
+        `Prometheus returned non-success status: ${metricsData.status}`);
+    const deleteRecordsCount = metricsData.data.result.length > 0
+        ? parseFloat(metricsData.data.result[0].value[1])
```
francoisferrand, line 406: is there really a single series?

- either we should verify > 0 on every series
- or tweak the prometheus query to aggregate all queries (e.g. `sum(kafka_cleaner_kafka_requests{...})`), to ensure there is really just one

tests/functional/ctst/common/common.ts
SylvainSenechal: It's not super nice, the real general idea is "wait the kafka cleaner interval multiplied a few times"
which should be similar but without the + 5000

tests/functional/ctst/common/common.ts
SylvainSenechal: I see, im gonna use the black list again, I think we have a problem with the backbeat-metrics topic that needs to be added to zenko operator then, or black listed here

tests/functional/ctst/common/common.ts
SylvainSenechal: Gonna rerun a bit, but yeah the test seems to run within 1 minute, which is in line with cleaning ~1200 kafka messages, and having the kafka cleaner run every minute from the beginning of the test. If it takes more than a few second, it probably meanas there is an issue with the kafka cleaner 🤔 

But I think previously it was just taking a long time because of the messed up timeout + something wrong about the splice indexing in the old code

tests/functional/ctst/common/common.ts
SylvainSenechal: yeah let's use sum

tests/functional/ctst/common/common.ts
SylvainSenechal: I don't think the initial delay matters, if its not cleaned on the first iteration it will be on the following one 🤔

tests/functional/ctst/common/common.ts
SylvainSenechal: yes we can do it

tests/functional/ctst/common/common.ts
SylvainSenechal: this is a refactoring to hopefully make it clearer

But what needs to be seen is the bug in the old code with the splice on topic : 
- previousOffsets was defined before the while loop
- there was slice on the topic list in the while loop
- newOffset were calculated from the sliced topic list in the while loop 

After the first slice : accessing old/newOffset by index was comparing wrong topics

francoisferrand: 👍 That comment is the underlying bug causing this PR : this should be stated in PR and commit message...

You did not refactor or tweak the delay to make it clearer : you identified this bug with splice causing mismatch of index comparison, and consequently fixed it (with a refactor maybe, but not so much given the explanation...)

tests/functional/ctst/common/common.ts
```diff
@@ -311,84 +311,105 @@ Then('i {string} be able to add user metadata to object {string}',
...
 
-                const newOffsets = await getTopicsOffsets(topics, kafkaAdmin);
+                const newOffsets = await getTopicsOffsets(remainingTopics, kafkaAdmin);
+                const nextTopics: string[] = [];
 
-                for (let i = 0; i < topics.length; i++) {
-                    this.logger.debug('Checking topic', { topic: topics[i] });
+                for (let i = 0; i < remainingTopics.length; i++) {
+                    const topic = remainingTopics[i];
+                    const oldPartitions = previousOffsets.get(topic)!;
```
francoisferrand:
```suggestion
                    const previousTopicOffsets = previousOffsets.get(topic)!;
```

tests/functional/ctst/common/common.ts
```diff
@@ -311,84 +311,105 @@ Then('i {string} be able to add user metadata to object {string}',
 Then('kafka consumed messages should not take too much place on disk', { timeout: -1 },
```
francoisferrand, line 311: not new, but this name is weird. This step does not check disk, only that all consumed messages have been removed....

```suggestion
Then('kafka consumed messages should be removed', { timeout: -1 },
```

francoisferrand (approved):

PR #2408, opened 2026-05-11
ctst: skip trailing 1s sleep when poll condition is met

delthas:
## Summary

Several CTST polling helpers wait one extra interval after the success condition is met, before the `while` loop re-evaluates and exits. Every successful call pays that sleep as a flat tax. This PR guards the sleep with the success flag so we exit immediately.

## What changes

- `verifyObjectLocation` — used by the `object "*" should be "*" and have storage class "*"` step (~80 call sites across feature files)
- `isObjectRehydrated` — used by "blob for object X must be rehydrated"
- `waitForZenkoToStabilize` / `waitForDataServicesToStabilize` — deploy stabilization

## Why it saves time

Each call previously sleeps 1s after success → 1s/call × hundreds of invocations per CI run, all pure overhead. With these guards, every successful poll returns as soon as the matching response arrives. No behavior change on the unhappy path: the sleep still paces retries when the condition isn't yet met.

Out of scope: `addTransitionWorkflow` also sleeps 5s on success, but its comment indicates that's intentional propagation wait — left as-is.

Issue: ZENKO-5271

### DIFF
--- a/tests/functional/ctst/steps/utils/kubernetes.ts
+++ b/tests/functional/ctst/steps/utils/kubernetes.ts
@@ -260,7 +260,6 @@ export async function waitForZenkoToStabilize(
     // So, this function will first wait till we detect a reconciliation
     // (deploymentInProgress = true), and then wait for the status to be available
     const startTime = Date.now();
-    let status = false;
     let deploymentFailure: ZenkoStatusValue = {
         lastTransitionTime: '',
         message: '',
@@ -287,7 +286,7 @@ export async function waitForZenkoToStabilize(
     world.logger.debug('Waiting for Zenko to stabilize');
     const zenkoClient = createKubeCustomObjectClient(world);
 
-    while (!status && Date.now() - startTime < timeout) {
+    while (Date.now() - startTime < timeout) {
         const zenkoCR = await zenkoClient.getNamespacedCustomObject({
             group: 'zenko.io',
             version: 'v1alpha2',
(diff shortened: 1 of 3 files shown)


tests/functional/ctst/steps/utils/kubernetes.ts
```diff
@@ -343,6 +343,7 @@ export async function waitForZenkoToStabilize(
             available.status === 'True'
         ) {
             status = true;
+            break;
```
SylvainSenechal: i guess can be return directly ?

tests/functional/ctst/steps/utils/kubernetes.ts
```diff
@@ -414,7 +415,9 @@ export async function waitForDataServicesToStabilize(world: Zenko, timeout = 15
             }
         }
 
-        await Utils.sleep(1000);
+        if (!allRunning) {
```
SylvainSenechal: its nit but maybe some refactor instead of this, probably for the other changes too : 

if all running : return True
sleep

} end loop

throw error

SylvainSenechal (approved):

tests/functional/ctst/steps/utils/kubernetes.ts
delthas: Good call — done in 027554fc, plus applied the same early-return pattern across the four functions per your hint on the other thread.

tests/functional/ctst/steps/utils/kubernetes.ts
delthas: Refactored all four poll loops to the early-return shape (`if (success) return; sleep; } throw`) in 027554fc. Dropped the function-scope flag variable in each. Type-check clean.

maeldonn (approved):

francoisferrand:
@delthas There is a conflict with #2399 changes on waterflow branch, please update/fix the waterflow branch...

PR #2409, opened 2026-05-11
ctst: parallelize cleanupAccount IAM operations

delthas:
## Summary

`cleanupAccount` (used by the `@BP-ASSUME_ROLE_USER_CROSS_ACCOUNT` `After` hook) makes ~12 sequential IAM round-trips per invocation. The hook fires 1260 times per full e2e run for ~1066s cumulative wall-clock — one of the largest single contributors to suite runtime.

This change parallelizes the work within each phase using `Promise.all`:

- **Listing** users/groups/roles in one round-trip (3 → 1).
- **Detach phase**: all per-entity policy detaches across users, groups and roles fan out concurrently (outer + inner).
- **Delete-policies phase**: all `deletePolicy` calls in parallel.
- **Delete-entities phase**: all role/group/user deletions in parallel.

Phase ordering is preserved (detach → delete-policies → delete-entities → `deleteAccount`). `Promise.all` matches the existing throw-on-first-error semantics — a failed op rejects the phase, the outer `try/catch` logs a warning and the hook completes normally.

Typical per-scenario fan-out is small (1-2 users, 1 role, 1 policy), so backend concurrency stays modest.

## Expected impact

Cleanup path drops from ~600ms of serial round-trips per call to ~150ms. For the ~80% of invocations that need cleanup, ballpark **~8 min saved per full e2e run**.

Issue: ZENKO-5273

### DIFF
--- a/tests/functional/ctst/common/utils.ts
+++ b/tests/functional/ctst/common/utils.ts
@@ -242,83 +242,82 @@ export async function cleanupAccount(world: Zenko, accountName: string) {
     try {
         Identity.useIdentity(IdentityEnum.ACCOUNT, accountName);
 
-        // List and detach policies for each user
-        const allUsers = await listAllEntities<User>(IAM.listUsers, 'Users');
-        for (const user of allUsers) {
-            const allUserPolicies = await listAttachedPolicies<AttachedPolicy>(
-                params => IAM.listAttachedUserPolicies({ userName: user.UserName, ...params }),
-            );
-            for (const policy of allUserPolicies) {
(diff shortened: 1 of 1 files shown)


SylvainSenechal (approved):
Ok, just before merging, take a look at the ci, if you see weird error "tag hr> thing", it usually happens when there is too much pressure on the server, I've had this when switching from 4 to 8 workers, and I wonder if it may happens here if there is too much stuff being parallelized

maeldonn (approved):

PR #2410, opened 2026-05-11
Add SMT plugin for Kafka Connect S3 object key partitioning

delthas:
## Summary

- New Maven module under `solution/kafka-connect/smt/` providing `com.scality.kafka.connect.transforms.TransformObjectKey`, a Kafka Connect Single Message Transform that rewrites the oplog topic message key from the MongoDB change-stream's `documentKey._id`, stripping the Scality master/version encoding produced by arsenal.
- `solution/kafka-connect/Dockerfile` gains an `smt-build` multi-stage that compiles the SMT JAR and copies it into `/usr/local/share/kafka/plugins/` next to the existing MongoDB source connector.
- The Kafka version targeted by the SMT is threaded from `solution/deps.yaml` (`kafka.tag: 2.13-3.1.2`) via `kafka_build_vars.sh` → `env.kafka_version` → build-arg `KAFKA_VERSION` → `mvn -Dkafka.version=`, so the SMT compiles against the same Kafka release the kafka image ships.

## Context

Backbeat [BB-768](https://scality.atlassian.net/browse/BB-768) fixes a long-standing partitioning bug on the oplog Kafka topic: today the message key is derived from `fullDocument.value.key`, but `fullDocument` is `null` on `update` and `delete` events (since BB-355 removed `change.stream.full.document=updateLookup` to avoid per-update round-trips). The result is that updates/deletes for an S3 object collapse onto a single partition while inserts spread across the topic — breaking per-object ordering across op types.

The Backbeat-side fix is to project `documentKey._id` (always populated, regardless of op type) and let an SMT strip the master/version encoding so master + all versions of the same logical S3 object land on the same partition. That SMT must live in Connect's plugin path before the Backbeat PR can ship — which is what this PR delivers.

Validated end-to-end on a local stack (mongo 6.0 replica set + kafka 7.1.7 + Connect with this image): inserts, updates, and deletes for 4 distinct S3 objects each routed to a single partition per object, matching the BB-768 acceptance criterion.

Issue: ZENKO-5274

[BB-768]: https://scality.atlassian.net/browse/BB-768?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


delthas:
Replaced by https://github.com/scality/backbeat/pull/2744

PR #2416, opened 2026-05-15
Improve cache usage

francoisferrand:
The GHA cache was sitting near the 10 GB cap with several hundred MB of identical content duplicated across every PR ref. PRs were starting cold on every build, causing extra delay and possibly flakiness in setup, especially when stability of upstream registries/services is degraded.
This was caused by issues in our setup, but also 2 specific constraints of GHA cache:
- Workflow can only read caches written on its own ref or on the repo's **default branch**. Sibling PR refs are invisible to each other, even with the same cache key.
- Entries are removed from cache when idle for more than 7 days, or when cache is full (least-recently-used dropped first)

To fix this rework setup and docker builds in multiple ways:
- **Pull images from the registry instead of rebuilding.**. Images (kafka, mongo...) are content-addressed with a version and (git) tree hash, so the registry image _is_ the primary cache. Build becomes a no-op when there are no change, otherwise fallback to GHA cache as before.
- **Cache warmer to seed shared (default-branch) cache.**. `cache-warmer` dispatched from `postmerge.yaml` runs on the **default branch** so its writes are visible to every PR. Runs subset of builds/setups, to warm the GHA-scope fallbacks (kafka/mongo), per-tool caches (kustomize/tilt), and yarn cache.
- **Install tools instead of caching deb blobs.**. Dropped the 700 MB `~/deb_cache` and 391 MB `/usr/local/bin` blob caches in favour of approved setup actions and per-tool version-keyed `actions/cache` entries.
- **Build doc in a container**. Documentation requires lots of dependencies, which bloat the cache in every PR even though they never change. Build in a container -which was actually already there- and use the registry itself as the cache in this case.
- **Cleanup aggressively.** New `cache-cleanup` workflow prunes on `delete` (PR branch, Bert-E's `w/*` and `q/*` branches) and on `pull_request: closed` (the review workflow's Bun cache under `refs/pull/<N>/merge`).

Issue: ZENKO-5278


delthas (approved):

benzekrimaha (approved):

maeldonn (approved):
LGTM

PR #2420, opened 2026-05-21
mongodb: vendor chart via git subtree from bitnami/charts

delthas:
## Summary

Switch `solution-base/mongodb/Makefile` from `helm fetch` against `charts.bitnami.com` to **vendoring the chart sources from the `bitnami/charts` GitHub repository via `git subtree`**, mirroring the pattern that ZENKO-5110 already established for vendored Bitnami *images* in `solution-base/images/Makefile`. All local modifications previously expressed as `.patch` files now live as ordinary git commits on top of the subtree merge, ordered chronologically by when each was originally introduced.

## Target branch

Targets **`development/2.15`** (not 2.14). This change is a prerequisite for integrating the upstream `pgrep -x` fix ([ZENKO-5276](https://scality.atlassian.net/browse/ZENKO-5276)) and reverting the installer-side workaround (ARTESCA-17516) — neither of which is needed on 2.14. 2.14 keeps the working `helm fetch` flow; there's no reason to churn it. 2.15 already carries the ZENKO-5110 image-vendoring work, so the chart-vendoring switch is a natural fit alongside it.

## Why

Bitnami stopped publishing the `mongodb-sharded` chart OCI artifact in August 2025 ([bitnami/charts#35164](https://github.com/bitnami/charts/issues/35164)). The last published version is `9.4.12` and any future commit merged into `bitnami/charts` `main` — including our upstream `pgrep -x` liveness probe fix in [bitnami/charts#36488](https://github.com/bitnami/charts/pull/36488) — will never reach the Helm registry. The chart source on GitHub is still updated and tagged for every release, so it is now the authoritative place to pull from.

Using `git subtree` (over a `curl | tar` snapshot) gives us:

- Real three-way merge conflicts when bumping — `git subtree merge --squash` surfaces upstream-vs-local conflicts directly, instead of silent `patch` overwrites or stale `.patch` files.
- One mental model with `solution-base/images/Makefile`, which already uses the same pattern for the Bitnami container-image vendoring under ZENKO-5110.
- Subsequent bumps (e.g. ZENKO-5276 → `mongodb-sharded/9.4.14`) become a single `CHART_VERSION` change + `make vendor-sync`.

## What changes

- **`solution-base/mongodb/Makefile`** — full rewrite. New targets `create-remote`, `fetch-remote`, `update-vendor-branch`, `vendor-sync`, `deps`. The single maintainer command is `make -C solution-base/mongodb vendor-sync` (~4 minutes — fetches `bitnami/charts` history, runs `git subtree split`, merges as squash, re-resolves `common`).
- **`solution-base/mongodb/charts/mongodb-sharded/`** — re-introduced as a `git subtree --squash` of upstream tag `mongodb-sharded/9.4.4`. The `common` library chart is resolved via `helm dependency build` and extracted into `charts/common/` so `solution-base/build.sh`'s `helm template` finds it as a sub-chart.
- **11 patch files migrated to 11 individual commits**, ordered chronologically (oldest local-tweak first → newest at top), so future `git log` archaeology under the prefix reflects the actual evolution of these tweaks.
- **`solution-base/mongodb/patches/` deleted entirely.**
- **`solution-base/mongodb/how_to_upgrade.md`** rewritten for the new chart flow, preserving the existing Image Build section.
- **`.github/workflows/end2end.yaml`** — `check-mongo-patches` job removed (with patches gone there's nothing to verify; conflicts now surface as ordinary git merge conflicts during `make vendor-sync`).
- **`.github/copilot-instructions.md`** — Chart upgrade-path row updated to describe the subtree flow.

## What does NOT change

- Pinned chart version stays at **9.4.4** — pure infrastructure switch. [ZENKO-5276](https://scality.atlassian.net/browse/ZENKO-5276) becomes a `CHART_VERSION` bump + `make vendor-sync` whose merge brings in upstream's `pgrep -x` liveness probe fix.
- `solution-base/deps.yaml` is untouched — image references are independent and addressed under ZENKO-5110.
- `solution-base/build.sh` is untouched — it expects `charts/common/` as an extracted directory, which the new Makefile preserves.

## Verification

- `helm template` against this branch produces manifests equivalent to the previous `helm fetch` 9.4.4 + `make patch` flow (modulo helm-generated random Secret values per invocation).
- `make -C solution-base/mongodb vendor-sync` is idempotent against `mongodb-sharded/9.4.4`: re-running on a clean tree produces no diff.
- The vendored chart tree is byte-identical to the prior `helm fetch` + patch result.

## History

Earlier revisions implemented the same goal via `curl | tar` of the GitHub release tarball. After discussion with @francoisferrand, switched to `git subtree` to match the `solution-base/images/` precedent and to make future upstream merges produce real merge conflicts rather than silent overwrites. The subtree variant was first explored on a parallel draft PR (#2423, now closed). Subsequently re-targeted from `development/2.14` to `development/2.15` per review feedback (this is a prerequisite for downstream work that only lands on 2.15+).

Issue: ZENKO-5281

[ZENKO-5276]: https://scality.atlassian.net/browse/ZENKO-5276?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ZENKO-5276]: https://scality.atlassian.net/browse/ZENKO-5276?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


solution-base/mongodb/Makefile
```diff
@@ -12,11 +15,13 @@ HELM=helm
...
-		--version ${CHART_MONGO_SHARDED_VERSION} \
-		--untar \
-		--untardir ${CHART_DIR}
+	@mkdir -p ${CHART_DIR}/mongodb-sharded
+	@curl -sfL ${CHART_URL} \
+		| tar xz --strip-components=3 -C ${CHART_DIR}/mongodb-sharded \
+			--wildcards --exclude='*/bitnami/mongodb-sharded/CHANGELOG.md' \
+			'*/bitnami/mongodb-sharded/*'
+	@${HELM} dependency build ${CHART_DIR}/mongodb-sharded
+	@cd ${CHART_DIR}/mongodb-sharded/charts && for tgz in *.tgz; do tar xzf "$$tgz" && rm "$$tgz"; done
```
francoisferrand: instead of curl / replace and manually handling diff, should we take the chance to move to `git subtree`, as used for the mongodb Dockerfile? This allows to actually import upstream (one commit), make our changes, and later just merge upstream changes (as in "git merge") ?

francoisferrand: why are there still tar files in the chart: did we not untag during download, line 19 ? are there TAR files in the tar ?

delthas: It's the `common` chart from Bitnami. In the packaged chart it's already in extracted form, but here we run `helm dependency build` so it's downloaded as an archive. To get the same output I'm extracting it here (in a generic manner, not just limited to `common`.

delthas:
Moved from vendoring from git download, to actual git subtree.

solution-base/mongodb/Makefile
delthas: Good suggestion — done. Force-pushed the branch with a `git subtree`-based implementation that mirrors `solution-base/images/Makefile`:

- `make -C solution-base/mongodb vendor-sync` is the single maintainer command (fetch → `git subtree split` → `git subtree merge --squash` → re-resolve `common`).
- The 11 patches under `solution-base/mongodb/patches/` are migrated to 11 individual commits on top of the subtree merge, ordered chronologically by when each was originally introduced. Future `git log` archaeology under the prefix now reflects how these tweaks evolved.
- `solution-base/mongodb/patches/` deleted, `make patch` removed, `check-mongo-patches` CI job removed — upstream-vs-local conflicts now surface as ordinary git merge conflicts during `make vendor-sync`.
- ZENKO-5276 (bump to 9.4.14) becomes a single `CHART_VERSION` change + `make vendor-sync`.

PR body updated to describe the new approach. Cost worth knowing: `make vendor-sync` is a ~4-minute coffee break (subtree split walks 36 599 upstream commits).

delthas:
First commits are setting up the subtree, then each commit is literally moving the patch to a commit, then last commit is cleaning up patches.

solution-base/mongodb/Makefile
```diff
@@ -1,23 +1,49 @@
...
+
+# Maintainer-only: bump CHART_VERSION above, then `make vendor-sync` to merge
+# upstream into our prefix. Resolves merge conflicts against our local commits
+# the same way any git merge does.
+vendor-sync: update-vendor-branch
+	git subtree merge --prefix=$(REPO_CHART_PREFIX) $(VENDOR_BRANCH) --squash
+	@$(MAKE) deps
+
+# Resolve and extract the `bitnami/common` library chart so it appears as a
+# subdirectory (required by `solution-base/build.sh`'s `helm template`).
```
francoisferrand, line 45: should this be done here (and vendor the result) ? Or should it be done at build time, updating solution-base/build.sh ?

solution-base/mongodb/Makefile
```diff
@@ -1,23 +1,49 @@
...
-PATCH_DIR:="${ROOT_DIR}/patches"
-PATCH_FILES:=$(shell ls -d ${PATCH_DIR}/*)
-HELM=helm
+# Repo-relative path of the vendored chart (needed by `git subtree`).
+REPO_CHART_PREFIX  := solution-base/mongodb/charts/$(CHART)
+# Path relative to this Makefile's directory (where `helm` runs).
+LOCAL_CHART_PREFIX := charts/$(CHART)
 
-.PHONY: fetch patch
+VENDOR_BRANCH := vendor/$(CHART_REF)
```
francoisferrand: could this conflict with the vendor branch used for the image?
maybe best to use `chart` in the branch name, to avoid any confusion?

```suggestion
VENDOR_BRANCH := vendor/charts/$(CHART_REF)
```

solution-base/mongodb/how_to_upgrade.md
```diff
@@ -2,49 +2,65 @@
...
-MongoDB 8.0.10 to 9.3.6):
+The `mongodb-sharded` Helm chart is vendored from the `bitnami/charts`
+GitHub repository via `git subtree`. Bitnami stopped publishing
+`mongodb-sharded` OCI chart artifacts in August 2025 (see
+[bitnami/charts#35164](https://github.com/bitnami/charts/issues/35164));
+the chart source on GitHub is still updated and tagged for every
+release, and is now the authoritative source for new versions.
+
+Our local modifications live as **ordinary git commits on top of the
+subtree merge**, not as separate `.patch` files. Future upgrades merge
```
francoisferrand: > not as separate `.patch` files

This is relevant on the PR (it changes the way we do things), but we don't really care in the future: so best remove it, to avoid introducing any hint that we could have .patch files (human or AI...)

```suggestions
Our local modifications live as **ordinary git commits on top of the
subtree merge**. Future upgrades merge
```

solution-base/mongodb/Makefile
francoisferrand: (would create more change though, maybe in a separate PR if relevant ?)

.github/copilot-instructions.md
```diff
@@ -26,7 +26,7 @@ When reviewing a PR, analyze changes against the following. Post inline comments
...
 | **Dockerfiles** (`solution/kafka/`, `solution/kafka-connect/`...) | Base images pinned by tag or digest, no secrets baked in, no unnecessary `COPY . .`, reasonable layer count, user is non-root where possible. |
 | **Helm charts & K8s manifests** (`solution-base/mongodb/charts/`, `monitoring/`) | Resource requests/limits set, label selectors match, no hard-coded namespaces, `securityContext` present, ServiceAccount scoping minimal. Any breaking chart-value renames documented in an upgrade note. |
-| **Chart upgrade path** (`solution-base/mongodb/patches/`, `how_to_upgrade.md`) | If chart version or MongoDB version changes, upgrade notes are updated and patches still apply cleanly. |
+| **Chart upgrade path** (`solution-base/mongodb/Makefile`, `how_to_upgrade.md`) | If chart version or MongoDB version changes, upgrade notes are updated and the `git subtree merge --squash` produced by `make vendor-sync` resolves cleanly against our local commits. |
```
francoisferrand: is this something that can be checked by claude code review?

- I don't think it has the workspace or permissions to do the required git commands or modify workspace...
- if we can run commands to do the check, best to do it with some script in end2end.yaml

(I think claude code review may be able to check some things like "changes to folder xxx are in a "merge commit" or subtree merge are squashde ; but checking that vendoring has been properly built/commited will probably not work, and best be addressed with a gha job)

francoisferrand (commented):
does this waterflow cleanly ?

if there is any change, is it really worth doing on 2.14 instead of just keeping "what works", and targeting 2.15 ?

(AFAIU, this is mostly a pre-requisite to integrate upstream fix and revert our installer patch ; but this is not needed on 2.14 ?)

solution-base/mongodb/Makefile
delthas: Done — renamed to `vendor/charts/$(CHART_REF)` so chart vendor branches (`vendor/charts/mongodb-sharded/9.4.4`) can't be confused with the image vendor branches (`vendor/mongodb-sharded/8.0/debian-12`) from `solution-base/images/`.

solution-base/mongodb/how_to_upgrade.md
delthas: Done — dropped the `.patch` reference; the doc now just states that local modifications live as ordinary git commits on top of the subtree merge.

solution-base/mongodb/Makefile
delthas: Keeping it vendored (committed) for now to keep this PR's diff small. Can create a separate PR later.

.github/copilot-instructions.md
delthas: Alright, dropped it.

delthas:
> does this waterflow cleanly? if there is any change, is it really worth doing on 2.14 instead of just keeping "what works", and targeting 2.15?

Agreed — re-targeted this PR to `development/2.15`.

It also fits 2.15 better: 2.15 already carries the ZENKO-5110 image-vendoring work, so the chart-vendoring switch sits naturally alongside it. I rebuilt the branch on 2.15 (fresh subtree + cherry-picked the local-modification commits) and merged `how_to_upgrade.md` so the existing Image Build section is preserved next to the new chart-vendoring steps. Chart tree is byte-identical to the prior `helm fetch` + patch result.

solution-base/mongodb/Makefile
```diff
@@ -1,23 +1,49 @@
...
-PATCH_DIR:="${ROOT_DIR}/patches"
-PATCH_FILES:=$(shell ls -d ${PATCH_DIR}/*)
-HELM=helm
+# Repo-relative path of the vendored chart (needed by `git subtree`).
+REPO_CHART_PREFIX  := solution-base/mongodb/charts/$(CHART)
+# Path relative to this Makefile's directory (where `helm` runs).
+LOCAL_CHART_PREFIX := charts/$(CHART)
 
-.PHONY: fetch patch
+VENDOR_BRANCH := vendor/charts/$(CHART_REF)
```
francoisferrand: this is actually a _tag_ : so we will always use a new vendored branch (re-split etc), is that not an issue?

don't we need a "stable" vendor branch (even if it pulls different tags from upstream) to merge? Or  will the repeated merging from different tags work just as well?

(having a stable branch would also avoid the split?)

francoisferrand (commented):
LGTM, just wondering about the "tag" vendor branch

solution-base/mongodb/Makefile
delthas: We don't need a stable branch because the subtree internally uses commit refs, not branch names / tag names.

Still, moved to a stable branch name.

francoisferrand (approved):

maeldonn (approved):

PR #2421, opened 2026-05-21
Added script to verify all tests scenario are associated with a version tag

SylvainSenechal:
Issue: ZENKO-5282

Some cleanup of a hook defined in cli-testing.

- Adding a check to enforce that all scenarios are tagged with a version, this will be nice for documentation
- Wanted to migrate the hook that conditionally runs tests depending on their versions tag + the version of the zenko but realize at the moment it really is useless : 
    - For hotfix on older branch : the older branch has an older VERSION file AND older test files, so newer tests simply don't exist there, so TARGET_VERSION doesn't matter.
    - It could be useful if we wanted to add tests for newer versions and not run them yet (like writting a new test for 2.16.0 when we are still in 2.15.0, so that it would not be run while the version remains 2.15.0). But in practice this is not gonna be used : when we add a test, we always make sure that they work, as tests and features are pretty much shipped together

So the hook is simply removed for now.

### DIFF
--- a/.github/workflows/end2end.yaml
+++ b/.github/workflows/end2end.yaml
@@ -356,6 +356,9 @@ jobs:
       - name: Check for unused step definitions
         working-directory: tests/functional
         run: yarn unused-steps
+      - name: Check that all scenarios have a version tag
+        working-directory: tests/functional
+        run: yarn ensure-version-tags
 
   end2end-pra:
     needs: [build-kafka, build-mongodb]
--- a/tests/functional/ctst/common/hooks.ts
+++ b/tests/functional/ctst/common/hooks.ts
@@ -23,7 +23,6 @@ import { startDLQConsumer, stopDLQConsumer } from 'steps/utils/kafka';
 
 import 'cli-testing/hooks/KeycloakSetup';
 import 'cli-testing/hooks/Logger';
-import 'cli-testing/hooks/versionTags';
 
 // HTTPS should not cause any error for CTST
 process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0';
--- a/tests/functional/ctst/scripts/ensureVersionTags.ts
+++ b/tests/functional/ctst/scripts/ensureVersionTags.ts
@@ -1,0 +1,75 @@
+/**
+ * Verifies that every Scenario in the ctst feature files
+ * has a semver version tag (e.g. @2.6.0), either on the Feature,
+ * an enclosing Rule, or directly on the scenario.
+ *
+ * Exits with code 1 and lists offenders if any are missing.
+ */
+
+/* eslint-disable no-console */
+
+import { Parser, AstBuilder, GherkinClassicTokenMatcher } from '@cucumber/gherkin';
+import { IdGenerator } from '@cucumber/messages';
+import * as fs from 'fs';
+import * as path from 'path';
+
+const VERSION_TAG_REGEX = /^@\d+\.\d+\.\d+$/;
+const FEATURES_DIR = path.resolve(__dirname, '../features');
+
+function collectFeatureFiles(dir: string): string[] {
+    const files: string[] = [];
+    for (const entry of fs.readdirSync(dir, { withFileTypes: true })) {
+        const full = path.join(dir, entry.name);
+        if (entry.isDirectory()) {
+            files.push(...collectFeatureFiles(full));
+        } else if (entry.name.endsWith('.feature')) {
+            files.push(full);
+        }
+    }
+    return files;
+}
+
+let totalScenarios = 0;
+const offenders: string[] = [];
+const parser = new Parser(new AstBuilder(IdGenerator.incrementing()), new GherkinClassicTokenMatcher());
+
+for (const file of collectFeatureFiles(FEATURES_DIR)) {
+    const rel = path.relative(FEATURES_DIR, file);
+    const ast = parser.parse(fs.readFileSync(file, 'utf8'));
+    const feature = ast.feature;
+    if (!feature) { continue; }
+
+    const featureHasVersion = feature.tags.some(t => VERSION_TAG_REGEX.test(t.name));
+
+    for (const child of feature.children) {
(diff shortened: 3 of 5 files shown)


tests/functional/ctst/scripts/ensureVersionTags.ts
SylvainSenechal, line 1: Actually no tests were breaking this rule, but wanna enforce it, it's important to know "this test was shipped with a feature on version 2.15.1"

Tested it in codespace by removing tags from some scenarios : 

Image

tests/functional/ctst/common/hooks.ts
```diff
@@ -23,7 +23,6 @@ import { startDLQConsumer, stopDLQConsumer } from 'steps/utils/kafka';
 
 import 'cli-testing/hooks/KeycloakSetup';
 import 'cli-testing/hooks/Logger';
-import 'cli-testing/hooks/versionTags';
```
SylvainSenechal, line 26: wrote why im doing this in the top comment of the pr

tests/functional/package.json
```diff
@@ -7,7 +7,7 @@
     "node": ">=24"
   },
   "dependencies": {
-    "@cucumber/cucumber": "^12.7.0",
+    "@cucumber/cucumber": "^12.9.0",
```
SylvainSenechal, line 10: kinda useless because of the caret but still did it as im adding the dev dependencies

tests/functional/ctst/scripts/ensureVersionTags.ts
DarkIsDude, line 1: We usually put them in .github/scripts ? Also we should have tests for it 🙏

tests/functional/ctst/scripts/ensureVersionTags.ts
```diff
@@ -0,0 +1,75 @@
...
+                if (!ruleChild.scenario) { continue; } // skip Background inside Rule
+                totalScenarios++;
+                const hasVersion = featureHasVersion || ruleHasVersion
+                    || ruleChild.scenario.tags.some(t => VERSION_TAG_REGEX.test(t.name));
+                if (!hasVersion) {
+                    offenders.push(`  ${rel}:${ruleChild.scenario.location.line} — "${ruleChild.scenario.name}"`);
+                }
+            }
+        }
+        // Background nodes are intentionally ignored
```
DarkIsDude, line 64: You have twice the same comment https://github.com/scality/Zenko/pull/2421/changes#diff-637785e467dea30f38c5f19d551eefd93a7feffae2da263d44066c7d8c8c2adcR55. It's maybe better to add the why, than just describing the code (if relevant).

tests/functional/ctst/scripts/ensureVersionTags.ts
```diff
@@ -0,0 +1,75 @@
...
+    if (!feature) { continue; }
+
+    const featureHasVersion = feature.tags.some(t => VERSION_TAG_REGEX.test(t.name));
+
+    for (const child of feature.children) {
+        if (child.scenario) {
+            totalScenarios++;
+            const hasVersion = featureHasVersion || child.scenario.tags.some(t => VERSION_TAG_REGEX.test(t.name));
+            if (!hasVersion) {
+                offenders.push(`  ${rel}:${child.scenario.location.line} — "${child.scenario.name}"`);
```
DarkIsDude, line 49: Does this `  ` is needed ?

tests/functional/ctst/scripts/ensureVersionTags.ts
```diff
@@ -0,0 +1,75 @@
...
+            }
+        } else if (child.rule) {
+            // Version tag can be inherited from Feature or Rule
```
DarkIsDude, line 52:
```suggestion
```

tests/functional/ctst/scripts/ensureVersionTags.ts
```diff
@@ -0,0 +1,75 @@
...
+            const ruleHasVersion = child.rule.tags.some(t => VERSION_TAG_REGEX.test(t.name));
+            for (const ruleChild of child.rule.children) {
+                if (!ruleChild.scenario) { continue; } // skip Background inside Rule
+                totalScenarios++;
+                const hasVersion = featureHasVersion || ruleHasVersion
+                    || ruleChild.scenario.tags.some(t => VERSION_TAG_REGEX.test(t.name));
+                if (!hasVersion) {
+                    offenders.push(`  ${rel}:${ruleChild.scenario.location.line} — "${ruleChild.scenario.name}"`);
```
DarkIsDude, line 60: You can create a new function to do that instead of duplicating the logic (`ruleChild` vs `child`).

DarkIsDude (changes requested):
`Adding a check to enforce that all scenarios are tagged with a version, this will be nice for documentation` for which documentation ?

https://github.com/scality/Zenko/pull/2421/changes#r3281331971 reading that. Why ? We have git history for that if needed ?

.github/workflows/end2end.yaml
```diff
@@ -356,6 +356,9 @@ jobs:
       - name: Check for unused step definitions
         working-directory: tests/functional
         run: yarn unused-steps
+      - name: Check that all scenarios have a version tag
```
francoisferrand, line 359: why do we want that?

this "tagging" was introduced when CTST was used in Artesca, and embedded the tests : so there was one version of CTST and tests targeting multiple zenko versions.

the tests are now managed in `git` -so the information is really available via git blame or others-, I don't see the benefit of manually adding the tag on every test?

SylvainSenechal: Well, yes, there is another version of this pr where I just remove all tags + the useless before hook 
I kinda liked having directly visible tags attached to the scenarios as its faster to read than commits but you're right

SylvainSenechal:
too complex for what it brings => 
In this pr I will simply remove the useless hook : https://github.com/scality/Zenko/pull/2422
And no need for a script to enforce tagged versioned scenarios

PR #2422, opened 2026-05-21
move keycloak hook seeder from cli-testing to zenko

SylvainSenechal:
Issue: ZENKO-5283

- fix codespace node version : 22 => 24
- bump cucumber to 13.0 (new worker implementation, and new summary design (we will see how it looks. edit: looks alright, very similar to what we have before, but potentially more customizable if we want later))
- removed useless @tag hook
- And the bigger diff : moved keycloak seeder hook in this repo

Big diff but its a 98% copy of whats in cli-testing

Once this is merged, together with the other pr adding the new method to load parameters directly from within the test, we will be able to do some refactoring around some hard coded values and weird patterns like this one : 
image
to replace it with a proper config

Reduce dependency on cli-testing


tests/functional/ctst/common/hooksKeycloakSetup.ts
SylvainSenechal, line 1: source : https://github.com/scality/cli-testing/blob/development/1.3/hooks/KeycloakSetup.ts

tests/functional/ctst/steps/utils/keycloak.ts
SylvainSenechal, line 1: source : https://github.com/scality/cli-testing/blob/development/1.3/utils/Keycloak.ts

tests/functional/ctst/steps/utils/keycloak.ts
```diff
@@ -0,0 +1,407 @@
...
+        logger?: Werelogs.RequestLogger,
+    ) {
+        this.logger = logger || new Werelogs.Logger('Keycloak').newRequestLogger();
+        this.realm = parameters.realm;
+        this.host = parameters.host;
+        this.username = parameters.username;
+        this.password = parameters.password;
+        this.clientId = parameters.clientId;
+
+        logger?.debug('Keycloak parameters', {
```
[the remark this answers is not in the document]
SylvainSenechal: yeah we are leaking 'password' 🤷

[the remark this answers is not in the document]
francoisferrand: don't do it : it's password today, but you don't know what it may be as time passes and needs evolve...

tests/functional/ctst/common/hooks.ts
```diff
@@ -21,9 +21,7 @@ import {
 import { createKubeCustomObjectClient, waitForZenkoToStabilize } from 'steps/utils/kubernetes';
 import { startDLQConsumer, stopDLQConsumer } from 'steps/utils/kafka';
 
-import 'cli-testing/hooks/KeycloakSetup';
 import 'cli-testing/hooks/Logger';
-import 'cli-testing/hooks/versionTags';
```
francoisferrand, line 26: I'm kind of confused: what were these used for?
or were they already useless?

tests/functional/ctst/steps/utils/keycloak.ts
```diff
@@ -0,0 +1,407 @@
...
+        try {
+            const users = await this.kcAdminClient.users.find({
+                realm: this.realm,
+                username,
+                exact: true,
+            });
+
+            if (!users || users.length === 0) {
+                throw new Error(`User not found: ${username}`);
+            }
```
delthas, line 229: Could we perhaps throw on users.length !== 1 just in case (we've had had duplicate name issues in the past)

delthas (approved):

tests/functional/ctst/common/hooks.ts
SylvainSenechal: - Keycloak : for setting up keycloak users for the tests (kept and moved to this repo)
- version tag : to skip tests based on their @version tag, and the value of process.env.TARGET_VERSION : removed, we talked about it a bit in the other pr that I closed but it's really not useful now

maeldonn (approved):

PR #2431, opened 2026-06-05
feat(mongodb): alert on createIndexes command failures

delthas:
## Summary
- Add `MongoDbCreateIndexesFailed` Prometheus alert (`monitoring/mongodb/alerts.yaml`) on `mongodb_ss_metrics_commands_createIndexes_failed` from mongodb_exporter (Percona, exposes `serverStatus.metrics.commands.*`).
- Fires when any MongoDB pod records a failed `createIndexes` command in a 10-minute window. Catches index-build refusals such as MongoDB 7.1+ `indexBuildMinAvailableDiskSpaceMB` rejections, and any other `createIndexes` failures regardless of which client triggered them.
- Companion fixture added to `alerts.test.yaml`.

A narrower Backbeat-side alert on the same scenario lives in the lifecycle conductor under [BB-784](https://scality.atlassian.net/browse/BB-784) (Backbeat PR #2753). Part of the [Empty bucket feature epic](https://scality.atlassian.net/browse/ARTESCA-16740).

The compaction/fragmentation alert (also part of ZENKO-5285's description) is deferred — needs more thought on the right signal.

Issue: ZENKO-5285

[BB-784]: https://scality.atlassian.net/browse/BB-784?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2433, opened 2026-06-08
feat(mongodb): alert when compaction is needed

delthas:
Follow-up to [ZENKO-5285](https://scality.atlassian.net/browse/ZENKO-5285) (PR #2431), which bundled two alerts in its description and only shipped the first (createIndexes-failed). This adds the second — a fragmentation / compaction-needed signal — that @DarkIsDude explicitly asked be filed as a follow-up.

## Two commits, deliberately split

1. **`mongodb: enable dbstatsfreestorage collector in exporter`** — one-line `values.yaml` change adding `--collector.dbstatsfreestorage` to `metrics.extraArgs`. The chart's exporter no longer uses `--collect-all` (dropped in 8414833 for ZENKO-5281), so each sub-collector is opted into individually in `extraArgs` (already lists `dbstats`, `diagnosticdata`, `replicasetstatus`, `topmetrics`). Without `dbstatsfreestorage`, `freeStorageSize` / `totalFreeStorageSize` only appear as part of the clunky per-host `mongodb_dbstats_raw_<host>_*` expansion instead of as top-level `mongodb_dbstats_*` series.

2. **`mongodb: alert when compaction is needed`** — the alert proper.

## The alert

```promql
mongodb_dbstats_totalFreeStorageSize > 0.3 * mongodb_dbstats_fsTotalSize
for: 1h
```

Per-(pod, database) granularity. Severity warning. Threshold exposed as the `compactionFreeStorageRatioThreshold` x-input (default 0.3).

Expressing the threshold as a fraction of **filesystem capacity** (not of the DB's own storage) lets it scale across cluster sizes:

- 100 GB FS → fires around 30 GB reclaimable
- 1 TB FS → fires around 300 GB reclaimable
- 10 TB FS → fires around 3 TB reclaimable

## Heuristic notes

A previous draft combined three legs (FS-pressure + fragmentation ratio + absolute floor). @francoisferrand's review pointed out that the fragmentation-ratio leg added noise without much signal and the FS-pressure leg could mask real waste on under-used disks. The single fs-scaled threshold above folds both concerns into one condition.

Per-DB granularity means the alert tells you `pod={{ "{{ $labels.pod }}" }}` and `database={{ "{{ $labels.database }}" }}` but does **not** identify the specific collection. To find that, an operator runs `collStats` on the alerting DB. The exporter has a `--collector.collstats` flag that could expose per-collection visibility, but at Artesca-scale cardinality (thousands of buckets per DB) that's expensive — deferred.

## Safety against missing / zero values

- Missing metric for a `(pod, database)` tuple → PromQL produces empty → no alert.
- `totalFreeStorageSize = 0` (no fragmentation) → `0 > 0.3 * fsTotalSize` is always false.
- `fsTotalSize = 0` would degenerate (`0.3 * 0 = 0`, then any positive `freeStorage` would fire), but it's physically impossible for an attached PVC.

## Follow-up (not in this PR)

Matching `compactionFreeStorageRatioThreshold` config option in ZKOP, to expose the knob to ops. Not filed as a ticket yet; will track when this lands.

## Related

- [ZENKO-5293](https://scality.atlassian.net/browse/ZENKO-5293) — this ticket
- [ZENKO-5285](https://scality.atlassian.net/browse/ZENKO-5285) — parent alert ticket; createIndexes-failed half shipped in #2431
- [BB-784](https://scality.atlassian.net/browse/BB-784) — Backbeat-side LifecycleIndexCreationFailed (scality/backbeat#2753)
- [BB-783](https://scality.atlassian.net/browse/BB-783) — putBucketIndexesFailed metric (scality/backbeat#2751)
- Parent epic ARTESCA-16740 — Empty bucket feature

Issue: ZENKO-5293

[ZENKO-5285]: https://scality.atlassian.net/browse/ZENKO-5285?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ZENKO-5293]: https://scality.atlassian.net/browse/ZENKO-5293?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/monitoring/mongodb/alerts.yaml
+++ b/monitoring/mongodb/alerts.yaml
@@ -27,6 +27,9 @@ x-inputs:
   - name: replicationLagOplogSizeThreshold
     type: config
     value: 0.5
+  - name: reclaimableStorageWarningThreshold
+    type: config
+    value: 0.3
 
 groups:
 - name: MongoDb
@@ -279,6 +282,17 @@ groups:
     annotations:
       description: "MongoDB pod `{{ $labels.pod }}` reported failed createIndexes commands in the last 5 minutes."
       summary: MongoDB index creation failure
+
+  - alert: MongoDbCompactionNeeded
+    expr: |
+      mongodb_dbstats_totalFreeStorageSize{namespace="${namespace}",pod=~"${service}.*"}
+        > ${reclaimableStorageWarningThreshold} * mongodb_dbstats_fsTotalSize{namespace="${namespace}",pod=~"${service}.*"}
+    for: 1h
+    labels:
+      severity: warning
+    annotations:
+      description: "MongoDB pod `{{ $labels.pod }}` database `{{ $labels.database }}` has accumulated reclaimable storage exceeding ${reclaimableStorageWarningThreshold} of the filesystem capacity. Consider running compaction to recover disk space."
+      summary: MongoDB compaction needed
 
   - alert: MongoDbRSNotSynced
     expr: |
--- a/monitoring/mongodb/alerts.test.yaml
+++ b/monitoring/mongodb/alerts.test.yaml
@@ -698,3 +698,37 @@ tests:
       - alertname: MongoDbCreateIndexesFailed
         eval_time: 16m
         exp_alerts: []
+
+  - name: MongoDbCompactionNeeded
+    interval: 5m
+    input_series:
+      # Compaction-needed DB: 4 GB reclaimable on a 10 GB filesystem (40% > 30%)
+      - series: mongodb_dbstats_fsTotalSize{namespace="zenko", pod="data-db-mongodb-sharded-shard0-data-0", database="needs-compaction"}
+        values: 1e10x13
+      - series: mongodb_dbstats_totalFreeStorageSize{namespace="zenko", pod="data-db-mongodb-sharded-shard0-data-0", database="needs-compaction"}
+        values: 4e9x13
+      # Healthy DB on same pod: 1 GB reclaimable on the same 10 GB filesystem (10% < 30%)
+      - series: mongodb_dbstats_fsTotalSize{namespace="zenko", pod="data-db-mongodb-sharded-shard0-data-0", database="healthy-db"}
+        values: 1e10x13
+      - series: mongodb_dbstats_totalFreeStorageSize{namespace="zenko", pod="data-db-mongodb-sharded-shard0-data-0", database="healthy-db"}
+        values: 1e9x13
+    alert_rule_test:
+      # for: 1h not yet satisfied
+      - alertname: MongoDbCompactionNeeded
+        eval_time: 10m
+        exp_alerts: []
+      - alertname: MongoDbCompactionNeeded
+        eval_time: 55m
+        exp_alerts: []
+      # for: 1h satisfied; only the needs-compaction DB fires
+      - alertname: MongoDbCompactionNeeded
+        eval_time: 65m
+        exp_alerts:
+          - exp_labels:
+              severity: warning
+              namespace: zenko
+              pod: data-db-mongodb-sharded-shard0-data-0
+              database: needs-compaction
+            exp_annotations:
+              description: "MongoDB pod `data-db-mongodb-sharded-shard0-data-0` database `needs-compaction` has accumulated reclaimable storage exceeding 0.3 of the filesystem capacity. Consider running compaction to recover disk space."
+              summary: MongoDB compaction needed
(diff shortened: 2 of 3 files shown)


delthas:
@francoisferrand Heuristics are up for discussion -- I chose a mix of:
- Disk is being significantly used (to avoid saving 10GB on a disk that has 10TB free space)
- There is a signficiant number of GBs to save (to avoid alarms on tiny collections)
- There is a significant ratio of data to save (to avoid running compact just to save 10GB on a 10TB DB)

monitoring/mongodb/alerts.yaml
```diff
@@ -27,6 +27,15 @@ x-inputs:
...
     value: 0.5
+  - name: compactionDiskUsageThreshold
+    type: config
+    value: 0.7
+  - name: compactionFreeStorageRatioThreshold
+    type: config
+    value: 0.3
+  - name: compactionFreeStorageAbsoluteThreshold
+    type: config
+    value: 10 * 1024 * 1024 * 1024 # 10 GiB, in bytes
```
francoisferrand: are we confident about these `levels` ?
- good defaults *everywhere*
- alert when really needed
- not too noisy

these alerts need to be "low touch", and fire only when an action is required...

monitoring/mongodb/alerts.yaml
```diff
@@ -270,6 +279,27 @@ groups:
...
+    expr: |
+      (
+        mongodb_dbstats_fsUsedSize{namespace="${namespace}",pod=~"${service}.*"}
+          / mongodb_dbstats_fsTotalSize{namespace="${namespace}",pod=~"${service}.*"}
+      ) > ${compactionDiskUsageThreshold}
+      and
+      (
+        mongodb_dbstats_totalFreeStorageSize{namespace="${namespace}",pod=~"${service}.*"}
+          / mongodb_dbstats_totalSize{namespace="${namespace}",pod=~"${service}.*"}
+      ) > ${compactionFreeStorageRatioThreshold}
```
francoisferrand: not sure about this rule : percentage of disk (FS) and percentage of reclaimable (I gather totalFreeStorageSize is what we can reclaim?) could fire in many case, but often when not relevant ?

e.g. we are interested when the reclaimable amount is significant vs the available mongodb (free/total) size I guess?
* i.e. your last condition, or possibly a variable where the threshold is a % of fs(free/total)size?
* first condition seems fine, to avoid alerts when there is lot of space (though that is debatable, not sure it is really useful or we want to delay in that case)
* I don't see at all what the benefit of % of reclaimable space is ?

monitoring/mongodb/alerts.yaml
francoisferrand: (note: since we add config, shoud add matching options in ZKOP afterwards, to allow setting them...)

francoisferrand (changes requested):
- not sure about thresholds/computation: are these relevant default?
- not sure we should merge in 2.15, or target 2.16 and get time to "preview" this and avoid shipping alerts which would lead to support calls...

monitoring/mongodb/alerts.yaml
delthas: Simplified the heuristic in 2790943 — only one knob left now (`compactionFreeStorageRatioThreshold`, default 0.3). Threshold scales naturally with cluster size (30% of fsTotalSize), so it should be reasonable across small dev clusters and large prod ones without tuning.

ZKOP: created https://scality.atlassian.net/browse/ZKOP-558

monitoring/mongodb/alerts.yaml
delthas: Good critique — dropped both the FS-usage leg and the % reclaimable leg. The rule is now just `totalFreeStorageSize > 0.3 * fsTotalSize` for 1h.

That folds your suggestion (absolute threshold as a fraction of fsTotalSize) into the only condition, which I agree is the most meaningful one: how much disk capacity is sitting as reclaimable WT free space. Both other legs were noise relative to that core signal.

delthas:
Keeping on 2.15 for now as 2.16 doesnt exist.

delthas:
Intentionally set 30% by default (conservative), up for discussion

francoisferrand (approved):

monitoring/mongodb/alerts.yaml
```diff
@@ -27,6 +27,9 @@ x-inputs:
   - name: replicationLagOplogSizeThreshold
     type: config
     value: 0.5
+  - name: compactionFreeStorageRatioThreshold
```
francoisferrand: maybe `reclaimableStorageWarningThreshold` is mot appriate & consistent with others?

delthas: Done

benzekrimaha (approved):

PR #2438, opened 2026-06-12
Move test resource setup from Python to Typescript

SylvainSenechal:
Issue: ZENKO-5077

- Moving the configuration of various resources needed for the tests (accounts, locations, buckets etc) from python to typescript. Integrating it directly with our existing functional tests library (no new javascript project). 
This change wasn't absolutely needed, but it was very cheap and easy to do with Claude, it goes toward aligning the codebase on a single programming language and a single typescript project, to ease future maintenance.
It also open doors for later progressive refactoring of test configuration, as the code is now shared, cucumber/mocha test could directly call functions defined in this configuration folder
- Bump functional test transitive dependencies to reduce dependabot noise


tests/functional/configuration_setup/cleans3c.py
SylvainSenechal, line 1: this one is completely removed, was not used

tests/functional/testResourcesSetup/setup/buckets/ring.ts
```diff
@@ -0,0 +1,91 @@
...
+        credentials: {
+            accessKeyId: env.RING_S3C_ACCESS_KEY,
+            secretAccessKey: env.RING_S3C_SECRET_KEY,
+        },
+        endpoint: env.RING_S3C_ENDPOINT,
+        region: 'us-east-1',
+        forcePathStyle: true,
+    });
+
+    const versionedBucket = env.RING_S3C_INGESTION_SRC_BUCKET_NAME!;
```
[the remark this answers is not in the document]
SylvainSenechal: outdated

tests/functional/testResourcesSetup/setup/locations.ts
```diff
@@ -0,0 +1,105 @@
...
+
+        if (!env.ENABLE_RING_TESTS && location.locationType === 'location-scality-ring-s3-v1') {
+            continue;
+        }
+
+        if (location.locationType === 'location-scality-crr-v1') {
+            if (!env.DEPLOY_CRR_LOCATIONS) {
+                continue;
+            }
+            const accountName = location.name === env.CRR_DESTINATION_LOCATION_NAME
```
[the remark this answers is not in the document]
SylvainSenechal: outdated

tests/functional/testResourcesSetup/setup/buckets/aws.ts
SylvainSenechal, line 1: General thought : Not addressed in this pr, but wondering if some of these setup (no just this file, whole resources setup) could be done in simpler before hook from within the tests

I guess for some things like locations, it's not advisable because of reconciliation, or even not possible maybe for accounts that we wanna create before anything ? 
But maybe for buckets creation its possible (not even sure as some/all of these buckets are the buckets of the locations)

tests/functional/testResourcesSetup/setup/buckets/ring.ts
```diff
@@ -0,0 +1,81 @@
...
+        forcePathStyle: true,
+    });
+
+    if (!env.RING_S3C_INGESTION_SRC_BUCKET_NAME || !env.RING_S3C_INGESTION_SRC_NON_VERSIONED_BUCKET_NAME) {
+        throw new Error('RING_S3C_INGESTION_SRC_BUCKET_NAME and RING_S3C_INGESTION_SRC_NON_VERSIONED_BUCKET_NAME are required when Ring S3C credentials are provided');
+    }
+
+    const versionedBucket = env.RING_S3C_INGESTION_SRC_BUCKET_NAME;
+    const nonVersionedBucket = env.RING_S3C_INGESTION_SRC_NON_VERSIONED_BUCKET_NAME;
+    const objectCount = env.RING_S3C_INGESTION_NON_VERSIONED_OBJECT_COUNT_PER_TYPE ?? 0;
```
maeldonn: Behavioral regression: the Python original validated this env var was set (would throw on missing). It now silently defaults to `0`, skipping pre-population entirely.

tests/functional/testResourcesSetup/config.ts
```diff
@@ -0,0 +1,137 @@
...
+    AZURE_SECRET_KEY?: string;
+    AZURE_CRR_BUCKET_NAME?: string;
+    AZURE_ARCHIVE_BUCKET_NAME?: string;
+    AZURE_ARCHIVE_BUCKET_NAME_2?: string;
+    AZURE_ARCHIVE_QUEUE_NAME?: string;
+}
+
+export function loadConfig(configFile: string): E2EConfig {
+    const content = fs.readFileSync(configFile, 'utf8');
+    return yaml.load(content) as E2EConfig;
```
maeldonn, line 94: Behavioral regression: the Python original validated this config against a JSON schema. The `as E2EConfig` cast does no runtime validation, so a malformed `resources.yaml` crashes later with a cryptic `undefined` error.

maeldonn (approved):

maeldonn (commented):
Some regressions. Need to know if it's intentional or not.

tests/functional/testResourcesSetup/config.ts
SylvainSenechal: ah yeah i see, with python we had schema.py, then called validate(e2e_config, yaml.full_load(schema.e2e_config_schema))

I think it's ok without its a bit overkill here. If there is a spelling error it will be spotted in ci.
There is also a solution where we completely remove the template file, and directly write in json the resources, but for now it may be a bit complex to do because it's built from environment variable.

tests/functional/testResourcesSetup/setup/clients/pensieveClient.ts
```diff
@@ -0,0 +1,113 @@
...
+            method,
+            headers: {
+                'Content-Type': 'application/json',
+                'X-Authentication-Token': this.token,
+            },
+            body: JSON.stringify(body),
+        });
+        if (!response.ok) {
+            const text = await response.text();
+            const requestBody = JSON.stringify(body, null, 2);
```
benzekrimaha: can we avoid dumping the whole request body, especially since location payloads include secrets ?

tests/functional/testResourcesSetup/setup/clients/k8s.ts
```diff
@@ -0,0 +1,53 @@
...
+        apiVersion: 'v1',
+        kind: 'Secret',
+        metadata: { name, labels: { type: 'end2end' } },
+        stringData,
+    };
+    try {
+        await coreClient.createNamespacedSecret({ namespace, body: secret });
+        console.log(`Created k8s secret: ${name}`);
+    } catch (err: unknown) {
+        if ((err as { response?: { statusCode?: number } }).response?.statusCode === 409) {
```
benzekrimaha: CreateNamespacedSecret response processor throws new ApiException(response.httpStatusCode, ...) for non-2xx responses. I think our check should use err instanceof ApiException && err.code === 409, not err.response.statusCode === 409.
I checked the documentation as well and seems to be retuning code not statusCode , to be safe we can handle both

tests/functional/testResourcesSetup/setup/clients/pensieveClient.ts
SylvainSenechal: yes I removed it

tests/functional/testResourcesSetup/setup/clients/k8s.ts
SylvainSenechal: Yeah actually it was the reason why the ci was failing on pra tests

benzekrimaha (approved):

PR #2439, opened 2026-06-15
re-attribute mocha hook failures in junit merge

francoisferrand:
Mocha reports a failed beforeEach/afterEach as a synthetic testcase named '"<when>" hook for "<test>"' and emits nothing for the hook on a passing run. action-junit-report pairs a failure with a later pass by name/classname/file, so a hook-failure entry can never match its retry: a single hook flake fails the build on every rerun, and the only workaround is to manually delete the failed attempt's reports from the artifacts server.

Rewrite hook-failure testcases in the merge step to carry the name/classname of the test they belong to, so a retried pass cancels them out.

Issue: ZENKO-5296

### DIFF
--- a/.github/scripts/merge-junit-reports.py
+++ b/.github/scripts/merge-junit-reports.py
@@ -9,15 +9,39 @@ This ensures flaky test detection works properly in CI when tests are retried.
 
 import xml.etree.ElementTree as ET
 import os
+import re
 import sys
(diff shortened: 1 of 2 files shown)


maeldonn (approved):

benzekrimaha (approved):
LGTM

PR #2440, opened 2026-06-15
monitoring: check dashboard build fine

francoisferrand:
- upgrade grafanalib to latest version
- use requirements.txt to track these dependencies
- re-generate in CI to verify the json is up to date with the py source
- remove unused mongodb exporter metrics

Issue: ZENKO-5099


.github/workflows/end2end.yaml
```diff
@@ -126,17 +126,44 @@ env:
...
+          cache: pip
+          cache-dependency-path: monitoring/requirements.txt
+      - name: Setup venv
+        run: |
+          python -m venv .venv
+          echo "$PWD/.venv/bin" >> $GITHUB_PATH
+      - name: Install dependencies
+        working-directory: monitoring
+        run: pip install -r requirements.txt
+      - name: Generate dashboards
```
DarkIsDude, line 150: Why do we re-generate them in the CI instead of just generating and not commiting them ?

DarkIsDude (approved):

.github/workflows/end2end.yaml
francoisferrand: - It allows testing/deploying more easily : i.e. the pre-generated dashboard.json is already available for loading on a platform (manually, from grafana UI) if needed. For example for troubleshooting with CS team.
- the CI check verifies they are indeed correctly generated (i.e. proper dependencies, ...)

delthas (approved):

PR #2447, opened 2026-06-22
Add CRR cascaded tests

SylvainSenechal:
Issue: ZENKO-5263

Crr cascaded tests, didn't add very complex scenarios, let me know if you can come up with other test ideas but really not sure. I think there is only one happy path to test, and for the potential problematic situation, there is loop which I tested, and the rest not really sure we can test easily (like concurrency on write and stale micro version id) 

Wont work in ci, as deps.yaml isn't updated with cloudserver/backbeat.
BUT : "It works on my ~computer~ codespace"


.github/scripts/end2end/configure-e2e.sh
```diff
@@ -96,6 +96,12 @@ CRR_SOURCE_LOCATION_NAME=${CRR_SOURCE_LOCATION_NAME} \
 CRR_DESTINATION_LOCATION_NAME=${CRR_DESTINATION_LOCATION_NAME} \
 CRR_SOURCE_ACCOUNT_NAME=${CRR_SOURCE_ACCOUNT_NAME} \
 CRR_DESTINATION_ACCOUNT_NAME=${CRR_DESTINATION_ACCOUNT_NAME} \
+CRR_CASCADE_LOCATION_A_NAME=${CRR_CASCADE_LOCATION_A_NAME} \
```
SylvainSenechal: I know, in the past, we talked about avoiding creating too many locations but 🤔 
I don't know, I think it's probably fine ? 
- we still don't have that many locations
- It's probably not a bad idea to create many locations in ci, to see if it actually creates issues or not

tests/functional/ctst/steps/crrCascade.ts
```diff
@@ -0,0 +1,147 @@
...
+    'the object should replicate to location {string} within {int} seconds',
+    { timeout: 600_000 },
+    async function (this: Zenko, location: string, timeoutSeconds: number) {
+        const bucket = this.getSaved<Record<string, string>>('cascadeBuckets')[location];
+        const objectName = this.getSaved<string>('cascadeObjectName');
+        const deadline = Date.now() + timeoutSeconds * 1000;
+        while (Date.now() < deadline) {
+            Identity.useIdentity(IdentityEnum.ACCOUNT, location);
+            try {
+                await this.createS3Client().send(new HeadObjectCommand({ Bucket: bucket, Key: objectName }));
```
SylvainSenechal: Kept it light and very functional for this test : only check that the object is here, not checking further thing like specific object metadata/status. Could add, not sure it's super useful for functional tests, although I know sometimes we did it in the past

.github/scripts/end2end/configure-e2e.sh
maeldonn: It would be useful to share locations across tests. Could you make the name more generic ?

tests/functional/ctst/world/Zenko.ts
```diff
@@ -965,6 +965,7 @@ export default class Zenko extends World<ZenkoWorldParameters> {
             credentials: {
                 accessKeyId: credentials.accessKeyId,
                 secretAccessKey: credentials.secretAccessKey,
+                sessionToken: credentials.sessionToken || undefined,
```
maeldonn: `AWSCredentials.sessionToken?: string | undefined`

```suggestion
                sessionToken: credentials.sessionToken,
```

tests/functional/ctst/features/replication/crrCascade.feature
```diff
@@ -0,0 +1,37 @@
...
+        And a versioned bucket exists in location "crr-location-a"
+        And a versioned bucket exists in location "crr-location-b"
+        And a versioned bucket exists in location "crr-location-c"
+        And replication is configured from location "crr-location-b" to "crr-location-c"
+        And replication is configured from location "crr-location-c" to "crr-location-a"
+        And replication is configured from location "crr-location-a" to "crr-location-b"
+        When an object "cascade-loop-obj" is put in location "crr-location-a"
+        Then the object should replicate to location "crr-location-b" within 300 seconds
+        And the object should replicate to location "crr-location-c" within 300 seconds
+        And the object at location "crr-location-a" should never have replication status PENDING within 30 seconds
```
francoisferrand, line 39: should also check the "final" replication status on each location : SUCCESS / REPLICA / REPLICA  ?

francoisferrand, line 39: should we have more complex scenario, where we write on multiple points in the loop?
may not be able to control how he operations race to ensure a specific scenario, but worth adding to ensure the system is "eventually" consistent and stable?

tests/functional/ctst/features/replication/crrCascade.feature
```diff
@@ -0,0 +1,37 @@
...
+    Scenario: Cascade replication nominal case : A -> B -> C
+        Given cascade replication accounts are registered
+        And a versioned bucket exists in location "crr-location-a"
+        And a versioned bucket exists in location "crr-location-b"
+        And a versioned bucket exists in location "crr-location-c"
+        And replication is configured from location "crr-location-b" to "crr-location-c"
+        And replication is configured from location "crr-location-a" to "crr-location-b"
+        When an object "cascade-obj" is put in location "crr-location-a"
+        Then the object should replicate to location "crr-location-b" within 300 seconds
+        And the object should replicate to location "crr-location-c" within 300 seconds
```
francoisferrand, line 20: should also check the "final" replication status on each location : SUCCESS / REPLICA / REPLICA ?

tests/functional/testResourcesSetup/resources.yaml.template
```diff
@@ -133,3 +136,30 @@ locations:
       secretKey: ""
       bucketMatch: false
       repoId: []
+  - name: "${CRR_LOCATION_A_NAME}"
+    locationType: "location-scality-crr-v1"
```
francoisferrand, line 140: nit: do we really need multiple _locations_ to test the feature? could we not test with a single location and multiple buckets, yet get a similar result & good confidence?

tests/functional/ctst/steps/crrCascade.ts
francoisferrand: I think we should test `replication-status` ; but not use if this can be check here (may be race conditions -esp. when using cascade- so the status may have intermediate value...)

tests/functional/ctst/steps/crrCascade.ts
```diff
@@ -0,0 +1,151 @@
...
+                await new Promise(resolve => setTimeout(resolve, 3000));
+            }
+        }
+        assert.fail(`Timeout: object '${objectName}' not found in bucket '${bucket}' after ${timeoutSeconds}s`);
+    },
+);
+
+Then(
+    'the object at location {string} should never have replication status PENDING within {int} seconds',
+    { timeout: 120_000 },
```
francoisferrand, line 155: kind of redundant, deadline/timeout is implemented in the loop itself ?

SylvainSenechal: yeah but prefer to keep it, it could avoid ending up with 5h ci in case of a bug

tests/functional/testResourcesSetup/resources.yaml.template
SylvainSenechal: I think I discussed this with Maël too and I was thinking the same thing, technically it should be possible since the destination bucket is not bounded to the location, but in the end I think I prefer working with multiple locations as it forces us to build more realistic test scenarios

tests/functional/ctst/steps/crrCascade.ts
SylvainSenechal: yeah there are 2 things to check : 

  Metadata: { 'crr-location-c-replication-status': 'PENDING' },
  ReplicationStatus: 'REPLICA',

ReplicationStatus will always be REPLICA, because its based on the new isReplica boolean
But in metadata, depending how fast we check, it will be either PENDING or REPLICA, as a cascade transition will turn a replica into an object thats pending a replication.

I'm adding some check here, and i will add a step at the end to check all the status

tests/functional/ctst/features/replication/crrCascade.feature
SylvainSenechal: Yes I'm gonna add a step at the end to check all the status

tests/functional/ctst/features/replication/crrCascade.feature
SylvainSenechal: I came up with a more complex scenario which I think is decent : 

- Loop config
- Concurrent writes on all location, writing a specific random metadata marker
=> Assert that they all converge on the same metadataMarker

Tbh, still not the perfect test but I think worth doing

tests/functional/ctst/features/replication/crrCascade.feature
SylvainSenechal, line 1: Realising here I have 3 scenarios, and each new scenario pretty much covers the case of the previous scenario so in theory we could just run the last one to test the feature, I think still good to have a few scenarios

tests/functional/ctst/features/replication/crrCascade.feature
```diff
@@ -0,0 +1,57 @@
...
+        And the object at location "crr-location-a" should never have replication status PENDING within 30 seconds
+        When I wait 15 seconds
+        Then the cascade replication states should be settled
+
+    @2.15.0
+    @PreMerge
+    @ReplicationTest
+    @CRRCascade
+    @haha
```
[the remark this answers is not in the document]
SylvainSenechal: hold on hold on

tests/functional/ctst/features/replication/crrCascade.feature
```diff
@@ -0,0 +1,57 @@
...
+        And a versioned bucket exists in location "crr-location-a"
+        And a versioned bucket exists in location "crr-location-b"
+        And a versioned bucket exists in location "crr-location-c"
+        And replication is configured from location "crr-location-b" to "crr-location-c"
+        And replication is configured from location "crr-location-a" to "crr-location-b"
+        When an object "cascade-obj" is put in location "crr-location-a"
+        Then the object should replicate to location "crr-location-b" within 300 seconds
+        And the object should replicate to location "crr-location-c" within 300 seconds
+        When I wait 15 seconds
+        Then the cascade replication states should be settled
```
francoisferrand, line 22: instead/in addition to verify things are "settled" (which may pass by luck, in a dynamic system) : should/coudl we try to "count" replications? i.e. use metrics/logs/kafka topic to verify no extra replication was triggered...

(not sure if possible, i.e. how to observe this specifically when other tests may be running...)

tests/functional/ctst/steps/crrCascade.ts
francoisferrand: the problem is that the value is hard to set : set it long and it is mostly useless (may trigger is test is genuinely stuck, but only after a loooong time) but set it short and it may prevent the `waitSeconds` from working properly (and thus create a false positive)...

if we do this at least we should assert `waitSeconds` is below this timeout (120), to easily detect that inconsistency

tests/functional/ctst/steps/crrCascade.ts
```diff
@@ -0,0 +1,266 @@
...
+        }));
+    },
+);
+
+When('an object {string} is put in location {string}',
+    async function (this: Zenko, objectName: string, location: string) {
+        const cascadeBuckets = this.getSaved<Record<string, string>>('cascadeBuckets');
+        const marker = Utils.randomString().toLowerCase();
+        Identity.useIdentity(IdentityEnum.ACCOUNT, location);
+        await this.createS3Client().send(new PutObjectCommand({
```
francoisferrand: these tests rely on each (new) putObject call being made with a new marker : so this code happens in multiple places → should introduce a helper to put an object with a (new) marker, and addToSaved' it consistently

tests/functional/ctst/steps/crrCascade.ts
```diff
@@ -0,0 +1,266 @@
...
+        await Promise.all(
+            Object.entries(cascadeBuckets).flatMap(([location, bucket]) =>
+                Array.from({ length: writesPerLocation }, () => {
+                    const marker = Utils.randomString().toLowerCase();
+                    return clientByLocation[location].send(new PutObjectCommand({
+                        Bucket: bucket,
+                        Key: objectName,
+                        Body: new Uint8Array(0),
+                        Metadata: { marker },
+                    }));
```
francoisferrand: should save the marker (for consistency)

francoisferrand (approved):

delthas:
Can rebase on top of 2.16 (since #2378 was merged). Includes CRR changes.

tests/functional/ctst/features/replication/crrCascade.feature
```diff
@@ -0,0 +1,56 @@
+Feature: CRR Cascade Replication
+
+    Cross-Region Replication Cascade allows replication of an object from
+    a first location to carry on through multiple locations without
+    stopping at first destination.
+
+    @2.15.0
```
delthas: All those new tests should be marked 2.16 since that's where CRR will land.

delthas (approved):

SylvainSenechal:
Some comments, will do a last pass after bumping the components

maeldonn (approved):
Don't forget to update to @2.16 and also to check the final replication status

SylvainSenechal:
> Don't forget to update to @2.16 and also to check the final replication status

yeah 2.16 is done, also for final status its done with "Then the cascade replication states should be settled"

delthas:

delthas:
Heads-up: the deterministic (not flaky — retries won't help) `end2end-pra` "PRA (nominal case)" and `Kafka Cleaner` CTST failures on this PR are a regression in backbeat >= 9.5.0-preview.2: alert threshold defaults like `10*60` render invalid Go-template annotations, prometheus-operator rejects the rule and aborts the entire Prometheus config sync, so the metrics these tests poll never get scraped. Tracked in [BB-824](https://scality.atlassian.net/browse/BB-824), fix in scality/backbeat#2778 — bumping backbeat to 9.5.0-preview.7 once released will unblock it. (Full analysis: scality/Zenko#2461 (comment thread).)

PR #2448, opened 2026-06-23
fix codespace yarn setup prompt

SylvainSenechal:
Issue: ZENKO-5301

Issue probably introduced when I changed the resources setup from python to typescript.

In codespace, after the cluster is deployed, we (the codespace, it's not the developer doing this manually 🧐 ) need to run the typescript resources setup script, but it's blocked with : 
image

This shouldn't require manual intervention, and anyways, this screenshot is from the codespace creation log, it's not an interactive shell, we can't 'Y' that thing

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -112,6 +112,7 @@
     "containerEnv": {
         "ZENKO_MONGODB_DATABASE": "zenko-database"
     },
+    "onCreateCommand": "sudo corepack disable",
     // Use 'postCreateCommand' to run commands after the container is created.
     "postCreateCommand": ".devcontainer/setup.sh"
 }


.devcontainer/devcontainer.json
```diff
@@ -113,5 +113,5 @@
         "ZENKO_MONGODB_DATABASE": "zenko-database"
     },
     // Use 'postCreateCommand' to run commands after the container is created.
-    "postCreateCommand": ".devcontainer/setup.sh"
+    "postCreateCommand": "sudo corepack enable && corepack prepare yarn@1.22.22 --activate && .devcontainer/setup.sh"
```
francoisferrand: we don't actually use corepack:
- should the change not be to _disable_ corepack ?
- can't it be done during the build (in Dockerfile), instead of doing it on startup?

francoisferrand: had similar issue on my laptop, c.f. https://scality.slack.com/archives/C0A1GUKGLNS/p1773765068189439

SylvainSenechal: Let me see, looking on some threads i saw some people using an env var to disable it

.devcontainer/devcontainer.json
```diff
@@ -110,7 +110,8 @@
         }
     },
     "containerEnv": {
-        "ZENKO_MONGODB_DATABASE": "zenko-database"
+        "ZENKO_MONGODB_DATABASE": "zenko-database",
+        "COREPACK_ENABLE_DOWNLOAD_PROMPT": "0"
```
francoisferrand: does this disable "corepack" (i.e. same as `corepack disable`) ? or does it disable the prompt?

SylvainSenechal: it doesn't disable the prompt, and we aren't using corepack currently so we can disable it. I guess both solution are ok though

benzekrimaha (approved):

francoisferrand (approved):

PR #2451, opened 2026-06-29
Add Cucumber tests for multi-destination CRR

maeldonn:
Cover V2 multi-rule replication configs, mixed CRR + cloud destinations, overlap validation, disabled rules, and V1 plus legacy comma-StorageClass backward compatibility. Factor shared replication polling and replica-match helpers out of the existing single-destination steps so both single- and multi-destination scenarios reuse them.

Issue: ZENKO-5253

### DIFF
--- a/tests/functional/ctst/features/replication/replication.feature
+++ b/tests/functional/ctst/features/replication/replication.feature
@@ -1,22 +1,21 @@
-Feature: Live Object Replication
-    This feature tests live replication of objects across different sizes,
-    verifying that oplog location stripping does not affect replication
-    correctness.
-
-    The objectRefreshSizeThresholdMB is set to 2 in the Zenko CR, meaning
-    objects >= 2 MiB have their location stripped from the MongoDB change
-    stream event. The replication consumer detects this and re-fetches
-    metadata from the source before replicating.
+Feature: Bucket Replication
+    End-to-end coverage of Zenko's bucket replication, covering both
+    cloud destinations (AWS / GCP / Azure backends) and CRR loopback
+    destinations (replication between Zenko accounts).
 
     @2.14.0
     @PreMerge
     @ReplicationTest
+    # Targets the location-stripping threshold: objectRefreshSizeThresholdMB
+    # is set to 2 in the Zenko CR, so objects >= 2 MiB have their location
+    # stripped from the MongoDB change stream event. The replication consumer
+    # detects this and re-fetches metadata from the source before replicating.
     Scenario Outline: Object of <objectDescription> replicates with location stripping threshold
         Given an existing bucket "repl-strip-<sizeBytes>" "with" versioning, "without" ObjectLock "without" retention mode
         And a replication configuration to "awsbackendmismatch" location
         And 1 objects "repl-strip-obj" of size <sizeBytes> bytes
         Then the object replication should "succeed" within 300 seconds
-        And the replicated object should be the same as the source object
+        And the replicated object should match the source on every configured destination
 
         Examples:
             | sizeBytes | objectDescription                  |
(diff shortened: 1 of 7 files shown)


tests/functional/ctst/features/replication/replication.feature
```diff
@@ -24,3 +23,156 @@ Feature: Live Object Replication
...
+    # Multi-destination replication: a single replication configuration
+    # carries several rules targeting distinct destinations, optionally with
+    # overlapping prefixes resolved by Priority. The legacy V1 format and
+    # the non-standard comma-separated StorageClass form remain supported
+    # for backward compatibility and round-trip unchanged through
+    # getBucketReplication.
+    Scenario: V2 replication fans out to two cloud destinations
+        Given an existing bucket "multi-cloud-2" "with" versioning, "without" ObjectLock "without" retention mode
+        And a multi-destination replication configuration with rules:
+            | id     | prefix | priority | location              |
```
SylvainSenechal, line 39: didnt know about this table thing, its nice

tests/functional/ctst/features/replication/replication.feature
```diff
@@ -24,3 +23,156 @@ Feature: Live Object Replication
...
+
+    @2.15.0
+    @PreMerge
+    @ReplicationTest
+    Scenario: V2 rules with overlapping prefixes to different destinations all apply
+        Given an existing bucket "multi-overlap" "with" versioning, "without" ObjectLock "without" retention mode
+        And a multi-destination replication configuration with rules:
+            | id     | prefix | priority | location              |
+            | broad  |        | 1        | awsbackendmismatch    |
+            | narrow | docs-| 2        | awsbackend    |
```
SylvainSenechal: alignement

tests/functional/ctst/steps/replication.ts
```diff
@@ -1,18 +1,342 @@
...
+                secretAccessKey: locationsConfigs[location].details.credentials.secretKey,
+            },
+            tls: false,
+            maxAttempts: 1,
+            forcePathStyle: true,
+        }),
+    };
+}
+
+async function putBucketReplicationRaw(
```
SylvainSenechal: No change required but just putting it out there : 
It's starting to get a little smelly, we have 2/3 ways to putBucketReplication, and more generally, not just for replication tests but for any tests, we tend to have 2/3 ways to do the same thing. Probably means the existing functions are too hard to reuse, or the code structure is not good

tests/functional/ctst/features/replication/replication.feature
```diff
@@ -24,3 +23,156 @@ Feature: Live Object Replication
...
+    Scenario: V2 replication fans out to two cloud destinations
+        Given an existing bucket "multi-cloud-2" "with" versioning, "without" ObjectLock "without" retention mode
+        And a multi-destination replication configuration with rules:
+            | id     | prefix | priority | location              |
+            | rule-a |        | 1        | awsbackendmismatch    |
+            | rule-b |        | 2        | awsbackend    |
+        And 1 objects "multi-cloud-obj" of size 1024 bytes
+        Then the multi-destination object replication should "succeed" within 300 seconds
+        And the replicated object on "awsbackendmismatch" should match the source
+        And the replicated object on "awsbackend" should match the source
```
SylvainSenechal: nit but nice if possible : 
Here you are referencing awsbackend and awsbackendmismatch, because they appeared in the rules table defined on the second step.

Would it be possible to have the variables defined once in an example table, and have the steps reference it instead.

tests/functional/ctst/features/replication/replication.feature
SylvainSenechal: Although, to be honest, looking at the code associated with it, I think it may have been easier and more cucumber like to implement it this way 

Given a replication rule <id> with prefix <prefix> and priority <priority> on location <location>
Given a replication rule <id> with prefix <prefix> and priority <priority> on location <location>
When i apply the given rules on my bucket
....

tests/functional/ctst/steps/replication.ts
```diff
@@ -1,18 +1,342 @@
...
+    status?: string;
+    location: string;
+    bucket?: string;
+    account?: string;
+    deleteMarkerReplication?: string;
+};
+
+const DEFAULT_ROLE = 'arn:aws:iam::root:role/s3-replication-role';
+
+interface CRRAccountInfo {
```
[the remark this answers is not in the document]
SylvainSenechal: 🤔 this could be fine if they have different meaning, I would move the const and type definition higher in the file though

tests/functional/ctst/steps/replication.ts
```diff
@@ -210,3 +447,134 @@ When('the destination bucket on the location is created again', async function (
     });
     await awsS3Client.send(versioningCommand);
 });
+
+Given('a multi-destination replication configuration with rules:',
```
SylvainSenechal: I haven't evaluated if this is hard to do or not, maybe its not even possible without creating a mess, but I see we have 3 steps that are basically :

Given a valid multi destination
Given an invalid multi destination
given a legacy replication configuration

Maybe the legacy config should stay isolated as it is but for the valid/invalid multi replication : Should they be under the same step, with a valid/invalid variable passed as a parameter ? It could allow some decent refactoring to inline the helper function buildConfigFromRows, maybe also inline enabledLicationsFromRows, and probably easier to see the difference between the 2 valid/invalid situations.

tests/functional/ctst/features/replication/replication.feature
maeldonn: Updated.

tests/functional/ctst/steps/replication.ts
```diff
@@ -210,3 +447,134 @@ When('the destination bucket on the location is created again', async function (
...
+            locations,
+        );
+    });
+
+Then('the replicated object on {string} should match the source',
+    async function (this: Zenko, location: string) {
+        await assertReplicaMatchesSource(this, location);
+    });
+
+Then('the object should not be replicated to {string}',
```
SylvainSenechal, line 499: We have this step already : 
the multi-destination object replication should {string} within {int} seconds

and this function pollReplicationOutcome

Don't we have a way to merge into a single step with a variable "the object should (not?) be replicated to .." ?

SylvainSenechal: I would inline readPerDestinationStatus, but still think it could use the pollReplicationOutcome with replicationOutcome = "never happen". The two function are just doing headObjects and checking its result

SylvainSenechal (commented):
For me it's pretty much ok, I left some comment that you can check, I think it's ok if the suggested changes aren't applied but there may be some refactoring possible to make it a bit easier to read

tests/functional/ctst/steps/replication.ts
maeldonn: Merged the two `S3.putBucketReplication` helpers into one `putBucketReplicationRaw` in utils.ts, cascade path stays separate.

tests/functional/ctst/features/replication/replication.feature
maeldonn: The rules in a `putBucketReplication` call are a set applied atomically as one config, not a sequence of setup steps. The table reflects that semantically, and mirrors the shape of the actual `ReplicationConfiguration.Rules` array, so the feature reads close to what's actually sent. The line-per-rule form would imply sequencing that doesn't exist and drift further from the API shape as we add columns.

tests/functional/ctst/steps/replication.ts
maeldonn: Did a bit of refactor

tests/functional/ctst/features/replication/replication.feature
maeldonn: Added `the replicated object should match the source on every configured destination` which reads back the saved locations

tests/functional/ctst/steps/replication.ts
maeldonn: Inlined readPerDestinationStatus. Kept it separate from pollReplicationOutcome since other rules complete here, so the top-level is COMPLETED while only the disabled location stays undefined. "never happen" would mean two different things.

delthas:
Can rebase on top of 2.16 (since #2378 was merged). Includes CRR changes.

SylvainSenechal:
> Can rebase on top of 2.16 (since #2378 was merged). Includes CRR changes.

rebase from my branch, which i also just rebased from 2.16

SylvainSenechal (approved):
Just nit on linter warning single line if and alignement of rules array

tests/functional/ctst/features/replication/replication.feature
```diff
@@ -1,26 +1,167 @@
...
+        And the replicated object should match the source on every configured destination
 
         Examples:
             | sizeBytes | objectDescription                  |
             |         0 | 0 bytes (zero-byte)                |
             |   1048576 | 1 MiB (below threshold)            |
             |   2097152 | 2 MiB (at threshold, stripped)     |
             |   4194304 | 4 MiB (above threshold, stripped)  |
+
+    @2.15.0
```
delthas: CRR v2 will be shipped in 2.16 -> all these new tests should be tagged 2.16.0

delthas (approved):
LGTM apart from comment

delthas:

delthas:
Heads-up: the deterministic (not flaky — retries won't help) `end2end-pra` "PRA (nominal case)" and `Kafka Cleaner` CTST failures on this PR are a regression in backbeat >= 9.5.0-preview.2: alert threshold defaults like `10*60` render invalid Go-template annotations, prometheus-operator rejects the rule and aborts the entire Prometheus config sync, so the metrics these tests poll never get scraped. Tracked in [BB-824](https://scality.atlassian.net/browse/BB-824), fix in scality/backbeat#2778 — bumping backbeat to 9.5.0-preview.7 once released will unblock it. (Full analysis: scality/Zenko#2461 (comment thread).)

PR #2452, opened 2026-06-29
🐛 (ZENKO-5303) make voting and priority right

DarkIsDude:
https://github.com/scality/Zenko/pull/2449

UPSTREAM : https://github.com/bitnami/containers/pull/95156

It fixes a bug in the MongoDB sharded image bootstrap (`libmongodb.sh`) that can leave every replica set with **only one voting member**, turning that single node into a hard single point of failure: if it goes down, the shard (and therefore the datastore) goes read‑only and Zenko is down.

---

### A 30‑second MongoDB primer (for non‑Mongo readers)

Our data lives in **replica sets**: groups of MongoDB nodes (here, 3 per shard and 3 for the config server) that each hold a copy of the data.

- One node is the **PRIMARY** (takes writes); the others are **SECONDARIES** (copies).
- To stay alive, a replica set must keep a **majority of votes** online and be able to **elect** a PRIMARY. Two settings per member matter:
  - **`votes`** (0 or 1): can this member vote in an election? Majority is counted over the *sum of votes*.
  - **`priority`** (0+): can this member be *elected* PRIMARY? `priority: 0` means "never become PRIMARY".
- A healthy 3‑node set has **3 members with `votes: 1, priority: 1`**. It survives losing any one node: the remaining 2 votes are still a majority, and a surviving member can be elected PRIMARY.

A member that is only partially configured — `votes: 1, priority: 0`, or worse `votes: 0, priority: 0` — still holds data but **cannot help keep the set alive**.

---

### The bug

When a SECONDARY first joins, the bootstrap script does this (`mongodb_configure_secondary`):

1. `rs.add(... votes: 0, priority: 0)` — add the node *without* voting power so it doesn't disturb the existing majority while it copies data (this is MongoDB's recommended safe procedure).
2. **Confirm** the node is now listed in the replica set (`mongodb_node_currently_in_cluster`).
3. Wait for it to finish its initial sync (reach `SECONDARY` state).
4. **Grant voting rights**: reconfigure it to `votes: 1, priority: 1`.

The confirmation in step 2 reads `rs.status()` and greps the output for the node:

```bash
result=$(mongodb_execute ... <<< "rs.status().members")
grep -q "'$node:$port'" <<<"$result"
```

The problem: with the broken image, `mongodb_execute` is a thin wrapper around `debug_execute`, which **throws away stdout** unless `BITNAMI_DEBUG=true` (it isn't, by default):

```bash
debug_execute() { if is_boolean_yes "${BITNAMI_DEBUG:-false}"; then "$@"; else "$@" >/dev/null 2>&1; fi; }
```

So `result` is **always empty**, the grep always fails, and `mongodb_node_currently_in_cluster` **always returns false**. That makes step 2 (`mongodb_wait_confirmation`) time out and the bootstrap aborts with:

```
ERROR ==> Unable to confirm that <node> has been added to the replica set!
```

The container exits, Kubernetes restarts the pod, and on the **second** boot the data directory already exists, so the bootstrap takes the "deploy with persisted data" path and **skips replica set configuration entirely**. The node is left frozen at `votes: 0, priority: 0`. Steps 3–4 never run, so the SECONDARY is never promoted.

Net result: only the bootstrap PRIMARY (`*-0`) ends up with a vote → **1 voter per replica set → single point of failure**.

---

### Root cause: an incomplete fork of the Bitnami image

Bitnami stopped publishing the `mongodb-sharded` image, so we vendored its scripts into the repo (ZENKO-5110, #2366). The vendoring happened in **two** commits, and they are **not equal**:

| Commit | Branch | `libmongodb.sh` | `mongodb_execute()` defs | Result |
|---|---|---|---|---|
| [`0ae8c8a3`](https://github.com/scality/Zenko/commit/0ae8c8a3e4e92a7545f454c6f1a064bce792fdf1) | development/2.14 | **1712 lines** | **2** | ✅ works (→ 2.14.5) |
| [`2f6c7e42`](https://github.com/scality/Zenko/commit/2f6c7e42dd879975a83868c2e07786f60338f8fe) | development/2.15 | **1669 lines** | **1** | ❌ broken (→ 2.15.1) |

[**Compare**](https://github.com/scality/Zenko/compare/0ae8c8a3e4e92a7545f454c6f1a064bce792fdf1...2f6c7e42dd879975a83868c2e07786f60338f8fe): the only relevant difference is the **last 43 lines** of `libmongodb.sh`.

Upstream's `libmongodb.sh` is assembled by **concatenating script fragments**, and it deliberately defines `mongodb_execute` **twice**:

1. Early in the file — the legacy wrapper that **discards stdout** (`debug_execute mongodb_execute_print_output "$@"`).
2. As a **final appended fragment** (it carries its own `# Copyright …` header and `# shellcheck disable=SC2148` — the tell‑tale "no shebang" marker of a separate concatenated file) — the real version that calls `mongosh` directly and **returns output**.

In bash **the last definition wins**, so upstream's *effective* `mongodb_execute` is #2 (returns output) — which is exactly what `mongodb_node_currently_in_cluster` needs.

The **2.15 re‑vendoring (`2f6c7e42`) truncated the file at 1669 lines and dropped that final fragment.** Only the output‑discarding wrapper was left, silently reverting `mongodb_execute` and breaking the confirmation check. The 2.14 vendoring had copied the whole file, so 2.14.5 worked. We didn't *add* a fix in 2.14 — we just vendored completely there, and lost it in 2.15.

It was easy to miss because dropping a *duplicate* function definition leaves valid bash that runs fine; the only signal was the line count (1712 vs 1669), and the failure only surfaces as a silent bootstrap race that is invisible until a node dies.

#### How we confirmed it (two clusters, same Mongo version)

A cluster on image base `2.14.5` was healthy (3 voters); a cluster on `2.15.1` was broken (1 voter). We confirmed the chain end‑to‑end from the live `rs.conf()` (broken cluster: secondaries at `votes: 0, priority: 0`; healthy cluster: `votes: 1`) and the pod boot logs (broken cluster fails at *"Unable to confirm…"*; healthy one gets past it). The diff between the two images' `libmongodb.sh` was exactly the 43‑line fragment above.

---

### The fix

Two commits:

1. **`🐛 add missing libmongo.sh fork`** — restores the dropped 43‑line fragment, so the file matches upstream again. `mongodb_execute` is once more the output‑returning definition, `mongodb_node_currently_in_cluster` can read `rs.status()`, and the bootstrap no longer aborts before granting voting rights. This is the root‑cause fix (faithful re‑sync with upstream). (in the first PR)

2. **`🐛 ensure secondary keeps votes and priority after restart`** — makes the voting‑rights grant **idempotent and re‑runnable**, so a node can no longer be left stranded without votes:
   - new helper `mongodb_secondary_node_has_voting_rights` checks whether the member already has `votes > 0 && priority > 0`;
   - `mongodb_configure_secondary` now grants voting rights **whenever they are missing — even if the node is already in the cluster**, instead of only on the freshly‑added path. So if a previous attempt added the node at `votes/priority 0` and then failed (or was restarted) before promotion, the next run finishes the job and converges it to `votes: 1, priority: 1`;
   - the misleading error message (`"did not get marked as secondary"` printed for the voting step) is corrected to `"did not get granted voting rights"`.

After this change a fresh deployment reliably ends with all members at `votes: 1, priority: 1` (true HA), and a member that is already in the replica set but under‑privileged is repaired rather than silently left non‑voting.

---

Fixes ZENKO-5302.

- **Already‑broken clusters won't fully self‑heal from the image alone.** On a pod restart the bootstrap takes the persisted‑data path, which skips replica‑set (re)configuration entirely, so an existing cluster already stranded at `votes: 0` still needs a **one‑time manual** `rs.reconfig()` on each replica set (`shard-N` and `configsvr`) to set members 1 and 2 to `votes: 1, priority: 1`. The script change guarantees correct behaviour for **new** bootstraps and for any path where `mongodb_configure_secondary` runs against a member that lacks voting rights.
- Should we keep the second fix ? We can use a dedicated PR for that as we already have this issue on older cluster (4.2.3 have it).


solution-base/images/mongodb-sharded/debian-12/rootfs/opt/bitnami/scripts/libmongodb.sh
```diff
@@ -1667,3 +1700,46 @@ mongodb_execute_print_output() {
...
+#   $1 - User to run queries
+#   $2 - Password
+#   $3 - Database where to run the queries
+#   $4 - Host (default to result of get_mongo_hostname function)
+#   $5 - Port (default $MONGODB_PORT_NUMBER)
+#   $6 - Extra arguments (default $MONGODB_SHELL_EXTRA_FLAGS)
+# Returns:
+#   None
+########################
+mongodb_execute() {
```
francoisferrand, line 1729: already merged : please rebase on top of actual target branch

solution-base/images/mongodb-sharded/debian-12/rootfs/opt/bitnami/scripts/libmongodb.sh
```diff
@@ -951,6 +951,32 @@ EOF
     grep -q "ok: 1" <<<"$result"
 }
 
+########################
+# Get if secondary node already has voting rights
```
francoisferrand, line 958: nit:
```suggestion
# Check if secondary node already has voting rights
```

solution-base/images/mongodb-sharded/debian-12/rootfs/opt/bitnami/scripts/libmongodb.sh
```diff
@@ -1187,7 +1213,15 @@ mongodb_configure_secondary() {
...
         mongodb_wait_confirmation "$node" "$port"
+    fi
 
+    # Grant voting rights to the node if it does not have them yet. This must be
+    # done even when the node is already in the cluster: a previous attempt may
+    # have added it with votes/priority 0 and failed (or been restarted) before
+    # granting voting rights, leaving the node stuck without them.
+    if mongodb_secondary_node_has_voting_rights "$node" "$port"; then
+        info "Node already has voting rights"
+    else
```
francoisferrand, line 1230: this will not behave properly when we have 9-replicas, where 2 nodes are always "non voting". This is expected, and we must not try to change that (in particular, the voting nodes must be in precisely the expected DC for proper redundancy)

→ the change will make the startup slower on these extra secondaries -(re-)trying to make them voting-, possibly with the extra risk of _changing_ the set of voters...

francoisferrand (commented):
The change itself looks ok; however I am not sure we should do this.

I think the configuration was done only on first startup _by design_ : this ensures that after first startup, each instance will just use the configuration it has and let (human) operator manage it.

This means there is a gap in the chart indeed: if the instance is restarted during this startup, it may be left in an incorrect state... But this is a tradeoff: either we enforce the state of replicas (i.e. your change) -even though we don't actually know the intent of the human deploying the chart- and risk setting the wrong state for more advanced setups ; or we keep the existing/upstream approach to automatically handle the nominal path only, and leave recovery for humans...

→ if we were writing an _operator_, I would say it should recover automatically -and we just need to add whatever necessary in the CR to make intent clear
→ however this is a chart -mostly static, with not much way to get instant- so I would rather stay conservative
→ practically we never experienced this issue until we removed part of the chart (what you already fixed) : so the risk seems very low to keep it as upstream, and we can add an extra check in _installer_ to validate the overall state during/after chart deployment for extra safety?

delthas (commented):
- Makefile clone optimization: LGTM
- Granting voting rights to existing replicas: will break on > 7 replicas (as mentioned by François). Not sure how to proceed

solution-base/images/mongodb-sharded/debian-12/rootfs/opt/bitnami/scripts/libmongodb.sh
```diff
@@ -951,6 +951,32 @@ EOF
...
+#########################
+mongodb_secondary_node_has_voting_rights() {
+    local -r node="${1:?node is required}"
+    local -r port="${2:?port is required}"
+    local result
+
+    debug "Checking voting rights of the node"
+    result=$(
+        mongodb_execute_print_output "$MONGODB_INITIAL_PRIMARY_ROOT_USER" "$MONGODB_INITIAL_PRIMARY_ROOT_PASSWORD" "admin" "$MONGODB_INITIAL_PRIMARY_HOST" "$MONGODB_INITIAL_PRIMARY_PORT_NUMBER" <<EOF
+rs.conf().members.filter(m => m.host === '$node:$port' && m.votes > 0 && m.priority > 0).length === 1
```
maeldonn: nit:
```suggestion
rs.conf().members.some(m => m.host === '$node:$port' && m.votes > 0 && m.priority > 0)
```

maeldonn (commented):
+1 for @francoisferrand

DarkIsDude:
@francoisferrand @maeldonn @delthas the upstream has been merged, what should we do ?

solution-base/images/mongodb-sharded/debian-12/rootfs/opt/bitnami/scripts/libmongodb.sh
francoisferrand: @DarkIsDude did you try that case?

Looking at the code below, it seems that the script would exit if it fails to configure secondary node voting: does that function already silently handle the case where we reached the max number of voters already?

francoisferrand (changes requested):
Granting voting rights to existing replicas will break on > 7 replicas : need to add extra condition to either
- skip `mongodb_configure_secondary_node_voting` if there are already too many voters
- or detect (and ignore) the "too many voters" error
- or ignore the `mongodb_configure_secondary_node_voting` error if we see (after the call/error) that there are 7 voters already

(in particular, must check the behavior with the 9-replicas setup we have: to ensure the change would not change the voters: which are manually set to be precisely in the expected datacenter/room, and thus meet the availability constraints specified by customer)

SylvainSenechal (commented):
. nvm

solution-base/images/mongodb-sharded/debian-12/rootfs/opt/bitnami/scripts/libmongodb.sh
DarkIsDude: Yes — tested on a real 9-replica shard, both with the current development/2.15 image and with this branch. Short answer to your question: no, it is not silently handled — it exits. But that is pre-existing behaviour, not something this PR introduces.

First: your review caught a real bug, just not the one we were discussing. `mongodb_secondary_node_has_voting_rights()` grepped mongosh's output for `true`, and mongosh always prints a connection banner containing `directConnection=true`. So the check always returned true, the grant branch became dead code, and no secondary ever got voting rights. Observed directly: a fresh secondary joined with a free voter slot available and stayed at votes=0 forever (9 members / 6 voters). By extension a fresh install would leave only pod-0 — which self-initiates — with a vote. Fixed using an explicit HAS_VOTES_YES/HAS_VOTES_NO sentinel. Re-tested after the fix: same scenario now yields votes=1 priority=1 → 7 voters.

Secondly: "the extra risk of changing the set of voters": The only path where this branch differs from development/2.15 is data dir empty and node already in rs.conf() — i.e. PVC loss on an existing member.

Last one: "does that function silently handle reaching the max number of voters?" No. Reproduced, scaling 3 → 9:
```
mongod:  Replica set configuration contains 8 voting members, but must be at least 1 and no more than 7
pod-7:   ERROR ==> Secondary node did not get marked as secondary     (after 24 × 5s of retries)
         → exit 1 → container restart
```
It does recover: on restart the data dir is no longer empty, config is skipped, and the member settles as a working non-voting secondary. Net cost is one crash + ~3 min per extra member, and the final topology is the expected 9 members / 7 voters.
This branch behaves identically (9 members / 7 voters, one crash each on pods 7 and 8), with a more accurate message: Secondary node did not get granted voting rights rather than the old, misleading "did not get marked as secondary".
Happy to add a members.filter(m => m.votes > 0).length < 7 guard to skip the doomed promotion entirely if you'd like that in this PR. Put the behaviour is unchanged.

DarkIsDude:
> Granting voting rights to existing replicas will break on > 7 replicas : need to add extra condition to either
> 
> (in particular, must check the behavior with the 9-replicas setup we have: to ensure the change would not change the voters: which are manually set to be precisely in the expected datacenter/room, and thus meet the availability constraints specified by customer)

Tested on a real 9-replica shard, on both 2.15 and this branch.

Good catch, but the check was worse than suspected: it grepped mongosh output for true, which always matches the directConnection=true banner — so no secondary ever got votes at all. Fixed in f313b494, plus 5f71df4 for the same bug in mongodb_is_secondary_node_ready.

The >7 crash is real but pre-existing (reproduced on 2.15): ~2 min of retries, exit 1, then the restart finds a non-empty data dir, skips config, and the member settles as non-voting — final topology is still 9 members / 7 voters.

Voters can't drift in the nominal case: configure_replica_set only runs when the data dir is empty (libmongodb-sharded.sh:74), so restarts and rolling updates never touch rs.conf(); and at 7 voters the ceiling rejects any promotion.

The only real risk is "empty disk + ≤6 voters" (6+1=7 is accepted), i.e. node replacement, since storage is node-local. We can add your idea — a pre-check on the voter count rather than parsing the error string if you prefer but this will differ from the upstream ?

solution-base/images/Makefile
```diff
@@ -25,15 +24,20 @@ vendor_branch = vendor/$1/$(call bitnami_path,$1)
...
+	# Shallow-fetch only the pinned ref; full upstream history is unnecessary for a --squash merge.
+	git fetch --depth 1 --filter=blob:none $(BITNAMI_REMOTE) $(call bitnami_ref,$*)
 	-git branch -D $(call vendor_branch,$*)
-	git subtree split --prefix=bitnami/$*/$(call bitnami_path,$*) $(call bitnami_ref,$*) -b $(call vendor_branch,$*)
+	sha=$$(git rev-parse FETCH_HEAD) && \
+	wt=$$(mktemp -d) && \
+	trap 'git worktree remove --force "$$wt" 2>/dev/null' EXIT && \
+	git worktree add --no-checkout --detach "$$wt" "$$sha" && \
+	git -C "$$wt" sparse-checkout set --no-cone bitnami/$*/$(call bitnami_path,$*) && \
+	git -C "$$wt" checkout && \
```
francoisferrand: why do we need a checkout?

solution-base/images/Makefile
```diff
@@ -25,15 +24,20 @@ vendor_branch = vendor/$1/$(call bitnami_path,$1)
...
-	# Fetch full history from Bitnami main so subtree split can see full subtree history.
-	git fetch $(BITNAMI_REMOTE) main
-
-update-vendor-branch-%: fetch-remote
+update-vendor-branch-%: create-remote
+	# Shallow-fetch only the pinned ref; full upstream history is unnecessary for a --squash merge.
+	git fetch --depth 1 --filter=blob:none $(BITNAMI_REMOTE) $(call bitnami_ref,$*)
 	-git branch -D $(call vendor_branch,$*)
-	git subtree split --prefix=bitnami/$*/$(call bitnami_path,$*) $(call bitnami_ref,$*) -b $(call vendor_branch,$*)
+	sha=$$(git rev-parse FETCH_HEAD) && \
```
francoisferrand: no need for Git, this is actually `$(call bitnami_ref,$*)`.
So best to do it the other way:

```
sha=$(call bitnami_ref,$*) &&
git fetch --depth 1 --filter=blob:none $(BITNAMI_REMOTE) $${sha} &&
```

solution-base/images/mongodb-sharded/debian-12/rootfs/opt/bitnami/scripts/libmongodb.sh
francoisferrand: > Happy to add a members.filter(m => m.votes > 0).length < 7 guard to skip the doomed promotion entirely if you'd like that in this PR. Put the behaviour is unchanged.

Ack for the unchanged behavior, but I guess it's best to fix it, right?
(should probably be upstreamed as well)

francoisferrand (changes requested):
The history of this PR seems weird : there are commits to remove "vendored image dirs ahead of subtree bootstrap"

The whole point of using git subtree is that we don't revenger from scratch (= loose local changes), but can instead repeatedly merge with the usual git semantics

→ please remove the history, and update upstream with merge only (bootstrap has been done already, no need to redo it

DarkIsDude:
> The history of this PR seems weird : there are commits to remove "vendored image dirs ahead of subtree bootstrap"
> 
> The whole point of using git subtree is that we don't revenger from scratch (= loose local changes), but can instead repeatedly merge with the usual git semantics
> 
> → please remove the history, and update upstream with merge only (bootstrap has been done already, no need to redo it

Thanks for the question. The bootstrap was never done or working. The subtree was never bootstrapped for solution-base/images/* but for solution-base/mongodb/charts/mongodb-sharded. The initial copy was just a plain copy 😬. I did a reword in any case to make it more clear.

solution-base/images/Makefile
DarkIsDude: `git subtree` checks the prefix on disk. This split only reads history, but it still refuses to start if the directory is not materialized. So without the checkout:

```
git worktree add --no-checkout --detach "$wt" "$sha"
git -C "$wt" subtree split --prefix=bitnami/mongodb-sharded/8.0/debian-12 "$sha" -b vendor/...
fatal: 'bitnami/mongodb-sharded/8.0/debian-12' does not exist; use 'git subtree add'
```

solution-base/images/Makefile
DarkIsDude: `bitnami_ref` is a SHA only for `mongodb-sharded`; `mongodb-exporter` and `os-shell` are pinned to main. `git rev-parse FETCH_HEAD` is there to collapse both forms into a concrete commit, which the two later steps need.

With `sha=$(call bitnami_ref,$*)`, the branch-name case breaks. Reproduced on a repo shaped like this one (no local main, remote named `bitnami-containers`).

We can drop that if we pin three image to sha. Not sure it's useful now.

solution-base/images/mongodb-sharded/debian-12/rootfs/opt/bitnami/scripts/libmongodb.sh
DarkIsDude: I'm not sure we should fix it.

- It's pure cosmetic.
- We'll add in the logic (on our side and maybe upstream) a logic that is specific to Mongo. If Mongo change this limits or change the behaviour, no one will notice that here.
- It's at boot time, so race condition can occur (don't know if they can occur) : two extra pods bootstrapping at the same time both read 6 voters, both proceed, and one still fails;

If we really wants to fix it, we should read the output from Mongo, parse it and stop retrying. Again, not sure is useful right now.

solution-base/images/Makefile
```diff
@@ -1,21 +1,20 @@
...
-BITNAMI_mongodb_sharded_REF := 48a109547d39cd8cf8a5d4058d832ecb5844829e
+BITNAMI_mongodb_sharded_REF := 657585595c550d4dc107a4e6cd3a598a9d284eec
 
 BITNAMI_mongodb_exporter_PATH := 0/debian-12
-BITNAMI_mongodb_exporter_REF := $(BITNAMI_UPSTREAM_MAIN_REF)
+BITNAMI_mongodb_exporter_REF := main
 
 BITNAMI_os_shell_PATH := 12/debian-12
-BITNAMI_os_shell_REF := $(BITNAMI_UPSTREAM_MAIN_REF)
+BITNAMI_os_shell_REF := main
```
delthas: These two refs are floating, which is a problem on its own — but it also points at the bigger issue with how the three prefixes were bootstrapped.

Context first: no subtree metadata exists for `solution-base/images/*` anywhere in 2.15's history, only the charts prefix has it —

```
$ git log --grep='git-subtree-dir' origin/development/2.15 | grep git-subtree-dir | sort -u
git-subtree-dir: solution-base/mongodb/charts/mongodb-sharded
```

so a bootstrap genuinely was required here, and the churn is content-neutral (net diff over `solution-base/images` is just this Makefile and `libmongodb.sh`).

The problem is *where* two of the three were bootstrapped. mongodb-exporter and os-shell were added at June-2026 `main` — the value of these very lines — and `c5abc5961f` then reverts their content back to the older vendored state: exporter `0.51.0`→`0.49.0`, os-shell yq `4.53.3`→`4.52.4`, scuttle `1.1.15`→`1.1.14`, `IMAGE_REVISION 68`→`61`, plus re-deleting upstream's `checksums/*.sha256`. The net diff stays zero, but the recorded baseline no longer matches the tree, so future syncs will either silently drop those upstream deltas or conflict on exactly those lines — with the upstream bump permanently misfiled as a Zenko customization.

The commits whose subdir content actually matches what we vendor, anchored on the `image.created` / `IMAGE_REVISION` values in our own Dockerfiles:

| Prefix | Bootstrap at | Upstream release |
|---|---|---|
| `mongodb-sharded/debian-12` | `657585595c550d4dc107a4e6cd3a598a9d284eec` | 8.0.13-debian-12-r1 (#86011) |
| `mongodb-exporter/debian-12` | `3af1e651db624f9938f407de85387bc7fe721d49` | 0.49.0-debian-12-r1 (#91364) |
| `os-shell/debian-12` | `f6aa42546a035ccefc7a2823674e622cc9ccaaba` | 12-debian-12-r61 (#91552) |

Verified at the object level: every `prebuildfs/**` and `rootfs/**` blob is byte-identical to what we vendor today, and those upstream trees contain no `checksums/*.sha256` — so those deletions aren't local tweaks either.

mongodb-sharded is already correctly baselined, so `3b033cb25e` can stay as-is. Related, and in its defence: the old pin `48a109547d39` is the commit that *deleted* `bitnami/mongodb-sharded/8.0/debian-12` upstream (replaced by a README pointing at 8.2), so the ref bump in `346d789711` is a genuine fix.

Redone at the right commits, `c5abc5961f` shrinks to the real delta — digest-pinned `minideb`, the `/bin/bash /usr/sbin/…` prefixes, sharded's `MONGODB_VERSION` arg and `fastdl.mongodb.org` binaries, the `docker-compose.yml` removal, and the stripped exec bits. Two hunks for os-shell, three for exporter. Restoring the 755 modes would additionally remove the need for the `/bin/bash` prefixes and the `RUN find … -exec chmod a+x` line.

And these two lines should be pinned to shas like `BITNAMI_mongodb_sharded_REF` above, otherwise the recorded split moves under us whenever upstream `main` advances.

solution-base/images/Makefile
```diff
@@ -25,15 +24,20 @@ vendor_branch = vendor/$1/$(call bitnami_path,$1)
...
 	@git remote get-url $(BITNAMI_REMOTE) >/dev/null 2>&1 || git remote add $(BITNAMI_REMOTE) $(BITNAMI_REPO)
 
-fetch-remote: create-remote
-	# Fetch full history from Bitnami main so subtree split can see full subtree history.
-	git fetch $(BITNAMI_REMOTE) main
-
-update-vendor-branch-%: fetch-remote
+update-vendor-branch-%: create-remote
+	# Shallow-fetch only the pinned ref; full upstream history is unnecessary for a --squash merge.
+	git fetch --depth 1 --filter=blob:none $(BITNAMI_REMOTE) $(call bitnami_ref,$*)
```
delthas, line 32: This is the root cause of the subtree trouble, and it's why the charts prefix needs none of this ceremony.

`solution-base/mongodb/Makefile:35` fetches **full** history deliberately:

```make
# Fetch full main + the target chart tag so `git subtree split` can synthesize
# the chart's history.
git fetch --no-tags $(BITNAMI_REMOTE) main
```

With full history, `git subtree split` is deterministic and *chained* — re-splitting a newer upstream ref regenerates the previous split commit as an ancestor of the new one, so the `git-subtree-split:` hash recorded in history always resolves locally. That's why `e0089dd84c` is a real incremental sync (`Squashed '…' changes from d2eb70018d..cd87843fdd`) with nothing pushed anywhere.

`--depth 1` breaks it: each shallow split is an unrelated single root, so the recorded hash can never be regenerated. `subtree merge --squash` re-resolves it through `process_subtree_split_trailer`, and since `cmd_merge` runs without a repository argument it cannot even attempt a fetch — the split exists only on a local `vendor/*` branch, and `git ls-remote --heads origin 'refs/heads/vendor/*'` is empty. Reproduced on a scratch repo:

```
$ git subtree merge --prefix=images/img vendor2 --squash
fatal: could not rev-parse split hash fc8ac3d4… from commit f06304e9…
```

So `make vendor-sync` fails on any other clone, and here too as soon as a pin moves — immediate for the two `main` refs above.

Fix is one line: drop `--depth 1`, keep `--filter=blob:none`. A blobless full-history fetch still regenerates the identical split commit and the merge resolves, so most of the speed win survives and this Makefile ends up matching the charts one. The ~4-minute split cost is already documented in `how_to_upgrade.md`.

delthas (commented):
Two threads on `solution-base/images/Makefile` covering the bootstrap baselines and the `vendor-sync` breakage.

solution-base/images/Makefile
delthas: Droppable — `git subtree split` operates on commit objects and works fine in a `--no-checkout` worktree. The `sparse-checkout set` + `checkout` pair only bounds lazy blob fetching under `--filter=blob:none`, and once `--depth 1` goes away (thread on line 29) the whole worktree dance can probably go with it, since the charts Makefile splits directly from the toplevel.

On the other suggestion — `sha=$(call bitnami_ref,$*)` instead of `git rev-parse FETCH_HEAD` — that doesn't work as written, because `bitnami_ref` can be a branch name (`main` today for two of the three images) rather than a sha. It works once all three refs are pinned to shas, which is worth doing regardless.

delthas:
Sorry I had to defer to Claude largely here for the comments because the subtree stuff is hard to explain easily

solution-base/images/Makefile
DarkIsDude: @delthas seems to go for the pin https://github.com/scality/Zenko/pull/2452#discussion_r3814249104 it you want to share your opinion

solution-base/images/Makefile
francoisferrand: 👍 for the makefile bug, but not clear to me where to go from there ( too long/complicated explications, and I'm lost in ordering of multiple threads...)

- with this fix, can we directly perform the update from bitnami upstream, or do we still need the "re-bootstrap" commits ? (= do we still need the "rebootstrap / merge / restore changes" commits)
- for the SHA1 : it should actually be embedded in git history (i.e. the git (subtree) merge), and the only moment where we run the makefile target is to merge upstream...i.e. pick up newest changes. My understanding is thus that these command are not about being reproducible (except when upstream did not change), but really about doing the upgrade / upstream sync. Is this it, or am I missing something?

DarkIsDude: Fixed, but not by dropping --depth 1 — I couldn't reproduce it as an ancestry problem. Two independent --depth 1 splits give unrelated roots and merge --squash still works: it resolves the recorded trailer as an object and diffs two trees, never consulting ancestry.
The real cause is reachability — the vendor/* branches were local-only, which is the empty ls-remote you saw. So update-vendor-branch-% now pushes each split to refs/heads/vendor-baseline/<image>/<sha> and vendor-sync-% fetches them before merging. All three baselines are published and match the trailers in history.
Keeping --depth 1 for cost: subtree split doesn't restrict its rev-list to the prefix, so full history walks all 485k commits of the containers monorepo. Happy to switch if you think unrelated roots are a problem in themselves.

Yes, it's the goal of the fix to be able to run it without redoing the bootstrap (the goal is to do what was the first goal that was not done finally).

For the SHA1 the goal of the makefile is mainly to update, but indeed not reproductible. I changed the makefile to make it reproductible 🙏

solution-base/images/Makefile
DarkIsDude: Dropped

solution-base/images/Makefile
DarkIsDude: Re-bootstrapped both prefixes at the shas from your table, and confirmed the mismatch at tree level first: exporter was recorded at 0.51.0/rev 6 against 0.49.0/rev 1 vendored, os-shell at yq 4.53.3/rev 68 against 4.52.4/rev 61. Sharded already matched, so it stayed as-is.

All three recorded baselines now match their pinned refs, and the restore commit shrank from 30 files (+101/-100) to 17 (+77/-18) — no more deleting upstream checksums, no more reverting version bumps. Real delta came out at exporter +3/-3 and os-shell +2/-2, as you said it would. Final tree is byte-identical to before the rewrite.

Also took the last paragraph: restored the 755 modes on all 13 files, which let the four `/bin/bash /usr/sbin/…` prefixes, the `/bin/bash …/postunpack.sh` one and the `RUN find … -exec chmod a+x` line go. Checked that the chmod line was really redundant first — the four scripts needing +x are exactly the four at 755 upstream, the three sourced libraries stay 644. Exporter and os-shell now differ from upstream by a single line each, the minideb digest pin.

solution-base/images/mongodb-exporter/debian-12/prebuildfs/usr/sbin/install_packages
francoisferrand, line 1: are all the permission changes expected?

(i.e. change in upstream associated with the `find /opt/bitnami/scripts -name "*.sh" -exec chmod a+x {} +` removal?)

solution-base/images/Makefile
francoisferrand: - I think @benzekrimaha did the initial bootstrap : if she still has the branch (locally), she could just push it so we don't need the bootstrap?
- I changed the makefile to make it reproductible : I don't understand the benefit, especially when it contradicts the goal of the makefile (i.e. to update the charts). Can you please elaborate?

solution-base/images/Makefile
```diff
@@ -1,21 +1,18 @@
...
-BITNAMI_mongodb_sharded_REF := 48a109547d39cd8cf8a5d4058d832ecb5844829e
+BITNAMI_mongodb_sharded_REF := 657585595c550d4dc107a4e6cd3a598a9d284eec
 
 BITNAMI_mongodb_exporter_PATH := 0/debian-12
-BITNAMI_mongodb_exporter_REF := $(BITNAMI_UPSTREAM_MAIN_REF)
+BITNAMI_mongodb_exporter_REF := 3af1e651db624f9938f407de85387bc7fe721d49
 
 BITNAMI_os_shell_PATH := 12/debian-12
-BITNAMI_os_shell_REF := $(BITNAMI_UPSTREAM_MAIN_REF)
+BITNAMI_os_shell_REF := f6aa42546a035ccefc7a2823674e622cc9ccaaba
```
francoisferrand, line 13: since you now stick the SHA1 ("_changed the makefile to make it reproductible_"), what is now the process to _bump_ ? (i.e. fetch upstream, update the SHA1 in makefile, ...)
→ shoud update the docs/readme

solution-base/images/mongodb-sharded/debian-12/rootfs/opt/bitnami/scripts/libmongodb.sh
francoisferrand: - It's not just purely cosmetic : it means a pod restarts, which delays "deployment" and increases the install duration (since installer waits).
- logic that is specific to Mongo : deploying mongo is specific to mongo. But this limit has not changed in many years, I don't expect it to change, and anyway the correct way requires to handle the error instead of just pre-filtering

Please create a followup BUGFIX ticket (in Zenko) to document the problem, and we may pick it up (or not) later.

francoisferrand (approved):
there is no rush there, should it not go in 2.16 instead?

solution-base/images/mongodb-exporter/debian-12/prebuildfs/usr/sbin/install_packages
DarkIsDude: Yes, comes from the upstream. The initial bootstrap (I missed it) changed that on our side only and apply some fix... It was an error, the goal is to fix it and be as much as possible closer from upstream. Spotted by Thomas 🙏 in a previous comment.

solution-base/images/mongodb-sharded/debian-12/rootfs/opt/bitnami/scripts/libmongodb.sh
delthas: I see this is resolved. What's the followup ticket?

solution-base/images/Makefile
delthas: I'd say update the refs, then run `make vendor-sync`, similarly to charts where we bump `CHART_VERSION` and run `make`?

DarkIsDude: indeed, thanks 🙏

solution-base/images/Makefile
DarkIsDude: Good question. The second one is maybe the easiest one. To update the chart, we now need to change the ref that we want to use. Before the makefile was not working at all (with the same sha for mongo or not). This is what I means by "reproductible".

The second one is maybe more complicated. From my understanding, the first bootstrap was not done with a substree but with a classic git commit. I checked with Maha when I first started and even on her side the makefile was not working anymore. So even with an another branch or tree, the bootstrap is still needed to retrieve everything.

solution-base/images/mongodb-sharded/debian-12/rootfs/opt/bitnami/scripts/libmongodb.sh
delthas: Bump

delthas (approved):
👍

maeldonn (approved):

PR #2456, opened 2026-06-30
Log cronjob pod diagnostics on failure/hang in CTST

delthas:
## What

Add best-effort pod diagnostics + a heartbeat to `createJobAndWaitForCompletion` (`tests/functional/ctst/steps/utils/kubernetes.ts`), the shared helper that triggers and waits for cronjobs (count-items) in quota / utilization / storage-usage CTST scenarios.

## Why

When the triggered job fails or hangs, the step previously logged only `"Job failed"` with the job object — never *why* (no pod logs, exit code, or `BackoffLimitExceeded` reason). The count-items pod logs lived only in the kind-logs artifact, which on a cancelled/hung run is frequently truncated or never uploaded, leaving the failure undiagnosable from the run itself. The watch also had no heartbeat, so a stalled job produced dead air until the 20-min Before-hook timeout.

## Changes (all best-effort, wrapped so logging never breaks the test)

- **Heartbeat** — every 30 s while waiting, re-read the Job and log its real status (active/succeeded/failed, conditions) + pod state, so a silent stall is visible instead of dead air.
- **Pod diagnostics on failure** — on job failure (and once on first detected container crash/restart), list the job's pods and log their status + container logs (current + previous), so the crash reason lands directly in the GitHub step log.
- **Lock visibility** — log lock acquisition / cross-worker contention at info/warn.

CTST already runs at debug log level in CI, so these surface in the step log with no artifact dependency.

`tsc --build` and `eslint` pass on the changed file.

Issue: ZENKO-5308

### DIFF
--- a/tests/functional/ctst/steps/utils/kubernetes.ts
+++ b/tests/functional/ctst/steps/utils/kubernetes.ts
@@ -102,6 +102,84 @@ export async function createJobAndWaitForCompletion(
 
     const lockFilePath = path.join('/tmp', `${jobName}.lock`);
 
+    let expectedJobName: string | undefined;
+    let heartbeat: ReturnType<typeof setInterval> | undefined;
+    let dumpedPodLogs = false;
+    let watchStart = Date.now();
+
+    // Best-effort: log the job's status plus its pods' status, and the pod logs
+    // when a container has failed or restarted, so the reason a cronjob run failed
+    // or hung is visible in the test step log itself rather than only in artifacts.
+    const logJobDiagnostics = async (reason: string, forceLogs = false): Promise<void> => {
+        if (!expectedJobName) {
+            return;
+        }
+        try {
+            const current = await batchClient.readNamespacedJob({
+                name: expectedJobName,
+                namespace: 'default',
+            });
+            world.logger.info('cronjob status', {
+                jobName,
+                reason,
+                instance: expectedJobName,
+                elapsedSec: Math.round((Date.now() - watchStart) / 1000),
+                active: current.status?.active,
+                succeeded: current.status?.succeeded,
+                failed: current.status?.failed,
+                conditions: current.status?.conditions,
+            });
+        } catch (e) {
+            world.logger.warn('Failed to read cronjob status', { jobName, e });
+        }
+        try {
+            const coreClient = createKubeCoreClient(world);
+            const pods = await coreClient.listNamespacedPod({
+                namespace: 'default',
+                labelSelector: `batch.kubernetes.io/job-name=${expectedJobName}`,
+            });
+            for (const pod of pods.items || []) {
+                const podName = pod.metadata?.name;
+                const containerStatuses = pod.status?.containerStatuses || [];
+                const restarted = containerStatuses.some(s => (s.restartCount || 0) > 0);
+                const crashed = restarted
+                    || containerStatuses.some(s => s.state?.terminated || s.lastState?.terminated);
+                world.logger.warn('cronjob pod status', {
+                    jobName,
+                    reason,
+                    podName,
+                    phase: pod.status?.phase,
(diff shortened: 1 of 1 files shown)


delthas:
This additional debug logging helped find the s3utils issue.

tests/functional/ctst/steps/utils/kubernetes.ts
```diff
@@ -117,13 +176,29 @@ export async function createJobAndWaitForCompletion(
         }
         tries--;
         if (!lockAquired) {
+            if ((600 - tries) % 15 === 0) {
```
SylvainSenechal: isnt (tries % 15) === 0 the same ?

tests/functional/ctst/steps/utils/kubernetes.ts
```diff
@@ -117,13 +176,29 @@ export async function createJobAndWaitForCompletion(
...
+                    waitedSec: Math.round((Date.now() - lockStart) / 1000),
+                });
+            }
             await Utils.sleep(1000);
         }
     }
 
-    try {
-        world.logger.debug(`Acquired lock for job: ${jobName}`);
+    if (lockAquired) {
```
SylvainSenechal: If I read correctly

- When we fail to acquire the lock, we still run the cronjob
- This was already the case before but we didn't log anything when we couldn't acquire the log

Is it reasonable to run the cronjob if we can't acquire the lock ? or should the test be failed in that case ?

tests/functional/ctst/steps/utils/kubernetes.ts
```diff
@@ -146,10 +221,37 @@ export async function createJobAndWaitForCompletion(
...
+                        });
+                        world.logger.info('Still waiting for cronjob to complete', {
+                            jobName,
+                            instance: expectedJobName,
+                            elapsedSec: Math.round((Date.now() - watchStart) / 1000),
+                            active: current.status?.active,
+                            succeeded: current.status?.succeeded,
+                            failed: current.status?.failed,
+                            conditions: current.status?.conditions,
+                        });
```
SylvainSenechal: shouldn't this be grouped with collectJobPodDiagnostic ?

SylvainSenechal (approved):

tests/functional/ctst/steps/utils/kubernetes.ts
SylvainSenechal: or maybe not on first iteration

tests/functional/ctst/steps/utils/kubernetes.ts
delthas: Yes you read correctly. Thinking about it.

tests/functional/ctst/steps/utils/kubernetes.ts
delthas: Done — switched to `tries % 15 === 0` (equivalent since 600 is a multiple of 15, and it doesn't fire on the first iteration either).

tests/functional/ctst/steps/utils/kubernetes.ts
delthas: Good call — changed it to throw/fail instead of silently proceeding when the lock can't be acquired after ~10 min, so a stuck or leaked lock surfaces as a hard failure rather than double-running count-items. (It's a per-runner file lock serializing count-items across the parallel cucumber workers; with the s3utils/arsenal fix count-items is fast again, so contention should be rare.)

tests/functional/ctst/steps/utils/kubernetes.ts
delthas: Done — folded the job-status read+log into the diagnostics helper (now `logJobDiagnostics`), so the heartbeat and the failure path both call one function that logs job status + pod status/logs.

benzekrimaha (approved):

PR #2457, opened 2026-06-30
Stop CTST gracefully at a cucumber-level time budget

delthas:
## What

Add an in-process **time budget** so CTST stops gracefully before the GitHub step timeout would hard-kill it. Touches only `tests/functional/ctst/common/hooks.ts` and `.github/scripts/end2end/run-e2e-ctst.sh`.

## Why

The only stop for a hung `ctst-end2end-sharded` run is the step `timeout-minutes`, which **hard-kills** the process. cucumber-js v13 installs no signal handler and has no whole-run timeout, so on a kill it never reaches `testRunFinished` — `report.xml` is written empty (`Error parsing report.xml: no element found`), teardown is skipped, and the archive step has nothing to publish.

## How

- `run-e2e-ctst.sh` exports `CTST_DEADLINE_EPOCH_MS` (default 180 min, via `CTST_MAX_RUNTIME_MIN`) and a marker path, and clears any stale marker.
- An early `Before` hook (registered before the expensive `@Quotas`/count-items hooks) skips any scenario that starts past the deadline — returning `'skipped'` short-circuits the remaining hooks/steps — and drops the marker file.
- cucumber then finishes normally: it writes `report.xml`/`report.html`/`report.ndjson` and runs teardown. The runner script exits 1 if the marker is present, so a timed-out run is still red while keeping a clean report.

When `CTST_DEADLINE_EPOCH_MS` is unset (local runs), the hook is a no-op.

## Relationship to ZENKO-5306

Companion to the step-level timeout. ZENKO-5306 lowers the step `timeout-minutes` to 190 as a hard backstop: 180 min (graceful, this PR) < 190 min (hard backstop) < the old 360-min job cap.

`tsc --build`, `eslint`, and `bash -n` pass.

Issue: ZENKO-5309

### DIFF
--- a/tests/functional/ctst/common/hooks.ts
+++ b/tests/functional/ctst/common/hooks.ts
@@ -53,6 +53,19 @@ AfterAll(async function () {
     await stopDLQConsumer();
 });
 
+// Fail any scenario that starts past the global time budget, so the run stops
+// (and writes its reports) before the step timeout hard-kills it.
+const ctstMaxRuntimeMs = 180 * 60 * 1000;
+const ctstStartMs = Date.now();
+Before((scenario: ITestCaseHookParameter) => {
+    if (Date.now() > ctstStartMs + ctstMaxRuntimeMs) {
+        throw new Error(
+            `CTST global time budget (${ctstMaxRuntimeMs / 60000} min) exceeded; failing scenario `
+            + `"${scenario.pickle.name}" (started too late to run within the budget)`,
+        );
+    }
+});
+
 Before(async function (this: Zenko, scenario: ITestCaseHookParameter) {
     this.resetSaved();
     Identity.resetIdentity();


.github/scripts/end2end/run-e2e-ctst.sh
```diff
@@ -23,6 +23,17 @@ cd "$FUNCTIONAL_TESTS_DIR"
 mkdir -p ctst/reports
 
 export SDK=true # Cli-testing also has a cli mode, not really used in practice
+
+# Gracefully stop before the GitHub step timeout hard-kills the process:
+# scenarios that start after this deadline are skipped (see the time-budget
+# Before hook), so Cucumber finishes normally and writes its reports. The hook
+# drops a marker file; we fail the run if it is present.
+CTST_MAX_RUNTIME_MIN=${CTST_MAX_RUNTIME_MIN:-180}
+export CTST_DEADLINE_EPOCH_MS=$(( $(date +%s%3N) + CTST_MAX_RUNTIME_MIN * 60 * 1000 ))
```
francoisferrand: best to minimize content of wrapper script : should do this computation in javascript

```suggestion
export CTST_MAX_RUNTIME_MIN=${CTST_MAX_RUNTIME_MIN:-180}
```

.github/scripts/end2end/run-e2e-ctst.sh
```diff
@@ -33,4 +44,13 @@ yarn cucumber-js \
     --format pretty \
     --format html:ctst/reports/report.html \
     --format junit:ctst/reports/report.xml \
-    --format message:ctst/reports/report.ndjson
+    --format message:ctst/reports/report.ndjson || rc=$?
+
+if [ -f "$CTST_TIMEOUT_MARKER" ]; then
+    echo "::error::CTST exceeded its ${CTST_MAX_RUNTIME_MIN}-minute time budget; remaining scenarios were skipped."
```
francoisferrand: do we need to display a log *here* ?
we could display the same from within CTST (first time -or everytime- we hit the deadline...), and avoid the whole "marker" handling: which seems overly complex

tests/functional/ctst/common/hooks.ts
```diff
@@ -53,6 +54,23 @@ AfterAll(async function () {
...
+Before(() => {
+    const deadline = Number(process.env.CTST_DEADLINE_EPOCH_MS);
+    if (deadline && Date.now() > deadline) {
+        const marker = process.env.CTST_TIMEOUT_MARKER;
+        if (marker) {
+            try {
+                fs.writeFileSync(marker, 'timeout');
+            } catch { /* best-effort: marker is advisory */ }
+        }
+        return 'skipped';
```
francoisferrand: what does this "skipped" do?
these (skipped) tests need to be marked as _somewhat failed_, to ensure we don't pass CI unexpectedly: maybe "skipped" does that, but it must a skip like "I could not execute it" (=potentially failed) rather than "test was intentionally disabled/skipped" (=ignore)

.github/scripts/end2end/run-e2e-ctst.sh
```diff
@@ -33,4 +44,13 @@ yarn cucumber-js \
...
+    if [ "$rc" -eq 0 ]; then
+        rc=1
```
francoisferrand: instead of wrapper script, there is already some code to "tweak" the return code (and make it pass always), c.f. tests/functional/ctst/cucumber.config.cjs :

```
if (process.env.CI_PASS_ON_TEST_FAILURE === 'true') {
    const _exit = process.exit;
    process.exit = function exit(code) {
        _exit(code === 1 ? 0 : code);
    };
    process.on('beforeExit', () => {
        if (process.exitCode === 1) {
            process.exitCode = 0;
        }
    });
}
```

→ instead of this, you could thus tweak the result to actually return something other than 0 if we want to skip further analysis (e.g. test results merging...)
→ however, why do we need to return error 1 in this case? Timeout is likely an issue in the tests, shouldn't we handle this like all other failed tests (i.e. ignore the result and let followup steps check the report for failure)

.github/scripts/end2end/run-e2e-ctst.sh
francoisferrand: even CTST_MAX_RUNTIME_MIN default value could be set directly in the javascript code?

francoisferrand (changes requested):

tests/functional/ctst/common/hooks.ts
SylvainSenechal: I think it's an official cucumber keyword, but yeah this should probably use fail instead 🤔

.github/scripts/end2end/run-e2e-ctst.sh
SylvainSenechal: no need imo, I would display a log in the hook, for each test skipped, we can access the name of the scenario and just write "scenario xxx skipped because of global timeout"

tests/functional/ctst/common/hooks.ts
```diff
@@ -53,6 +54,23 @@ AfterAll(async function () {
     await stopDLQConsumer();
 });
 
+// Time budget: scenarios that start after the deadline are skipped so Cucumber
+// finishes normally (writing its reports) before the GitHub step timeout would
+// hard-kill the process. The marker file lets the runner script fail the run.
+Before(() => {
```
SylvainSenechal: I would simplify this : 
- hardcode here max duration allowed for all ctst tests
- send as env variable a timestamp of when we started running ctst (+ verify if maybe this variable is already exposed by cucumber)  
- just verify time.now < timestampStart + maxDurationAllowed

SylvainSenechal (changes requested):

.github/scripts/end2end/run-e2e-ctst.sh
delthas: Done — the deadline math and the default (180 min) now live in the hook (JS); the wrapper only exports a single start timestamp (`CTST_START_EPOCH_MS`).

.github/scripts/end2end/run-e2e-ctst.sh
delthas: Dropped the marker entirely. The hook now throws per over-budget scenario with the scenario name in the message, so each one is reported/logged individually.

tests/functional/ctst/common/hooks.ts
delthas: Switched from `"skipped"` to throwing (fail), so these are real failures — CI won't pass unexpectedly, and they surface via the junit report like any other failure.

.github/scripts/end2end/run-e2e-ctst.sh
delthas: Good point — removed the forced `exit 1` and the marker. Over-budget scenarios now just fail in the report and go through the normal report analysis + `CI_PASS_ON_TEST_FAILURE` handling like any other test failure; the wrapper no longer touches the exit code.

tests/functional/ctst/common/hooks.ts
delthas: Done, exactly this: max duration (default 180 min) is hardcoded in the hook (overridable via `CTST_MAX_RUNTIME_MIN`), `run-e2e-ctst.sh` exports `CTST_START_EPOCH_MS` once (shared by all parallel workers), and the hook checks `now > start + max`.

delthas: Follow-up: dropped the env override and just hardcoded the 180-min budget in the hook — it was a new var nothing else used, so it's a plain constant now, as you suggested.

delthas:
Largely simplified based on your suggestions

.github/scripts/end2end/run-e2e-ctst.sh
```diff
@@ -23,6 +23,10 @@ cd "$FUNCTIONAL_TESTS_DIR"
 mkdir -p ctst/reports
 
 export SDK=true # Cli-testing also has a cli mode, not really used in practice
+
+# Shared start time for the global CTST time budget (enforced in the Before hook).
+export CTST_START_EPOCH_MS=$(date +%s%3N)
```
francoisferrand: can't we set this in javascript as well? e.g. a global `const`, just like `ctstMaxRuntimeMs` ?

```suggestion
const ctstStartMs = new Date();
```

and if cucmber-js forks the process, this could be done in the first process:

```
const ctstStartMs = Number(process.env.CTST_START_EPOCH_MS) || new Date():
process.env.CTST_START_EPOCH_MS = ctstStartMs; // Export for forked processes
```

francoisferrand (approved):

.github/scripts/end2end/run-e2e-ctst.sh
delthas: Done — captured it in JS (`const ctstStartMs = Date.now()`) at hook load and dropped the shell export entirely, so run-e2e-ctst.sh is no longer touched. I went with the plain module-load const rather than the env-propagation variant: under `--parallel` each worker captures its own start, but workers spawn within ~1-2s of each other, so the per-worker skew is negligible against the 180-min budget.

SylvainSenechal (approved):

PR #2461, opened 2026-07-08
Add CTST coverage for Days=0 lifecycle expiration

delthas:

Adds end-to-end (CTST) coverage for S3 lifecycle Expiration rules with `Days=0` ("delete all matching objects as soon as backbeat runs"), and bumps the component images that ship the feature.

- Bumps `solution/deps.yaml`: backbeat `9.5.0-preview.1` → `9.5.0-preview.3` (BB-791), cloudserver `9.4.0-preview.2` → `9.4.0-preview.4` (CLDSRV-928, arsenal 8.5.6), pensieve-api `1.10.1` → `1.11.0-preview.1` (PSVAPI-128).
- Adds `features/lifecycleExpiration.feature` with two `@PreMerge` scenarios using the direct S3 `PutBucketLifecycleConfiguration` API:
  - non-versioned bucket + `Expiration Days=0` → all objects deleted;
  - versioned bucket + `Expiration Days=0` + `NoncurrentVersionExpiration NoncurrentDays=0` → all versions and delete markers removed.
- Factors the shared lifecycle-config PUT-with-retry logic out of the transition workflow helper into `putBucketLifecycleConfigurationWithRetry`, and adds an `addExpirationWorkflow` helper.

`Days=0` makes objects immediately eligible (backbeat lists with no `BeforeDate`), so the test needs no time-progression or one-day-earlier tricks and is not tagged `@Flaky`.

The backbeat index-skip heuristic (BB-779) is intentionally out of scope — it is not yet merged and is covered by backbeat unit tests, not e2e.

ZENKO-5314

The new scenarios follow `tests/functional/ctst/HOW_TO_WRITE_TESTS.md`: unique per-scenario buckets, external-state-only assertions (via `ListObjectVersions`), poll-based assertions (idempotent after waits), automatic teardown via the global `After` hook, and no `@Flaky`.

Issue: ZENKO-5314

### DIFF
--- a/tests/functional/ctst/features/lifecycleExpiration.feature
+++ b/tests/functional/ctst/features/lifecycleExpiration.feature
@@ -1,0 +1,17 @@
+Feature: Lifecycle expiration
+
+    @2.16.0
+    @PreMerge
+    @Expiration
+    @Lifecycle
+    Scenario Outline: Days=0 expiration empties a "<versioningConfiguration>" bucket
+    Given a "<versioningConfiguration>" bucket
+    And 5 objects "expire-obj" of size 100 bytes
+    When i set a lifecycle expiration of 0 days for the "<scope>"
+    Then the bucket should contain 0 objects within 180 seconds
+
+    Examples:
+        | versioningConfiguration | scope                           |
+        |           Non versioned | current version                 |
+        |               Versioned | current and noncurrent versions |
+        |               Suspended | current and noncurrent versions |
--- a/solution/deps.yaml
+++ b/solution/deps.yaml
@@ -6,7 +6,7 @@ backbeat:
   dashboard: backbeat/backbeat-dashboards
   image: backbeat
   policy: backbeat/backbeat-policies
-  tag: 9.5.0-preview.4
+  tag: 9.5.0-preview.7
   envsubst: BACKBEAT_TAG
 busybox:
   image: busybox
--- a/solution/deps.yaml
+++ b/solution/deps.yaml
@@ -6,7 +6,7 @@ backbeat:
   dashboard: backbeat/backbeat-dashboards
   image: backbeat
   policy: backbeat/backbeat-policies
-  tag: 9.5.0-preview.1
+  tag: 9.5.0-preview.4
   envsubst: BACKBEAT_TAG
 busybox:
   image: busybox
@@ -16,7 +16,7 @@ cloudserver:
   sourceRegistry: ghcr.io/scality
   dashboard: cloudserver/cloudserver-dashboards
   image: cloudserver
-  tag: 9.4.0-preview.2
+  tag: 9.4.0-preview.4
   envsubst: CLOUDSERVER_TAG
 drctl:
   sourceRegistry: ghcr.io/scality
@@ -80,7 +80,7 @@ mongodb-connector:
 pensieve-api:
   sourceRegistry: ghcr.io/scality
   image: pensieve-api
-  tag: 1.10.1
+  tag: 1.11.0-preview.1
   envsubst: PENSIEVE_API_TAG
 rclone:
   sourceRegistry: rclone
--- a/tests/functional/ctst/steps/lifecycleExpiration.ts
+++ b/tests/functional/ctst/steps/lifecycleExpiration.ts
@@ -1,0 +1,39 @@
+import { Then, When } from '@cucumber/cucumber';
+import { S3, Utils } from 'cli-testing';
+import { ListObjectVersionsOutput } from '@aws-sdk/client-s3';
+import assert from 'assert';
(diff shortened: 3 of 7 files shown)


delthas:
CI failing due to cloudserver bump introducing unrelated issue, tracked in https://github.com/scality/cli-testing/pull/98. Please review regardless! :) (The introduced tests pass.)

tests/functional/ctst/features/lifecycleExpiration.feature
```diff
@@ -0,0 +1,21 @@
...
+
+    @2.16.0
+    @PreMerge
+    @Expiration
+    @Lifecycle
+    Scenario: Days=0 expiration deletes all objects in a non-versioned bucket
+    Given a "Non versioned" bucket
+    And 5 objects "expire-obj" of size 100 bytes
+    When i set a lifecycle expiration of 0 days
+    Then the bucket should contain 0 objects within 120 seconds
```
DarkIsDude: Where does comes from 120 and 180 ?

DarkIsDude (approved):
To be honest it's clear 🙏. Thanks !

tests/functional/ctst/features/lifecycleExpiration.feature
delthas: These are poll deadlines for backbeat to finish expiring (the step returns as soon as the bucket is empty, it doesn't wait the full time). They're sized to the lifecycle conductor cadence × the number of passes each case needs, plus margin for CI variance:

- **120s (non-versioned):** a single pass deletes the current objects.
- **180s (versioned):** up to 3 passes — `Expiration Days=0` creates delete markers, `NoncurrentVersionExpiration NoncurrentDays=0` removes the now-noncurrent versions, then the orphan delete markers are cleaned.

In the passing CI run they actually completed in ~60s / ~70s (≈133s for the versioned case on a real Artesca cluster), so the values leave roughly 2× headroom.

tests/functional/ctst/features/lifecycleExpiration.feature
```diff
@@ -0,0 +1,21 @@
...
+
+    @2.16.0
+    @PreMerge
+    @Expiration
+    @Lifecycle
+    Scenario: Days=0 expiration deletes all versions and delete markers in a versioned bucket
+    Given a "Versioned" bucket
```
francoisferrand: nit:
- should we also test version-suspended bucket?
- can we/should we have a single scenario with different parameter (non versioned, versioned, version-suspended)? may be complicated to merge the setup/when steps though...

francoisferrand (approved):

tests/functional/ctst/features/lifecycleExpiration.feature
delthas: Done in ce20f6436c — collapsed into a single `Scenario Outline` over **Non versioned / Versioned / Suspended**, with the rule scope (current version vs current and noncurrent versions) as an `Examples` column, so the setup/when steps stay unified. Added the suspended-bucket coverage you suggested, and validated all three modes empty within the 180s deadline on a real cluster (suspended in ~90s).

delthas:

delthas:
CI investigation summary (run 29758508050, all 8 attempts failed): the three deterministic failures have **two distinct root causes**, both introduced by the deps bumps in this PR.

**test_iam_policies (end2end-2-shards-http)** — cloudserver >= 9.4.0-preview.3 (CLDSRV-932, scality/cloudserver#6208) now validates `x-amz-content-sha256` against the actual request body. The mocha helper `makeUpdateRequest` never passed the body to `aws4.sign`, so it signed the empty-body hash while sending the real payload → all 10 RestoreObject IAM-policy tests failed `400 !== 403` (rejected before policy evaluation). Same class of bug as the cli-testing quota helper fixed in v1.3.1. Fixed by 423f8c98 on this branch.

**PRA nominal case (end2end-pra) + Kafka Cleaner (ctst-end2end-sharded)** — backbeat >= 9.5.0-preview.2 (BB-740) ships lifecycle alert threshold defaults like `10*60` that render as `{{ 10*60 | humanizeDuration }}` in alert annotations — an invalid Go template (`unexpected "*" in operand`). prometheus-operator rejects the rule group and **aborts the entire Prometheus config sync**, so the PRA federation scrape config and the kafka-cleaner ServiceMonitor never reach Prometheus. kafka-cleaner actually works fine (broker logs show DeleteRecords every minute) — its metrics are just never scraped; platform alerting is also silently dead for the whole run. Tracked in [BB-824](https://scality.atlassian.net/browse/BB-824), fix in scality/backbeat#2778. This PR will need a backbeat bump to 9.5.0-preview.7 once released — until then end2end-pra and the Kafka Cleaner scenario will keep failing deterministically, retries won't help.

PR #2462, opened 2026-07-09
fix(mongodb): lower indexBuildMinAvailableDiskSpaceMB to 5 GiB

delthas:
## Summary

Lower `MONGODB_INDEX_BUILD_MIN_DISK_SPACE_MB` from `10000` to `5000` in `solution-base/build.sh`.

#2436 (ZENKO-5295) added `--setParameter indexBuildMinAvailableDiskSpaceMB=10000` to config servers and shard data servers. MongoDB reads this value as MiB, so it requires `10000 * 1048576` = ~9.77 GiB of **free** space before it will start any index build.

That's fine on every production SKU (the smallest Mongo data volume across all installer profiles is 119Gi), but the `demo` / `demolvm` profiles provision only a **10Gi** data volume (**8Gi** config). After ext4 overhead a fresh volume has ~9.5 GiB / ~7.5 GiB free — below the 9.77 GiB threshold — so on fresh demo/CI/AWS installs mongod refuses **every** `createIndexes`, the `artesca-data-ops-vault-create-indexes` job crash-loops, and the `MongoDbCreateIndexesFailed` alert fires.

`5000` (~4.88 GiB) fits under both demo volumes while keeping meaningful headroom on all production SKUs.

## Related

- #2436 / ZENKO-5295 — introduced the 10 GiB threshold
- RD-2017 — reported the symptom (CrashLoopBackOff + MongoDbCreateIndexesFailed)

Issue: ZENKO-5315


solution-base/build.sh
```diff
@@ -63,7 +63,10 @@ MONGODB_SHARDSERVER_RAM_LIMIT="MONGODB_SHARDSERVER_RAM_LIMIT"
 MONGODB_SHARDSERVER_EXTRA_FLAGS="MONGODB_SHARDSERVER_EXTRA_FLAGS"
 MONGODB_SHARDSERVER_RAM_REQUEST="MONGODB_SHARDSERVER_RAM_REQUEST"
 MONGODB_MONGOS_RAM_REQUEST="MONGODB_MONGOS_RAM_REQUEST"
-MONGODB_INDEX_BUILD_MIN_DISK_SPACE_MB=10000
+# Must stay below the free space of the smallest provisioned Mongo volume
+# (the demo/demolvm profiles use a 10Gi data / 8Gi config volume, ~9.5/7.5 GiB
+# free after fs overhead), otherwise mongod refuses every index build.
```
DarkIsDude:
```suggestion
# Demoprofiles use a 10Gi data, must be below
```

DarkIsDude (approved):

benzekrimaha (approved):

solution-base/build.sh
francoisferrand: "demo profile" is not a concept in zenko (OSS), should not be mentioned...

PR #2463, opened 2026-07-13
Set default memory resources for the ZooKeeper quorum

delthas:

Bumps zenko-operator to v1.8.14 and sets `defaults.zookeeperResources` in `solution/zenkoversion.yaml`, alongside the existing `defaults.kafkaResources`. Without it the quorum pods run BestEffort and the zookeeper image sizes its JVM heap as a percentage of the host memory instead of the container limit (RD-1962 finding, confirmed fleet-wide on ~240 sosreports).

Fixes ZENKO-5316.

zenko-operator v1.8.14 (scality/zenko-operator#624) is what applies the field to the ZookeeperCluster; the bump commit lands before the field is set. Sizing rationale and fleet data: see the review thread.

Issue: ZENKO-5316

### DIFF
--- a/solution/zenkoversion.yaml
+++ b/solution/zenkoversion.yaml
@@ -138,6 +138,9 @@ spec:
         limitCPU: 3
       cruiseControl:
         limitMemory: 3Gi
+    zookeeperResources:
+      requestMemory: 1Gi
+      limitMemory: 1Gi
   featureFlags:
     lifecycleOptimization: true
     manageKafkaTopics: true
--- a/solution/deps.yaml
+++ b/solution/deps.yaml
@@ -131,7 +131,7 @@ zenko-operator:
   sourceRegistry: ghcr.io/scality
   dashboard: zenko-operator/zenko-operator-dashboards
   image: zenko-operator
-  tag: v1.8.13
+  tag: v1.8.14
   envsubst: ZENKO_OPERATOR_TAG
 zookeeper:
   sourceRegistry: ghcr.io/adobe/zookeeper-operator


solution/zenkoversion.yaml
```diff
@@ -138,6 +138,9 @@ spec:
         limitCPU: 3
       cruiseControl:
         limitMemory: 3Gi
+    zookeeperResources:
+      requestMemory: 2Gi
+      limitMemory: 2Gi
```
francoisferrand: zookeeper is used for kafka quorum, but also internally for many things in backbeat (depending on the features) : as cache, status....
→ are we sure about these limits?

francoisferrand:
is this really the issue we had in production? If Zookeeper were actually "easting" the memory, it would have affected the system already (i.e. prevented others from using memory); or is there something else causing this to happen _now_ ?

delthas:
Requested @DarkIsDude in place of Maha Benzekri, who is currently on PTO.

solution/zenkoversion.yaml
delthas: Fair challenge — we grounded the value with data rather than picking it by feel. Consolidated analysis:

## Method

Scality sosreports embed a 24h hourly snapshot of the platform's Prometheus (`sos_commands/metalk8s/metrics/`, one JSON file per metric). We extracted ZooKeeper memory from:

- the three RD-1962 sosreports (4.2.2 + 4.3.0-rc.1 AWS pair, plus the 4.3.0 HW lab **under perf load**), and
- the fleet sosreport inventory: **~240 distinct clusters** (customers + QA, ARTESCA 2.x → 4.3, deduped by installation id, most recent report per cluster).

Per quorum pod: `jvm_memory_bytes_committed/max{area="heap"}` (what the heap flags resolved to), `jvm_memory_pool_bytes_used{pool="G1 Old Gen"}` (post-collection survivors ≈ live set), container RSS, znode count / data size.

## What the data shows

- **Where the problem comes from**: zenko 2.13.9 swapped `pravega/zookeeper:0.2.15` (stock fixed ~1 GB `-Xmx`) for the adobe 3.8.4 image whose zkEnv.sh sets `-XX:InitialRAMPercentage=30 -XX:MaxRAMPercentage=70` (ZENKO-5198 — originally a CI/JDK fix). Without a cgroup limit, those percentages resolve against **host** RAM. The fleet split is perfectly clean: every cluster on ≤ 2.13.5-1 commits ~1 GB flat; every cluster on ≥ 2.13.9 commits 30% of node RAM — up to **315 GB committed / 736 GB max heap** on 1 TB nodes, worst observed pod at **149.8 GB RSS**, for a coordination service holding ~64 KB of znodes.
- **What ZK actually keeps alive**: G1 Old Gen (survivors after collection) **never exceeds 15 MB on any cluster**, including the HW lab under bench load. GC works fine — the multi-GiB "heap used" readings are young-gen garbage piling up because a huge Eden rarely triggers a collection.
- **The proven baseline**: 207 clusters ran for years on the old fixed ~1 GB heap, zero known ZK OOMs. Per-pod RSS there: median 0.4 GB, p99 0.6 GB, all-time max 1.1 GB. Note that max is just the 1 GB heap fully touched plus JVM overhead — under any fixed heap, RSS converges to ~Xmx over uptime regardless of need, so observed RSS is an artifact of the old Xmx, not evidence of demand.

## Why 1Gi, and why it's safe

The honest sizing anchor is **live set + burst headroom** (request buffers: `jute.maxbuffer` 1 MB default × ~20 observed connections, snapshot serialization, watch fires — tens of MB at fleet scale), not observed RSS.

The image's RAMPercentage flags are container-aware, so with a 1Gi limit the same zkEnv line resolves to a ~717 MB max heap:

- ≈ **48×** the worst live set ever observed on any cluster;
- above everything the bounded fleet ever actually used (median pod touched 0.4 GB over years of production);
- ~300 MB of the limit left for non-heap (observed metaspace + buffers: ~40 MB);
- failure mode of undersizing is merely more frequent sub-millisecond young GCs — three orders of magnitude below ZK session timeouts;
- `requests == limits` makes the pods Guaranteed QoS and the quorum footprint visible to the scheduler (the original RD-1962 crash was a scheduling-headroom problem).

maeldonn (approved):

francoisferrand (approved):

francoisferrand:
> Where the problem comes from: zenko 2.13.9 swapped pravega/zookeeper:0.2.15 (stock fixed ~1 GB -Xmx) for the adobe 3.8.4 image whose zkEnv.sh sets -XX:InitialRAMPercentage=30 -XX:MaxRAMPercentage=70

this is the information which was missing! thanks!

delthas:

SylvainSenechal (approved):

PR #2469, opened 2026-07-27
Update cucumber to 13.2.0, move runOnceAcrossWorkers in Zenko

SylvainSenechal:
Issue: ZENKO-5328

Cucumber 13.2.0 introduce a new method to run beforeAll/afterAll hook a single time across all workers : 
- Use it instead of our custom runOnceAcrossWorkers method when possible
- When not possible, use runOnceAcrossWorkers which is now defined in Zenko instead of cli-testing


tests/functional/ctst/common/utils.ts
```diff
@@ -362,66 +359,39 @@ export async function prepareMetricsScenarios(
...
     } = options;
 
-    if (!fs.existsSync(filePath)) {
-        fs.writeFileSync(filePath, JSON.stringify({
-            ready: false,
-        }));
-    } else {
-        initiated = true;
-    }
+    await runOnceAcrossWorkers(
```
SylvainSenechal, line 362: Same result but instead of custom lockfile mechanism, use common method now to avoid having multiple ways to do the same thing.

Also I was able to remove proper-lockfile dependency thanks to this

tests/functional/ctst/common/WorkerCoordination.ts
SylvainSenechal, line 1: This file is a simplified copy of https://github.com/scality/cli-testing/blob/development/1.3/utils/WorkCoordination.ts

tests/functional/ctst/world/Zenko.ts
```diff
@@ -664,39 +663,13 @@ export default class Zenko extends World<ZenkoWorldParameters> {
...
-                            minTimeout: 1000,
-                            maxTimeout: 5000,
-                        }
-                    });
-
-                    try {
+                await runOnceAcrossWorkers(
+                    { lockName: `account-init-${accountName}`, logger: CacheHelper.logger },
+                    async () => {
                         await SuperAdmin.createAccount({ accountName });
```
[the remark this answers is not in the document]
SylvainSenechal: Ye, the old code was hammering concurrent requests, now we have a runOnceAcrossWorkers, so if we end up with a entityAlreadyExists, it means something really went wrong 🧐

delthas (approved):
Good work :)

benzekrimaha (approved):
LGTM

PR #2471, opened 2026-07-28
MongoDbPodRecovering alert firing once per reporting pod instead of once per recovering member

benzekrimaha:
## Context

RD-1612 (Bergen Kommune, ARTESCA 4.0.4): after a node reinstall left one shard0 member stuck in RECOVERING, `MongoDbPodRecovering` fired **three times** — once per shard0 pod — while the other two members were a healthy PRIMARY and SECONDARY.

## Root cause

`mongodb_rs_members_state` has one series per (reporting pod, member): every pod's exporter reports the state of every replica-set member. A single recovering member therefore matches `== 3` once per reporting pod, and the annotation prints `{{ $labels.pod }}` — the reporter, not the recovering member.

## Fix

- Aggregate with `min by (namespace, rs_nm, member_idx)` so exactly one alert fires per recovering member, and rebuild `pod` from the first DNS segment of `member_idx` so the description names the right pod. Also keeps the alert firing when the recovering member's own exporter is unscrapeable (peers still report it).
- Rewrite the unit tests with realistic series (every pod reporting every member): the nominal case expects exactly one alert per severity, and a second scenario covers the unscrapeable recovering member. The new tests fail against the old rule — reproducing the customer's symptom — and pass with the fix; the previous test used a single synthetic series and could not catch the fan-out.

Issue: ZENKO-5332

## Also included: fix for the pre-existing `check-workflows` failure on development/2.13

The `check-workflows` job was already broken on this branch (the 5 "Promote artifacts" tests failed on any PR), which blocked CI here. Included the fix in this PR to avoid a separate PR. Root cause is environment drift, since everything the tests use is pinned:

- The catthehacker act runner image (pulled with `forcePull`) now ships node 24 in its toolcache. `action-gh-release@v2.5.0` never calls `getReleaseByTag`: it scans `listReleases` **before** creating the release, and the second mocked `listReleases` reply contained the draft release, so the action took the update-existing-release path and issued an `updateRelease` call no mock matches. This used to pass by accident (unmatched request → fast 404 → the action fell back to creating the release); on node 24 the unmatched request hangs until an undici headers timeout and the step fails. The second reply is now draft-free so the action takes the create path with every request matched.
- `scality/action-artifacts@v4` (floating tag) is now a node24 action, which the pinned act 0.2.75 refuses to run: bumped act to 0.2.84 (same as development/2.14) and adapted the step-result parsing to the timing suffix newer act appends to status lines.

All 23 workflow tests pass locally.

Issue: ZENKO-5317


delthas (approved):

SylvainSenechal (approved):

SylvainSenechal:
Can you do waterfall on 3 branches ? we have 2.16 now, I thought waterfall was only 2 branches so 2.14 and 2.15 max 🤔

benzekrimaha:
Integration branch `w/2.14` created with conflicts resolved (VERSION bumped to 2.14.7, kept the 2.14 release workflow mocks).

benzekrimaha:
Integration branch `w/2.15` created (VERSION bumped to 2.15.6).

benzekrimaha:
@bert-e create_pull_requests

PR #2479, opened 2026-08-04
Stop asserting exact location-version-id equality after replication

delthas:
The `Bucket Replication location stripping` scenario asserted that the `<location>-version-id` stamped in source metadata equals the destination's current version id. Replication is at-least-once, so that is not an invariant the product offers, and the test fails on correct behaviour.

**What it caught.** In one census run the object was replicated **twice**: two oplog populators were alive during a rolling restart and each allocated a Kafka-Connect connector for the same bucket, producing three entries for one source version. The status processor kept the first stamp — *"entry replication is already COMPLETED for this location, skipping metadata update"* — while the destination retained the version written by the second replication. Both objects were present and identical in content; only the recorded version id disagreed. Replication itself had succeeded, in 59 s.

**The change.** Assert on content rather than identity: compare ETags (normalising the surrounding quotes, as `get-object-attributes.ts:68` already does) and require only that the version-id stamp is present. `ContentLength` was already compared. The neighbouring `scal-version-id`, `scal-replication-status` and `<location>-replication-status` assertions are untouched — `scal-version-id` records the *source* version, which both copies share, so the double replication does not affect it.

**Census confirmation.** Across the two 30-run censuses since:

- location stripping failures: **2/30 → 0/30**
- the changed assertion ran **420 times with zero failures** in the latest census

Issue: ZENKO-5339

### DIFF
--- a/tests/functional/ctst/steps/replication.ts
+++ b/tests/functional/ctst/steps/replication.ts
@@ -207,12 +207,18 @@ async function assertReplicaMatchesSource(world: Zenko, location: string): Promi
     const sourceObj = safeJsonParse<SourceObjectReplicationMeta>(sourceResponse.stdout || '{}');
     assert(sourceObj.ok);
     assert.strictEqual(sourceObj.result?.ContentLength, replicaObj.ContentLength);
+    const sourceETag = sourceObj.result?.ETag?.replace(/^"|"$/g, '');
+    const replicaETag = replicaObj.ETag?.replace(/^"|"$/g, '');
+    assert.ok(sourceETag, 'source object has no ETag');
+    assert.strictEqual(sourceETag, replicaETag);
     // CRR loopback only writes the per-destination status; cloud backends
     // also stamp version-id / scal-version-id.
     if (!crrCtx) {
-        assert.strictEqual(
+        // replication is at-least-once, so the stamped version id may name an
+        // earlier replica than the one currently at the destination
+        assert.ok(
             sourceObj.result?.Metadata?.[`${location}-version-id`],
-            replicaObj.VersionId,
+            `source metadata has no ${location}-version-id stamp`,
         );
         assert.strictEqual(
             sourceObj.result?.VersionId,
--- a/tests/functional/ctst/steps/replication.ts
+++ b/tests/functional/ctst/steps/replication.ts
@@ -214,12 +214,20 @@ async function assertReplicaMatchesSource(world: Zenko, location: string): Promi
     // CRR loopback only writes the per-destination status; cloud backends
     // also stamp version-id / scal-version-id.
     if (!crrCtx) {
-        // replication is at-least-once, so the stamped version id may name an
-        // earlier replica than the one currently at the destination
-        assert.ok(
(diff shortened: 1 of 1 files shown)


SylvainSenechal (approved):
Ok for me although when you say this : 

"What it caught. In one census run the object was replicated twice: two oplog populators were alive during a rolling restart and each allocated a Kafka-Connect connector for the same bucket, producing three entries for one source version" :


- It would be nice to at least have a ticket to think about this behavior, and decide if its an acceptable design or if its possible to fix it on the backbeat side
- I think i would still keep the versionID assertion, and maybe when they are not equal, instead of failing the test, log a warning "test replication find non matching version ID likely due to backbeat oplog populator restarts..."

delthas:
Both points done.

**Version-id check kept as a warning** (`7dee82f80f`). It still compares the stamped `<location>-version-id` against the destination's current version; on mismatch it logs a warning naming the likely cause instead of failing:

```
replicated object version id does not match the source stamp
  location, stampedVersionId, destinationVersionId,
  hint: the object was likely replicated more than once, for instance by two
        oplog populators both allocating a connector for the bucket after a restart
```

That keeps the signal visible without failing on something at-least-once replication is allowed to do.

**Ticket filed**: BB-834, for the populator behaviour itself — fence bucket-to-connector ownership so a terminating and a starting instance cannot both allocate the same bucket. It records the traced occurrence (three replication entries for one source version, the status processor dropping the second stamp with *"already COMPLETED for this location, skipping metadata update"*), and explicitly notes that after this PR a mismatch no longer fails CI, so nothing will catch it regressing. Suggested looking at it together with the populator's scan/watch ordering hole, since both are about ownership across a restart.

Worth adding for context: the later 30-run census did not reproduce it, so there is no rate estimate beyond that single occurrence. The trigger is rollout frequency, which in CI is inflated by the operator re-rendering every backbeat deployment on each location CRUD (ZKOP-568).

DarkIsDude (commented):
Same as Sylvain. I'm not sure it's the right approach. We should fix that in backbeat instead of here and keep the hard comparison here (it's what we want finally). Or it's a product decision and we are align with it, then your fix is the right one (but it seems not the case, at least in the scenario tested where one replication should happens), or it's not and we should fix it in backbeat.

Moreover, I would approve this PR if the occurence is high enough to be painful. Seems not the case regarding your last comment (30 runs didn't reproduce it).

We can also ask to the team what do they think, but IMO it's not the right approach here.

DarkIsDude:
@delthas also in your ticket https://scality.atlassian.net/browse/BB-834 add a link to this thread to have the full context ?

delthas:
@DarkIsDude you're right on the frequency, and I checked rather than assumed. The single occurrence came from one census run (`Bucket Replication location stripping`, 4 MiB row); no census since has reproduced it. The 30-run July census still had the hard equality assert in place and produced 6 replication-related failures, all of them timeouts, with no version-id assertion failure among the annotations. So the rate is low enough that relaxing a CI assertion is not worth it.

Closing in favour of BB-834.

delthas:
Also I linked the thread in the ticket.

PR #2480, opened 2026-08-04
Poll until the cascade settles, then assert it stays settled

delthas:
Reduced to one topic. The other two are split out as #2486 (diagnostics and step budgets) and #2487 (the loop guard).

The step waited a blind 15s and asserted once. That single observation had to be lucky twice: late enough for every location to have settled, and early enough that a loop writing back to the source had not already been overwritten. In the July census it was late by 21s — location C reached REPLICA *after* the assertion had failed — and the same shape can pass by observing a loop before its write-back lands.

So it now polls until every location reaches its expected state, then holds: keeps checking for a further window and fails if anything moves again. @SylvainSenechal — this is the two-phase version we landed on: the "nothing more happens while watching" part you wanted is the hold phase, and the polling only replaces the blind sleep before it.

**Now two steps**, per @DarkIsDude's point about the "and" in the name — they answer different questions and each carries its own budget:

```gherkin
Then the cascade replication states should settle within 120 seconds
And the cascade replication states should stay settled for 15 seconds
```

**The check moved out of the step body** into `cascadeUnsettledReason`, which both steps share.

**Error handling is unchanged.** An earlier revision of this PR retried failed reads to the deadline; that was scope creep and is removed. A failed read fails the step, exactly as before. The wider inconsistency in this file — `should have the expected tags` has no retry, `should replicate to location` retries only 404 — predates this PR and is better fixed deliberately than as a side effect here.

On the per-poll S3 client: left as is, per @SylvainSenechal — the client lifecycle wants its own refactor.

Not carried over: the budget guard from #2486. Its two steps and these two would be better off sharing one convention, but that is a follow-up once one of them lands rather than a reason to couple the PRs.

Issue: ZENKO-5340

### DIFF
--- a/tests/functional/ctst/features/replication/crrCascade.feature
+++ b/tests/functional/ctst/features/replication/crrCascade.feature
@@ -18,8 +18,8 @@ Feature: CRR Cascade Replication
         When an object "cascade-obj" of 0 bytes is put in location "crr-location-a"
         Then the object should replicate to location "crr-location-b" within 280 seconds
         And the object should replicate to location "crr-location-c" within 280 seconds
-        When I wait 15 seconds
-        Then the cascade replication states should be settled
+        Then the cascade replication states should settle within 120 seconds
+        And the cascade replication states should stay settled for 15 seconds
 
     @2.16.0
     @PreMerge
@@ -37,8 +37,8 @@ Feature: CRR Cascade Replication
         Then the object should replicate to location "crr-location-b" within 280 seconds
         And the object should replicate to location "crr-location-c" within 280 seconds
         And the object at location "crr-location-a" should never have replication status PENDING within 30 seconds
-        When I wait 15 seconds
-        Then the cascade replication states should be settled
+        Then the cascade replication states should settle within 120 seconds
+        And the cascade replication states should stay settled for 15 seconds
 
         Examples:
             | description      | objectName                | bodySize |
@@ -63,8 +63,8 @@ Feature: CRR Cascade Replication
         When tags are put on the object "cascade-tag-loop-obj" in location "crr-location-a"
         Then the object at location "crr-location-c" should have the expected tags within 280 seconds
         And the object at location "crr-location-a" should never have replication status PENDING within 30 seconds
-        When I wait 15 seconds
-        Then the cascade replication states should be settled
+        Then the cascade replication states should settle within 120 seconds
+        And the cascade replication states should stay settled for 15 seconds
 
     @2.16.0
     @PreMerge
@@ -93,5 +93,5 @@ Feature: CRR Cascade Replication
         And replication is configured from location "crr-location-b" to "crr-location-a"
         When an object "bidirectional-obj" of 42 bytes is put in location "crr-location-a"
         Then the object should replicate to location "crr-location-b" within 280 seconds
-        When I wait 15 seconds
-        Then the cascade replication states should be settled
+        Then the cascade replication states should settle within 120 seconds
+        And the cascade replication states should stay settled for 15 seconds
--- a/tests/functional/ctst/steps/crrCascade.ts
+++ b/tests/functional/ctst/steps/crrCascade.ts
@@ -214,34 +214,63 @@ Then(
     },
 );
 
+// undefined when every location is in its expected state
+async function cascadeUnsettledReason(world: Zenko): Promise<string | undefined> {
+    const cascadeBuckets = world.getSaved<Record<string, string>>('cascadeBuckets');
+    const objectName = world.getSaved<string>('cascadeObjectName');
+    const sourceLocation = world.getSaved<string>('cascadeSourceLocation');
+
+    for (const [location, bucket] of Object.entries(cascadeBuckets)) {
+        Identity.useIdentity(IdentityEnum.ACCOUNT, location);
+        const res = await world.createS3Client().send(
+            new HeadObjectCommand({ Bucket: bucket, Key: objectName }),
+        );
+
+        const expectedStatus = location === sourceLocation ? 'COMPLETED' : 'REPLICA';
+        if (res.ReplicationStatus !== expectedStatus) {
+            return `Expected ReplicationStatus '${expectedStatus}' at '${location}', `
+                + `got '${res.ReplicationStatus}'`;
+        }
+
+        const pendingBackends = Object.entries(res.Metadata ?? {})
+            .filter(([key, value]) => key.endsWith('-replication-status') && value === 'PENDING');
+        if (pendingBackends.length > 0) {
+            return `Location '${location}' still has PENDING backends: ${
+                pendingBackends.map(([k, v]) => `${k}=${v}`).join(', ')
+            }`;
+        }
+    }
+    return undefined;
+}
+
 Then(
-    'the cascade replication states should be settled',
-    { timeout: 30_000 },
-    async function (this: Zenko) {
-        const cascadeBuckets = this.getSaved<Record<string, string>>('cascadeBuckets');
-        const objectName = this.getSaved<string>('cascadeObjectName');
-        const sourceLocation = this.getSaved<string>('cascadeSourceLocation');
+    'the cascade replication states should settle within {int} seconds',
+    { timeout: 180_000 },
+    async function (this: Zenko, settleSeconds: number) {
+        const deadline = Date.now() + settleSeconds * 1000;
+        let unsettled: string | undefined;
+        while (Date.now() < deadline) {
+            unsettled = await cascadeUnsettledReason(this);
+            if (unsettled === undefined) {
+                return;
+            }
+            await new Promise(resolve => setTimeout(resolve, 3000));
+        }
+        assert.fail(`cascade states did not settle within ${settleSeconds}s. ${unsettled}`);
+    },
+);
 
-        for (const [location, bucket] of Object.entries(cascadeBuckets)) {
-            Identity.useIdentity(IdentityEnum.ACCOUNT, location);
-            const res = await this.createS3Client().send(
-                new HeadObjectCommand({ Bucket: bucket, Key: objectName }),
-            );
-
-            const expectedStatus = location === sourceLocation ? 'COMPLETED' : 'REPLICA';
+Then(
+    'the cascade replication states should stay settled for {int} seconds',
+    { timeout: 60_000 },
+    async function (this: Zenko, holdSeconds: number) {
+        const deadline = Date.now() + holdSeconds * 1000;
+        while (Date.now() < deadline) {
+            await new Promise(resolve => setTimeout(resolve, 1000));
+            const changed = await cascadeUnsettledReason(this);
             assert.strictEqual(
-                res.ReplicationStatus,
-                expectedStatus,
-                `Expected ReplicationStatus '${expectedStatus}' at '${location}', got '${res.ReplicationStatus}'`,
-            );
-
-            const pendingBackends = Object.entries(res.Metadata ?? {})
-                .filter(([key, value]) => key.endsWith('-replication-status') && value === 'PENDING');
-            assert.strictEqual(
-                pendingBackends.length, 0,
-                `Location '${location}' still has PENDING backends after settling: ${
-                    pendingBackends.map(([k, v]) => `${k}=${v}`).join(', ')
-                }`,
+                changed, undefined,
+                `cascade states were not settled during the ${holdSeconds}s observation window. ${changed}`,
             );
         }
     },
--- a/tests/functional/ctst/steps/crrCascade.ts
+++ b/tests/functional/ctst/steps/crrCascade.ts
@@ -14,6 +14,8 @@ import { Identity, IdentityEnum, Utils } from 'cli-testing';
 import Zenko from 'world/Zenko';
 
 const STEP_TIMEOUT_MS = 300_000;
+const POLL_MS = 3_000;
+const STABILITY_INTERVAL_MS = 1_000;
 
 export interface CRRAccountInfo {
     AccessKeyId: string;
@@ -131,7 +133,7 @@ Then(
             if (found) {
                 return;
             }
-            await new Promise(resolve => setTimeout(resolve, 3000));
+            await Utils.sleep(POLL_MS);
         }
         assert.fail(
             `Timeout: tag 'cascade-test-tag=${tagValue}' not found at '${location}' after ${timeoutSeconds}s`,
@@ -161,7 +163,7 @@ Then(
                 );
                 lastStatus = res.ReplicationStatus ?? 'unset';
                 if (res.ReplicationStatus === 'PENDING') {
-                    await new Promise(resolve => setTimeout(resolve, 3000));
+                    await Utils.sleep(POLL_MS);
                     continue;
                 }
                 assert.strictEqual(
@@ -179,7 +181,7 @@ Then(
                     throw err;
                 }
                 lastStatus = 'not found';
-                await new Promise(resolve => setTimeout(resolve, 3000));
+                await Utils.sleep(POLL_MS);
             }
         }
         assert.fail(
@@ -209,7 +211,7 @@ Then(
                 `Object at '${location}' was found with ReplicationStatus=PENDING, ` +
                 'indicating the cascade loop wrote back to the source.',
             );
-            await new Promise(resolve => setTimeout(resolve, 1000));
+            await Utils.sleep(STABILITY_INTERVAL_MS);
         }
     },
 );
@@ -254,7 +256,7 @@ Then(
             if (unsettled === undefined) {
                 return;
             }
-            await new Promise(resolve => setTimeout(resolve, 3000));
+            await Utils.sleep(POLL_MS);
         }
         assert.fail(`cascade states did not settle within ${settleSeconds}s. ${unsettled}`);
     },
@@ -266,7 +268,7 @@ Then(
     async function (this: Zenko, holdSeconds: number) {
         const deadline = Date.now() + holdSeconds * 1000;
         while (Date.now() < deadline) {
-            await new Promise(resolve => setTimeout(resolve, 1000));
+            await Utils.sleep(STABILITY_INTERVAL_MS);
             const changed = await cascadeUnsettledReason(this);
             assert.strictEqual(
                 changed, undefined,
@@ -342,7 +344,7 @@ Then(
                 );
                 return;
             }
-            await new Promise(resolve => setTimeout(resolve, 3000));
+            await Utils.sleep(POLL_MS);
         }
 
         assert.fail(`Cascade locations did not converge to the same marker within ${timeoutSeconds}s`);


tests/functional/ctst/features/replication/crrCascade.feature
SylvainSenechal, line 1: The intent of 
"When I wait 15 seconds
 Then the cascade replication states should be settled"

Was not really i wanna check replication status asap, but it was intentionally more about : "we have have finished the cascading replication and verified the replication status are ok in these steps : And the object should replicate to location "crr-location-c" within 300 seconds, now we wanna make sure after waiting a bit that there is nothing more happening that would change the status of the object again"

SylvainSenechal: So yeah I think the 
"should be settled **within** 120 seconds" is not the right intent
What I wanted with this test is more
"should be settled **after** 120 seconds of observation, nothing changed"

SylvainSenechal (changes requested):
Ok for the second point about "the failure message was wrong".

Ok with the "loop guard accused the product of a bug it did not have", although this issue seems to arise only when the replication takes too long, if I understand correctly this would never be surfaced if replication was fast (and it's alwasys been fast ~5-20 seconds in my codespace with no flaky), so I still think the focus should be more on backbeat pod crash mishandled

For the first point "settle step did not poll", well as I wrote in the other comment, the settle step wasnt really about doing a polling, but more about asserting that after some time we have a stable situation, although maybe its fine to also do polling, although the previous steps are already doing polling too, yeah I guess we can keep the polling in the settle step as in this last step we observe that status of all location (for const [location, bucket]...), but I don't think this step should return as soon as we have no pending backend, because the point of this test was to make sure nothing more happens while watching for a few seconds (but maybe this test is debatable, its a bit over testing and it could probably be removed)

delthas:
Sorry about the prev slop on this comment.

TL;DR:

Moved to wait for initial REPLICA for 120s, then ensure it stays REPLICA for 15s.

This way we have more time for the initial REPLICA, and we keep the "ensure it stays REPLICA" part but not for too long, otherwise it would add too much CI time.

tests/functional/ctst/steps/crrCascade.ts
```diff
@@ -107,7 +107,7 @@ When('tags are put on the object {string} in location {string}',
 
 Then(
     'the object at location {string} should have the expected tags within {int} seconds',
-    { timeout: 300_000 },
+    { timeout: 330_000 },
```
DarkIsDude: Regarding this fix, if we have `within 500 seconds` we are back to the same issue. Should we keep 300 seconds as timeout and add an assert to make sure we are always under 280s ? Also why increasing this one and not decreasing the other one ?

delthas:
Ticket for the `backbeat-gc` stall mentioned above: **BB-835** — https://scality.atlassian.net/browse/BB-835

The log pins it to a race rather than a crash: on revoke with entries still in flight, `BackbeatConsumer._onRebalance` defers the un-assign until the queue drains, and if the next generation grants the partitions back in the meantime the deferred callback un-assigns them anyway.

```
01:42:18.914  rdkafka.revoke   parts=[0,1,2,3,4]  running=5 ledger=5
01:42:19.046  rdkafka.assign   parts=[0,1,2,3,4]
01:42:19.059  processing queue drained, un-assigning
```

The consumer stays a group member owning all five partitions with no local assignment, so no further rebalance ever rescues it — 188 of 199 messages unconsumed, 2h35m idle.

tests/functional/ctst/steps/crrCascade.ts
```diff
@@ -183,49 +188,88 @@ Then(
...
                 new HeadObjectCommand({ Bucket: bucket, Key: objectName }),
             );
-            assert.notStrictEqual(
-                result.ReplicationStatus,
-                'PENDING',
-                `Object at '${location}' was found with ReplicationStatus=PENDING, ` +
-                'indicating the cascade loop wrote back to the source.',
-            );
+            if (settledAs === undefined) {
+                if (result.ReplicationStatus !== 'PENDING') {
```
DarkIsDude: What is the goal of those line ? Why `unset` ? Maybe I don't understand very well the 191 comments ?

tests/functional/ctst/steps/crrCascade.ts
```diff
@@ -183,49 +188,88 @@ Then(
...
-        for (const [location, bucket] of Object.entries(cascadeBuckets)) {
-            Identity.useIdentity(IdentityEnum.ACCOUNT, location);
-            const res = await this.createS3Client().send(
-                new HeadObjectCommand({ Bucket: bucket, Key: objectName }),
-            );
+        // returns '' when every location is in its expected state
+        const unsettledReason = async (): Promise<string> => {
+            for (const [location, bucket] of Object.entries(cascadeBuckets)) {
+                Identity.useIdentity(IdentityEnum.ACCOUNT, location);
+                const res = await this.createS3Client().send(
```
DarkIsDude: Do we need to create a client each time ?

tests/functional/ctst/steps/crrCascade.ts
DarkIsDude: My question is : why do we manage the `unknown` statue ?

tests/functional/ctst/steps/crrCascade.ts
DarkIsDude: The goal is to catch a 403 on it right ? If we have a 403, this one is not in the try/catch, how to try again in this case ?

tests/functional/ctst/steps/crrCascade.ts
```diff
@@ -183,49 +188,88 @@ Then(
...
+                break;
+            }
+            await new Promise(resolve => setTimeout(resolve, 3000));
+        }
+        assert.strictEqual(
+            unsettled, '',
+            `cascade states did not settle within ${settleSeconds}s. ${unsettled}`,
+        );
+
+        // the cascade must then stop moving: a loop that writes back to the
```
DarkIsDude: To ease readability, should this one be a dedicated test ?  as you have an AND in your sentence, I guess yes.

tests/functional/ctst/steps/crrCascade.ts
```diff
@@ -183,49 +188,88 @@ Then(
...
         const objectName = this.getSaved<string>('cascadeObjectName');
         const sourceLocation = this.getSaved<string>('cascadeSourceLocation');
 
-        for (const [location, bucket] of Object.entries(cascadeBuckets)) {
-            Identity.useIdentity(IdentityEnum.ACCOUNT, location);
-            const res = await this.createS3Client().send(
-                new HeadObjectCommand({ Bucket: bucket, Key: objectName }),
-            );
+        // returns '' when every location is in its expected state
+        const unsettledReason = async (): Promise<string> => {
```
DarkIsDude: We usually try to move function outside step, will ease readability and avoid too long step 🙏

tests/functional/ctst/features/replication/crrCascade.feature
```diff
@@ -63,8 +61,7 @@ Feature: CRR Cascade Replication
         When tags are put on the object "cascade-tag-loop-obj" in location "crr-location-a"
         Then the object at location "crr-location-c" should have the expected tags within 300 seconds
         And the object at location "crr-location-a" should never have replication status PENDING within 30 seconds
```
DarkIsDude, line 65: That's not really try anymore, `should never have replication status PENDING` after reaching an end state ?

DarkIsDude: Also whats happens if the object always have a PENDING state ?

tests/functional/ctst/steps/crrCascade.ts
DarkIsDude, line 178: We should also here update the lastStatus. If the HEAD fail, the lastStatus should reflect it ?

DarkIsDude (changes requested):
Not easy easy easy to review as we mix 3 differents topics in the same PR. Even if I can understand your point (I'm in), I'm not sure it's the right move. With claude, saying "open 3 PR's for each topics" if a costless move. Think it twice next time.

Anyway, good job and work

tests/functional/ctst/steps/crrCascade.ts
SylvainSenechal: I wanna do something (some day) about the clients : initialize them cleanly somewhere at the begging of the ci test, but it requires some refactoring work so thats why its sitting as it is right now. I think its ok to keep for the moment

tests/functional/ctst/steps/crrCascade.ts
delthas: Moved to #2486, and taken your way rather than mine: the timeout stays at 300s and the feature budgets drop to 280s, so the headroom is visible in the feature file. The step asserts the relationship on entry, so `within 500 seconds` fails immediately naming both numbers.

On "why increasing this one and not decreasing the other one" — you were right, decreasing is the better half. Increasing only moved the cliff.

tests/functional/ctst/steps/crrCascade.ts
delthas: Moved to #2486. `unset` is the display value when the object has no `x-amz-replication-status` header at all, so the timeout message can say what was actually seen instead of printing `undefined`. It is only ever a diagnostic string, never compared against.

tests/functional/ctst/steps/crrCascade.ts
delthas: Kept as is, per @SylvainSenechal's answer.

tests/functional/ctst/steps/crrCascade.ts
delthas: Done here — two steps now:

```gherkin
Then the cascade replication states should settle within 120 seconds
And the cascade replication states should stay settled for 15 seconds
```

tests/functional/ctst/steps/crrCascade.ts
delthas: Done here — the check is now `cascadeUnsettledReason` at module scope, shared by both steps.

tests/functional/ctst/features/replication/crrCascade.feature
delthas: Moved to #2487, and renamed exactly as you suggest: `should not return to replication status PENDING`.

delthas: Now also handling objects staying in PENDING state, will now fail.

tests/functional/ctst/steps/crrCascade.ts
delthas: Moved to #2486 — `lastStatus` is now set to `not found` in the catch, so a HEAD that 404s is reflected rather than leaving a stale earlier status in the message.

tests/functional/ctst/steps/crrCascade.ts
```diff
@@ -198,34 +198,74 @@ Then(
...
+            }
             assert.strictEqual(
-                pendingBackends.length, 0,
-                `Location '${location}' still has PENDING backends after settling: ${
-                    pendingBackends.map(([k, v]) => `${k}=${v}`).join(', ')
-                }`,
+                changed, '',
+                `cascade states changed during the ${holdSeconds}s observation window. ${changed}`,
             );
         }
```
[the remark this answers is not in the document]
delthas: Good catch, fixed in 238468fa37 — same vacuous-pass shape this PR is meant to remove, reintroduced by the retry. Taken as suggested, counting successful checks and asserting at least one ran.

delthas:
Reworked and force-pushed to a single commit: this PR is now only the settle step, with the diagnostics in #2486 and the loop guard in #2487. Every thread above has a reply pointing at where its answer landed.

tests/functional/ctst/steps/crrCascade.ts
delthas: Dropped the retry I had added here — this PR leaves 403 handling exactly as it was, so a failed read still fails the step.

Retrying a 403 would only hide an `AccessDenied` behind a 120s timeout.

tests/functional/ctst/steps/crrCascade.ts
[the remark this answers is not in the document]
delthas: Fixed differently in 177728a268: rather than counting successful checks, the swallowed-error path is gone entirely. The retry that created it was scope creep on my part — this PR now leaves read-error handling as it was, so there is no `null` return and no way for the hold loop to complete without asserting.

tests/functional/ctst/steps/crrCascade.ts
```diff
@@ -198,34 +198,63 @@ Then(
...
+    },
+);
+
+Then(
+    'the cascade replication states should stay settled for {int} seconds',
+    { timeout: 60_000 },
+    async function (this: Zenko, holdSeconds: number) {
+        const deadline = Date.now() + holdSeconds * 1000;
+        while (Date.now() < deadline) {
+            await new Promise(resolve => setTimeout(resolve, 1000));
```
DarkIsDude:
```
            await Utils.sleep(POLL_MS);
```

tests/functional/ctst/steps/crrCascade.ts
```diff
@@ -198,34 +198,63 @@ Then(
...
+    { timeout: 180_000 },
+    async function (this: Zenko, settleSeconds: number) {
+        const deadline = Date.now() + settleSeconds * 1000;
+        let unsettled = 'no location was checked';
+        while (Date.now() < deadline) {
+            unsettled = await cascadeUnsettledReason(this);
+            if (!unsettled) {
+                return;
+            }
+            await new Promise(resolve => setTimeout(resolve, 3000));
```
DarkIsDude:
```
            await Utils.sleep(POLL_MS);
```

DarkIsDude (approved):

tests/functional/ctst/steps/crrCascade.ts
delthas: Done, and taken across the whole file — it had seven of these, and it was the last file in the suite not using `Utils.sleep`. Separate commit so it stays out of the settle change.

Named after the existing settle-then-stability step in `steps/mongodbExporter.ts`: `POLL_MS` (3s) while waiting for a state to arrive, `STABILITY_INTERVAL_MS` (1s) while watching that it does not change.

tests/functional/ctst/steps/crrCascade.ts
```diff
@@ -193,39 +196,68 @@ Then(
...
+
+Then(
+    'the cascade replication states should settle within {int} seconds',
+    { timeout: 180_000 },
+    async function (this: Zenko, settleSeconds: number) {
+        const deadline = Date.now() + settleSeconds * 1000;
+        let unsettled = 'no location was checked';
+        while (Date.now() < deadline) {
+            unsettled = await cascadeUnsettledReason(this);
+            if (!unsettled) {
```
SylvainSenechal: nit: since we are in typescript, maybe there are better ways to express this situation than doing !unsettled on a promise(string)

SylvainSenechal (approved):

tests/functional/ctst/steps/crrCascade.ts
delthas: Agreed — the empty string was doing double duty. It returns `string | undefined` now, so settled is `undefined` and both call sites compare explicitly instead of relying on truthiness. Folded into the settle commit.

tests/functional/ctst/steps/crrCascade.ts
```diff
@@ -193,39 +196,68 @@ Then(
...
+                pendingBackends.map(([k, v]) => `${k}=${v}`).join(', ')
+            }`;
+        }
+    }
+    return undefined;
+}
+
+Then(
+    'the cascade replication states should settle within {int} seconds',
+    { timeout: 180_000 },
```
benzekrimaha, line 250: the fixed cucumber timeout silently caps the `{int}` parameter: a future scenario writing `settle within 200 seconds` gets killed by cucumber at 180s with a generic timeout error instead of this step's own diagnostic. a one-line guard at step entry makes the contract explicit:

```js
assert(settleSeconds * 1000 < 180_000, `settle budget ${settleSeconds}s exceeds the step timeout`);
```

(same for the hold step below, 60s cap vs its `{int}`)

tests/functional/ctst/steps/crrCascade.ts
```diff
@@ -193,39 +196,68 @@ Then(
...
+    async function (this: Zenko, settleSeconds: number) {
+        const deadline = Date.now() + settleSeconds * 1000;
+        let unsettled: string | undefined = 'no location was checked';
```
benzekrimaha: nit: the initializer is only reachable when `settleSeconds <= 0`; with the guard above it becomes dead and this could just be a plain declaration. fine either way.

tests/functional/ctst/steps/crrCascade.ts
```diff
@@ -193,39 +196,68 @@ Then(
...
+        while (Date.now() < deadline) {
+            await Utils.sleep(STABILITY_INTERVAL_MS);
+            const changed = await cascadeUnsettledReason(this);
             assert.strictEqual(
-                pendingBackends.length, 0,
-                `Location '${location}' still has PENDING backends after settling: ${
-                    pendingBackends.map(([k, v]) => `${k}=${v}`).join(', ')
-                }`,
+                changed, undefined,
+                `cascade states changed during the ${holdSeconds}s observation window. ${changed}`,
```
benzekrimaha: nit: “changed during the observation window” presumes the settle step ran first , if this step is ever used alone, the reason could be a state that was never settled to begin with. a wording like “was not settled” covers both; not worth more than that.

tests/functional/ctst/steps/crrCascade.ts
delthas: Good catch, and it's the same invariant #2486 added for the tags and replicate steps in this file — those now carry `STEP_TIMEOUT_MS` plus exactly the guard you describe.

Deliberately not extending it here though: I filed **ZENKO-5344** for it, because the gap is suite-wide rather than local. 16 steps declare a cucumber timeout and bound their own polling separately — 6 in this file, 6 in `pra.ts`, 2 in `replication.ts`, 2 in `cronjobs/countItems.ts` — and 6 of those delegate the polling to a helper that hardcodes its budget in another file, so the pair isn't visible from either side. `replication.ts:468` is the nearest live case: 600s timeout against a `within 500 seconds` that already exists in the feature files.

Doing settle and hold here would close two of sixteen and leave the audit needing to revisit this file anyway.

tests/functional/ctst/steps/crrCascade.ts
delthas: Dropped — it's now a plain `let unsettled: string | undefined;`. You're right that the only path reaching it is a non-positive budget, and on any normal exit the loop has assigned a reason.

tests/functional/ctst/steps/crrCascade.ts
delthas: Taken — it now reads `cascade states were not settled during the Ns observation window`. That matters more than it looks: the settle and hold steps are separate in this PR, so hold genuinely can run alone, and the old wording asserted an ordering it can't check.

PR #2486, opened 2026-08-07
Keep the cascade step budgets under their timeout, and report the last observed status

delthas:
Split out of #2480, which mixed three topics. This is the diagnostics one.

**The failure message was wrong.** The replication step polls HEAD until the object reaches REPLICA and treats a 404 as "not there yet". On timeout it reported the object as *missing* — true only if the last HEAD had 404ed. An object sitting at PENDING for the whole budget was reported as absent, which sends the investigation looking for a lost object instead of a stalled replication. It now names what the last HEAD actually saw: a status, `unset` when the header is absent, or `not found` when it 404ed.

**The step could not report at all.** The feature file asked for exactly the step's own cucumber timeout, so cucumber killed the step at the same moment its deadline passed and the `assert.fail` never ran — the run showed a bare step timeout with no diagnostic.

Taking @DarkIsDude's suggestion: the timeout stays at 300s and the budgets drop to 280s, so the headroom lives in the feature file rather than in a raised timeout. The step asserts the relationship on entry, so `within 500 seconds` fails in milliseconds naming both numbers instead of silently losing the diagnostic again. No shared helper — it is two call sites in one file, and the value is per-step.

Out of scope: `all cascade locations should converge...` takes a 300s budget against a 600s timeout, so it already has headroom and is untouched.

The two commits are one failure mode from both sides: a better message that never runs is useless, and surviving long enough to print the wrong message is no better.

Issue: ZENKO-5340

### DIFF
--- a/tests/functional/ctst/steps/crrCascade.ts
+++ b/tests/functional/ctst/steps/crrCascade.ts
@@ -13,6 +13,8 @@ import assert from 'assert';
 import { Identity, IdentityEnum, Utils } from 'cli-testing';
 import Zenko from 'world/Zenko';
 
+const STEP_TIMEOUT_MS = 300_000;
+
 export interface CRRAccountInfo {
     AccessKeyId: string;
     SecretAccessKey: string;
@@ -107,8 +109,12 @@ When('tags are put on the object {string} in location {string}',
 
 Then(
     'the object at location {string} should have the expected tags within {int} seconds',
-    { timeout: 300_000 },
+    { timeout: STEP_TIMEOUT_MS },
     async function (this: Zenko, location: string, timeoutSeconds: number) {
+        assert.ok(
+            timeoutSeconds * 1000 < STEP_TIMEOUT_MS,
+            `budget of ${timeoutSeconds}s exceeds the step timeout of ${STEP_TIMEOUT_MS / 1000}s`,
+        );
         const bucket = this.getSaved<Record<string, string>>('cascadeBuckets')[location];
         const objectName = this.getSaved<string>('cascadeObjectName');
         const tagValue = this.getSaved<string>('cascadeTagValue');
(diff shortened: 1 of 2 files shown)


tests/functional/ctst/steps/crrCascade.ts
```diff
@@ -135,19 +141,25 @@ Then(
...
+            timeoutSeconds * 1000 < STEP_TIMEOUT_MS,
+            `budget of ${timeoutSeconds}s exceeds the step timeout of ${STEP_TIMEOUT_MS / 1000}s`,
+        );
         const bucket = this.getSaved<Record<string, string>>('cascadeBuckets')[location];
         const objectName = this.getSaved<string>('cascadeObjectName');
         const deadline = Date.now() + timeoutSeconds * 1000;
         Identity.useIdentity(IdentityEnum.ACCOUNT, location);
         const client = this.createS3Client();
         const expectedContentLength = this.getSaved<number>('cascadeExpectedContentLength');
+        let lastStatus = 'not found';
```
DarkIsDude: should it be `not found` or `unknown` ?

DarkIsDude (approved):

tests/functional/ctst/steps/crrCascade.ts
delthas: Good catch — `unknown` now, since `not found` is what an actual 404 sets a few lines down. It only shows if no HEAD completes at all, but that is exactly when the message should not be claiming an observation.

SylvainSenechal (approved):
I think we have multiple similar patterns in this ci, where the step has a timeout, and also does some kind of infinite polling loop. Worth checking if similar case could also be made clearer

delthas:
@SylvainSenechal good call — filed as **ZENKO-5344**: https://scality.atlassian.net/browse/ZENKO-5344

Counting them, 16 steps declare a cucumber timeout and bound their own polling separately: `crrCascade.ts` (6), `pra.ts` (6), `replication.ts` (2), `cronjobs/countItems.ts` (2). Two of them were genuinely broken — the ones this PR fixes. The closest remaining is `steps/replication.ts:468`, which declares 600s and already receives `within 500 seconds` from the feature files, so it works today on 100s of headroom.

Six of the 16 delegate their polling to a helper that hardcodes the budget elsewhere (`steps/utils/utils.ts:446`, 5 min), so for those the relationship spans two files and neither side shows whether the pair is safe.

The ticket also records that a single suite-wide constant was tried here and dropped, since those 16 budgets range from 30s to 20 minutes.

PR #2487, opened 2026-08-07
Stop the cascade loop guard accusing the product, and stop it passing vacuously

delthas:
Split out of #2480, which mixed three topics. This is the loop guard.

The step polls the source for 30s asserting it is never PENDING, to catch a cascade that writes back round the loop. It had a failure at each end.

**It accused the product of a bug it did not have.** The source is legitimately PENDING until its own outbound replication status is written back, so a PENDING observed *before* the source has settled says nothing about the loop. When replication was slow the guard failed there and reported a write-back that had not happened. It now waits for the source to reach a non-PENDING status first, and only judges PENDING observations after that.

**It passed when the object was PENDING the whole time.** With the gate above, an object that never settles never reaches the assertion — so a step named "should never have replication status PENDING" passed for an object that was PENDING for every single poll. That is the one outcome it could least afford to wave through: it means the outbound status was never written back at all. It now fails, with a message that says so rather than blaming the loop.

The step is renamed to `should not return to replication status PENDING`, which is what it actually asserts.

Worth a reviewer's eye: the window stays at the existing 30s. If the source's write-back turns out to be slower than that under churn, this converts a silent pass into a failure — it is one number in the feature file, and the failure message distinguishes "never settled" from "returned to PENDING" so which case fired is unambiguous.

Issue: ZENKO-5340

### DIFF
--- a/tests/functional/ctst/steps/crrCascade.ts
+++ b/tests/functional/ctst/steps/crrCascade.ts
@@ -174,7 +174,7 @@ Then(
 );
 
 Then(
-    'the object at location {string} should never have replication status PENDING within {int} seconds',
+    'the object at location {string} should not return to replication status PENDING within {int} seconds',
     { timeout: 120_000 },
     async function (this: Zenko, location: string, waitSeconds: number) {
         const cascadeBuckets = this.getSaved<Record<string, string>>('cascadeBuckets');
@@ -203,6 +203,11 @@ Then(
             }
             await new Promise(resolve => setTimeout(resolve, 1000));
         }
+        assert.ok(
+            settledAs,
+            `Object at '${location}' never left ReplicationStatus=PENDING within ${waitSeconds}s, `
+            + 'so its outbound replication status was never written back',
+        );
     },
 );
 
--- a/tests/functional/ctst/steps/crrCascade.ts
+++ b/tests/functional/ctst/steps/crrCascade.ts
@@ -183,16 +183,24 @@ Then(
         const deadline = Date.now() + waitSeconds * 1000;
         Identity.useIdentity(IdentityEnum.ACCOUNT, location);
         const client = this.createS3Client();
+        // only a PENDING seen after the source settled means the cascade looped
+        let settledAs: string | undefined;
         while (Date.now() < deadline) {
             const result = await client.send(
                 new HeadObjectCommand({ Bucket: bucket, Key: objectName }),
             );
-            assert.notStrictEqual(
-                result.ReplicationStatus,
-                'PENDING',
-                `Object at '${location}' was found with ReplicationStatus=PENDING, ` +
-                'indicating the cascade loop wrote back to the source.',
-            );
+            if (settledAs === undefined) {
+                if (result.ReplicationStatus !== 'PENDING') {
+                    settledAs = result.ReplicationStatus ?? 'unset';
+                }
+            } else {
+                assert.notStrictEqual(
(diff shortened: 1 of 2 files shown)


DarkIsDude (approved):

tests/functional/ctst/steps/crrCascade.ts
SylvainSenechal, line 1: Side note: I think I introduced a similar pattern here in the old mocha test, can probably leave for now though : https://github.com/scality/Zenko/pull/2447/changes/7d1bd5531ae8ece3ab21c69938b2c42d499fefb8

SylvainSenechal (changes requested):
Wait im surprised about this bug though. 

In the test scenario, there is this step "object should replicate to location...", which does : 
```
        while (Date.now() < deadline) {
            try {
                const res = await client.send(
                    new HeadObjectCommand({ Bucket: bucket, Key: objectName }),
                );
                if (res.ReplicationStatus === 'PENDING') {
                    await new Promise(resolve => setTimeout(resolve, 3000));
                    continue;
                }
                assert.strictEqual(
                    res.ReplicationStatus, 'REPLICA',
                    `Expected ReplicationStatus to be REPLICA at '${location}', got '${res.ReplicationStatus}'`,
                ); 
```

So normally, after this step is finished, we have verified that the object is not pending but a replica, and I don't understand why it would go back to being pending (unless there is a bug) 🤔 

Because in the pr message, you say "The source is legitimately PENDING until its own outbound replication status is written back, so a PENDING observed before the source has settled says nothing about the loop" : But if you look at the while loop code above, as long as the status is pending (which include outbound replication), we continue looping, and only stop to go in the next scenario step once its a replica

EDIT: reading this pr : https://github.com/scality/Zenko/pull/2486/changes

I see you dealt with a bug related to a step passing (with object still pending), while it should've failed, because of a mishandled timeout. But I wonder if this pr is still useful after the fix from the other pr 🤔  

Edit 2: actually im not sure what i said on edit 1 is true, I think this timeout pr doesn't fix any bug and only adds debugging clarity 🤔

delthas:
Closing this — you're right to push back, and the evidence doesn't support it.

**Where the premise differs from what I wrote**, since it's worth recording: the two preceding steps head the *destination* buckets (`crr-location-b`, `crr-location-c`) and assert those are REPLICA. This step heads the *source* (`crr-location-a`), which nothing before it checked. The source's status only becomes COMPLETED when backbeat's replication status processor consumes the A→B status message and writes back to A's metadata — and that write-back races the B→C replication, since both start from "B has the object". So "C is REPLICA" does not imply "A is COMPLETED", and the window is bounded by the status-processor hop, which BB-833 measures unassigned for 45-90s after a rollout.

**But that stays theory.** Only one row of behaviour actually changes here — a source still PENDING when the step starts, which then settles inside the window, no longer failing on the first poll. Everything else (a PENDING after settling, a PENDING for the whole window) fails either way. And that row has never occurred: zero in the 30-run August census, zero in the July annotations.

Same standard as #2479 — not worth changing a CI assertion for a case we have never seen, so I'd rather not spend your review time on it.

One thing left behind for whoever next touches this file: the failure message asserts `indicating the cascade loop wrote back to the source`, which is only one of the two causes the step can't distinguish. If it ever fires on a delayed write-back it will send someone hunting the wrong subsystem. Too small for its own PR; worth folding in next time the file is edited.

Branch stays in place if BB-833 or a future census changes the picture.

PR #2490, opened 2026-08-11
Guard the step budgets that come from the feature file

delthas:
Follow-up to #2486, which guarded two steps this way. This does the remaining five.

A step that takes `within {int} seconds` from the feature file and enforces it against a fixed cucumber timeout gets killed at that timeout when the budget does not fit, losing its own failure message. `assertBudgetFitsStepTimeout` now lives in `common/utils.ts` and takes the timeout as an argument; all seven steps call it on entry, and each timeout is named after the step it governs.

Every budget in `features/` already fits. The tightest is the replication outcome step, 500s against 600s.

Steps whose budget is a constant in the same file stay in ZENKO-5344.

Issue: ZENKO-5344

### DIFF
--- a/tests/functional/ctst/common/utils.ts
+++ b/tests/functional/ctst/common/utils.ts
@@ -1,3 +1,4 @@
+import assert from 'assert';
 import { exec } from 'child_process';
 import http from 'http';
 import { createHash } from 'crypto';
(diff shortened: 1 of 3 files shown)


SylvainSenechal (approved):

DarkIsDude (approved):
thanks

delthas:
!done 2h

delthas:
[`ZENKO-5344`](https://scality.atlassian.net/browse/ZENKO-5344): logged `2h`; transitioned to Done (`Fixed`)

PR #2492, opened 2026-08-12
🐛(ZENKO-5346) remove duplicated word in ConsumerLagWarning description

DarkIsDude:

The `ConsumerLagWarning` alert description in `monitoring/kafka/alerts.yaml` read "has been more more than ... seconds". This text is rendered as-is in Prometheus/Alertmanager notifications, so the typo is user-visible. This removes the duplicated "more".

Fixes ZENKO-5346.

Text-only change to an alert annotation — no rule expression, threshold, or label touched.

### DIFF
--- a/tests/ctst/features/azureArchive.feature
+++ b/tests/ctst/features/azureArchive.feature
@@ -273,26 +273,34 @@ Feature: Azure Archive
         |               Versioned |           2 |      30000 |      10 |
         |               Suspended |           2 |      30000 |      10 |
 
-    @2.7.0
-    @PreMerge
-    @Flaky
-    @AzureArchive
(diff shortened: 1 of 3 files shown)


francoisferrand (approved):

francoisferrand:
probably not worth fixing in 2.11, won't ever be released I think : 2.14 or maybe 2.15 is likely enough... or even just fix on latest version, not very important.

delthas (approved):
Would have been a good candidate for Pod UI agent thing

tests/ctst/features/azureArchive.feature
DarkIsDude, line 1: Backport from more recent branches

PR #2513, opened 2026-08-18
Remove VERSION file

francoisferrand:
Drops the committed `VERSION` file. The next release version is now derived from the branch name (`development/X.Y` or `hotfix/X.Y.Z`) and the existing tags, so there's no need to bump a file before each release — no extra PR to open, review, and clear CI, and no waterflow merge conflicts on `VERSION` anymore. Also fixes the case where a freshly created development branch carries over a stale `VERSION` inherited from its parent branch.

Nothing new conceptually: waterflow and bert-e already enforce the relationship between the branch name and the versions that ship from it (a development/2.15 branch releases 2.15.x tags). This just leans on that invariant directly instead of tracking a redundant copy in a file.

The release workflow now takes a `type` (release / preview / rc) instead of an explicit tag, and computes the version itself before building the ISO, tagging, creating the GitHub release, and promoting the artifacts.

A side effect is that we don't know the actual version in pre-merge anymore - so we need to re-build the ISO during the release. Only the ISO build is enough, i.e. packaging the already-built images, so this is cheap enough, and implemented by extracting the build-iso job in a separate workflow (used both in premerge and release).

Issue: ZENKO-5279


version.sh
SylvainSenechal, line 1: Reading this I realize, in this repo we write everything in bash by default. Wouldn't python be more readable 🤔

SylvainSenechal (approved):
You can do a "yarn upgrade" in /workflows to close 3 dependabots prs. Same for functional tests although you didn't touch these tests here

delthas (approved):
Nice!

PR #2529, opened 2026-09-07
Cover D/R deletion and in-place overwrite in the PRA test

delthas:
The PRA scenario deleted nothing, and its only bucket was versioned — so an object was never overwritten in place, and never deleted without a version to remove. This adds those paths and points the solution at the builds that make them pass.

### Scenarios

| scenario | before | this PR |
| --- | --- | --- |
| `PRA (nominal case)` — `Versioned` | ✓ | ✓, plus a version deletion and an in-place overwrite |
| `PRA (nominal case)` — `Non versioned` | — | restored to the Examples table |
| `PRA (suspended null version deletion)` | — | new scenario |
| `PRA (unversioned deletion)` | — | new scenario |

The `Non versioned` row was in the table until `137e69dd91` swapped it for `Versioned`, in the same commit that added the cross-site restore steps. Nothing records why — not the commit message, not the pull request, not the ticket — and the restore helper it extracted keeps the version id optional, so it runs against a bucket that has none. Since the swap, nothing has restored or failed over an object whose bucket holds no version to carry its state. Both rows run again.

The two deletions get scenarios of their own rather than more steps in the nominal one: a failing step abandons the rest of its scenario, and `@PRA` is assigned at most one pickle at a time, so scenarios cannot fight over the two clusters.

What each of those two covers. A **suspended** bucket's delete creates a null delete marker, and the marker itself is not replicated — it carries the bucket's location constraint, which can never be a cold one. What removes the object on the sink is the *other* write the delete makes: cloudserver removes the null version the marker replaces, and that removal carries the cold object's own metadata. An **unversioned** bucket holds no version to remove at all, so its delete takes away the object's only document — a different metadata event from the version removal the nominal scenario covers.

Deleting an unversioned object already passes on `development/2.16`, but for the wrong reason — the delete flag lands on the master, the document a version-less `HeadObject` reads, so the object reads as absent while its document stays behind forever. That row is a regression guard, not a demonstration.

### What the bumps point at

| component | fix | state |
| --- | --- | --- |
| zenko-operator | scality/zenko-operator#631 — the source pipeline that makes deletions visible | open, based on scality/zenko-operator#641 |
| backbeat | scality/backbeat#2833 — the mongo-processor's D/R mode | open |
| backbeat | scality/backbeat#2842 — omitting the metrics block the D/R sink has no consumer for | merged |
| zenko-drctl | `v1.1.0-preview.1` — carries DRCTL-83, the objects topic on both D/R CRs, and the merged DRCTL-79 that sets the backbeat image on the sink | released |

The two backbeat PRs stay pinned as one build, since the sink needs both and #2833 has not merged. The bump commit is replaced by released versions before this merges, so expect it to be amended until then.

### A known race, not fixed here

Bucket definitions reach the sink through the Kafka Connect entity sync; object entries reach it through the mongo-processor. Nothing orders the two. An object entry that arrives before its bucket gets `NoSuchBucket`, and the offset is committed regardless — so that object's metadata is dropped for good, and the scenario fails on an assertion no amount of waiting can satisfy.

We have not tried to fix it here. The entity path this races is expected to be retired.

Issue: ZKOP-562


solution/deps.yaml
```diff
@@ -2,11 +2,11 @@
 # to sort keys, use the following command
 # yq eval 'sortKeys(.)' -i deps.yaml
 backbeat:
-  sourceRegistry: ghcr.io/scality
+  sourceRegistry: ghcr.io/scality/playground/delthas
```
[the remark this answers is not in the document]
delthas: :clown_face:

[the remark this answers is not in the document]
delthas: :new_moon_with_face:

.github/scripts/end2end/configs/zenko.yaml
```diff
@@ -25,7 +25,17 @@ spec:
       logLevel: debug
   backbeat:
     triggerExpirationsOneDayEarlierForTesting: ${EXPIRE_ONE_DAY_EARLIER}
-    backlogMetricsIntervalSeconds: 5
+    configurationOverrides:
+      # every consumer whose lag the lifecycle conductor waits on, and the
```
DarkIsDude: I don't find very intuitive to read neither understand here.

tests/functional/ctst/features/pra.feature
```diff
@@ -87,6 +87,71 @@ Feature: PRA operations
...
+    When i delete object "objsdel-1" on "Primary" site
+    Then object "objsdel-1" should "not" exist on "DR" site
+
+    @2.6.0
+    @PreMerge
+    @Dmf
+    @PRA
+    @ColdStorage
+    Scenario: PRA (unversioned deletion)
+    # An unversioned bucket holds no version to remove, so a delete takes away
```
DarkIsDude: Why the comment here and before ? They describe the tests and the design ?

DarkIsDude (approved):

.github/scripts/end2end/configs/zenko.yaml
delthas: Dropped

It's just that the field was dropped by https://scality.atlassian.net/browse/ZKOP-561 and so we must use that instead. It's a dumb migration.

tests/functional/ctst/features/pra.feature
delthas: The point is that the test is distinct from just the "unversioned" test row above, so it must be kept. Moved to the PR body.

tests/functional/ctst/steps/pra.ts
```diff
@@ -428,10 +435,84 @@ Given('access keys for the replicated account', { timeout: 360000 }, async () =>
...
+        const expected = this.getSaved<string>('objectETag');
+        assert(expected, 'no write saved an etag to compare against');
+
+        const timeout = 300000;
+        const start = Date.now();
+        let seen;
+
+        while (Date.now() - start < timeout) {
+            const res = await headObjectOnSite(this, objectName, site);
+            if (!res.err) {
```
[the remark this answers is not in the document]
delthas: Good catch. Factored the check into a `throwUnlessMissing` helper so both steps share it instead of repeating the guard.

delthas:
Superseded by #2532 — same commits, reopened on a branch named for its own ticket (ZENKO-5355) rather than borrowing ZKOP-562's.

PR #633, opened 2019-04-03
documentation: ZENKO-1687_update_Grafana_documentation

wabernat:
This PR adds details to the Application-Layer Grafana instructions, and updates some troubleshooting information.  

This fixes ZENKO-1687

PR #2532, opened 2026-09-14
Extend PRA coverage to deletions, overwrites and non-versioned buckets

delthas:
The PRA scenario deleted nothing, and its only bucket was versioned — so an object was never overwritten in place, and never deleted without a version to remove. This adds those paths and points the solution at the builds that make them pass.

### Scenarios

| scenario | before | this PR |
| --- | --- | --- |
| `PRA (nominal case)` — `Versioned` | ✓ | ✓, plus a version deletion |
| `PRA (nominal case)` — `Non versioned` | — | restored to the Examples table |
| `PRA (nominal case)` — `Suspended` | — | new row |
| `PRA (null version in a versioned bucket)` | — | new scenario |
| `PRA (in-place overwrite)` | — | new scenario |

The `Non versioned` row was in the table until `137e69dd91` swapped it for `Versioned`, in the same commit that added the cross-site restore steps. Nothing records why — not the commit message, not the pull request, not the ticket — and the restore helper it extracted keeps the version id optional, so it runs against a bucket that has none. Since the swap, nothing has restored or failed over an object whose bucket holds no version to carry its state. Both rows run again.

Deletion rides the Examples matrix rather than getting scenarios of its own: the steps are identical, and what differs is the bucket they run against. A **versioned** bucket removes a version; an **unversioned** one has no version to remove, so the delete takes away the object's only document; a **suspended** one creates a null delete marker, and the marker itself is not replicated — it carries the bucket's location constraint, which can never be cold. What removes the object there is the *other* write the delete makes: cloudserver removes the null version the marker replaces, and that removal carries the cold object's own metadata. Three different metadata events, one set of steps.

Two cases cannot be parameterised, so they run as their own scenarios rather than being repeated identically in every row. The **in-place overwrite** needs a non-versioned bucket, which no other row provides. The **null version in a versioned bucket** is about ordering: an object written before versioning is enabled stays a bare master with no version id, and stays one after the bucket becomes `Enabled` — the state cloudserver describes as *"a single null version [that] looks like a master key"*. Both scenarios install D/R first and write afterwards, so the write travels the change stream rather than the bootstrap dump.

Deleting an unversioned object already passes on `development/2.16`, but for the wrong reason — the delete flag lands on the master, the document a version-less `HeadObject` reads, so the object reads as absent while its document stays behind forever. That row is a regression guard, not a demonstration.

### What the bumps point at

| component | fix | state |
| --- | --- | --- |
| zenko-operator | scality/zenko-operator#631 — the source pipeline that makes deletions visible | open, based on scality/zenko-operator#641 |
| backbeat | scality/backbeat#2833 — the mongo-processor's D/R mode | open |
| backbeat | scality/backbeat#2842 — omitting the metrics block the D/R sink has no consumer for | merged |
| zenko-drctl | `v1.1.0-preview.1` — carries DRCTL-83, the objects topic on both D/R CRs, and the merged DRCTL-79 that sets the backbeat image on the sink | released |

The two backbeat PRs stay pinned as one build, since the sink needs both and #2833 has not merged. The bump commit is replaced by released versions before this merges, so expect it to be amended until then.

### A known race, not fixed here

Bucket definitions reach the sink through the Kafka Connect entity sync; object entries reach it through the mongo-processor. Nothing orders the two. An object entry that arrives before its bucket gets `NoSuchBucket`, and the offset is committed regardless — so that object's metadata is dropped for good, and the scenario fails on an assertion no amount of waiting can satisfy.

We have not tried to fix it here. The entity path this races is expected to be retired.

Issue: ZENKO-5355


delthas:
@DarkIsDude Moved #2529 here (to get the right ticket ref in the branch)

.github/scripts/end2end/configs/zenko.yaml
```diff
@@ -25,7 +25,15 @@ spec:
...
-    backlogMetricsIntervalSeconds: 5
+    configurationOverrides:
+      backbeat-lifecycle-producer:
+        KAFKA_BACKLOG_METRICS_INTERVALS: "5"
+      backbeat-lifecycle-bucket-processor:
+        KAFKA_BACKLOG_METRICS_INTERVALS: "5"
+      backbeat-lifecycle-object-processor:
+        KAFKA_BACKLOG_METRICS_INTERVALS: "5"
+      backbeat-replication-data-processor:
+        KAFKA_BACKLOG_METRICS_INTERVALS: "5"
```
francoisferrand, line 36: is it not the same? or do we have a bug?

solution/deps.yaml
francoisferrand, line 130: we can't bump ZKOP to 1.9 at this point, analytics is still in flux...

delthas: Keeping this thread open until we decide what to do

solution/deps.yaml
```diff
@@ -2,11 +2,11 @@
 # to sort keys, use the following command
 # yq eval 'sortKeys(.)' -i deps.yaml
 backbeat:
-  sourceRegistry: ghcr.io/scality
+  sourceRegistry: ghcr.io/scality/playground/delthas
```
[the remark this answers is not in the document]
delthas: :new_moon_with_face:

solution/deps.yaml
```diff
@@ -131,7 +131,7 @@ zenko-operator:
   sourceRegistry: ghcr.io/scality
   dashboard: zenko-operator/zenko-operator-dashboards
   image: zenko-operator
-  tag: v1.8.16
+  tag: cf7e60c45ce51a598ef5c39ea3a58057c86627ac
```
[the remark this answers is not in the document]
delthas: :new_moon_with_face:

.github/scripts/end2end/configs/zenko.yaml
delthas: I added the previous value in https://github.com/scality/zenko-operator/pull/626 and it was removed as part of your PR in https://github.com/scality/zenko-operator/pull/633 moving to generic overrides. So I'm updating the CR. (But is this what you're asking?)

francoisferrand: my bad, I forget about this.

But should this not go in a separate PR? That is a fix we must do as soon as we upgrade ZKOP -even on 1.8 branch- (i.e. we are already at risk if we need an urgent zkop bump - easy to fix, but still...), and thus may not be the same timeline as this PR, which depends on a PR in zkop 1.9

tests/functional/ctst/features/pra.feature
```diff
@@ -87,6 +87,61 @@ Feature: PRA operations
...
+    # An object overwritten in place, which only a non-versioned bucket allows,
+    # must carry its new metadata to the DR site
+    Given a "Non versioned" bucket on "Primary" site
+    And a transition workflow to "e2e-cold" location
+    And 1 objects "objow" of size <objectSize> bytes on "Primary" site
+    Then object "objow-1" should "" be "transitioned" and have the storage class "e2e-cold" on "Primary" site
+    And object "objow-1" should "" be "transitioned" and have the storage class "e2e-cold" on "DR" site
+    When i overwrite object "objow-1" with 200 bytes on "Primary" site
+    Then object "objow-1" should "" be "transitioned" and have the storage class "e2e-cold" on "Primary" site
+    And object "objow-1" should have the last written etag on "DR" site
```
francoisferrand: nit?
```suggestion
	And object "objow-1" should "" be "transitioned" and have the storage class "e2e-cold" on "DR" site
    And object "objow-1" should have the last written etag on "DR" site
```

tests/functional/ctst/features/pra.feature
```diff
@@ -87,6 +87,61 @@ Feature: PRA operations
...
+
     Examples:
     | versioningConfiguration | objectCount | objectSize |
```
francoisferrand, line 98: this should not apply to your new `Given a "Non versioned" bucket on "Primary" site` test: that test is only for non-versioned buckets?

tests/functional/ctst/features/pra.feature
```diff
@@ -87,6 +87,61 @@ Feature: PRA operations
...
-    |               Versioned |           2 |        100 |
\ No newline at end of file
+    |           Non versioned |           2 |        100 |
+    |               Versioned |           2 |        100 |
```
francoisferrand, line 100: if we introduce non-versioned, should we do the same for version-suspended and null-versions?

In particular, can we not run the whole scenario in these case instead of the 2 cases you add below?

.github/scripts/end2end/configs/zenko.yaml
delthas: Moved to #2535

tests/functional/ctst/features/pra.feature
delthas: Done — added a `Suspended` row and dropped both standalone deletion scenarios. The steps were identical; only the bucket differed.

tests/functional/ctst/features/pra.feature
delthas: Right — it hardcoded its own non-versioned bucket, so it ran identically in every row. Moved it out to its own scenario, `PRA (in-place overwrite)`.

tests/functional/ctst/features/pra.feature
delthas: Taken, in the new scenario.

.github/scripts/end2end/configs/zenko.yaml
delthas: Split out: #2535. It has to carry the operator bump with it — `v1.8.16` has only the old field and every 1.8 release after your #633 has only the override, so neither change works alone.

tests/functional/ctst/features/pra.feature
delthas: Moved to a separate scenario

tests/functional/ctst/features/pra.feature
delthas: Done, added a `Suspended` row

delthas: Added null-versions too, as `PRA (null version in a versioned bucket)`. It cannot be a row: the point is that the object predates versioning being enabled, so it stays a bare master with no version id inside an `Enabled` bucket. Needs a new `the bucket versioning is set to` step.

delthas: Added null-versions too, as a separate test, since we need to create an object before enabling versioning.

benzekrimaha (approved):
LGTM , don't forget to put the bb tag version before merging

PR #2533, opened 2026-09-14
Serialise drctl boolean flags with the "=" form

delthas:

CTST built every `zenko-drctl` command line with `--flag value`. pflag boolean flags do
not consume the next argument, so `--force-rotate-service-credentials false` set the flag
to **true** and left `"false"` as a positional argument, silently ignored.

Every `Given a DR installed` therefore force-rotated the sink's service credentials,
instead of only the install that follows an uninstall. Each rotation rolls the sink's
backbeat pods and can stall D/R replication, which shows up as intermittent PRA failures:
on run [34585935213](https://github.com/scality/Zenko/actions/runs/34585935213) the six
installs produced six rotations on the sink against one on the source, and both failing
scenarios were replicated deletes that never landed.

Emit booleans as `--flag=value`, and cover `paramToCli` in the tests for
`common/utils.ts`, renamed from `parseGoDuration.test.ts` now that it holds cases for a
second helper.

Fixes ZENKO-5356.


francoisferrand (approved):

tests/functional/ctst/common/utils.ts
```diff
@@ -135,6 +135,29 @@ export function parseGoDuration(duration: string): number {
...
+/**
+ * Serialises an options object into arguments for a Go CLI built with cobra/pflag.
+ * @param {Record<string, unknown>} params - the options to serialise
+ * @return {string} - the arguments, space-separated
+ */
+export function paramToCli(params: Record<string, unknown>): string {
+    const command: string[] = [];
+    Object.keys(params).forEach(key => {
+        const value = params[key];
+        if (value !== undefined && value !== null) {
```
francoisferrand: nit: prefer early exit, to avoid nested `if`s
(and strict equality to undefined or null is actually loose equality to `null`...)

```suggestion
        if (value == null) {
            return;
        }
```

benzekrimaha (approved):

delthas:
!done 2h

delthas:
[`ZENKO-5356`](https://scality.atlassian.net/browse/ZENKO-5356): logged `2h`; transitioned to Done (`Fixed`)

PR #2535, opened 2026-09-16
Set the backlog metrics interval through the generic override

delthas:
`backbeat.backlogMetricsIntervalSeconds` is a dedicated CR field the operator no longer has: scality/zenko-operator#626 added it, scality/zenko-operator#633 reverted it right after `v1.8.16` in favour of the generic `backbeat.configurationOverrides` map.

The two spellings never coexist in a released operator:

| operator | `backlogMetricsIntervalSeconds` | `backbeat.configurationOverrides` |
| --- | --- | --- |
| `v1.8.16` | works | absent — the field exists only on `ZenkoSpecSorbet` |
| `v1.8.17` | removed | works |

So the config and the operator pin have to move together, and bumping the operator without this change fails the whole suite on a rejected field. Hence both in one commit.

The override also needs a backbeat that can take it. The map arrives as environment variables, and 9.5 answers most of those names from an entrypoint that rewrites `conf/config.json` in place — a directory the operator mounts read-only, so the container dies before it serves anything. `KAFKA_BACKLOG_METRICS_INTERVALS` is one of the 57 names rewritten that way, which is what failed this branch eight runs in a row. 9.6 deleted the entrypoint and reads those names from the configuration schema itself (BB-808), so the bump to `9.6.0-preview.1` comes first here. That the operator passes these to a backbeat that cannot take them is ZKOP-605.

`KAFKA_BACKLOG_METRICS_INTERVALS` is backbeat's declared env alias for `kafka.backlogMetrics.intervalS` (`intervalS: joi.number().default(60).meta({ env: 'INTERVALS' })`), so the value reaches the same key with the same unit — seconds — as the removed field did, with joi coercing `"5"` to the number the schema declares. It is set on the four consumers whose lag the lifecycle conductor waits on.

| pin | before | after |
| --- | --- | --- |
| backbeat | `9.5.1` | `9.6.0-preview.1` |
| zenko-operator | `v1.8.16` | `v1.8.17` |

Split out of #2532, which carries the same change but is blocked on a zenko-operator 1.9 PR and so cannot land on this timeline.

Issue: ZENKO-5359


francoisferrand (approved):

benzekrimaha (approved):

delthas:
!done 5h

PR #2537, opened 2026-09-17
Bump jmx-javaagent to 1.5.0

francoisferrand:
banzaicloud stopped publishing jmx-javaagent after 0.16.1 (April 2022);
adobe's fork of koperator ships up-to-date images, and koperator itself
now defaults to 1.5.0.

`jmx_exporter` 1.x renames the JVM metrics to follow the OpenMetrics
convention of putting the unit last, so the kafka and kafka-connect
dashboards move to `jvm_memory_used_bytes` / `jvm_memory_max_bytes`. It also
appends `_total` to any COUNTER whose configured name does not already end
in it, which catches the three broker panels built on koperator's generic
`kafka_$1_$2_$3_count` rules. Custom dashboards and alerts on the old names
need the same treatment.

Zookeeper is not affected: it exposes metrics through its own bundled
PrometheusMetricsProvider, not through the javaagent, and keeps the old
names.

Issue: ZENKO-4844


monitoring/kafka-connect/kafka-connect.dashboard.py
DarkIsDude, line 1: LGTM. But claude says :

```
jmx_exporter 1.x (client_java PrometheusTextFormatWriter.resolveExpositionName) appends _total to every COUNTER whose configured name doesn't already end in it; 0.16.1 emitted the rule name verbatim. koperator 0.31.0's default broker config (used here — zenko-operator's buildMonitoringConfig leaves KafkaJMXExporterConfig empty) has four type: COUNTER rules named kafka_$1_$2_$3_count, which now expose as ..._count_total:

- monitoring/kafka/dashboard.json:1240 — Request handler idle %
- monitoring/kafka/dashboard.json:4225 — log flush rate
- monitoring/kafka/dashboard.json:4439 — leader election rate

The PR only accounts for the JVM metric renames. Series already named ..._total (kafka_server_brokertopicmetrics_bytesin_total, etc.) are unaffected, which is why this slipped through.
```

I checked the doc https://github.com/prometheus/jmx_exporter/blob/1.5.0/docs/content/1.5.0/http-mode/rules.md. Can we check ? Similar to Claude feedback.

DarkIsDude (approved):

SylvainSenechal (approved):
3 names need to be appended with _total

monitoring/kafka-connect/kafka-connect.dashboard.py
francoisferrand: Good catch, confirmed and fixed. I reproduced it by registering the three MBeans against koperator 0.31.0's broker config and scraping both agents: 0.16.1 emits `..._count`, 1.5.0 emits `..._count_total`. The cause is the four generic `kafka_$1_$2_$3_count` fall-through rules, which are the only `type: COUNTER` rules in that config whose name does not already end in `_total`.

Blast radius is exactly the three panels you list, now updated. Everything else checks out: the `kafka_connect_*_count` series come from our own `kafka-connect.yaml`, which has zero COUNTER rules (all GAUGE/UNTYPED), and I verified those names are byte-identical across both agents. Counters already named `_total` are unchanged too, and no alert rule references any of these.

monitoring/kafka-connect/kafka-connect.dashboard.py
```diff
@@ -222,12 +222,12 @@ class Metrics:
...
-        'jvm_memory_bytes_used', 'area', 'instance', job='${job}', namespace='${namespace}'
+    jvm_memory_used_bytes = metrics.Metric(
+        'jvm_memory_used_bytes', 'area', 'instance', job='${job}', namespace='${namespace}'
     ).with_defaults('instance=~"$instance"')
 
-    jvm_memory_bytes_max = metrics.Metric(
-        'jvm_memory_bytes_max', 'area', 'instance', job='${job}', namespace='${namespace}'
+    jvm_memory_max_bytes = metrics.Metric(
+        'jvm_memory_max_bytes', 'area', 'instance', job='${job}', namespace='${namespace}'
     ).with_defaults('instance=~"$instance"')
```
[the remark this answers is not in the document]
francoisferrand: Thanks for checking this. I re-scraped jmx-javaagent 0.16.1 and 1.5.0 with both the Zenko Kafka Connect config and the koperator broker config, and both versions still emit only jvm_gc_collection_seconds_count and jvm_gc_collection_seconds_sum, with no jvm_gc_duration_seconds metric, so I left these queries unchanged.

PR #2543, opened 2026-09-21
Bump Azurite to 3.37.0, refactor Azure client

SylvainSenechal:
ISSUE: ZENKO-5330

- Bumping Azurite from 3.35 to 3.37
- Using this as an opportunity to bump all azure libraries
- Also reducing Cli-Testing dependency : Moving the AzureClient from Cli-Testing to Zenko
- Refactoring around the Clients folder organisation to group them : Long term effort to have one place to store k8s,pensieve,s3,iam,azure clients, Because today it's a bit all over the place.


tests/functional/clients/azure.ts
SylvainSenechal, line 1: This is moslty a copy of https://github.com/scality/cli-testing/blob/development/1.3/utils/AzureHelper.ts

francoisferrand: what is the point of copying it here?

francoisferrand (approved):

tests/functional/clients/azure.ts
SylvainSenechal: Should be easier to have the clients we use directly defined on this repo. 
If we ever need to modify them, we won't need to make a pr in cli-testing

Here I'm not sure I would have been able to bump Azurite to 3.37 without making changes in cli testing for example

benzekrimaha (approved):

